Issue #289, opened 2014-10-30
LinkTo problem parsing in atom with @ in stream name


gregoryyoung:
@ in name causes exception in atom code.

```
[PID:11504:009 2014.10.28 17:55:59.971 ERROR QueuedHandlerThreadP] Error while processing message ReadStreamEventsBackward InternalCorrId: f9434083-92e4-46b9-8325-068fdc7b0386, CorrelationId: f9434083-92e4-46b9-8325-068fdc7b0386, EventStreamId: $streams, FromEventNumber: -1, MaxCount: 20, ResolveLinkTos: True, RequireMaster: False, ValidationStreamVersion:  in queued handler 'StorageReaderQueue #1'.
System.Exception: link not in proper format.
   at EventStore.Core.Services.Transport.Http.Convert.ToEntry(ResolvedEvent eventLinkPair, Uri requestedUrl, EmbedLevel embedContent, Boolean singleEntry) in c:\EventStore\src\EventStore.Core\Services\Transport\Http\Convert.cs:line 252
   at EventStore.Core.Services.Transport.Http.Convert.ToStreamEventBackwardFeed(ReadStreamEventsBackwardCompleted msg, Uri requestedUrl, EmbedLevel embedContent, Boolean headOfStream) in c:\EventStore\src\EventStore.Core\Services\Transport\Http\Convert.cs:line 87
   at With
```

seburgi:
Could you please get this fix into the 3.1.0 release. I think it is not uncommon to use '@' in a stream name. Thanks!

gregoryyoung:
I believe it is in 3.1.0 @pieter can you verify?

On Wed, Jul 15, 2015 at 5:08 PM, Sebastian Burgstaller <
notifications@github.com> wrote:

> Could you please get this fix into the 3.1.0 release. I think it is not
> uncommon to use '@' in a stream name. Thanks!
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/EventStore/EventStore/issues/289#issuecomment-121663366>
> .
>



-- 
Studying for the Turing test

pgermishuys:
@gregoryyoung this is not in 3.1.0. this should be cherry picked.

seburgi:
Great, thanks!

PR #290, opened 2014-10-30
fixes is #289

gregoryyoung:
atom was parsing badly when stream had a @ in it. Also added tests


seburgi:
Would have been great to have this in 3.0.2, we are still on 3.0.0 because this problem makes it impossible for us to upgrade. Please provide a commercial release containing this fix soon.

jen20:
@seburgi this will be released in 3.1.0, which will hit over the next week or so.

seburgi:
Great news, thanks!

PR #347, opened 2015-01-04
Connections

gregoryyoung:
This PR supports connection strings for client api.

There is some weirdness in this commit due to not wanting to introduce a breaking change and should be reviewed. 

There are still overloads of create that work with connectionsettings/clustersettings. The clustersettings type and the overloads that use it have been marked as depreceated though. Likely docs need updating here...

For the new overloads there are 3 primary cases that are handled

Connect to single node:
     Create(new Uri("tcp://user:password@145.33.22.11:44"))

Connect to dns cluster
     Create(new Uri("discovery://whatevername:3030"))

Both also can take ConnectionSettings. If you put things in the uri they override the connection settings

Connect to gossip seeds (or either of the above)
    Create(ConnectionSettings) and Create("Connection String")

either of these will take the information from the settings/string. either of the previous examples can also be done in this way. Ideally we will remove the things on the connectionsettings that exist as well on the uri (default credentials) as its less confusing. Again this is worth discussing before accepting


gregoryyoung:
one more change coming.

PR #350, opened 2015-01-08
Override of chunk size in embedded builder DSL

thefringeninja:
There's currently no way to override the chunk size to avoid the OOM error as discussed here: https://groups.google.com/forum/#!topic/event-store/bER82NC1wDY

PR #360, opened 2015-01-12
Override of chunk size in embedded builder DSL

thefringeninja:


jen20:
This replaces #350 (to target a different branch). After discussion of the use cases this seems reasonable.

PR #362, opened 2015-01-12
Update UI with confirmation of shutdown changes

pgermishuys:
This PR contains the confirmation of the Shutdown as well as the removing of the 'query' navigation item with the same logic as removal of the projection navigation item.


jen20:
Manually merged.

jen20:
Thanks @pgermishuys!

PR #368, opened 2015-01-13
Added Maximum Subscribers to persistant subscribtion

mipi3:
I have a situation when I want to have only one subscriber for the group. The easiest way to do that is to check the subscriber count on the EventStore side.

### DIFF
--- a/src/EventStore.ClientAPI/ClientOperations/ConnectToPersistentSubscriptionOperation.cs
+++ b/src/EventStore.ClientAPI/ClientOperations/ConnectToPersistentSubscriptionOperation.cs
@@ -71,6 +71,12 @@ namespace EventStore.ClientAPI.ClientOperations
                     result = new InspectionResult(InspectionDecision.EndOperation, "SubscriptionDropped");
                     return true;
                 }
+                if (dto.Reason == ClientMessage.SubscriptionDropped.SubscriptionDropReason.SubscriberMaxCountReached)
+                {
+                    DropSubscription(SubscriptionDropReason.NotFound, new Exception("Maximum subscribptions reached."));
+                    result = new InspectionResult(InspectionDecision.EndOperation, "SubscriptionDropped");
+                    return true;
+                }
                 DropSubscription((SubscriptionDropReason) dto.Reason, null, _getConnection());
                 result = new InspectionResult(InspectionDecision.EndOperation, "SubscriptionDropped");
                 return true;
--- a/src/EventStore.ClientAPI/ClientOperations/CreatePersistentSubscriptionOperation.cs
+++ b/src/EventStore.ClientAPI/ClientOperations/CreatePersistentSubscriptionOperation.cs
@@ -23,6 +23,7 @@ namespace EventStore.ClientAPI.ClientOperations
         private readonly int _checkPointAfter;
         private readonly int _minCheckPointCount;
         private readonly int _maxCheckPointCount;
+        private readonly int _maxSubscriberCount;
 
         public CreatePersistentSubscriptionOperation(ILogger log,
                                        TaskCompletionSource<PersistentSubscriptionCreateResult> source,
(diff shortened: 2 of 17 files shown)


gregoryyoung:
Mindaugas i will review this tonight.

Cheers,

Greg

gregoryyoung:
So I will merge this but a comment:

https://github.com/mipi3/EventStore/blob/maxsubscribers/src/EventStore.ClientAPI/ClientOperations/ConnectToPersistentSubscriptionOperation.cs#L76

This probably shouldn't return "NotFound" but deleted shouldn't either (I am changing the deleted as well).

Aside from that it looks reasonable. One question, is this something that could benefit somehow with the monitoring currently in dev (and moving forward)

gregoryyoung:
Thanks!

mipi3:
No problem :) 

Sorry, by your question you mean, would there be some benefit in adding maximum subscribers to ui in competing consumers section, right?

mipi3:
actually I noticed this feature only yesterday, and its very cool. I can now see all the streams and check weather these are active (have any subscribtions) or not and also see the stats on throughput.

mipi3:
I would also benefit from getting information about stream and it's groups directly from a tcp connection... I have a subscriber who on every message have to make a decision weather to create a stream X with persistant subscribtion and send a "SubscribeToPersistentSubscribtion(X)" competing command. If the X stream has 0 subscribers (subscribers drop connection after some idle timeout) it has to send this command over again.

mipi3:
Also there would be benefit if there would be a way to create persistant connection by pattern. For example CreatePersistentSubscription("someipaddress-?", params). After that when I do AppendTo("someipaddress-?", event, etc...) EventStore creates Persistant subscrptions and connections....

gregoryyoung:
There is also a restful api for this see controller

gregoryyoung:
This sounds like a bad idea. Can you explain more about what you are trying
to accomplish?

Also if using only one connection at most on a persistent subscription why
not just use a catchupsubscription?

On Wed, Jan 14, 2015 at 9:24 AM, Mindaugas Pielikis <
notifications@github.com> wrote:

> I would also benefit from getting information about stream and it's groups
> directly from a tcp connection... I have a subscriber who on every message
> have to make a decision weather to create a stream X with persistant
> subscribtion and send a "SubscribeToPersistentSubscribtion(X)" competing
> command. If the X stream has 0 subscribers (subscribers drop connection
> after some idle timeout) it has to send this command over again.
>
> —
> <https://github.com/EventStore/EventStore/pull/368#issuecomment-69878376>.
>



-- 
Studying for the Turing test

mipi3:
Yes, I know sounds complicated :)

The reason for that is rather simple. The server or cluster needs to connect to many devices. The connections per device can be **maximum one**. These connections can be kept for some time and these **must be reliable**. 

So each connection has a one consumer that consumes command queue say Device-{ip}. And there goes commands that a consumer must process while keeping the connection with a device open. These **consumers are created dynamically** when a command MakeConnectionWith(Device-{ip}) is received. These consumersWithConnections must be reliable. So it just dies if something goes wrong (as in Erlang) or after some idle timeout (no work to do, close the connection). When this happens the Device-{ip} stream is left without it's consumer. So now there should be the supervisor that checks if there is the subsribtion for that connection or not. And if there isn't it sends the MakeConnectionWith(ip). To be sure that only one subscriber is subscribed we have max constraint now in EventStore. And it works basically.

A little bit of details. So I have two main streams here

* DeviceCommandsStream - here goes all commands to all devices. Consumer redirects the command to Device-{ip} and checks if there is a subscribtion and consumer on that stream. If not it sends MakeConnectionWith(Device-{ip})

* ChannelConsumerStream - here goes the MakeConnectionWith(Device-{ip}). Each node listens to it to dynamically create consumers for the Device-{ip}.

* and alot of Device-{ip} streams

I hope it's more clear now.

gregoryyoung:
Maybe doing it like in erlang might be better an raise a dead event back to
the supervisor?

How many subscriptions are you planning on having?

mipi3:
Yes, Erlang is in our plans. We have it already integrated with ES, it acts as a server for devices. But for a client to server connection there is a lot of C# already so maybe when we will hit some problems or bottlenecks we will move to Erlang with interop or without.

We are planning to have something about hundreds of thousands, milliion, something like that. I have a concern about threading issues here.

gregoryyoung:
Millions of persistent subscriptions? I think there are better ways of
doing this

mipi3:
What is the max reasonable number of persistent connections in EventStore?

gregoryyoung:
Few hundred. My guess is there is a better way to do what you want. Im
heading to kaunas now but will be around later this week

mipi3:
Few hundred... ooops :) 
It would be great to have a beer :) :+1:

gregoryyoung:
The idea would be to have one handler handle many connections

mipi3:
Yeah that's right, good idea. We can save our model with supervisor, and make one subscriber for all connections, make a category out of Device-{ip}. The next level could be some partitioning to spread connections over a cluster. The next level could be the move to ErlangVm. Great plan :)

gregoryyoung:
Or hash them into say 16 streams then have 16 nodes gor connections

mipi3:
Yeah this would work. No need for one-device-stream-one-counsumer as I stated, can be many-devices-byhash-one-consumer, this will remove the stress from es. cool :)

mipi3:
Thanks!

PR #374, opened 2015-01-15
Cachedchunk

gregoryyoung:
This is changes to pre-cache chunks to avoid some latency at 99.99 when a new chunk gets created. This needs to spend some time in the power outage cluster


jen20:
This needs some attention to see what's conflicting, I suspect some of the stuff around the persistent subscriptions shouldn't be in here?

gregoryyoung:
I think it was a compile error in mono that I fixed :)

On Saturday, March 14, 2015, James Nugent <notifications@github.com> wrote:

> This needs some attention to see what's conflicting, I suspect some of the
> stuff around the persistent subscriptions shouldn't be in here?
>
> —
> <https://github.com/EventStore/EventStore/pull/374#issuecomment-80503020>.
>


-- 
Studying for the Turing test

jen20:
Ah that would make sense. Will merge this manually then.

Issue #371, opened 2015-01-14
move internal stuff in clientapi


gregoryyoung:
its kind of haphazard now

PR #376, opened 2015-01-15
move internal stuff to internal in client api

gregoryyoung:
addresses #371 

Makes it easier to rationalize about visibility of things. All internal things are in internal, public things in root.

Also moved projection manager to .Projections


jen20:
:+1:

Issue #369, opened 2015-01-14
ReadIndex should use temp file during merge


gregoryyoung:
Right now it holds unsafe for too long. Using a temp file will allow the unsafe to only be entered during the move which should be near instantaneous

PR #378, opened 2015-01-17
Index

gregoryyoung:
For #369 

Removes merging.m which doesnt really help anything and causes vast numbers (comparatively) of corrupted index rebuilds from back up

Needs running on power pull cluster


jen20:
Manually merged to `release-oss-3.0.2`

PR #380, opened 2015-01-20
update esquery to support result/resultstream

gregoryyoung:
Added result vs resultstream result coming from projections


gregoryyoung:
Only result works. but after esquery at least works with single result queries, the projection backend no longer seem to support others?

PR #385, opened 2015-01-28
Add --development-mode flag to disable caching

### LINKED ISSUE #384: Add a `--development` mode that turns off cache headers
Have watched a number of people get trolled by caching in the UI, I think it's worth adding a `--development` flag that sets events to non-cacheable.

On second thoughts, given the usage some of these "not for production" flags get, it probably needs to be called `--this-is-not-for-production-and-I-understand-Im-doing-a-bad-thing-but-please-dont-cache`.

jen20:
This disables caching of events if the --development-mode flag is passed by sending a `Cache-Control: max-age=0, no-cache, must-revalidate` header where requests may previously have been cacheable.

Fixes #384.


gregoryyoung:
Looks reasonable

gregoryyoung:
manually merged.

PR #389, opened 2015-02-04
Hdr

gregoryyoung:
First iteration in bringing in hdr histogram. This PR supports it in test client for showing a histogram on read/write floods.


jen20:
:+1: looks good to me

PR #440, opened 2015-04-05
fix typo

t3hnar:


gregoryyoung:
This is a breaking change of anyone using dev and competing consumers. Not a huge problem but one to keep in mind

PR #441, opened 2015-04-05
add expiring of messages to reads (can be improved)

gregoryyoung:
Limits the reader workers in situations where storage may get very very slow (such as remote storage taking 1 minute/read)


jen20:
Add initial version of expiring reads to deal with slow storage

PR #444, opened 2015-04-12
Byecustomqueue

gregoryyoung:
removes custom queue implementations for mono as mono now uses MS reference sources


gregoryyoung:
uhh and appears to add index depth option which is in another PR

PR #447, opened 2015-04-16
Fix race condition in EventStorePersistentSubscription

GojiSoft:
Queue processing (ProcessQueue) may be triggered before _subscription has been set i.e. ProcessQueue may be called whilst _subscription is still null. The fix delays queue processing until _subscription has been set and tries to do this in a way that does not add performance overhead to the normal path (i.e. _subscription has been set). If the subscription has not been initialized, the first attempt to process the queue will result in a single callback being registered to start processing the queue without any blocking.

### DIFF
--- a/src/EventStore.ClientAPI/Common/Utils/Threading/ManualResetEventSlimExtensions.cs
+++ b/src/EventStore.ClientAPI/Common/Utils/Threading/ManualResetEventSlimExtensions.cs
@@ -1,0 +1,34 @@
+﻿using System.Threading;
+using System.Threading.Tasks;
+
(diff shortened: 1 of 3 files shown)


jen20:
@gregoryyoung This looks good to me, any thoughts?

gregoryyoung:
This looks reasonable.

gregoryyoung:
Trying to manually merge gave some interesting test failures here though (I think unrelated)

gregoryyoung:
ok did manual merge (ugly and had to fix some other things) tests running now hopefully can merge in a few

gregoryyoung:
hmm @jen20 I have some unrelated test failures still mostly around users and atom feed

Issue #448, opened 2015-04-19
PersistentSubscription: Retry 10 events causes subscription to block


lscpike:
To reproduce the issue:
- Create a persistent subscription on a stream with > 10 events.
- Connect to the subscription with default buffer size of 10.
- NACK + Retry all messages received.
- You will stop receiving messages after the first 10 retries.

I actually noticed this in the code first. The code path through PersistentSubscriptionClientCollection.RemoveProcessingMessage results in the in flight message being removed form the PersistenSubscriptionClient but AvailableSlots is not incremented.

I'm happy to put together a pull request for this, but it looks suspiciously like there was a reason for this I can't see. Any comments?

gregoryyoung:
It should be. Include tests if a pr. If not I'll be on it maybe tomorrow but likely in a week or two due to baby. If you pr don't feel bad if it sits for a bit before review.

PR #450, opened 2015-04-20
Fix issue where retries don't get correctly moved from the PersistentSubscriptionClient causing an eventual client block.

lscpike:
I raised issue #448 for this originally. Here is my fix with tests.

## The problem
An explicit retry would call a different remove method on the PersistentSubscriptionClient which didn't increment AvailableSlots. With enough retries the client would forever be marked as unable to send.

## Solution
I've removed this method as there seemed to be no real use, and instead use the other method which correctly tracks the event.

There was a similar issue for timed out retries and parks. For parking, the asynchronous callback was not removing the event from the client at all. I added this as well as triggering another push to clients. This required a locking (which I think should have been there before.) 

I can see the case for not doing this when timing out. If an event has timed out it is likely the client is dead and therefore pushing new events to it could be bad. I think that may require discussion.


gregoryyoung:
Thanks!

PR #474, opened 2015-05-07
Make tests work on Windows. 

lscpike:
The only way I could get the tests to run on Windows was to make this change. PortHelper is using 2 different http prefixes.

With a urlacl of http://127.0.0.1:45000/ the attempt to bind to http://+:45000/ gets permission denied. With a urlacl of http://+:45000/ the attempt to bind to http://127.0.0.1:45000/ gets permission denied. 
With both urlacls then http.sys doesn't route correctly and lots of tests get "Server returned 503 (Service Unavailable)"


gregoryyoung:
:thumbsup: avoids run as admin!

PR #476, opened 2015-05-07
Consumer Strategy

lscpike:
## Quick Background

We have an aggregate hosting platform built on top of ActiveMQ and MySQL that we're migrating to EventStore for it's much better fit and impressive performance. One of our requirements is high availability which we support through active/active operation where aggregate instances can be run on any host. We want to avoid aggregate instances moving randomly between hosts, which incurs a large overhead through concurrency retries and state loading. We currently take advantage of the JMS spec for message groups to pin an instance to a consumer until the consumer dies. (See http://activemq.apache.org/message-groups.html for more info.) This avoids unnecessary overhead but maintains excellent resilience!

The new persistent subscription feature in EventStore provides the foundations for everything we need. However, the current round robin and prefer single strategies will lead to a non-optimal distribution of load for this fairly common usage. Therefore, our plan is to take this excellent platform and contribute a new event distribution strategy that behaves similarly to JMS message groups. 

## The Changes

This pull request contains a refactor of the existing 2 strategies into a framework that will allow us to complete the design. The framework takes a named consumer strategy from the persistent subscription settings. The name is then used to construct the correct strategy as an IPersistentSubscriptionConsumerStrategy.

## Technical details
* The ClientApi interface is backwards and forwards compatible via the protobuf contracts. 
* Where the EventStore does not support a named strategy an error is returned.
* On EventStore upgrade the existing settings are updated to use the new named strategy.
* There are tests covering the round robin and single strategies already. All tests pass.
* **Bug Fix**: The UpdatePersistentSubscriptionOperation wasn't updating the max subscriber count. I'm not sure if this was on purpose, but it seemed more like an incomplete change.

### DIFF
--- a/src/EventStore.ClientAPI.Embedded/EventStoreEmbeddedNodeConnection.cs
+++ b/src/EventStore.ClientAPI.Embedded/EventStoreEmbeddedNodeConnection.cs
@@ -429,6 +429,7 @@ namespace EventStore.ClientAPI.Embedded
 
             var corrId = Guid.NewGuid();
 
+
             _publisher.PublishWithAuthentication(_authenticationProvider, credentials, source.SetException, user => new ClientMessage.CreatePersistentSubscription(
                 corrId, 
                 corrId,
@@ -443,11 +444,11 @@ namespace EventStore.ClientAPI.Embedded
                 settings.HistoryBufferSize,
                 settings.LiveBufferSize,
                 settings.ReadBatchSize,
-                settings.PreferRoundRobin,
                 (int)settings.CheckPointAfter.TotalMilliseconds,
                 settings.MinCheckPointCount,
                 settings.MaxCheckPointCount,
                 settings.MaxSubscriberCount,
+                settings.NamedConsumerStrategy,
                 user,
                 credentials == null ? null : credentials.Username,
                 credentials == null ? null : credentials.Password));
@@ -481,11 +482,11 @@ namespace EventStore.ClientAPI.Embedded
                 settings.HistoryBufferSize,
                 settings.LiveBufferSize,
                 settings.ReadBatchSize,
-                settings.PreferRoundRobin,
                 (int) settings.CheckPointAfter.TotalMilliseconds,
                 settings.MinCheckPointCount,
                 settings.MaxCheckPointCount,
                 settings.MaxSubscriberCount,
+                settings.NamedConsumerStrategy,
                 user,
                 credentials == null ? null : credentials.Username,
                 credentials == null ? null : credentials.Password));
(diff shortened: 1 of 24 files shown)


gregoryyoung:
This will take me a few days to review.

In primciple i see no issue with more confirgurable strategies

Given the size of change i will likely make some chabges on top so will use
here to discuss.

Greg

gregoryyoung:
After reading up am i correct that you basically want sticky sessions with
assured ordering within groups? It might be worth discussing a bit further
but i looked at something more generic that i think could support this
using causationid that we use some other places

Greg

lscpike:
The ordering isn't so much the concern as, thanks to the event store, we get a sequential version for all events. The design can be considered sticky sessions, but we're not offering a guarantee, only a best effort. This allows us to rebalance when deemed necessary, such as when new subscribers join. One major flaw in the active mq implementation!

Are you suggesting causation Id as the equivalent jms messagegroupid? Our usage kind of drives our implementation a little:
* Each aggregate will have its own stream of commands and events using a common category per aggregate type.
* The system projection will create us a stream per aggregate type.
* This category stream will have a single persistent subscription.
* Therefore each event has a implicit grouping by the original stream Id. 
* Each original stream id will be pinned to a persistent subscriber client giving better cache hits, less concurrency errors and hopefully optimal throughput.

This of course means the pinned consumer strategy is dependant on a projection to work. It would certainly be better to have an explicit group/aggregate Id.

We have an initial implementation already, but not good enough for a PR yet. I can push it to my fork tomorrow if you want some visibility. Thanks for taking a look at this.  Really appreciate the interest!

lscpike:
Here's a branch with the first iteration of what we want:

https://github.com/lscpike/EventStore/tree/pinnedconsumer

All comments welcome.

gregoryyoung:
Sorry for the delay on this I am still looking it over. To be fair we had a baby recently so I haven't been able to put the time in on this that I normally could

gregoryyoung:
So a quick question.

   
+    [ProtoMember(15, IsRequired = true, Name=@"subscriber_max_count", DataFormat = DataFormat.TwosComplement)]
+    public readonly int SubscriberMaxCount;
+  
+    [ProtoMember(16, IsRequired = false, Name=@"named_consumer_strategy", DataFormat = DataFormat.Default)]
+    public readonly string NamedConsumerStrategy;
+  

Why make them required? We have been trying to get away from required on anything as it makes versioning a lot easier. Would it make more sense to make required false and default to some very large number so we don't break the wire format on client api?

lscpike:
The max subscriber count was already part of the proto message. The code gen hadn't been run with it included, but it was going to conflict so I left it in and updated the code as required. I guess it was left out as an oversight. The max subscriber count in the create contract was complete.

The named consumer strategy is optional. The handler falls back to the original preferroundrobin if not present.

Congratulations on the baby! Much more important than our stuff.

gregoryyoung:
Its totally possible that its an oversight. I will correct it as part of this PR. I will continue looking through it so far it seems mostly :+1:

gregoryyoung:
I'm going to accept this and do the above and possibily 1-2 refactors in dev.

Thanks!
:+1:

PR #487, opened 2015-05-27
Fix assembly merge script to not use undefined Join-String

lscpike:
The cmdlet Join-String isn't standard and doesn't seem to exist in the script anywhere. I've converted the script to use the standard -join.

I also had to remove the /wildcards flag from the embedded client ilmerge. As the docs describe, wildcards and relative parent directories do not work together. I think this is there to support the internalize exclude list, but this works without the flag.

### DIFF
--- a/scripts/package-windows/merge-assemblies.ps1
+++ b/scripts/package-windows/merge-assemblies.ps1
@@ -33,9 +33,8 @@ Function Merge-ClusterNode
     }
 
     # Find other assemblies to merge (with some specifically excluded for e.g. native code)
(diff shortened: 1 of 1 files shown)


gregoryyoung:
Looks good though I will wait for @jen20 to comment.

btw you appear to be in london? if so I'll owe you a beer next time I'm out if you want :)

jen20:
Join-String is part of the powershell community extensions, aka, the stuff that exists because Microsoft didn't finish their work. It annoys me every time I need to create a build agent image and forget it.

:+1: on this.

lscpike:
I am indeed in London in a team with 3 others. I'm sure we'd all be up for a beer at some point! I'll drop you an email from my personal address.

PR #488, opened 2015-05-28
Stop PersistentSubscriptionService receiving duplicate messages.

lscpike:
I noticed this when fixing the UI listing of competing consumers. When starting up the BecomeMaster message is received twice causing the subscriptions to be loaded twice. A race results in duplicates.

It seems that the in memory bus is subscribed twice to the BecomeMaster message via a subscription to StateChangeMessage. This is because they are the same type id. I'm not sure of the design of the internal bus, but I've removed the subscription to the more specific types.

This might not be the correct approach though.


gregoryyoung:
seems reasonable (can do it either way) :+1:

PR #505, opened 2015-06-06
Make the RequestManagementService thread safe

lscpike:
The RequestManagementService subscribes to the MainBus which can dispatch from many threads in parallel. The class uses a Dictionary which is not thread safe.

I've changed to use a ConcurrentDictionary. I'm not sure of the performance implications.

PR #506, opened 2015-06-08
Fix incorrect multi-thread access to the main InMemoryBus

lscpike:
Replaces #505.

Seems the only incorrect access was from the IODispatcher. This seems to only be used by the PersistentSubscription code.

I've changed the code slightly to try and prevent further misuse.


gregoryyoung:
Seems reasonable :+1: Thanks!

PR #507, opened 2015-06-08
Fix batch fail for persistent subscriptions when using linked events

lscpike:
Same fix as for the single ack and fail.


gregoryyoung:
:+1: Looks good

PR #511, opened 2015-06-10
Make MessageHierarchy tolerate of missing assemblies

jageall:
Fixes an issue scanning for message types when running event store embedded.

If you have an assembly(A) loaded (maybe via manual loading into an app domain), and that assembly references another(B), unless you use a method or type that inherits from that a type in B, B does not need to be loaded. If you call GetTypes() and A has types that inherit or expose types from B, it will fail with a ReflectionTypeLoadException.

The scenario was discovered in was using embedded in a test assembly with xunit and running the tests using the visual studio xunit test runner which references a MS test assembly that it doesn't use in the app domain that the embedded instance is running in.


jen20:
Pulling this into release 3.0.6 rather than into dev - it will get merged back from there.

jen20:
Landed in 84ef2efe53ca2fc2bc50dee2d1fb94f0de9a22ee.

PR #514, opened 2015-06-11
Fix the periodic tick in the PersistentSubscriptionService

lscpike:
There's functionality to check for old messages to retry in the persistent subscriptions. This isn't working as the flag to guard against early ticks was incorrectly cleared.

I'm assuming this wasn't on purpose.


gregoryyoung:
:+1:

PR #527, opened 2015-06-18
Add new cluster node stat "proc_cpuScaled" which is "proc-cpu" but scaled by number of logical processors

GojiSoft:
Add new cluster node stat "proc_cpuScaled" which is "proc-cpu" but scaled by number of logical processors to give a value which bounded to 0-100%. The current "proc-cpu" figure has a range of 100% * number of logical processors which makes it tricky where downstream users may not know the number of logical processors or if the number of processors changes e.g. VM environment or hardware changes.

See the following discussion: https://github.com/EventStore/EventStore/issues/521


pgermishuys:
Looks good!

PR #534, opened 2015-07-01
Include metadata at EmbedLevel.Body

### LINKED ISSUE #533: ATOM feed with embed=body doesn't return the metadata created with the application/vnd.eventstore.events+json media type
Posted metadata using the application/vnd.eventstore.events+json media type.

Retrieved using the application/vnd.eventstore.atom+json content type.

The metadata field is not included in the entries.

pgermishuys:
Fixes #533


gregoryyoung:
@pieter what are the unmerited commits? Are they important?

jen20:
It needs further work to not prettify the body at `embed=body` but to do so at `embed=pretty` for metadata, but apart from that LGTM.

PR #545, opened 2015-07-13
Add support to run the run-node.sh script via absolute path

dnauck:
This small change add support to run the run-node.sh script via absolute path. Currently it is only possible to cd into the EventStore directory and call ```./run-node.sh``` otherwise ```LD_LIBRARY_PATH```was wrong and the ```clusternode``` binary was not found.

No you can use the ```run-node.sh```script in environments where you're not allowed to change path before executing the script.

### DIFF
--- a/scripts/package-mono/run-node.sh
+++ b/scripts/package-mono/run-node.sh
@@ -1,3 +1,4 @@
 ##!/usr/bin/env bash
+EVENTSTORE_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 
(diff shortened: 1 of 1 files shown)


jen20:
This will be going away in the official packages anyway as the js1 library gets installed in the correct place

dnauck:
Thanks @jen20 

I'll not be able to use you new debian packages; just upload the tar.gz and unpack it into my service hime dir are allowed ;)

Should i make a pull also agains the 3.1 release?

jen20:
Yes, was just about to ask you to target it at 3.1.0 as well. Thanks!

PR #553, opened 2015-07-16
[WIP] Refactor directory location logic

dnauck:
**This PR is still work in progress and starting point for discussions**

During debugging because of missing logging configuration with the latest v3.1.0 branch with @pgermishuys yesterday we found out that the mkbundle version of eventstore is unable to detect its application directory (where the binaries, web-ui and log.config are located).

mkbundle problem refs: [1] and [2] (i verified both also with latest Mono 4.0.2)

While reading the code i found out that we have location detection code spread all over the EventStore solution. So even if we have a fix for location detection code we cannot fix this easily.

This PR is an **attempt** to centralize the location detection code. The goal is to have only one part in the code that detects application dependencies like config files, web-ui files, plugins, etc. and calculates the correct precedence (e.g. local config file before global config file, etc.)

If you like the idea i can go on the parts i understand. On all other i need help from you.

What is missing:
* projection and prelude includes
* plugin loading/detection
* ...

Why on release-v3.1.0 branch and not dev?
Because these issues with web-ui, projection, prelude and log.config exist in 3.1

[1]: https://bugzilla.xamarin.com/show_bug.cgi?id=18768
[2]: http://mono.1490590.n4.nabble.com/How-to-determine-the-executable-directory-in-a-mkbundle-packaged-app-td4042049.html


dnauck:
Squashed into one commit as requested by @pgermishuys

PR #556, opened 2015-07-18
Separate what we gossip (advertise) vs what we bind to.

pgermishuys:
**New Options**
--ext-ip-advertise-as - External Ip Address Advertise As
--ext-tcp-port-advertise-as - External Tcp Port Advertise As
--ext-http-port-advertise-as - External Http Port Advertise As

This allows us to gossip (advertise) separate information from what we are binding to.


gregoryyoung:
I will add an issue for the others. :+1:

PR #559, opened 2015-07-20
Use correct path on executing eventstore

dnauck:


pgermishuys:
Only saw this now, this was already changed in the dist-packages branch. Sorry.

PR #577, opened 2015-08-04
Correct the replay parked event version tracking and pushing to clients.

gregoryyoung:
Enforce the max retry contract on explicit NAK retry.

### DIFF
--- a/src/EventStore.Core.Tests/Services/PersistentSubscription/PersistentSubscriptionTests.cs
+++ b/src/EventStore.Core.Tests/Services/PersistentSubscription/PersistentSubscriptionTests.cs
@@ -842,22 +842,31 @@ namespace EventStore.Core.Tests.Services.PersistentSubscription
                     .WithEventLoader(new FakeStreamReader(x => { }))
                     .WithCheckpointReader(reader)
                     .WithCheckpointWriter(new FakeCheckpointWriter(i => { }))
+                    .WithMaxRetriesOf(10)
                     .WithMessageParker(parker)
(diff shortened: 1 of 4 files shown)


gregoryyoung:
@lscpike I am going to bring this in. The not using rules on retries if explicitly retried was a feature not something missing (e.g. use unknown if you want the behaviour of defaults). I personally think this makes sense though.

lscpike:
Excellent thanks. #539 is also relevant to get retries working fully if you have 2 minutes.

gregoryyoung:
Tbf we should probably have named it "let server decide"

lscpike:
Without this behaviour it's difficult for clients to know when to give up and NAK park. It would be handy to send the retry count to the client but that's more protocol changes. Maybe we could make Unknown follow this path and explicit retry always retries. A little unnatural, but docs can clarify it.

gregoryyoung:
Thats what was happening. I kind of like the idea of making the limits
explicit as you have done. Max Retries is 10 never any more regardless of
what the client says.

On Tue, Aug 4, 2015 at 6:18 PM, lscpike <notifications@github.com> wrote:

> Without this behaviour it's difficult for clients to know when to give up
> and NAK park. It would be handy to send the retry count to the client but
> that's more protocol changes. Maybe we could make Unknown follow this path
> and explicit retry always retries. A little unnatural, but docs can clarify
> it.
>
> —
> <https://github.com/EventStore/EventStore/pull/577#issuecomment-127663364>
> .
>



-- 
Studying for the Turing test

PR #580, opened 2015-08-06
Idempotent User Creation Behaviour

pgermishuys:
When attempting to create an existing user with the same password - Reply 201 - Created
When attempting to create an existing user with a different password - Reply 409 - Conflict


gregoryyoung:
:+1:

PR #599, opened 2015-08-21
Add Competing Consumers

pgermishuys:
Manual Merge of
https://github.com/EventStore/EventStore/commit/0ed5854c43f1fd51129bc03eb49926a2dfc43bc7

**An additional change was made.**
a_nak_in_subscription_handler_in_autoack_mode_drops_the_subscription was
failing and reporting UserInitiated as the reason from dropping the
subscription instead of EventHandlerException which was what was happening
The fix was to report the dropped subscription reason before unsubscribing


gregoryyoung:
merging. looks good from looking through tests + a few random diffs from dev.

PR #610, opened 2015-08-25
Add Max Read Size to Client API

pgermishuys:
Manual merge of
https://github.com/EventStore/EventStore/commit/a18646dda17a01d855fedf68a4b5d8104bd4f2dd


gregoryyoung:
:+1:

CumpsD:
Ouch, just upgraded libraries and this one did hit me. I had this code:

```
    let getState applyEvents initState id es =
        readFromStream es id 0 Int32.MaxValue
        |> Async.map (fun (events, _, _) -> events)
        |> Async.map (applyEvents initState)
```

To read a stream from start to end to hydrate an entity, which suddenly broke

jen20:
For streams with lots of events it would have blown up in a much nastier manner (they all go in one reply package which have a natural size limit anyway).

CumpsD:
I'm reading them in slices of 4096 now, does this code have any drawbacks in your opinion?

https://github.com/exira/ges/blob/master/src/Exira.EventStore/EventStore.fs#L32-L62

PR #615, opened 2015-08-26
Mark Obsolete in favor of URI and ConnectionStrings

pgermishuys:
Use non-Obsolete Create method in Projections Tests

Manual merge of https://github.com/EventStore/EventStore/commit/d58dc42efa998348ed20a27f847701966b506a11


jen20:
I think we should consider whether we want to at least have one release of the Client API with connection strings in before marking the builders as obsolete - or even whether it's a good idea to obsolete them?

Any thoughts @gregoryyoung @pgermishuys?

gregoryyoung:
For me obsoleting them isn't a big deal. At worst people will get a compiler warning.

PR #616, opened 2015-08-26
Take event position into consideration when redirecting to stream

### LINKED ISSUE #295: uris in client
basic things like /streams/foo have changed and no longer work with ui (3.0.1)

pgermishuys:
Fixes #295 
The last issue mentioned on #295 was that using an event number is not taken into consideration when redirecting from /streams/{stream-name}/{event-number} to /web/index.html#/streams/{stream-name}


gregoryyoung:
Good one

PR #618, opened 2015-08-26
refactored PesistentSubscriptionConsumer strategies (RoundRobin and S…

riccardone:
I refactored RoundRobinPersistentSubscriptionConsumerStrategy and DispatchToSinglePersistentSubscriptionConsumerStrategy in order to use a Queue datastructure instead of a list. This remove the need to use a "Move to next" method, a class field _currentIndex and a class property CurrentIndex. I guess that performance are not negatively affected but this is something to test. I think that the new Round Robin implementation is executed in circular order. If we want to implement multithreading in the future we can just use the "Cuncurrent" version of the Queue data structure.
Hope this help.


jen20:
@gregoryyoung any thoughts on this?

riccardone:
Please let me know if any additional information is required

gregoryyoung:
Looks good :+1:

PR #623, opened 2015-08-27
Add index depth option

pgermishuys:
Manual merge of https://github.com/EventStore/EventStore/pull/443
One minor change to the manual merge is the setting of the index depth to 16 instead of the 12 in the PR in ClusterVNodeSettings


gregoryyoung:
:+1:

PR #628, opened 2015-08-28
Failed resolved links handled differently in PersistentSubscriptionService

lscpike:
### Issue
In the PersistentSubscriptionService the failed resolve would create a resolved event with Event set to the link and Link as null. This is in contrast to all other areas where Link is set to the link and Event is null. This was most likely because the ClientAPI checks for Event = null to detect the unsubscribe request. With the Event set to null the ClientApi will throw an error when the event is received. 

This was working in the obvious case, but when a historical bad link is read via the storage reader and resolved (not via the PersistentSubscriptionService resolve) the other format is used and the error case is hit. This stops the client from being able to process the stream as it drops the subscription on every attempt.

### Solution
I've changed the ResolvedEvent class to have clear factory methods for the 3 scenarios it appears to support. The PersistentSubscriptionService now uses the correct format. I've then fixed the ClientAPI to handle the correct format of Link != null and Event == null.

This is a change to the public interface but as I think the original behaviour was unclear and maybe unintended it is probably better to make the break sooner rather than later.

### Tests
To get all tests running on this branch I had to change specification_with_standard_projections_runnning.cs to use the ExtHttpEndPoint instead of IntHttpEndPoint. I also had to apply the fix from PR #474 to get the urlacls to work on Windows again (can this be included in this release branch please?).

There are 2 tests which are failing still for me (before and after my changes):
- a_failed_projection.when_starting.the_projection_status_becomes_running_enabled()
- a_completed_projection.when_starting.the_projection_status_becomes_running_enabled()

They both look incorrect to my ignorant eye.

### DIFF
--- a/src/EventStore.ClientAPI/EventStorePersistentSubscriptionBase.cs
+++ b/src/EventStore.ClientAPI/EventStorePersistentSubscriptionBase.cs
@@ -178,7 +178,7 @@ namespace EventStore.ClientAPI
                 ResolvedEvent e;
                 while (_queue.TryDequeue(out e))
                 {
-                    if (e.Event == null) // drop subscription artificial ResolvedEvent
+                    if (e.Equals(DropSubscriptionEvent)) // drop subscription artificial ResolvedEvent
                     {
                         if (_dropData == null) throw new Exception("Drop reason not specified.");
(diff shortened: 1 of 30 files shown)


gregoryyoung:
Thanks @lscpike

PR #630, opened 2015-08-31
Add new cluster node stat "proc_cpuScaled"

pgermishuys:
Original PR https://github.com/EventStore/EventStore/pull/527
What do you think @gregoryyoung and @jen20 ?


jen20:
:+1: this can be a useful metric

PR #635, opened 2015-09-02
Fix english

pgermishuys:
Pulling over the fantastic work @NetChris has done on the original "dev (now competing-etc)" branch


gregoryyoung:
:) Thanks again @NetChris this is now staged for release.

PR #648, opened 2015-09-02
Remove Obsolete as part of this release

pgermishuys:
Remove Obsolete as part of release-v3.2.0 (The first introduction of Connection Strings and URIs). This can be marked as Obsolete in a future release.


jen20:
:+1: better to have at least one release without the obsolete mark IMO

PR #660, opened 2015-09-07
Add support to detect SLES

dnauck:
Allow us to detect SuSE Linux Enterprise Server including version


dnauck:
An example SuSE-brand file looks like:

```
SLES
VERSION = 11
CO-BRANDS = SLE openSUSE
```

PR #662, opened 2015-09-07
Use correct casing for the WERROR_STRING variable

dnauck:
Missed this from the last commit


jen20:
I was wondering about this ;-)

PR #664, opened 2015-09-08
Add SLES 11 build of libjs1

dnauck:
Added a pre build version of the libjs1 for SuSE Linux Enterprise Server 11

This build was compiled from the oss-v3.2.0 branch on SLES 11 x64


jen20:
Sorry, this isn't one we want to support - we only intend on adding supported OS builds of js1

PR #665, opened 2015-09-08
Implemented 494 story about mapping ClientApi queries to DTO objects

riccardone:
I implemented the story following the already existent UserManager implementation. 
I created a DTO class called ProjectionDetails that reflect the JSon results that I found. I don't know exactly the meaning of each field so I was not able to describe it in the field comment.
I wrote some integration tests in the following class:
EventStore.Core.Tests/ClientAPI/Projections/list_projections.cs

The above changes break some tests that I fixed but I'm not able to run those tests right now because I don't have a test environment ready.
The tests are:
src/EventStore.Projections.Core.Tests/ClientAPI/specification_with_standard_projections_runnning.cs
src/EventStore.Projections.Core.Tests/ClientAPI/Cluster/specification_with_standard_projections_runnning.cs
and the fix that allow me to compile but probably need to be reviewd is:
all.Any(p => p.Name == "Faulted")

I also had to fix a problem with the EventStore.Padmin.Commands class that I end up fixing with the following code:
LogUnformatted(string.Join(" | ", manager.ListAllAsync(userCredentials).Result));
I did not found tests for this so I was not able to check properly this fix.

Cheers
Riccardo

### DIFF
--- a/src/EventStore.ClientAPI/Projections/ProjectionDetails.cs
+++ b/src/EventStore.ClientAPI/Projections/ProjectionDetails.cs
@@ -1,0 +1,217 @@
+﻿using System;
+
+namespace EventStore.ClientAPI.Projections
+{
+    /// <summary>
+    /// Provides the details for a projection.
+    /// </summary>
+    public sealed class ProjectionDetails
+    {
+        /// <summary>
+        /// The CoreProcessingTime
+        /// </summary>
+        public readonly int CoreProcessingTime;
+
+        /// <summary>
+        /// The projection version
+        /// </summary>
+        public readonly int Version;
+
+        /// <summary>
+        /// The Epoch
+        /// </summary>
+        public readonly int Epoch;
+
+        /// <summary>
+        /// The projection EffectiveName
+        /// </summary>
+        public readonly string EffectiveName;
+
+        /// <summary>
+        /// The projection WritesInProgress
+        /// </summary>
+        public readonly int WritesInProgress;
+
+        /// <summary>
+        /// The projection ReadsInProgress
+        /// </summary>
+        public readonly int ReadsInProgress;
+
+        /// <summary>
+        /// The projection PartitionsCached
+        /// </summary>
+        public readonly int PartitionsCached;
+
+        /// <summary>
+        /// The projection Status
+        /// </summary>
+        public readonly string Status;
+
+        /// <summary>
+        /// The projection StateReason
+        /// </summary>
+        public readonly string StateReason;
+
+        /// <summary>
+        /// The projection Name
+        /// </summary>
+        public readonly string Name;
+
+        /// <summary>
+        /// The projection Mode
+        /// </summary>
+        public readonly string Mode;
+
+        /// <summary>
+        /// The projection Position
+        /// </summary>
+        public readonly string Position;
+
+        /// <summary>
+        /// The projection Progress
+        /// </summary>
+        public readonly decimal Progress;
+
+        /// <summary>
+        /// LastCheckpoint
+        /// </summary>
+        public readonly string LastCheckpoint;
+
+        /// <summary>
+        /// The projection EventsProcessedAfterRestart
+        /// </summary>
+        public readonly int EventsProcessedAfterRestart;
+
+        /// <summary>
+        /// The projection StatusUrl
+        /// </summary>
+        public readonly Uri StatusUrl;
+
+        /// <summary>
+        /// The projection StateUrl
+        /// </summary>
+        public readonly Uri StateUrl;
+
+        /// <summary>
+        /// The projection ResultUrl
+        /// </summary>
+        public readonly Uri ResultUrl;
+
+        /// <summary>
+        /// The projection QueryUrl
+        /// </summary>
+        public readonly Uri QueryUrl;
+
+        /// <summary>
+        /// The projection EnableCommandUrl
+        /// </summary>
+        public readonly Uri EnableCommandUrl;
+
+        /// <summary>
+        /// The projection DisableCommandUrl
+        /// </summary>
+        public readonly Uri DisableCommandUrl;
+
+        /// <summary>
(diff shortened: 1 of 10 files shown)


pgermishuys:
:+1: thanks @riccardone 
I can run the tests this side, will look at this a bit later and provide feedback if not merging it in.

src/EventStore.ClientAPI/Projections/ProjectionDetails.cs
```diff
@@ -0,0 +1,217 @@
...
+{
+    /// <summary>
+    /// Provides the details for a projection.
+    /// </summary>
+    public sealed class ProjectionDetails
+    {
+        /// <summary>
+        /// The CoreProcessingTime
+        /// </summary>
+        public readonly int CoreProcessingTime;
```
jen20: Need to investigate whether this should be an `int` or some other type (perhaps `Timespan` given the name though I'm not sure what it actually represents without looking.

src/EventStore.ClientAPI/Projections/ProjectionDetails.cs
```diff
@@ -0,0 +1,217 @@
...
+
+        /// <summary>
+        /// The projection Progress
+        /// </summary>
+        public readonly decimal Progress;
+
+        /// <summary>
+        /// LastCheckpoint
+        /// </summary>
+        public readonly string LastCheckpoint;
```
jen20: This should be a `Position` in some cases I think?

src/EventStore.ClientAPI/Projections/ProjectionDetails.cs
```diff
@@ -0,0 +1,217 @@
...
+
+        /// <summary>
+        /// The projection DisableCommandUrl
+        /// </summary>
+        public readonly Uri DisableCommandUrl;
+
+        /// <summary>
+        /// The projection CheckpointStatus
+        /// </summary>
+        public readonly string CheckpointStatus;
```
jen20: Perhaps we should make an enumeration of the possible values here

src/EventStore.ClientAPI/Projections/ProjectionsClient.cs
```diff
@@ -50,23 +52,41 @@ public Task CreateTransient(IPEndPoint endPoint, string name, string query, User
...
+                    .ContinueWith(x =>
+                    {
+                        if (x.IsFaulted) throw x.Exception;
+                        var r = JObject.Parse(x.Result);
+                        return r["projections"] != null ? r["projections"].ToObject<List<ProjectionDetails>>() : null;
+                    });
         }
 
-        public Task<string> ListContinuous(IPEndPoint endPoint, UserCredentials userCredentials = null)
+        public Task<List<ProjectionDetails>> ListContinuous(IPEndPoint endPoint, UserCredentials userCredentials = null)
```
jen20: Do we think we need to continue providing a `Task<string>` implementation (deprecated) for a while? @gregoryyoung, @pgermishuys?

jen20:
Hi @riccardone, I've added a few comments, mostly about the types in the field and also a couple of questions for @gregoryyoung and @pgermishuys about whether we want to keep the old API too.

Looks good, thanks for your contribution!

riccardone:
Hi @jen20 
ok, as soon as we'll get some answers I will refactor to the code

src/EventStore.ClientAPI/Projections/ProjectionsClient.cs
pgermishuys: I don't see a problem with doing that for 1, maybe 2 releases.

src/EventStore.ClientAPI/Projections/ProjectionDetails.cs
pgermishuys: Agreed, I am happy with this being a string in this PR and have a separate one refactoring the CheckpointStatus.

src/EventStore.ClientAPI/Projections/ProjectionDetails.cs
pgermishuys: We could possibly setup another structure that shows Stream/Position and provide an array
Something along the lines of
```
StreamPosition[] streamPositions
```
Also, could be part of another PR.

src/EventStore.ClientAPI/Projections/ProjectionsClient.cs
gregoryyoung: I think even changing it straight is actually ok projections are known to
be in flight. Also anyone using those methods would likely be quite happy
and have little exposure to the string based ones.

On Tue, Sep 8, 2015 at 7:57 PM, James Nugent <notifications@github.com>
wrote:

> In src/EventStore.ClientAPI/Projections/ProjectionsClient.cs
> <https://github.com/EventStore/EventStore/pull/665#discussion_r38958155>:
>
> >          }
> >
> > -        public Task<string> ListContinuous(IPEndPoint endPoint, UserCredentials userCredentials = null)
> > +        public Task<List<ProjectionDetails>> ListContinuous(IPEndPoint endPoint, UserCredentials userCredentials = null)
>
> Do we think we need to continue providing a Task<string> implementation
> (deprecated) for a while? @gregoryyoung <https://github.com/gregoryyoung>,
> @pgermishuys <https://github.com/pgermishuys>?
>
> —
> <https://github.com/EventStore/EventStore/pull/665/files#r38958155>.
>



-- 
Studying for the Turing test

riccardone:
I added task<string> methods as requested. 
I found a class called ProjectionStatistics that seems very similar to the new ProjectionDetails. In the ProjectionStatistics class there is an enum Mode that we probably need to use in ProjectionDetails (but I did not add the dependency yet so let me know). I changed some of the types according to that class. 
@jen20 you mention some type changes but can you please confirm because you seemed unsure?

jen20:
:+1: thanks for implementing the @riccardone, I'll merge this now and then any further changes can be made once people have played with it

PR #681, opened 2015-09-14
Improve inefficient connection string parsing

pbolduc:
Although parsing connection strings is performed infrequently, it shouldn't be inefficient. 

In the Client API ConnectionSettings class, there are 23 public properties, 22 of which are multi-word.  The WithSpaces method will loop 334 times.  That is a lot of string concatenation.  There is no need to convert strings to upper case.  Dictionary can use a case insensitive equality comparer.


jen20:
:+1:

Issue #690, opened 2015-09-18
Connection to localhost not working


rbanks54:
(using v3.2.1)

I was trying to connect to localhost, but was having problems. I found the following worked

var c = EventStoreConnection.Create(new Uri("tcp://127.0.0.1:3300"))
var c = EventStoreConnection.Create(new Uri("tcp://x.localtest.me:3300"))

But these didn't
var c = EventStoreConnection.Create(new Uri("tcp://localhost:3300"))
var c = EventStoreConnection.Create(new Uri("tcp://[::1]:3300")))

Given a ping of localhost (on my machine) returns an IPv6 address, it's seems like either the server doesn't support IPv6 at this point in time, or that it only binds on IPv4 addresses.

Also, the assumption made by this line of code https://github.com/EventStore/EventStore/blob/release-v3.3.0/src/EventStore.ClientAPI/EventStoreConnection.cs#L122 would appear to be incorrect. It looks like we need to enumerate the list to find an IPv4 address when resolving names.

Or am I missing something? I could well be. :-)

gregoryyoung:
"Given a ping of localhost (on my machine) returns an IPv6 address, it's seems like either the server doesn't support IPv6 at this point in time, or that it only binds on IPv4 addresses."

Right now I believe it only binds IPv4

rbanks54:
Not knowing the internals, is that a big change to make?

pgermishuys:
It would require a bit of effort I would imagine,
For example check that our URIs are properly formatted for IPv6 (The Http Server that we spin up uses this).

pgermishuys:
Closed by #692

PR #692, opened 2015-09-18
Ensure an ipv4 address is used when resolving dns names for tcp #690

rbanks54:
I couldn't see an easy way to unit test this, since the IPaddress isn't exposed on the connection anywhere and I didn't want to go using reflection to call private methods in unit tests.
As such I've only manually tested it.


gregoryyoung:
This seems like a reasonable stop gap. However I want to discuss further how we want to support v6

PR #701, opened 2015-09-24
Exit the application when the Db or Log directories starts with a tilde

### LINKED ISSUE #693: [Mono] Tilde expansion
Slight gripe is that if you are not paying attention (like I was) and do `mono EventStore.ClusterNode.exe --db=~/tmp/eventStoreData --log=~/tmp/eventStoreLog` it doesn't do any tilde expansion and in fact creates a directory called "~" in the current directory (./bin for me).

Now this is obviously not desired, but more so it could be dangerous, as somebody *might* - without thinking - then try and delete this `rm -rf ~` which obviously would be disastrous wiping out their home directory. Obviously that didn't happen to me, oh no.

PS. `extundelete` works well in these scenarios ;)

pgermishuys:
Fixes #693
When the Db or Log paths starts with a tilde, the application will exit.
The error can be overridden using the --force option.

### DIFF
--- a/src/EventStore.ClusterNode/Program.cs
+++ b/src/EventStore.ClusterNode/Program.cs
@@ -51,6 +51,13 @@ namespace EventStore.ClusterNode
         {
             base.PreInit(options);
 
+            if (options.Db.StartsWith("~") && !options.Force){
+                throw new ApplicationInitializationException("The given database path starts with a '~'. We don't expand '~'. You can use --force to override this error.");
+            }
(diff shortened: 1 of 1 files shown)


gregoryyoung:
I'm not sure contains is the right thing here.

I could very well say I want a file greg~foo~bar

gregoryyoung:
or maybe make it overridable with --force like other things

gregoryyoung:
:+1:

Issue #700, opened 2015-09-24
Embedded Client NuGet package has wrong dependency on Client package


ArturDorochowicz:
EventStore.Client.Embedded 3.2.1 NuGet package has a dependency on EventStore.Client >= 3.0.0, but it doesn't work with this version due to namespace change ```EventStore.ClientAPI.Core``` -> ```EventStore.ClientAPI.Internal```.

In particular ```EmbeddedEventStoreConnection.Create()``` fails with ```TypeLoadException``` about missing ```IEventStoreTransactionalConnection```.

The change was introduced in SHA: ce79d904ffd7254a3be7e5fc80e3d7561bdbe423 Move Client API Internal stuff to /Internal.

An obvious workaround now is to update the client package right after installing the embedded client.

jen20:
We should take a hard dependency only on the same version (i.e. 3.2.1 server = 3.2.1 client only).

No idea how to specify that in NuGet but will look into it.

PR #702, opened 2015-09-24
Take a hard dependency on EventStore.Client of the same version

pgermishuys:
As a result of #700 
Packaging of the EventStore.Client* packages are done via a single script. With this change, the EventStore.Client.Embedded will take a strict dependency on the same version of the EventStore.Client package.

This will eliminate the issue from #700 for future versions.


jen20:
Fixes #700 for future versions.

PR #707, opened 2015-09-29
Split development mode

### LINKED ISSUE #706: --development-mode starts projections when running on disk but not in memory
While debugging an issue with @sbellware, I discovered that `--development-mode` will run the standard projections unless `--mem-db` is specified. Probably some form of race condition that we should look into.

We should also rename and split the flag into `--disable-http-caching` and `--start-standard-projections` so as to not end up with some mode that has some of the options but not all of them.

pgermishuys:
Fixes #706

- Development mode is now 2 separate flags, one for disabling http cache and
  another for starting standard projections
- Wait for the individual projections to be at the stopped state before
  attempting to enable them when the "StartStandardProjections" flag has
  been set to true
- Embedded Client also contains a **breaking** change where the **EnableDevelopmentMode** method has been replaced by 2 new methods, **StartStandardProjections** and **DisableHTTPCaching**. We could leave the EnableDevelopmentMode in there and just set the appropriate settings in there? thoughts @gregoryyoung and @jen20

### DIFF
--- a/src/EventStore.ClientAPI.Embedded/EmbeddedVNodeBuilder.cs
+++ b/src/EventStore.ClientAPI.Embedded/EmbeddedVNodeBuilder.cs
@@ -31,7 +31,8 @@ namespace EventStore.ClientAPI.Embedded
         private string _dbPath;
         private long _chunksCacheSize;
         private bool _inMemoryDb;
-        private bool _developmentMode;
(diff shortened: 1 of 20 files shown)


jen20:
Looks good to me - my guess is no-one is using this in embedded mode, but if they are they'll just get a compile time error when they upgrade?

PR #708, opened 2015-10-01
If an epoch cannot be found, start from beginning

### LINKED ISSUE #691: Epochs
We should keep more epochs on the master. Right now a node that is catching up fro very far behind if it loses connection and tries to reconnect will run into an issue of too many epochs. Increasing the number of epochs recognized by the master will help with this situation though not totally resolve it. Another option is in this case to start over in the pre-replica node that is catching up (though this can obviously lead to problems)

pgermishuys:
Fixes #691


jen20:
LGTM.

gregoryyoung:
I have been talking with @pgermishuys about this and will revert this for now.

PR #727, opened 2015-10-14
Set index path via options

pgermishuys:
Gives the user the option of specifying where the Index is stored
```
./clusternode --index ~/EventStoreDb/index
```
**PS: It does not append `index` to the given path**


gregoryyoung:
:+1:

PR #728, opened 2015-10-17
Use single hash in shebang

ruudud:
When running an executable, the kernel (and other executors) looks at the first
few bytes to determine the type of the file.
Some get confused when met by double hash.


jen20:
:+1: that's a bug.

PR #730, opened 2015-10-19
Honour projection authorisation

pgermishuys:
Move the authorisation check out of the Projection and into the
Projection Manager.

The authorisation check is performed inside of the Managed Projection, and therefor
poses a problem when the Projection Manager needs to perform a series of
steps after the Projection has completed it's work.

The use case that exposed a problem is the deletion of a projection which
replied with a NotAuthorised message to the user, but the Projection Manager continued
deleting the projection as it had no idea that a check failed inside of
the Managed Projection.


gregoryyoung:
:+1:

PR #731, opened 2015-10-19
Add Appveyor configuration file

jen20:


dnauck:
+1

Do you also plan to provide "nightly" builds?

jen20:
@dnauck the artefacts are available via Appveyor, yes. We've not yet discussed whether to provide official nightly builds. If we do we need to provide some way of making it incredibly explicit that those running them do at their own risk.

dnauck:
I'm using SemVer for this as it make it clear, e.g. 3.4.0-dev vs. 3.4.0, etc.

jen20:
That might work, it will need some scripts adapting, however. We may also have to adjust which attribute in the DLLs the version comes from in order to support a proper semver string.

PR #737, opened 2015-10-23
Http competing consumers

### LINKED ISSUE #309: Competing HTTP
I will be meeting on this today but am leaning towards using the same formatter as for atom to ease transition from one feed to another. Does anyone have any issues with this? 

Basically it would add 2 rel links to the feed (ack all) and a del link for ack on the item itself. The feed would look identical but would never be cacheable. This would allow for taking an existing atom client and moving it to competing in a few minutes at most.

pgermishuys:
Fixes #309

Merge of `chrisgreghttp`

**Creating a Subscription**
`PUT` to `/subscriptions/{stream}/{subscription}`

**Update a Subscription**
`POST` to `/subscriptions/{stream}/{subscription}`

**Getting Messages (Atom Feed)**
`GET` to `/subscriptions/{stream}/{subscription}/{count}?embed={embed}`
 - The last part of the uri ({count}?embed={embed}) is optional
 - Count defaults to 1
 - Embed defaults to EmbedLevel.None

**Acking a Single Message**
`POST` to `/subscriptions/{stream}/{subscriptions}/ack/{messageid}`

**Acking a set of Messages**
`POST` to `/subscriptions/{stream}/{subscriptions}/ack?ids={messageids}`

**Nacking a Single Message**
`POST` to `/subscriptions/{stream}/{subscriptions}/nack/{messageid}?action={action}`
 - Action defaults to NakAction.Unknown (Which currently results in a retry)
 - Possible values for action are (park, retry, unknown, skip, stop)

**Nacking a set of Messages**
`POST` to `/subscriptions/{stream}/{subscriptions}/nack?ids={messageid}&action={action}`
 - Action defaults to NakAction.Unknown (Which currently results in a retry)
 - Possible values for action are (park, retry, unknown, skip, stop)

**Replaying Parked Messages**
`POST` to `/subscriptions/{stream}/{subscriptions}/replayParked`

 **Getting information about all subscriptions**
`GET` to `/subscriptions`
Returns a list of subscriptions and contains an array of the request below (does not include the configuration)

 **Getting information about all the subscriptions for a stream**
`GET` to `/subscriptions/stream`
Returns a list of subscriptions and contains an array of the request below (does not include the configuration)

**Getting information about a specific subscription**
`GET` to `/subscriptions/{stream}/{subscriptions}/info`
This information includes (might include more in the future)

 - The configuration for the subscription
 - Event Stream Id
 - Group Name
 - Status (NotReady, Behind, OutstandingPageRequest, ReplayingParkedMessages, Live)
 - Average Items Per Second
 - Total Items Processed
 - Last Known EventNumber
 - Last Processed EventNumber
 - Total In Flight Messages

**NB: This requires some work in the UI to switch to using `/info` for information regarding the subscriptions**


gregoryyoung:
Is it worth putting in the change where by default a get to a stream returns a description document?

pgermishuys:
yeah @gregoryyoung, especially now with http competing it would be nice to be able to get the description document and say... I want the SSE link or the Atom feed.

pgermishuys:
In addition to adding HTTP Competing Consumers, a description document was also added.

When reading a stream through the HTTP API
e.g. 
`curl -i http://localhost:2113/streams/accounts` with no `Accept` header will result in the description document being presented to the consumer.

The consumer is able to query for the description document via the following media types. `application/vnd.eventstore.streamdesc+json` or `application/vnd.eventstore.competingatom+xml` and the description document will be in the following format.

```
{
  "title": "Description Document",
  "supportedMediaTypes": [
    {
      "uri": "http://localhost:2113/streams/accounts",
      "description": "alternate",
      "type": "application/xml"
    },
    {
      "uri": "http://localhost:2113/streams/accounts",
      "description": "alternate",
      "type": "application/atom+xml"
    },
    {
      "uri": "http://localhost:2113/streams/accounts",
      "description": "alternate",
      "type": "application/json"
    },
    {
      "uri": "http://localhost:2113/streams/accounts",
      "description": "alternate",
      "type": "application/vnd.eventstore.atom+json"
    },
    {
      "uri": "http://localhost:2113/streams/accounts",
      "description": "alternate",
      "type": "text/xml"
    },
    {
      "uri": "http://localhost:2113/streams/accounts",
      "description": "alternate",
      "type": "text/html"
    },
    {
      "uri": "http://localhost:2113/streams/accounts",
      "description": "alternate",
      "type": "application/vnd.eventstore.streamdesc+xml"
    },
    {
      "uri": "http://localhost:2113/streams/accounts",
      "description": "alternate",
      "type": "application/vnd.eventstore.streamdesc+json"
    },
    {
      "uri": "http://localhost:2113/subscriptions/accounts/samplegroup",
      "description": "alternate",
      "type": "application/vnd.eventstore.competingatom+xml"
    },
    {
      "uri": "http://localhost:2113/subscriptions/accounts/samplegroup",
      "description": "alternate",
      "type": "application/vnd.eventstore.competingatom+json"
    }
  ]
}
```
The description document can also be used to also indicate to the consumer that there are other forms of reading a stream as pointed out in this example where we include the `persistent subscriptions` for the stream. 

This can be useful, especially when we want to support for instance SSE and then adding that to the description document.

Is it better to just represent this document as plain text? I don't know if the json/xml representations are very useful as this is `actually` an error and needs to be looked at by a person.

gregoryyoung:
:+1:

PR #761, opened 2015-12-01
Make tests use UTC Now

pgermishuys:
Based on this commit (https://github.com/EventStore/EventStore/commit/8f7add6dc685b809ecfe47d0cc75497b6a8537b3), we need to be consistent with the dates being used


jen20:
:+1: seems legit.

PR #765, opened 2015-12-01
remove check for first/last on AlreadyCommitted for case where

### LINKED ISSUE #758: Idempotent writes causes ES to exit
I'm having difficulties with some idempotent writes to stream so I'm currently investigating why some of those requests (I'm using HTTP API) causes ES to die.

I was able to produce simple php script which kills eventstore (tried on 3.0.5 and 3.1.0). So basically what it does is that it shuffles order of the existing events (same guids) and POSTing them to /stream/testarticles (async requests). This should probably return 400 response instead of exiting?

Let me know if you need any more information.

https://gist.github.com/alesblaznik/2903ec0027696e40f5e3

```
[PID:07284:008 2015.11.23 22:49:08.848 ERROR StorageWriterService] Exception in writer.
System.ArgumentOutOfRangeException: LastEventNumber is less than FirstEventNumber
Parameter name: lastEventNumber
at EventStore.Core.Messages.StorageMessage+AlreadyCommitted..ctor (Guid correlationId, System.String eventStreamId, Int32 firstEventNumber, Int32 lastEventNumber) [0x00000] in <filename unknown>:0
at EventStore.Core.Services.Storage.StorageWriterService.ActOnCommitCheckFailure (IEnvelope envelope, Guid correlationId, CommitCheckResult result) [0x00000] in <filename unknown>:0
at EventStore.Core.Services.Storage.StorageWriterService.EventStore.Core.Bus.IHandle<EventStore.Core.Messages.StorageMessage.WritePrepares>.Handle (EventStore.Core.Messages.WritePrepares msg) [0x00000] in <filename unknown>:0
[PID:07284:008 2015.11.23 22:49:08.894 FATAL StorageWriterService] Unexpected error in StorageWriterService. Terminating the process...
System.ArgumentOutOfRangeException: LastEventNumber is less than FirstEventNumber
Parameter name: lastEventNumber
at EventStore.Core.Messages.StorageMessage+AlreadyCommitted..ctor (Guid correlationId, System.String eventStreamId, Int32 firstEventNumber, Int32 lastEventNumber) [0x00000] in <filename unknown>:0
at EventStore.Core.Services.Storage.StorageWriterService.ActOnCommitCheckFailure (IEnvelope envelope, Guid correlationId, CommitCheckResult result) [0x00000] in <filename unknown>:0
at EventStore.Core.Services.Storage.StorageWriterService.EventStore.Core.Bus.IHandle<EventStore.Core.Messages.StorageMessage.WritePrepares>.Handle (EventStore.Core.Messages.WritePrepares msg) [0x00000] in <filename unknown>:0
[PID:07284:008 2015.11.23 22:49:08.910 ERROR Application         ] Exiting with exit code: 1.
Exit reason: Unexpected error in StorageWriterService: LastEventNumber is less than FirstEventNumber
Parameter name: lastEventNumber
```

https://gist.github.com/alesblaznik/2903ec0027696e40f5e3

gregoryyoung:
many events with same ids are used

Should fix #758 worth adding a test (can't run tests here now)

### DIFF
--- a/src/EventStore.Core.Tests/ClientAPI/append_to_stream.cs
+++ b/src/EventStore.Core.Tests/ClientAPI/append_to_stream.cs
@@ -95,6 +95,67 @@ namespace EventStore.Core.Tests.ClientAPI
 
         [Test]
         [Category("Network")]
+        public void multiple_idempotent_writes()
+        {
+            const string stream = "multiple_idempotent_writes";
+            using (var store = BuildConnection(_node))
+            {
+                store.ConnectAsync().Wait();
+
+                var events = new[] { TestEvent.NewTestEvent(), TestEvent.NewTestEvent(), TestEvent.NewTestEvent(), TestEvent.NewTestEvent() };
+                Assert.AreEqual(3, store.AppendToStreamAsync(stream, ExpectedVersion.Any, events).Result.NextExpectedVersion);
+                Assert.AreEqual(3, store.AppendToStreamAsync(stream, ExpectedVersion.Any, events).Result.NextExpectedVersion);
+            }
+        }
+
+        [Test]
+        [Category("Network")]
+        public void multiple_idempotent_writes_with_same_id_bug_case()
+        {
+            const string stream = "multiple_idempotent_writes_with_same_id_bug_case";
+            using (var store = BuildConnection(_node))
+            {
+                store.ConnectAsync().Wait();
+                var x = TestEvent.NewTestEvent();
+                var events = new[] { x,x,x,x,x,x};
+                Assert.AreEqual(5,store.AppendToStreamAsync(stream, ExpectedVersion.Any, events).Result.NextExpectedVersion);
+            }
+        }
+
+        [Test]
+        [Category("Network")]
+        public void in_wtf_multiple_case_of_multiple_writes_expected_version_any_per_all_same_id()
+        {
+            const string stream = "in_wtf_multiple_case_of_multiple_writes_expected_version_any_per_all_same_id";
+            using (var store = BuildConnection(_node))
+            {
+                store.ConnectAsync().Wait();
+                var x = TestEvent.NewTestEvent();
+                var events = new[] { x, x, x, x, x, x };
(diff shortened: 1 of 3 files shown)


gregoryyoung:
Tests added though they deserve some discussion.


        [Test]
        [Category("Network")]
        public void in_wtf_multiple_case_of_multiple_writes_expected_version_any_per_all_same_id()
        {
            const string stream = "multiple_idempotent_writes_with_same_id_bug_case";
            using (var store = BuildConnection(_node))
            {
                store.ConnectAsync().Wait();
                var x = TestEvent.NewTestEvent();
                var events = new[] { x, x, x, x, x, x };
                Assert.AreEqual(5, store.AppendToStreamAsync(stream, ExpectedVersion.Any, events).Result.NextExpectedVersion);
                var f = store.AppendToStreamAsync(stream, ExpectedVersion.Any, events).Result;
                Assert.AreEqual(0, f.NextExpectedVersion);
            }
        }

        [Test]
        [Category("Network")]
        public void in_slightly_reasonable_multiple_case_of_multiple_writes_with_expected_version_per_all_same_id()
        {
            const string stream = "multiple_idempotent_writes_with_same_id_bug_case";
            using (var store = BuildConnection(_node))
            {
                store.ConnectAsync().Wait();
                var x = TestEvent.NewTestEvent();
                var events = new[] { x, x, x, x, x, x };
                Assert.AreEqual(5, store.AppendToStreamAsync(stream, ExpectedVersion.EmptyStream, events).Result.NextExpectedVersion);
                var f = store.AppendToStreamAsync(stream, ExpectedVersion.EmptyStream, events).Result;
                Assert.AreEqual(5, f.NextExpectedVersion);
            }
        }

Explain it all. When using ExpectedVersion.Any we can currently only hold a single place holder for which event it was. When all events have the same id we locate the first one. We operate properly when using an expected version.

For future additions we may want to consider checking for duplicate ids in the same write and return a bad request. For now this behaviour is documented and frankly don't do this. It will however no longer crash anything.

Cheers,

Greg

PR #767, opened 2015-12-04
Histogram Race Condition

thefringeninja:
you can run into this if you have multiple embedded servers in the same process (e.g. running lots of tests in parallel via TPL).


thefringeninja:
BTW you can work around this your own code by calling `HistogramService.StartJitterMonitor()` from a static constructor

PR #773, opened 2015-12-06
Failed resolved link issues part 2

lscpike:
I came across a new scenario which causes null reference exceptions from bad assumptions:

- A failed resolved link is found (in our case truncated old metadata)
- The client can cause the event to be parked.
- The message parker tries to to access the null ResolvedEvent.Event value.

Once I fixed this I found an issue in the projection system:
- The parked event gets picked up by the projection system.
- It correctly identifies it as failed resolved but assumes that the metadata contains a TF position.
- This results in it still accessing the null ResolvedEvent.Event.

My fix for this is to return the bad link TFPos. This seems like it should be safe but the projection system is large and I'm not very familiar.

Finally I noticed the issue I fixed in the ClientApi for failed resolved links causing dropped subscriptions also applies to the catch up subscription.


gregoryyoung:
This looks quite reasonable. :+1:

PR #774, opened 2015-12-06
Fixes for aligned chunk verification

lscpike:
Correct hash verification for scavenged chunks.
Update the chunk hash when the chunk version is downgraded by the unaligner tool.

I've provided these in case anybody else needs to unalign a v3 chunk to v2.


gregoryyoung:
:+1: adding. this was on this list but the unaligned branch has been abandoned.

PR #777, opened 2015-12-07
Pinned consumer strategy

lscpike:
This is the same as #776 but with the pinned consumer strategy that initiated the consumer strategy work.

I'm opening the PR as a courtesy for accepting the changes for the consumer strategy plugin. However, I'm not sure this strategy is suitable for general use. As we can use the plugin framework it's entirely up to you if you want to include this consumer strategy in the main release.

Some information on how it works:
- The underlying stream id of the resolved link event is extracted.
- A bucketed hash map of stream id to PersistentSubscriptionClient is kept.
- When a new event arrives it is sent to the client which it is already assigned to (or the new bucket gets assigned.)
- The strategy attempts to keep the number of buckets per client balanced at all time.
- If a client disconnects the buckets are reassigned.

This allows fairly good load balancing but *attempts* to ensure serial ordered processing of a sub-stream. Therefore processing is more efficient as concurrency clashes are reduced.

### DIFF
--- a/src/EventStore.ClientAPI/Common/SystemNames.cs
+++ b/src/EventStore.ClientAPI/Common/SystemNames.cs
@@ -139,6 +139,12 @@
         /// Distribute events to each client in a round robin fashion.
         /// </summary>
         public const string RoundRobin = "RoundRobin";
+
+        /// <summary>
+        /// Distribute events of the same streamId to the same client until it disconnects on a best efforts basis. 
+        /// Designed to be used with indexes such as the category projection.
+        /// </summary>
+        public const string Pinned = "Pinned";
     }
 
 }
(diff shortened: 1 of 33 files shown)


gregoryyoung:
This is fairly useful for many. Can you open up a quick PR in the docs and add the above description?

thinkbeforecoding:
Yes, I use the same kind of strategy in my messaging system. It avoids most concurrency problems at subscribers and offers generally good balancing.

lscpike:
I will do for the docs Greg. Assume it's incoming. Thank you for accepting it. I've closed the alternative PR.

gregoryyoung:
One question @lscpike 

-                    httpListener.Prefixes.Add(string.Format("http://+:{0}/", port));
+                    httpListener.Prefixes.Add(string.Format("http://127.0.0.1:{0}/", port));

What is the reasoning behind this change?

lscpike:
Sorry that was slightly unintentional to include here. It's the change from #474 that was never merged from the old dev branch.

gregoryyoung:
ok no worries :+1:

PR #778, opened 2015-12-08
Encode URIs consistently across APIs

pgermishuys:
- When URIs get sent to the client, there needs to be consistency across
the APIs (Atom, Competing Consumers, etc).

- Last Known Message gets used as the checkpoint in case there are no outstanding messages and affects the statistical information that gets used to display information to the user in the administration UI.
It gets set as part of pushing messages to the clients which only applies to TCP for now. This is now set as part of reading messages which specifically applies to HTTP Competing Consumers.


gregoryyoung:
:+1:

PR #782, opened 2015-12-18
Show scavenge status in the UI

### LINKED ISSUE #714: Scavenge Status
Currently there is no way to see the status of a scavenge process once it has been started.  It would be nice to be able to see if the GetEventStore was already in the process of scavenging.  This would prevent multiple scavenge requests from being issued.

Currently the only way to see the status and percentage of a scavenge request is to run the GetEventStore as a program and watch the output.

hayley-jean:
Fixes #714 by showing the scavenge status in the UI. I updated the scavenger to publish a message when it finishes scavenging a chunk and used that to display the progress in the UI, which ends up looking like this :

scavengestatus

On a related note: it doesn't look like there is any scavenge history being stored. Would it not be useful to create a stream called $scavenges and store the info in that stream?

### DIFF
--- a/src/EventStore.Core.Tests/Services/Storage/ReadIndexTestScenario.cs
+++ b/src/EventStore.Core.Tests/Services/Storage/ReadIndexTestScenario.cs
@@ -1,8 +1,11 @@
 ﻿using System;
 using EventStore.Common.Utils;
+using EventStore.Core.Bus;
 using EventStore.Core.Data;
 using EventStore.Core.DataStructures;
+using EventStore.Core.Helpers;
 using EventStore.Core.Index;
+using EventStore.Core.Messaging;
 using EventStore.Core.Services;
 using EventStore.Core.Services.Storage.ReaderIndex;
 using EventStore.Core.Tests.Fakes;
@@ -26,6 +29,8 @@ namespace EventStore.Core.Tests.Services.Storage
         protected TFChunkWriter Writer;
         protected ICheckpoint WriterCheckpoint;
         protected ICheckpoint ChaserCheckpoint;
+        protected IODispatcher IODispatcher;
+        protected InMemoryBus Bus;
 
         private TFChunkScavenger _scavenger;
         private bool _scavenge;
@@ -45,6 +50,9 @@ namespace EventStore.Core.Tests.Services.Storage
 
             WriterCheckpoint = new InMemoryCheckpoint(0);
             ChaserCheckpoint = new InMemoryCheckpoint(0);
+
+            Bus = new InMemoryBus("bus");
+            IODispatcher = new IODispatcher(Bus, new PublishEnvelope(Bus));
 
             Db = new TFChunkDb(new TFChunkDbConfig(PathName,
                                                    new VersionedPatternFileNamingStrategy(PathName, "chunk-"),
@@ -89,7 +97,7 @@ namespace EventStore.Core.Tests.Services.Storage
             {
                 if (_completeLastChunkOnScavenge)
                     Db.Manager.GetChunk(Db.Manager.ChunksCount - 1).Complete();
-                _scavenger = new TFChunkScavenger(Db, TableIndex, hasher, ReadIndex);
+                _scavenger = new TFChunkScavenger(Db, IODispatcher, TableIndex, hasher, ReadIndex, Guid.NewGuid(), "fakeNodeIp");
                 _scavenger.Scavenge(alwaysKeepScavenged: true, mergeChunks: _mergeChunks);
             }
         }
@@ -109,8 +117,8 @@ namespace EventStore.Core.Tests.Services.Storage
 
         protected abstract void WriteTestScenario();
 
-        protected EventRecord WriteSingleEvent(string eventStreamId, 
-                                               int eventNumber, 
+        protected EventRecord WriteSingleEvent(string eventStreamId,
+                                               int eventNumber,
                                                string data,
                                                DateTime? timestamp = null,
                                                Guid eventId = default(Guid),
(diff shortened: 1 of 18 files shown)


pgermishuys:
Looks good @hayley-jean

dnauck:
+1 .. nice ;)

src/EventStore.Core/TransactionLog/Chunks/TFChunkScavenger.cs
```diff
@@ -112,8 +117,8 @@ public long Scavenge(bool alwaysKeepScavenged, bool mergeChunks)
                             mergedSomething = true;
                         }
                     }
-                    Log.Trace("SCAVENGING: merge pass #{0} completed in {1}. {2} merged.",
-                              passNum, sw.Elapsed, mergedSomething ? "Some chunks" : "Nothing");
+                    Log.Trace("SCAVENGING: {0}", string.Format("merge pass #{0} completed in {1}. {2} merged.",
```
jen20: There should be no need for the `string.Format` here - IIRC the loggers format strings in the same manner.

src/EventStore.Core.Tests/Services/Storage/ReadIndexTestScenario.cs
```diff
@@ -109,8 +111,8 @@ public override void TestFixtureTearDown()
 
         protected abstract void WriteTestScenario();
 
-        protected EventRecord WriteSingleEvent(string eventStreamId, 
-                                               int eventNumber, 
+        protected EventRecord WriteSingleEvent(string eventStreamId,
```
jen20: Whitespace changes = correct, but probably should be on a different PR to avoid the noise here!

src/EventStore.Core/Services/Transport/Http/Controllers/InfoController.cs
```diff
@@ -22,6 +26,9 @@ public class InfoController : IHttpController, IHandle<SystemMessage.StateChange
         private readonly IOptions _options;
         private readonly ProjectionType _projectionType;
         private VNodeState _currentState;
+        private List<ScavengeStatus> _scavengeStatusMessages = new List<ScavengeStatus>();
```
jen20: Interesting design question here: is it useful to maintain a history of these (and if so should they be persisted in a stream?) - or do we actually only care about the status _right now_?

jen20:
Hi @hayley-jean! Thanks for submitting this - it looks good to me. I've left a few comments in line re: whitespace diffs and minor things, but the interesting question I'd like to bring @gregoryyoung and @pgermishuys on is whether it's useful to have the history or whether the only important status is the current one. Any thoughts?

pgermishuys:
I don't know that having a history of scavenges adds any value.

gregoryyoung:
@pieter I think it might be more "I ran an hour ago"

On Sat, Dec 19, 2015 at 7:50 PM, Pieter Germishuys <notifications@github.com
> wrote:

> I don't know that having a history of scavenges adds any value.
>
> —
> <https://github.com/EventStore/EventStore/pull/782#issuecomment-166019298>
> .
>



-- 
Studying for the Turing test

jen20:
@gregoryyoung In that case would we be better persisting them into say `$scavengeHistory` so it is retained across node restarts?

pgermishuys:
@gregoryyoung yeah, that's useful

hayley-jean:
@jen20 Thanks! I've removed that String.Format from the log message.

gregoryyoung:
Overall this looks pretty good (and likely can be released as a first attempt).

A few comments.

1) Start scavenge then go to stream browser then come back.
2) It would be useful to show how much space is saved (also sometimes chunks get "not scavenged"
3) Using a stream for this would probably be better. The way I see it working UI wise would be as follows:

a) bet effort to write (doesn't need to be perfect). I can show a few places where we do this such as in competing consumers using an IO dispatcher
b) Write event "ScavengeStarted" to start (with Guid of scavengeId)
c) Write events for scavenge
d) Write event ScavengeEnded to end (with Guid of scavengeId)

The UI could then just listen to the atom feed and draw/update the appropriate UI. You could also from any UI show the scavenge status of all nodes if done right (two streams) one for "start" and the other for the info on that scavenge .. include IP in the second one.

The side benefit of the scavenge stream is I can listen to it from other things than the UI.

As a side note I a glad to see out of the box it supports chunk compaction (3 chunks->1) as it wasn't mentioned previously.

Grreg

gregoryyoung:
If I wasn't clear on space saved etc:

Scavenge Chunks #0 - #0	00:00:24.0573685
Scavenge Chunks #1 - #1	00:00:23.1734647
Scavenge Chunks #2 - #2	00:00:22.9876880
Scavenge Chunks #3 - #3	00:00:22.8707708
Scavenge Chunks #4 - #4	00:00:20.5157329
Completed. Total Space Saved: 0	00:01:53.7066678

The chunks it came up with a disposition of "don't scavenge" because not enough space would be saved.

hayley-jean:
Cool, do you want to raise an issue for this as a separate enhancement, or would you prefer I do the work now as part of this one?

If you want to do the stream changes as a separate issue, I should still be able to update this to show how much space was saved and whether chunks were scavenged in the UI.

Dacke:
Question is there a .NET API or REST call that I can use to view the status of the scavenge operation?  Is that a planned feature enhancement?

gregoryyoung:
They are just in streams. $scavenge points to the individual
streams/scavenge

On Thu, Jan 28, 2016 at 2:28 PM, Eric Dorothy <notifications@github.com>
wrote:

> Question is there a .NET API or REST call that I can use to view the
> status of the scavenge operation? Is that a planned feature enhancement?
>
> —
> <https://github.com/EventStore/EventStore/pull/782#issuecomment-176208161>
> .
>



-- 
Studying for the Turing test

PR #791, opened 2016-01-18
Race condition when reconnecting to store

volak:
Added _isProcessing flag check before running a new catchup subscription, prevents ProcessLiveQueue and catch up from running at the same time creating a race condition in which 1 or more events can be processed twice.

ProcessReceive is capable of receiving 0 length messages, so don't disconnect the socket simply because of that.
Fixed couple of log messages

### DIFF
--- a/src/EventStore.ClientAPI/EventStoreCatchUpSubscription.cs
+++ b/src/EventStore.ClientAPI/EventStoreCatchUpSubscription.cs
@@ -169,8 +169,8 @@ namespace EventStore.ClientAPI
 
         private void OnReconnect(object sender, ClientConnectionEventArgs clientConnectionEventArgs)
         {
-            if (Verbose) Log.Debug("Catch-up Subscription to {0}: recovering after reconnection.");
-            if (Verbose) Log.Debug("Catch-up Subscription to {0}: unhooking from connection.Connected.");
+            if (Verbose) Log.Debug("Catch-up Subscription to {0}: recovering after reconnection.", IsSubscribedToAll ? "<all>" : StreamId);
+            if (Verbose) Log.Debug("Catch-up Subscription to {0}: unhooking from connection.Connected.", IsSubscribedToAll ? "<all>" : StreamId);
             _connection.Connected -= OnReconnect;
             RunSubscription();
         }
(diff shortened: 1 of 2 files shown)


gregoryyoung:
https://msdn.microsoft.com/en-us/library/w7wtt64b(v=vs.110).aspx

0 means that the socket has been shutdown (EOF)

volak:
Hmm well when I was diagnosing why the connection kept getting disconnected I found a number of disconnects happened just because bytesTransfered was 0 but socketError was SUCCESS.  I've been running the store client like this for a couple weeks no problem but that particular change doesn't matter so much.

gregoryyoung:
That means it lost connection (end of file) its especially likely if you
are running in a debugger.

On Sat, Jan 23, 2016 at 11:00 PM, Charles Solar <notifications@github.com>
wrote:

> Hmm well when I was diagnosing why the connection kept getting
> disconnected I found a number of disconnects happened just because
> bytesTransfered was 0 but socketError was SUCCESS. I've been running the
> store client like this for a couple weeks no problem but that particular
> change doesn't matter so much.
>
> —
> <https://github.com/EventStore/EventStore/pull/791#issuecomment-174225697>
> .
>



-- 
Studying for the Turing test

volak:
Ah - probably the debugger then - I just revered that change

gregoryyoung:
ok cool will review in morning.

the reason why you get it with debugger (easily reproducible) is the
debugger stops all threads and the server fails a heartbeat timeout. when
this happens it disconnects the socket (resulting in the 0 byte read).

Greg

On Sat, Jan 23, 2016 at 11:10 PM, Charles Solar <notifications@github.com>
wrote:

> Ah - probably the debugger then - I just revered that change
>
> —
> <https://github.com/EventStore/EventStore/pull/791#issuecomment-174226175>
> .
>



-- 
Studying for the Turing test

gregoryyoung:
as a side note you can disable this behaviour via command line on the
server.

On Sat, Jan 23, 2016 at 11:18 PM, Greg Young <gregoryyoung1@gmail.com>
wrote:

> ok cool will review in morning.
>
> the reason why you get it with debugger (easily reproducible) is the
> debugger stops all threads and the server fails a heartbeat timeout. when
> this happens it disconnects the socket (resulting in the 0 byte read).
>
> Greg
>
> On Sat, Jan 23, 2016 at 11:10 PM, Charles Solar <notifications@github.com>
> wrote:
>
>> Ah - probably the debugger then - I just revered that change
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/EventStore/EventStore/pull/791#issuecomment-174226175>
>> .
>>
>
>
>
> --
> Studying for the Turing test
>



-- 
Studying for the Turing test

PR #794, opened 2016-01-29
Adds logging for http requests and responses

### LINKED ISSUE #484: Option for logging http requests
Basically just adding a command line option that will dump http requests (maybe responses as well?) before processing them.

This is mostly useful in scenarios where you need to debug say a gossip or communication issue when setting up a cluster

hayley-jean:
Fixes #484 

Adds a new command line option, --log-http-requests, which turns on debug logging for http requests and responses.

The logged output looks like this:

loghttpsample

### DIFF
--- a/src/EventStore.ClientAPI.Embedded/EmbeddedVNodeBuilder.cs
+++ b/src/EventStore.ClientAPI.Embedded/EmbeddedVNodeBuilder.cs
@@ -33,6 +33,7 @@ namespace EventStore.ClientAPI.Embedded
         private bool _inMemoryDb;
         private bool _startStandardProjections;
         private bool _disableHTTPCaching;
+        private bool _logHttpRequests;
 
         private IPEndPoint _internalTcp;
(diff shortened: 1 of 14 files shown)


gregoryyoung:
We should probably log them as 1 log entry as opposed to multiple log entries. This will play better with analysis tools (even tools like grep or newrelic etc). @jen20 any opinions on how this would be best formatted?

Also we should probably include what ip address the request is coming from (I believe this is exposed)

hayley-jean:
@gregoryyoung Yup, the ip address for where the request is coming from is available, so I can easily add that to the logs

hayley-jean:
@gregoryyoung Single log like this? 
Or would you want each one to start on a new line so they line up nicely in the console?

loghttpsample2

hayley-jean:
I've updated this so that the http logs are output using a single call to Log.Debug each. 

Using grep to grab everything between "HTTP Request Received" and the next log line, I can get something like this:

image

PR #796, opened 2016-01-30
Fixing Occasional Race Condition With Authentication 

thefringeninja:


gregoryyoung:
This assumes that became master happens first when really you just need both events to happen.

PR #799, opened 2016-02-02
Shows stats about replicas in a cluster

### LINKED ISSUE #732: Cluster status UI
Used to be in manager, should add to nodes.

Would also be useful to give a rough prediction of how far back a node is in terms of catching up.

hayley-jean:
Fixes #732 

Adds a new route, /stats/replication, which returns the connection stats about subscription on the master node. 
The UI polls this route and uses it to calculate how far back a node is, and about how long it will take to catch up with the master. This information is then displayed at the bottom of the Cluster Status page.

The result looks like this: 
replicastatus

One thing to note: these stats are only available from the master node, accessing /stats/replication on a slave will return an empty array.

### DIFF
--- a/src/EventStore.Core.Tests/EventStore.Core.Tests.csproj
+++ b/src/EventStore.Core.Tests/EventStore.Core.Tests.csproj
@@ -459,6 +459,7 @@
     <Compile Include="Services\Transport\Http\http_service_should.cs" />
     <Compile Include="Services\Transport\Http\PortableServer.cs" />
     <Compile Include="Services\Transport\Http\ping_controller_should.cs" />
+    <Compile Include="Services\Transport\Http\when_getting_replica_stats_from_stat_controller.cs" />
     <Compile Include="MightyMooseIgnoreAttribute.cs" />
     <Compile Include="Services\Transport\Http\proxy_headers.cs" />
     <Compile Include="Services\Transport\Http\speed_test.cs" />
--- a/src/EventStore.Core.Tests/Services/Transport/Http/when_getting_replica_stats_from_stat_controller.cs
+++ b/src/EventStore.Core.Tests/Services/Transport/Http/when_getting_replica_stats_from_stat_controller.cs
@@ -1,0 +1,167 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
+using System.Threading;
+using EventStore.Core.Bus;
+using EventStore.Core.Messages;
+using EventStore.Core.Tests.Helpers;
+using EventStore.Transport.Http;
+using EventStore.Transport.Http.Client;
+using EventStore.Transport.Http.Codecs;
+using NUnit.Framework;
+
+namespace EventStore.Core.Tests.Services.Transport.Http
+{
+    [TestFixture]
+    public class when_getting_replica_stats_from_stat_controller : SpecificationWithDirectoryPerTestFixture
+    {
+        private readonly TimeSpan Timeout = TimeSpan.FromSeconds(60);
+        protected MiniClusterNode[] _nodes = new MiniClusterNode[2];
+        protected Endpoints[] _nodeEndpoints = new Endpoints[3];
+        protected HttpAsyncClient _client;
+        
+		private string _replicationUrl = "/stats/replication";
+        private string[] _urls = new string[2];
+        
+        private List<List<ReplicationMessage.ReplicationStats>> _responses;
+        
+        [TestFixtureSetUp]
+        protected void SetUp()
+        {
+            _client = new HttpAsyncClient();
+            
+            _nodeEndpoints[0] = new Endpoints(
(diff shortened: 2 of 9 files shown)


ahjohannessen:
Perhaps it would be better to show columns in this order `State`, `Status`, `TimeUTC`, `Checkpoints`, `Internal TCP`, `External TCP`, `Internal Http`, `External Http` and `Actions`.

Might also be an idea to collapse Internal Http & External Http into one Http column and show internal and external in one row:

```
  Int: IP:Port
  Ext: IP:Port
```
in one column. The same thing with `Tcp`.

ahjohannessen:
Last Updated could perhaps be closer to `Cluster Status` with a smaller font, e.g.

### Cluster Status
Last Updated: 14:40:39 - 2016-01-14

gregoryyoung:
Also the "show website" may not always work (think NAT) is it being used
off the advertise as address in gossip?

On Fri, Feb 5, 2016 at 12:03 PM, Alex Henning Johannessen <
notifications@github.com> wrote:

> Perhaps it would be better to show columns in this order State, Status,
> TimeUTC, Checkpoints, Internal TCP, External TCP, Internal Http, External
> Http and Actions.
>
> Might also be an idea to collapse Internal Http & External Http into one
> Http column and show internal and external in one row:
>
>   Int: IP:Port
>   Ext: IP:Port
>
> in one column. The same thing with Tcp.
>
> —
> <https://github.com/EventStore/EventStore/pull/799#issuecomment-180320854>
> .
>



-- 
Studying for the Turing test

ahjohannessen:
Yeah, `ping` and `gossip` are more interesting than `show website`

hayley-jean:
@ahjohannessen: Thanks! I agree that re-ordering the columns and collapsing the external/internal addresses makes more sense.

The show website button just opens the node's external http url. Do you want to reorder the buttons to ping, gossip and show website?

gregoryyoung:
@hayley external ip may not be routable consider that I have a NAT in front
of the nodes. There is also an "advertise as" address which may be
different.

On Fri, Feb 5, 2016 at 12:36 PM, Hayley Campbell <notifications@github.com>
wrote:

> @ahjohannessen <https://github.com/ahjohannessen>: Thanks! I agree that
> re-ordering the columns and collapsing the external/internal addresses
> makes more sense.
>
> The show website button just opens the node's external http url. Do you
> want to reorder the buttons to ping, gossip and show website?
>
> —
> <https://github.com/EventStore/EventStore/pull/799#issuecomment-180335095>
> .
>



-- 
Studying for the Turing test

ahjohannessen:
Isn't it enough just showing Internal Address: TCP, HTTP if it is only for identifying node? Personally, that would be enough information for me, even just IP address of node would do.

ahjohannessen:
If it only is node IP, then it would make sense to use that as the third or first column.

gregoryyoung:
I am refering to the clickable links.

On Fri, Feb 5, 2016 at 1:37 PM, Alex Henning Johannessen <
notifications@github.com> wrote:

> Isn't it enough just showing Internal Address: TCP, HTTP if it is only for
> identifying node? Personally, that would be enough information for me, even
> just IP address of node would do.
>
> —
> <https://github.com/EventStore/EventStore/pull/799#issuecomment-180360246>
> .
>



-- 
Studying for the Turing test

ahjohannessen:
Ah ok

hayley-jean:
All the info shown in the top table is taken from the gossip messages themselves.
So, as far as I can tell, the ip addresses shown in the top table are the ones set in the gossip advertise settings

gregoryyoung:
@hayley I am referring to the links ...

On Fri, Feb 5, 2016 at 3:02 PM, Hayley Campbell <notifications@github.com>
wrote:

> All the info shown in the top table is taken from the gossip messages
> themselves.
> So, as far as I can tell, the ip addresses shown in the top table are the
> ones set in the gossip advertise settings
>
> —
> <https://github.com/EventStore/EventStore/pull/799#issuecomment-180391126>
> .
>



-- 
Studying for the Turing test

hayley-jean:
@gregoryyoung The links for 'ping', 'show website' and 'show gossip' just link to the ip addresses shown in the table, so they would also be using the 'advertise as' addresses

Are those the links you're talking about?

gregoryyoung:
Yes. Should be ok then

On Mon, Feb 8, 2016 at 5:50 AM, Hayley Campbell <notifications@github.com>
wrote:

> @gregoryyoung <https://github.com/gregoryyoung> The links for 'ping',
> 'show website' and 'show gossip' just link to the ip addresses shown in the
> table, so they would also be using the 'advertise as' addresses
>
> Are those the links you're talking about?
>
> —
> <https://github.com/EventStore/EventStore/pull/799#issuecomment-181205969>
> .
>



-- 
Studying for the Turing test

hayley-jean:
After changing the order of the columns, the view now looks like this:

clusterinfo

ahjohannessen:
@hayley-jean much neater :+1: :)

PR #808, opened 2016-02-21
Use external heartbeat parameters in external tcp service

pgermishuys:
For the External TCP Service, the incorrectly used Internal TCP Heartbeat variables have been swopped out for the External TCP Heartbeat variables


gregoryyoung:
:+1:

PR #813, opened 2016-02-24
Fix projections stuck in preparing

pgermishuys:
Replace DebugLogger with Common Event Store Logging

When a node comes back up as Master, it re-subscribes and since the
cancellation token is never invoked, duplicate messages gets received
which puts projections into bad states

On the starting of services, reset the writers.

The writers seem to get themselves stuck and in a bad state when the nodes
shift statuses. Reset them as part of the node status shift.

Dont attempt to kill projections, but rather just clean them up in memory
such as removing them from the list. TODO: Check for proper cleanup

Handle reading from soft deleted streams
When soft deleting streams, the reader will attempt to read from 0 and
expect the first event number to be zero, but this is not the case in soft
deleted streams.


gregoryyoung:
:+1:

PR #817, opened 2016-02-26
Fix 404 errors in the UI when polling the $scavenges stream

hayley-jean:
Ensure the $scavenges stream always exists by writing a ScavengeIndexInitializedEvent to the it when the node starts up.

This is to fix an issue in the UI where it logs a constant stream of 404 errors when polling for the scavenge status and no scavenges have been run.


gregoryyoung:
@hayley-jean as I understand this is to remove the possibility of a 404 from the UI. As of now there will be an event written on every restart. If you are only trying to get rid of the 404 then you can set

https://github.com/EventStore/EventStore/blob/fix-scavenge-404s/src/EventStore.Core/Services/Storage/StorageScavenger.cs#L117

to -1 (ExpectedVersion.EmptyStream) and it will only succeed the first time it starts up. An alternative could be to set the stream metadata ({maxCount: 10000} as example) which would also keep the stream from 404ing without touching the data in the stream.

dnauck:
I like the stream meta data option without a new data event

PR #819, opened 2016-02-26
Introduce a System Ready message

pgermishuys:
Fixes https://github.com/EventStore/EventStore/issues/784
Keep track of the number of standard users that needs to be created and only send a
single user management service initlized message.

When skipping the standard users check, send the user management service
initialised message

When the user management service has been initialized, start the sub
systems and then send a system ready message.

The system ready message can be used by subscribers to know when a node
is ready to start handling requests

### DIFF
--- a/src/EventStore.Core.Tests/Helpers/MiniNode.cs
+++ b/src/EventStore.Core.Tests/Helpers/MiniNode.cs
@@ -155,13 +155,7 @@ namespace EventStore.Core.Tests.Helpers
         {
             StartingTime.Start();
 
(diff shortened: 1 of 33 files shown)


gregoryyoung:
:+1: This should simplify some situations for embedded client

gregoryyoung:
Upon review, do we want to add a helper event on clustervnode?

pgermishuys:
Good idea, should solve https://github.com/EventStore/EventStore/issues/784

gregoryyoung:
:+1:

PR #825, opened 2016-03-01
Change prelude js files to embedded resources

### LINKED ISSUE #752: Prelude javascript files should be emdedded resources
Currently Prelude javascript files are not embedded resources which means automated tests only work when shadow copying is turned off.
More info here: https://github.com/EventStore/EventStore/issues/751

hayley-jean:
Fixes #752 by changing the prelude js files from content files to embedded resources.
Also updates the V8 projections handler to get the prelude files from the embedded resources instead of the file system.


gregoryyoung:
This also precludes people from being able to edit prelude.js which can be useful in some scenarios. @hayley-jean we might want to add a flag for "use files" and still ship the files

PR #829, opened 2016-03-03
Use ClientResolvedEvent if OriginalPosition not null

### LINKED ISSUE #823: ResolvedEvent has null OriginalPosition in embedded catchup-subscription to All
Causes the event to be skipped and not forwarded to client. Have a pr, but not sure if it breaks anything else.

PeterHageus:
fixes #823, event not appearing in catchupsubscriptons on embedded
connection


gregoryyoung:
:+1:

PR #830, opened 2016-03-03
Update build status badges

pgermishuys:


gregoryyoung:
"Some checks haven’t completed yet"

I think they are probably ok.

PR #837, opened 2016-03-08
Make EventStoreCatchUpSubscription Async

samhjohnson:
EventStoreCatchUpSubscriptions blocked when calling ReadAllEventsForwardAsync on subscribing with catch up subscriptions. This allows for asynchronous initialization of catch up subscriptions preventing blocking calls and excessive thread usage.

In addition to this it also fixes a bug when disconnects occur where events can be delivered out of order on resubscribe.

### DIFF
--- a/src/EventStore.ClientAPI/EventStoreCatchUpSubscription.cs
+++ b/src/EventStore.ClientAPI/EventStoreCatchUpSubscription.cs
@@ -4,6 +4,7 @@ using EventStore.ClientAPI.Common.Utils;
 using EventStore.ClientAPI.Exceptions;
 using EventStore.ClientAPI.SystemData;
 using System.Collections.Concurrent;
+using System.Threading.Tasks;
 
 namespace EventStore.ClientAPI
 {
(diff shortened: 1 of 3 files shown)


samhjohnson:
Looks like the Travis build is not working correctly.

gregoryyoung:
I believe its timing out on tests. @hayleyjean is looking at it.

It will take us a bit to review this though it looks reasonable on first
quick look through.

Cheers,

Greg

On Tue, Mar 8, 2016 at 12:55 PM, samhjohnson <notifications@github.com>
wrote:

> Looks like the Travis build is not working correctly.
>
> —
> <https://github.com/EventStore/EventStore/pull/837#issuecomment-193726202>
> .
>



-- 
Studying for the Turing test

samhjohnson:
Ok my tests failing now will fix up.

pgermishuys:
This looks good to me. @samhjohnson mind squashing those commits?

samhjohnson:
BTW have squashed commits.

gregoryyoung:
:+1: looks reasonable and we will schedule some stress testing on it before release

gregoryyoung:
So we will be doing some stress testing on this. I was going through the code today and wondered what will happen if eventappeared (_,x) => Thread.Sleep(5000)

In the old version of the code, this was handled here: https://github.com/EventStore/EventStore/blob/oss-v3.5.0/src/EventStore.ClientAPI/EventStoreCatchUpSubscription.cs#L272 all processing by client was specifically dispatched to another thread so a bad client couldn't screw up internal processing of the client.

PR #844, opened 2016-03-10
Introduce a subset of codecs

pgermishuys:
When reading a single event (event/metadata), only a subset of the usual codecs are needed.
For instance, the atom codecs nor the stream description codecs are relevant.

Introduce an `Empty.Result` and return an `Empty.Xml` or `Empty.Json` result when this result is encountered in the codecs.

The change only impacts reading a single event.

Remove the tests that are no longer relevant and include a new set of tests for requests without accept headers.
Remove empty locks directory.


gregoryyoung:
Looks reasonable.

PR #848, opened 2016-03-11
Trigger cluster replication immediately after flush to speed up writes.

lscpike:
I think this may only be an issue on Windows. I don't have a Linux box available to test on. The Thread.Sleep(1) causes an average delay of (1/64)/2 seconds delay (8ms) in replication of a cluster. This results in irregular commits to a cluster to take in the region of 10 ms each.

I've removed the sleep and triggered replication from the checkpoint flush event. This reduces writes much closer to the raw flush costs at ~3ms.

I've replaced the existing WaitForFlush with an event as the Monitor.Pulse method has a race between the check for more data and the call to Wait. 

I also added a fake type so the queue monitoring on the dashboard shows the replication activity.

### DIFF
--- a/src/EventStore.Core/Services/Replication/MasterReplicationService.cs
+++ b/src/EventStore.Core/Services/Replication/MasterReplicationService.cs
@@ -63,6 +63,7 @@ namespace EventStore.Core.Services.Replication
         private volatile bool _newSubscriptions;
         private TimeSpan _noQuorumTimestamp = TimeSpan.Zero;
         private bool _noQuorumNotified;
+        private ManualResetEventSlim _flushSignal = new ManualResetEventSlim();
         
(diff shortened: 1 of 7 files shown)


gregoryyoung:
I was just looking through chaser for related reasons today

lscpike:
Bit trigger happy there GFY! Second PR to fix the build.

gregoryyoung:
Has already done locally :)

On Fri, Mar 11, 2016 at 2:18 PM, lscpike <notifications@github.com> wrote:

> Bit trigger happy there GFY! Second PR to fix the build.
>
> —
> <https://github.com/EventStore/EventStore/pull/848#issuecomment-195383453>
> .
>



-- 
Studying for the Turing test

PR #851, opened 2016-03-16
Fix flapping truncation tests

hayley-jean:
Increase the number of readers available in the reader pool in the ReadIndexTestScenario.
This fixes the ObjectPoolMaxLimitReachedException that was occasionally being thrown by TableIndex, causing it to timeout.


gregoryyoung:
:+1: good work!

Issue #480, opened 2015-05-19
Possible to set an invalid cluster config


grumpydev:
Ok, this one is very much down to my own stupidity, but after several hours of debugging it turns out that it's possible to specify multiple gossip seeds on the commandline, but set the cluster size to 1. Event Store will fire up without any warnings or errors, but then will never be able to sync up because each node creates its own db, then tries to replicate it with the other nodes.

Despite this being grade a stupidity on my part, might be useful if Event Store detected this invalid config and blew up.

![dumb](https://cloud.githubusercontent.com/assets/527013/7707747/4408966c-fe49-11e4-927e-f8ea4910b301.png)

jen20:
Hmm, not sure on this one. It is actually a valid use case to have a cluster size of 1 but gossip (with your manager). It might be worth a warning in the log though as it's likely quite uncommon.

grumpydev:
Fair enough, that's not a config I'm aware of - in our case it just cause the cluster to never come up properly - you may remember many moons ago I had the same issue on a 3 node cluster on azure, it was the same issue.

Do you specify managers using the same gossip seed commandline syntax? (we don't use a manager so I have no idea)

jen20:
Yes, generally a manager would be a gossip seed on the same box, and the managers would be configured to discover one another using DNS (or gossip seeds).

Fun story: gossip seeds were never supposed to exist and used to be called "Fake DNS" - the only reason they're there is so you can run three nodes on the same box for testing purposes. It used to be called "manager IP/port" and you only specified one, with all the cluster discovery happening in the manager, but because the manager isn't open source you needed a way of specifying more than one of them, hence gossip seeds.

I agree though a warning here is warranted - it would be very uncommon to set more gossip seeds than cluster size (it *is* possible if you want an additional external manager, as we did for example in the HA showcase).

pgermishuys:
fixed by #872

PR #872, opened 2016-03-31
#480 Fix: Check if the ClusterSize is compatible with the GossipSeed …

riccardone:
…value and if not throw an ApplicationInitializationException


gregoryyoung:
Hmm.

Why are you putting in a gossip seed at all if cluster size is one?

riccardone:
I guess that this is originate from a mistake that one user made time ago. There is an issue https://github.com/EventStore/EventStore/issues/480 where he described this as a difficult config mistake to discover

gregoryyoung:
hmm seems legit

PR #877, opened 2016-04-02
Remove not used WatchdogState option description and default value

pgermishuys:
Check if the property isn't null inside `EnsureCorrectType` (https://github.com/EventStore/EventStore/blob/f3ba760187ab296a3e3035d71066b786c768c123/src/EventStore.Common/Options/EventStoreOptions.cs#L145)


hayley-jean:
Looks good

PR #880, opened 2016-04-03
Increase the utility of PairingHeap

machonky:
Remove the need for #defines as well while increasing utility of the class with various node creation strategies.

### DIFF
--- a/src/EventStore.Core/DataStructures/PairingHeap.cs
+++ b/src/EventStore.Core/DataStructures/PairingHeap.cs
@@ -1,6 +1,4 @@
-﻿#define USE_POOL
-
-using System;
+﻿using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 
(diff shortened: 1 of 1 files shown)


gregoryyoung:
Have you benchmarked?

machonky:
Apologies - no.

gregoryyoung:
There is most certainly a cost to the abstraction (and this is a reasonably hot area of code). This is mostly used in the timing service. We have since drastically reduced the amount of work it was doing so this is likely ok. 

It might also be worth looking at a hashed wheel for our use case.

On a side note for these kinds of implementations might it be useful to extract a common assembly that contains just such thing?

machonky:
Agree about cost of abstraction. If it's a performance bottleneck, then a specific Object pool implementation is justified, otherwise code-re-use/maintainability approach wins (less code = better).
 
For me - new assemblies motivated only by proven re-use/decoupling need, otherwise build performance can suffer.

Perhaps greater value to project might be gained from some kind of performance suite that declares benchmarks for critical performance areas. Contributors can then test before & after code changes to determine a win. That said, benchmarks need Prod experience/judgement to be useful.

gregoryyoung:
We will still consider this as a PR, it just needs to be benchmarked so we know the overall cost of the introduction

machonky:
Sure. Look forward to outcome. Cheers.

PR #882, opened 2016-04-05
Add option aliases

pgermishuys:
This adds the ability to specify aliases for options.
As an example with this PR the option for `config` could also have an alias such as `configuration`.


hayley-jean:
Looks good

PR #892, opened 2016-04-14
Filter linked events when the includeLinks option is set

### LINKED ISSUE #890: fromAll linkTos
FromAll should support saying whether or not it wants the events resolved if it currently does not.

pgermishuys:
Fixes #890


hayley-jean:
Looks good

DaveKallidus:
Is this option going to be exposed in the CreateProjection part of the admin interface.

pgermishuys:
Options such as these are already available to set and is part of the projection's source.
e.g.
```
options({
    $includeLinks: false
})
fromAll().whenAny(function (s, e) {
    linkTo("no_links", e);
});
```
This particular PR just adds the necessary functionality so that the option actually does what it is intended to.

PR #894, opened 2016-04-15
Return a task when starting a persistent subscription

### LINKED ISSUE #881: Method ConnectToPersistentSubscription causes deadlock
Hi,

I have started to use EventStore recently in my project. Especially Persistent subscription. I encounter deadlock when I try to subscribe. I can 100% reproduce with following use-case:

- Connect to persistent subscription for the first time while it's not yet exists.
- Catch ArgumentException with message "Subscription not found" and create requested subscription.
- Connect to newly created pesistent subscription.

Deadlock occures in ConnectToPersistentSubscription method. Sometimes in happens during first try to connect. In other cases it deadlocks on second try.

I have investigated an implementation and noticed that StartSubscription method of EventStorePersistentSubscription class waits on a task. Which actually causes a deadlock.

I have worked it around by using a reflection for now to start that StartSubscription task and await it properly.

Is there any reason why that task is waited instead of making ConnectToPersistentSubscription method fully async? If not, would it be possible to have an async method ConnectToPersistentSubscriptionAsync? I believe I can create a pull-request with this fix implemented.

Cheers,
Andrii

hayley-jean:
Fixes #881 

The start method on persistent subscriptions now returns a task rather than calling task.Wait(), which used to cause deadlocks.

There is a new async method on EventStoreConnection for connecting to persistent connections that returns the subscription task, and the existing methods now wait on the subscription task before returning the subscription.

### DIFF
--- a/src/EventStore.ClientAPI/IEventStoreConnection.cs
+++ b/src/EventStore.ClientAPI/IEventStoreConnection.cs
@@ -312,7 +312,7 @@ namespace EventStore.ClientAPI
                 UserCredentials userCredentials = null);
 
         /// <summary>
-        /// Subscribes a persistent subscription (competing consumer) to the event store
+        /// Subscribes to a persistent subscription(competing consumer) on event store
         /// </summary>
         /// <param name="groupName">The subscription group to connect to</param>
         /// <param name="stream">The stream to subscribe to</param>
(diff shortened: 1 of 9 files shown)


src/EventStore.ClientAPI/IEventStoreConnection.cs
```diff
@@ -338,6 +338,33 @@ public interface IEventStoreConnection : IDisposable
...
+        /// <param name="bufferSize">The buffer size to use for the persistent subscription</param>
+        /// <param name="autoAck">Whether the subscription should automatically acknowledge messages processed.
+        /// If not set the receiver is required to explicitly acknowledge messages through the subscription.</param>
+        /// <remarks>This will connect you to a persistent subscription group for a stream. The subscription group
+        /// must first be created with CreatePersistentSubscriptionAsync many connections
+        /// can connect to the same group and they will be treated as competing consumers within the group.
+        /// If one connection dies work will be balanced across the rest of the consumers in the group. If
+        /// you attempt to connect to a group that does not exist you will be given an exception.
+        /// </remarks>
+        /// <returns>An <see cref="EventStoreSubscription"/> representing the subscription</returns>
```
pgermishuys: Reference should be `EventStorePersistentSubscription`

src/EventStore.ClientAPI/IEventStoreConnection.cs
```diff
@@ -338,6 +338,33 @@ public interface IEventStoreConnection : IDisposable
             int bufferSize = 10,
             bool autoAck = true);
 
+        /// <summary>
+        /// Asynchronously subscribes a persistent subscription (competing consumer) to the event store
```
pgermishuys: Would it be better if we instead of `Asynchronously subscribes a persistent subscription (competing consumer) to the event store` say `Asynchronously subscribes to a persistent subscription (competing consumer) to event store`?

pgermishuys:
See comments inline, good work on the tests

src/EventStore.ClientAPI/Internal/EventStoreNodeConnection.cs
```diff
@@ -352,29 +352,48 @@ private void EnqueueOperation(IClientOperation operation)
...
+            Ensure.NotNull(eventAppeared, "eventAppeared");
+
+            var subscription = new EventStorePersistentSubscription(
+                groupName, stream, eventAppeared, subscriptionDropped, userCredentials, _settings.Log,
+                _settings.VerboseLogging, _settings, _handler, bufferSize, autoAck);
+
+            return subscription.Start();
         }
-*/
+        /*
```
pgermishuys: Why is this commented out?

hayley-jean: That chunk of code was already commented out here https://github.com/EventStore/EventStore/blob/release-v3.6.0/src/EventStore.ClientAPI/Internal/EventStoreNodeConnection.cs#L359

PR #897, opened 2016-04-21
Use HttpClient instead of WebRequest

### LINKED ISSUE #875: WebRequest.Create
This code can block and hold up http requests being sent:

https://github.com/EventStore/EventStore/blob/release-v3.6.0/src/EventStore.Transport.Http/Client/HttpAsyncClient.cs#L25

hayley-jean:
Fixes #875 

Replace WebRequest with HttpClient in HttpAsyncClient and HttpSendService.


pgermishuys:
Good work @hayley-jean. Will merge this once reviewed properly.

PR #902, opened 2016-04-25
Add ExpectedVersion.StreamExists

### LINKED ISSUE #901: ExpectedVersion.StreamExists
I want to be able to send an event and specify that it should only succeed if there is an existing stream. It should also honour soft deletes.

Write ExpectedVersion.StreamExists (stream does not exist) -> concurrency exception
Write ExpectedVersion.StreamExists (at event 1) -> Accepted
Write ExpectedVersion.StreamExists (at event 7) -> Accepted
Write ExpectedVersion.StreamExists (only stream metadata) -> Accepted
Write ExpectedVersion.StreamExists (stream soft deleted) -> concurrency exception
Write ExpectedVersion.StreamExists (stream hard deleted) -> concurrency exception

hayley-jean:
Fixes #901 

Add ExpectedVersion.StreamExists to the ExpectedVersions enum with a value of -4.
This will only allow events to be appended to a stream if it exists, or if a metadata stream exists for it.
If the stream does not exist, it returns WrongExpectedVersion.
If the stream has been deleted (hard or soft) it returns DeletedStream.
After ensuring that the stream exists, this will behave the same way as ExpectedVersion.Any.

### DIFF
--- a/src/EventStore.ClientAPI/ExpectedVersion.cs
+++ b/src/EventStore.ClientAPI/ExpectedVersion.cs
@@ -6,18 +6,19 @@ namespace EventStore.ClientAPI
     /// <remarks>
     /// The use of expected version can be a bit tricky especially when discussing idempotency assurances given by the event store.
     /// 
-    /// There are four possible values that can be used for the passing of an expected version.
+    /// There are five possible values that can be used for the passing of an expected version.
     /// 
     /// ExpectedVersion.Any (-2) says that you should not conflict with anything.
     /// ExpectedVersion.NoStream (-1) says that the stream should not exist when doing your write.
     /// ExpectedVersion.EmptyStream (0) says the stream should exist but be empty when doing the write.
+    /// ExpectedVersion.StreamExists(-4) says the stream or a metadata stream should exist when doing your write.
     /// 
     /// Any other value states that the last event written to the stream should have a sequence number matching your 
     /// expected value.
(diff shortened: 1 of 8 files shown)


src/EventStore.Core.Tests/Http/Streams/when_posting_events_with_expected_versions.cs
```diff
@@ -0,0 +1,287 @@
...
+        [Test]
+        public void request_fails_with_wrong_event_number()
+        {
+            Assert.AreEqual(HttpStatusCode.BadRequest, _response.StatusCode);
+            Assert.AreEqual(WrongExpectedVersionDesc, _response.StatusDescription);
+        }
+    }
+
+    [TestFixture, Category("LongRunning")]
+    public class when_posting_event_with_expected_version_higher_than_the_number_of_events : ExpectedVersionSpecification
```
pgermishuys: It might be a good idea to keep these tests consistent with regards to naming and expectancies to those of the ClientAPI for example https://github.com/EventStore/EventStore/blob/699c17fc01ab633ceb14fac7a6e8ecd2c5535c18/src/EventStore.Core.Tests/ClientAPI/append_to_stream.cs#L259.

pgermishuys:
See comments about naming, other than that. It looks good @hayley-jean. Good work.

PR #921, opened 2016-05-16
Add application/octet-stream in HTTP API to support just data events

pgermishuys:
This PR adds support for data only events. This will require the user to provide the `ES-EventType` header and the user cannot have metadata associated with the Event.


hayley-jean:
When reading a stream that contains binary events with the body embedded, the events are returned as a utf8 string representation of the binary data.

@gregoryyoung any thoughts about this?

pgermishuys:
@hayley-jean we don't actually know what format those events are in, the best we can do is to just give the user some representation of it.
I expect consumers to use the Event Type here to indicate what type of data is contained in these events and perform a request for the contents of that event with the appropriate accept header.

gregoryyoung:
For embedded streams utf8 could work. This would give a side benefit of the UI somewhat working on these IIRC it uses embedded. The issue is we don't know whether this is XML or something else.

hayley-jean:
Looks good

PR #932, opened 2016-05-30
Add method to VNodeBuilder not to add the interface prefixes

pgermishuys:
There can be some additional work done around making sure that we align more with the regular node by looking [this](https://github.com/EventStore/EventStore/blob/release-v3.7.0/src/EventStore.ClusterNode/Program.cs#L150) can be pulled into the `VNodeBuilder`


gregoryyoung:
Looks good here.

you can merge if you like (dont want to merge if you have something to add before posting)

pgermishuys:
@gregoryyoung the PR has been updated to remove the `Runtime.IsMono` checks.
Previously the behaviour was...
**Windows**
- If the int/ext prefixes are empty, add one for what the int/ext http was set to

**Mono**
- If the int/ext prefixes are empty, add one for what the int/ext http was set to
- Add `http://localhost{intHttpPort}/` and `http://localhost:{intHttpPort}/`

With the updated PR Windows will behave like **Mono**.

PR #937, opened 2016-06-02
Ensure persistent subscription checkpoint version doesn't get corrupted

lscpike:
When an error occurs writing a persistent subscription checkpoint (CommitTimeout for example) it can result in the expected version for the next write to be incorrect. The state is now corrupt permanently until the node is cycled. 

This isn't apparent from the statistics as the writing of the checkpoint is not considered for the state of the persistent subscription. The result in a silent issue. This has lead to us replaying large numbers of events when the master node randomly changes.

I also noticed the checkpoint metadata will never be written as the version is never set to NoStream. I've left that for somebody else to decide how to handle given there will be people will un-truncated checkpoint streams in production.


gregoryyoung:
:+1:

PR #938, opened 2016-06-06
Only initialise projections when the core worker readers are ready

pgermishuys:
There are cases when the core worker readers takes a while to start up and
by that time the manager could have started the projections and messages
could have been written to the core worker's stream and they could have
been missed. 

The reason they can be missed is that when the core reader
starts, it will read the last event in the $projections-${worker_id} stream
and marks that as the starting point to read from.

The issue can result in some if not all projections to be stuck in the `preparing` state.


hayley-jean:
Looks good

PR #955, opened 2016-06-21
Set an exception on the task if one is thrown while processing the events

### LINKED ISSUE #945: DropSubscription doesn't get called on CatchupSubscriptions any longer
Currently in the 3.7.0 branch the client doesn't seem to be calling the DropSubscription (https://github.com/EventStore/EventStore/blob/release-v3.7.0/src/EventStore.ClientAPI/EventStoreCatchUpSubscription.cs#L343) any longer. Probably a regression from 3.6.3.

hayley-jean:
Fixes #945

### DIFF
--- a/src/EventStore.Core.Tests/ClientAPI/subscribe_to_stream_catching_up_should.cs
+++ b/src/EventStore.Core.Tests/ClientAPI/subscribe_to_stream_catching_up_should.cs
@@ -161,6 +161,25 @@ namespace EventStore.Core.Tests.ClientAPI
         }
 
         [Test, Category("LongRunning")]
+        public void call_dropped_callback_when_an_error_occurs_while_processing_an_event()
+        {
+            const string stream = "call_dropped_callback_when_an_error_occurs_while_processing_an_event";
+            using (var store = BuildConnection(_node))
+            {
+                store.ConnectAsync().Wait();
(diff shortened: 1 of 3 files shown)


src/EventStore.Core.Tests/ClientAPI/subscribe_to_stream_catching_up_should.cs
```diff
@@ -161,6 +161,44 @@ public void call_dropped_callback_after_stop_method_call()
...
+        public void call_dropped_callback_when_an_error_occurs_while_processing_an_event_for_stream()
+        {
+            const string stream = "call_dropped_callback_when_an_error_occurs_while_processing_an_event_for_stream";
+            using (var store = BuildConnection(_node))
+            {
+                store.ConnectAsync().Wait();
+                store.AppendToStreamAsync(stream, ExpectedVersion.Any, new EventData(Guid.NewGuid(), "event", false, new byte[3], null)).Wait();
+
+                var dropped = new CountdownEvent(1);
+                var subscription = store.SubscribeToStreamFrom(stream,
```
pgermishuys: @hayley-jean this variable is never used which will fail the build on travis.

hayley-jean: @pgermishuys Yup, fixing it now

PR #956, opened 2016-06-22
Use the advertised IP address in the atom links, if they are set

### LINKED ISSUE #948: Wrong HTTP port in Atom feed links
Given:
I have set ExtHttpPortAdvertiseAs to 2114, as my EventStore is fronted by an NGINX proxy (for authentication purposes) on that port. If I do a GET on a stream where the server happens to be a slave the redirect Location: is correctly constructed with 2114 as the port. 

When:
A GET on the feed on the master server

Expected:
Links in the JSON to prev/next/first/last with 2114 as port

Actual:
Links in there use port 2113 (my ExtHttpPrefixes is http://*:2113/), causing problems for the client when it follows those links.

hayley-jean:
Fixes #948 

The urls provided in the atom feed links will always use the advertised ip address and http ports, if they have been set. If not, the urls will be built from the request url as they were before.

The advertise ip and http port information is now included on GossipAdvertiseInfo so it can be passed into the HttpService, and later used to build the requestedUrl in HttpEntity.


pgermishuys:
@hayley-jean If we do merge this PR, i'd like to suggest that we rename `GossipAdvertiseInfo` to `AdvertiseInfo`.

PR #970, opened 2016-07-05
Track and deleting of emitted streams

### LINKED ISSUE #968: Emitted Stream to be a bit more lenient
In the case where a projection has been deleted and another projection has been created that emits to the same stream the validation `Multiple projections emitting to the same stream detected. Stream: 'StreamName'. Last event projection: '5'. Emitting projection: '6'` stops the new projection from emitting to that stream.

Things to look into can include introducing a `run-with-scissors` mode for projections that would disable these types of validations. Introduce the ability to track emitted streams and delete them when deleting a projection.

pgermishuys:
Fixes #968 
The emitted streams can be tracked if the user indicates that he/she wants
to do so. This is an additional parameter that can be passed when creating a projection.

The user also has the ability to indicate that he/she wants to delete the emitted streams of a projection if it has been tracked. This is an additional parameter that can be passed when deleting a projection.

For the above to work correctly (especially if the user creates another projection that emits to the previously deleted emitted streams), the `EmittedStream` needed to have it's logic adjusted for establishing whether an emitted stream is a new physical stream or not. Previously this was done by checking against the number of events in the read result. It has been adjusted to check whether the `LastEventNumber` of the read result is -1 (NoStream)

The ManagedProjection has also been reviewed and cleaned up as part of
this exercise.

The Projection Event Types have also extracted and moved into it's own class in a class called `ProjectionEventTypes`

### DIFF
--- a/src/EventStore.Core.Tests/Helpers/TestFixtureWithExistingEvents.cs
+++ b/src/EventStore.Core.Tests/Helpers/TestFixtureWithExistingEvents.cs
@@ -442,6 +442,12 @@ namespace EventStore.Core.Tests.Helpers
             }
             if (!_streams.TryGetValue(message.EventStreamId, out list) || list == null)
             {
+                if(message.ExpectedVersion == ExpectedVersion.Any)
+                {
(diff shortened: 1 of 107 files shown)


hayley-jean:
Looks good

PR #975, opened 2016-07-09
[WIP] Better queues

Scooletz:
This PR is still WIP. I'm interested in gathering some early info about the approach.
It addresses two things:

1. `QueuedHandler` performance tests that could have been suffering from the friction over one `Interlocked.Add` field. Now the cost is amortized
1. Providing a better multiple producer single consumer queue than `ConcurrentQueue` with some smart batching. Initial tests show that amortization brings performance benefits for multiple writers (10).

If we only had .NET 4.5 or more, it'd be faster with `Volatile.Read` and `Volatile.Write` ;-)


Scooletz:
I'll issue a separate one with tests only first.

PR #977, opened 2016-07-12
Better QueueSpeedTests that does not Interlock.Decrement for each mea…

Scooletz:
This PR addresses `QueuedHandler` performance tests that could have been suffering from the friction over one `Interlocked.Add` field. Now the cost is amortized by claiming a batch of messages to dispatch.


pgermishuys:
thanks @Scooletz 👍

hayley-jean:
Looks good

gregoryyoung:
We might also want to look at using: https://github.com/Azure/DotNetty/blob/dev/src/DotNetty.Common/Utilities/MpscLinkedQueue.cs#L16

Scooletz:
The PR was the initial improvement on tests. I'm playing with the impl
right now. It hurts a little bit that we don't have Volatile class as its
.net 4.0 ;-)

Will issue a separate PR when I get sth meaningful. It will require some
work on QueueStats as well.
12 lip 2016 18:55 "Greg Young" <notifications@github.com> napisał(a):

PR #979, opened 2016-07-12
Retry deleting the scavenge temp file

### LINKED ISSUE #964: Check file deletes in scavenge
While doing scavenge we currently eat the error deleting the file here: https://github.com/EventStore/EventStore/blob/release-v3.8.0/src/EventStore.Core/TransactionLog/Chunks/TFChunkScavenger.cs#L256

If this fails we should probably note how many have failed and bomb out the scavenge for safety (dont want to duplicate all the chunks). We should also probably note why it failed and not eat the exception.

hayley-jean:
Fixes #964 

If an exception occurs while deleting the scavenge temp file, the exception is logged and the delete is retried 5 times. If the deletion fails every time, the scavenge is stopped.


gregoryyoung:
:+1:

PR #981, opened 2016-07-14
[WIP] Much faster QueuedHandler

Scooletz:
This PR is still **WIP**.

This PR introduces much better multi producer single consumer queue without allocations after construction. It's a much better replacement for `ConcurrentQueue` in the `QueuedHandler`. Possibly a single producer single consumer queue (SPSC) could be used for other parts, but it's hard for me to reason about threading in different parts of the EventStore and any help would be greatly appreciated.

The MPSCMessageQueue provides a better consistent throughput and lower latency as:

1. It does no allocate after construction
1. It uses only one field in the array as `Message` is a reference type and can be written with `volatile` (acquire/release barrier)
1. Enables getting and dispatching messages in batches which amortizes the cost of accessing the concurrent structure on the consumers end.
1. It uses padding & mapping to the array address to limit false sharing (`ConcurrentQueue` doesn;t have this property)
1. It works only with `IntPtr.Size == 8` as it requires an equivalent of `Volatile.Reaf(ref long)`
1. Tests with 2 producers run on my 4 core machine showed great improvement. Tests with 10 looks not that good but as scheduling affects it...

Things to do & questions to be answered:
- [ ] Is the build process only `Mono` based and is run on Linux or do you have Windows build as well? I've changed only `MRES` impl.
- [ ] Clarify if the code can lie to `QueryStats` providing an estimated number of messages (true at the beginning of the batch processing).
- [ ] Clarify different structures that can be used in other places? Would providing SPSC be useful?
- [ ] Provide tests for the queue
- [ ] Possibly remove usages of `ConcurrentQueue`.


pgermishuys:
@Scooletz Would you mind doing the work for the PR against the current development branch which is v3.9.0?
Thank you for the contribution! much appreciated!

Scooletz:
@pgermishuys Ofc I can do it. I wasn't familiar with your branching strategy :)
I'll issue another PR then.

Scooletz:
A new PR will be issued.

PR #987, opened 2016-07-22
Copy metadata values including custom metadata

jageall:
Adds facility to create a new StreamMetadataBuilder from existing metadata.  Useful when wanting to update metadata with custom values


gregoryyoung:
:+1:

PR #999, opened 2016-08-04
Fault the projection on checkpoint write fail

pgermishuys:


kstrauss:
should the number of retries be setable from the command line? Also I don't see where the length of time before a timeout occurs can be set (i.e. I know I have slow storage/large volume and therefore the timeout period needs to managed).

pgermishuys:
The commit and prepare timeouts are settable from via options (http://docs.geteventstore.com/server/3.8.1/command-line-arguments/). See `CommitTimeoutMs`.

kstrauss:
Thanks for the insight. Perhaps I'll try using that. I had assumed that this was related to commitcount and running in a cluster.

PR #1004, opened 2016-08-10
Try reconnect to master the first time the connection is dropped

### LINKED ISSUE #611: Retry first on TCP drop?
In ReplicaService we currently on drop force an election

https://github.com/EventStore/EventStore/blob/release-v3.2.0/src/EventStore.Core/Services/Replication/ReplicaService.cs#L138

We might be better off to try reconnecting first as opposed to a cluster shift (say with a 1 second timeout). This would cause few election shifts that are unneeded though it would also increase time to failure detection (maybe a config point).

hayley-jean:
Fixes #611


pgermishuys:
Looks good to me, will merge once the tests pass.

gregoryyoung:
We should test with a network card turn off on this.

pgermishuys:
Will do @gregoryyoung

PR #1023, opened 2016-08-23
Continue scavenge if an exception is thrown while scavenging a chunk

### LINKED ISSUE #1006: Scavenge: don't end on error
Right now any error in any chunk will stop the scavenge process. We should probably report the error and continue to scavenge the rest of the chunks.

hayley-jean:
Fixes #1006 

If an error is encountered while scavenging a chunk, the chunk will be skipped and the scavenge will continue. The error is added to the `ScavengeChunkCompleted` event, and the temp file is deleted.

If we fail to delete the temp file after a number of tries, the scavenge is still stopped completely.


gregoryyoung:
:+1: when checks are finished.

PR #1041, opened 2016-09-03
Improve performance of ptable merges

pgermishuys:
This reintroduces the previous existsAt function that only checks for
the existence of a record instead of attempting to do an actual read. The read is only required when an upgrading the stream hash.

This should be included in a 3.9.1 release to reduce merge times and to avoid unnecessary reads. With the release of 4.0.0 and the deprecation of 32bit indexes, this PR along with https://github.com/EventStore/EventStore/pull/1021 should be reviewed and all the 32bit branches removed.


gregoryyoung:
:+1:

Issue #1045, opened 2016-09-14
RequestManager


gregoryyoung:
Verify that at least one of the N/2+1 ACKS is from self (eg don't allow N/2+1 slaves to ACK and consider it good).

PR #1048, opened 2016-09-14
Self ack

gregoryyoung:
Should be a fix for #1045. Needs review, could not run unit tests locally but travis seemed ok


gregoryyoung:
A comment on this is that we may want to consider setting _lastKnown to -1 instead of 0 as it is now as it could cause a stream with 1 event in it to not checkpoint (not that worried about it). I was unsure of the other repercussions of that so did with 0.

PR #1049, opened 2016-09-15
Return if lowest is 0

gregoryyoung:
Fixes an issue where if the subscription does a take over and is perfectly caught up, with nothing connected to it. The checkpoint would write to last known message which is 0


pgermishuys:
A comment on this is that we may want to consider setting _lastKnown to -1 instead of 0 as it is now as it could cause a stream with 1 event in it to not checkpoint (not that worried about it). I was unsure of the other repercussions of that so did with 0.

gregoryyoung:
Another alternative is to set lastKnownMessage appropriately but this can still cause a race condition

PR #1060, opened 2016-09-29
Support TLS 1.0/1.1/1.2 and SSL 3.0

hui000444:
SslProtocols.Default is stand for SSL 3.0 and TLS 1.0. Due to the security concerns, the enterprise server is usually disabled TLS1.0. So we want to enable the EventStore compatibility for TLS 1.1/1.2.


dnauck:
I think SSL 3.0 should be dropped (normally also TLS 1.0)

hui000444:
but in source code, it used SslProtocols.Default, means just support SSL3.0 and TLS1.0, it have conflict with you think.
Our production environment used EventStore and all server machine doesn't support TLS1.0, so we have problem with connection to EventStore.

jen20:
:+1: on this if we figure out what the build issues are.

dnauck:
Looks like the build server is running .NET 4.0

`Tls11` and `Tls12` was added in .NET 4.5 afaik.

See: https://msdn.microsoft.com/en-us/library/system.net.securityprotocoltype(v=vs.110).aspx

pgermishuys:
Hi @hui000444,
Apologies for only replying to this PR now, would you mind rebasing your branch off of 4.0.0. Your PR should be building fine on this branch.

hui000444:
Sorry for I merged wrong file, and I will close this PR and resend new same PR.
Thank you guys upgrade to 4.6 so fast, i know it's a large work. it's great.

PR #1068, opened 2016-10-05
Fix issue when merging mixed indexes where hash collisions occur

pgermishuys:
There is a case that is not being handled by the current implementation of the upgrade during a merge between a 32 bit index and a 64bit index.

The issue is encountered when hash collisions in the 32bit index occurs in the same index file.

The proposed solution is to read the entries from the 32bit index file up to where a different hash is encountered and then upgrade those entries to 64bit in a sorted list. The sorted list is then used to merge with the 64bit index.


gregoryyoung:
:+1:

PR #1069, opened 2016-10-06
During Scavenge operation count the number of times that it fails

riccardone:
code sample
int failedCount;
var spaceSaved = Scavenge(alwaysKeepScavenged, out failedCount);
if (failedCount > 0)
{
   Log.Trace("SCAVENGING: merge failed {0} times", failedCount);
}


pgermishuys:
Hi @riccardone thanks for the PR!
It seems like a large part of the PR is code reformatting, which makes it difficult to review and see what has changed. Would it be possible to make the change without reformatting the code?

PR #1075, opened 2016-10-14
Start the persistent subscription from the event after the checkpoint

hayley-jean:
This is to fix a bug where the checkpointed event was being reissued.

This PR also renames `_lastEventPulled` to `_nextEventToPullFrom` for clarity, as it is set to the previous read's `NextEventNumber`, not the number of the last event that was read.


gregoryyoung:
:+1:

PR #1076, opened 2016-10-14
Don't set the exception on the task completion source twice

hayley-jean:
Fixes an issue that could occur if an exception is thrown while reading events in a catchup subscription. The exception would be set on the completion source twice, resulting in an exception being thrown. 

Because the task has already returned, the exception on the task would not be checked by the client, and an exception would eventually be thrown later when the task gets disposed.


pgermishuys:
👍 will merge once all the tests pass

gregoryyoung:
:+1:

PR #1086, opened 2016-11-06
Add required format argument for debug messages

### LINKED ISSUE #1084: .NET client 3.9.2 log uses Debug instead of DebugFormat in some places
When subscription is recovering, we get messages like this:

```
[09:05:56 DBG] Catch-up Subscription to {0}: recovering after reconnection.
[09:05:56 DBG] Catch-up Subscription to {0}: unhooking from connection.Connected.
[09:05:56 DBG] Catch-up Subscription to {0}: recovering after reconnection.
[09:05:56 DBG] Catch-up Subscription to {0}: unhooking from connection.Connected.
```

Rest of the messages correctly display the subscription name

pgermishuys:
Fixes #1084 
Might be worthwhile going through and making sure these are the only occurrences of this.


gregoryyoung:
:+1:

I wish C# was more like F# in this regard and gave compile time errors.

PR #1087, opened 2016-11-07
Prevent possible errors when setting the task result for a catchup subscription

hayley-jean:
Use the Try* methods on the TaskCompletionSource in catchup subscriptions to prevent errors from being thrown when the task has been disposed.


gregoryyoung:
:+1: should be good.

PR #1096, opened 2016-11-16
Adding ConditionalAppendToStream API

SzymonPobiega:
The ConditionalAppendToStream returns the status instead of throwing an exception in cases of stream version mismatch or stream being deleted.

It may be useful in cases where there are a couple of different index streams that contain links to multiple source streams. When processing an index stream, the changes to source streams are applied only if the link leads to the latest version of the source stream and the stream itself exists.

### DIFF
--- a/src/EventStore.ClientAPI.Embedded/EmbeddedResponders.cs
+++ b/src/EventStore.ClientAPI.Embedded/EmbeddedResponders.cs
@@ -57,6 +57,59 @@ namespace EventStore.ClientAPI.Embedded
             }
         }
 
+        internal class ConditionalAppendToStream :
+            EmbeddedResponderBase<ConditionalWriteResult, ClientMessage.WriteEventsCompleted>
+        {
+            private readonly string _stream;
+
+            public ConditionalAppendToStream(TaskCompletionSource<ConditionalWriteResult> source, string stream)
+                : base(source)
+            {
+                _stream = stream;
+            }
+
+            protected override void InspectResponse(ClientMessage.WriteEventsCompleted response)
+            {
+                switch (response.Result)
+                {
+                    case OperationResult.Success:
+                        Succeed(response);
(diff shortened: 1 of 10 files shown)


src/EventStore.ClientAPI.Embedded/EmbeddedResponders.cs
```diff
@@ -57,6 +57,61 @@ protected override WriteResult TransformResponse(ClientMessage.WriteEventsComple
...
+            EmbeddedResponderBase<ConditionalWriteResult, ClientMessage.WriteEventsCompleted>
+        {
+            private readonly int _expectedVersion;
+            private readonly string _stream;
+
+            public ConditionalAppendToStream(TaskCompletionSource<ConditionalWriteResult> source, string stream, int expectedVersion)
+                : base(source)
+            {
+                _stream = stream;
+                _expectedVersion = expectedVersion;
```
pgermishuys: Assigned a value but never used. Currently failing on travis because it treats warnings as errors.

pgermishuys:
Thanks for the PR @SzymonPobiega! Travis is currently failing due to `treat warnings as errors`.

SzymonPobiega:
@pgermishuys I fixed that. Do you think it makes sense to add overloads of `ConditionalAppend` similar to the overloads of `Append`?

pgermishuys:
@SzymonPobiega thanks for fixing the build. I don't think so, no. I think the method you provided is fine.

pgermishuys:
@SzymonPobiega mind squashing your commits into a single commit so that we can merge? This is just so that we can keep the history clean.

Apologies for not asking this sooner.

SzymonPobiega:
Rebased against `release-4.0.0` and squashed. Should be good to go when the checks pass.

pgermishuys:
Thank you very much.

PR #1101, opened 2016-11-24
Upgrade V8

### LINKED ISSUE #926: Event Store crashes on projection using Date.toLocaleString (Windows only)
Original source of the issue https://groups.google.com/forum/#!topic/event-store/SZ31hiaQe30

> Using Date.toLocaleString in a query makes my EventStore crash. I'm running it from console in a Windows 8.1 and I've tried with versions 3.6.1 and 3.6.2. I don't think it has any impact but I've tried running the query from both Chrome and Internet Explorer.
> 
> The moment I run the query, EventStore crashes showing the usual Windows dialog stating that my "EventStore.ClusterNode.exe has stopped working" and immediately exiting in the console without showing any error message there. I've found no information regarding the crash in neither -cluster-node.log nor -cluster-node-err.log.
> 
> Others Date methods such as toDateString (which is what I ended up using) work without problem.

```
[
  {
    "eventId": "fbf4a1a1-b4a3-4dfe-a01f-ec52c34e16e4",
    "eventType": "testEvent",
    "data": {
      "Timestamp": "2016-05-24T09:38:27.8995302+02:00"
    }
  }
]
```

```
curl -i -d @testEvent.txt "http://127.0.0.1:2113/streams/newstream" -H "Content-Type:application/vnd.eventstore.events+json"
```

```
fromStream('newstream')
    .when({ $any: function(s,e) {

        var date = new Date(e.body['Timestamp']).toLocaleString();
        return date;
}}) 
```

### LINKED ISSUE #971: v8 project missing
When running `https://v8.googlecode.com/svn/trunk`,

```
Checking out V8 repository...
svn: E170013: Unable to connect to a repository at URL 'https://v8.googlecode.com/svn/trunk'
svn: E160013: '/svn/trunk' path not found
```

Clearly, the repository moved to another link. Please fix it or update the README :smiley:

pgermishuys:
Fixes #971 
Fixes #926
With the change v8 has undergone since 3.24.10 (the version of v8
projections used) there have been a lot of changes introduced and this
commit includes changes that upgrades the version of v8 to 5.2 which at
the point of starting the upgrade of v8 was the latest stable version of
v8 which was used by Chrome (Browser) found via OmahaProxy
(https://omahaproxy.appspot.com/)

Some of the notable changes introduced in later versions of v8 that
effected the upgrade process
are
- Isolate constructor is no longer parameterless
- ObjectTemplate only accepts primitive values only (numbers, strings,
  etc.)

The benefits include
- Taking advantage of bug fixes introduced in v8 since 3.24.10
- ECMAScript 6 support for Projections

Other notable changes
- Make use of the LLVM compiler toolchain on Linuxes

This is a relatively large change which will require the README to be updated to ensure that users are aware that currently v8 can only be built with Visual Studio 2015 installed on Windows, however on *nixes, LLVM is used.


andho:
Will the `v8-upgrade` be backported to any 3.x branches? Or will there be a fix for 3.x with the svn link pointing to the correct location of the appropriate v8 version?

pgermishuys:
@andho not currently. Projections was not a supported feature in 3.x versions of Event Store and therefore as it currently stands the upgrade will not be backported.

PR #1121, opened 2016-12-07
Use Environment.Exit when exiting to prevent hanging on shut down

hayley-jean:
Handle `Console.CancelKeyPress` in ClusterNode and TestClient to ensure they call Environment.Exit and shutdown properly.
Also add a new property `Initialized` to LogManager to determine whether to write output to the console or to the log.


gregoryyoung:
👍

PR #1124, opened 2016-12-14
Fixes #1034

### LINKED ISSUE #1034: Replica Status: negative total bytes sent
I noticed this today under cluster status

replica_status_odd_total_sent

Is this because of int vs. long in UI?

nissbran:
The eventstore made the wrong conversion and returned a negative value for totalbytessent in JSON response from the http://*:1114/stats/replication API call.


gregoryyoung:
:+1:

PR #1138, opened 2016-12-30
Add "DisableMessageTimeout" to persistent subscription settings

volak:
SSIA - instead of doing something silly like `.WithMessageTimeout(Timespan.FromDays(365))` the client can just disable timeouts all together giving him time to ACK to all the messages in his buffer without worrying about a time limit.

I think I may have to justify such a feature ... so here goes.

The idea of a message timeout doesn't gel with my implementation of ES.  `Exactly-once` delivery is not the desired result here, but it gets me closer.  I run ES with `WithMaxRetriesOf(0)` - I don't want another client getting a message simply because one client is running a little slow.  It could be he'll catch up at some point, or I could just add more clients to alleviate how many messages he gets.  But at the end of the day every event in his buffer should be his and no one elses. 

The idea of a message timeout, where a message sitting in the buffer of 1 client can be resent to another client is not ideal - so I set Retries to 0 to simply park the message.  

But when toying with #803 it came to my attention that once a message has timed out I can't ACK it successfully.  Be it a bug or w/e, according to the store by the time my slow client caught up to ACK the message its too late and the message is parked waiting to be delivered again.  However the message WAS successfully processed it just took too long.  Leaving me with the option to set a message timeout for a year+ or .. this PR :)

The situation message timeouts tries to address can be addressed by myself, therefore its a nice feature to be able to disable it.

Its my understanding that when a client disconnects entirely all unacked messages in his buffer are rerouted, this is just fine to me I don't need the store to be proactive about how slow my consumers are.

Implementation
-----------------
I implemented the feature by changing the settings objects to have `TimeSpan?` datatype for message timeout.  A null would mean timeouts are disabled.
After finally changing every place I had to I wondered if maybe you would prefer a new `bool timeoutDisabled` instead?  My changes were easier, but the `bool` would be better compatible with older versions I suppose.

* if you hate the change don't be afraid to tell me ( 😆 )


volak:
Oops for some reason my environment did not warn me about the unsupported `?.` operator - I'll commit a fix for the failed build in a few.

gregoryyoung:
OK I can kind of see value in this. As you mention this is achievable via setting the TimeSpan to an arbitrarily high value. Perhaps it would be better to just support that in the client API with a better named method? As example DoesNotTimeOut() which sets the TimeSpan to int64.maxvalue which is roughly 30k years?

gregoryyoung:
@volak to be fair the DoesNotTimeOut solution does fail in the case of being > 30k years. I have my doubts how much we will all care at that point.

volak:
@gregoryyoung Originally I thought setting timeout milliseconds to int.MaxValue - however from what I see the RetryableMessage holding structure is holding all the messages waiting with an expiration.
Wouldn't setting timeouts arbitrarily long in-theory slow this structure down as it has to sort through it?  I understand its balanced to put the most close expired message first, but still seems like a cost we can just avoid with a null check

gregoryyoung:
@volak it will only hold client buffer sizes + read buffer size messages. It should be fine.

volak:
Ok sounds good - did you want me to close this one and make a new PR?  Or is the change simple enough to do without my clumsiness?

pgermishuys:
@volak a separate pull request might be better.

volak:
Ok - I'll close this one and work on it tomorrow.  Seems like it will only require 1 line change in the client settings builder and I'll copy over the tests I wrote for timing out.

pgermishuys:
Thanks @volak

gregoryyoung:
@volak #1140

PR #1139, opened 2016-12-30
Handle timeouts in the Event By Type Reader

pgermishuys:
The event by type index event reader first attempts to read each of the
streams of the event types it cares about. e.g. "$et-eventTypeExample"
and after there are no more events, it will switch to an $all reader.

These readers will now handle the case where a message expiry could have
resulted in not returning a result for the reads that the reader sent
off.

Add additional tests to verify that another read is attempted for the
reads that timed out


gregoryyoung:
:+1:

PR #1140, opened 2016-12-30
Cctimeout

gregoryyoung:


volak:
@gregoryyoung I just spotted what seems to be an issue with this fix.  In the OutstandingMessageCache which is keeping track of messages expiring the only way a message is removed from `_byTime` is when it expires...
`Remove(Guid MessageId)` only removes objects from `_outstandingRequests` when the message actually expires its checked if its still in `_outstandingRequests` before yielding as an expired message.

Won't using `TimeSpan.MaxValue` make _byTime inflate forever?

gregoryyoung:
_byTime isn't removed on an ack?

volak:
Not that I can see - I think ACK just calls `Remove` which removes the message from `_outstandingRequests`  and `_bySequences` only.  

The only time something is removed from `_byTime` is when `GetMessagesExpiringBefore` detects an expiration

gregoryyoung:
put as a separate issue then as it was an existing problem we just had never run into someone using huge numbers of messages with very long timeouts.

volak:
Also... after some of my testing TimeSpan.MaxValue isn't compatible with the message format between client api and server.  Its getting casted to an `int` which ends up being `-2147483648` making my messages expire instantly 🤕 

TotalMilliseconds itself is `922337203685477` - ideal for expiration but too large for an `int`

gregoryyoung:
Yes have to make it lower (and gate too high timespans) 

required int32 message_timeout_milliseconds = 5;

volak:
Just a thought but you could make "infinite timeout" special number `-1`  'Cause who would want messages to expire in -1 milliseconds? haha.
A max timeout of 25 days is OKAY but theoretically possible to hit if a dev ops guy wasn't paying attention.  
Then you could probably put a max timeout of .. 30 minutes? or something.  Which would fix #1147 as `_inTime` wouldn't grow insanely large even at 2000 msgs/s

gregoryyoung:
@volak I know of nobody else using the "we want to set timeouts very long so they never happen use case" remember the worst case after 25 days is the message would be parked.

PR #1142, opened 2016-12-30
Tests 1140

volak:
Add 2 tests to client api and 1 test to core for testing message timeouts
There's a test to make sure a client gets messages after a theoretical timeout, and a test to make sure a client doesn't get a message after a timeout.

Don't know if they'll be used but I wrote them for #1138


hayley-jean:
Thanks for the PR @volak!

We've manually merged your tests into PR https://github.com/EventStore/EventStore/pull/1215

Issue #1147, opened 2017-01-03
PersistentSubscription: when message timeout is large _byTime grows large


volak:
The `OutstandingMessageCache` structure doesn't remove messages from `_byTime` upon ACKing.  It only removes them when the message expires causing an issue if you use a large message timeout such as #1140 

If a message "never" expires `_byTime` would just grow forever

gregoryyoung:
The reason I moved it from #1140 as its a different issue. Whether it was 30 years or 5 years at 2000 messages/second wouldn't much matter  :+1:

gregoryyoung:
The solution I think is to remove on read by remembering the expiry (otherwise it ends up in a scan)

hayley-jean:
This has been fixed by #1151

PR #1151, opened 2017-01-04
Competingtime

gregoryyoung:
This should be a fix for #1147

It required moving bytime to be a sorteddictionary on a dual key (time/message id) as time can be duplicated. A pairing heap as it was would not be able to handle the remove operation. Under anything near normal circumstances the perf difference should be minimal. As it was all encapsulated into outstandingmessagecache I just used tuples where needed. Two item named classes may be preferred.

Also this changes how timeouts are working more in line with what #1140 wanted.

Have tested (unit tests only in linux).
Have not stress tested live subscriptions
Have not tested the infinite timeout though it should work
Have not tested in windows at all

### DIFF
--- a/src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
+++ b/src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
@@ -69,7 +69,7 @@ namespace EventStore.Core.Services.PersistentSubscription
             _checkpointReader = new PersistentSubscriptionCheckpointReader(_ioDispatcher);
             _streamReader = new PersistentSubscriptionStreamReader(_ioDispatcher, 100);
             //TODO CC configurable
-            _tickRequestMessage = TimerMessage.Schedule.Create(TimeSpan.FromMilliseconds(1000), 
+            _tickRequestMessage = TimerMessage.Schedule.Create(TimeSpan.FromMilliseconds(1000),
                                                    new PublishEnvelope(_bus),
                                                    new SubscriptionMessage.PersistentSubscriptionTimerTick());
         }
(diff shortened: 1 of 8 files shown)


gregoryyoung:
Have a small change will push tomorrow.

gregoryyoung:
NM this is fine. To optimize is not worth it at this point.

In particular I was looking at walking the tree instead of using .First each time but as it becomes an enumerator (only way to access) you run into the changing enumerator problem. To work around this just isn't worth the effort.

src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
```diff
@@ -171,29 +171,31 @@ public void Handle(ClientMessage.CreatePersistentSubscription message)
...
+            CreateSubscriptionGroup(message.EventStreamId,
+                                    message.GroupName,
+                                    message.ResolveLinkTos,
                                     message.StartFrom,
                                     message.RecordStatistics,
                                     message.MaxRetryCount,
                                     message.LiveBufferSize,
                                     message.BufferSize,
                                     message.ReadBatchSize,
-                                    TimeSpan.FromMilliseconds(message.CheckPointAfterMilliseconds),
```
volak: @gregoryyoung  Here and line 188 - times are reversed.  message timeout being used as checkpoint and visa versa.

Side note in the web ui the timespan is being displayed as `-2147483648` when using `DontTimeout` - not a big deal

gregoryyoung: @volak updated. And yes there is some UI work associated.

PR #1159, opened 2017-01-05
Get rid of __MonoCS__

### LINKED ISSUE #1157: __MonoCS__ gone.
Need to move to some other define.

gregoryyoung:
Replaced with existing USE_UNIX_IO and MONO for other places.
Also added a few MONOCKECK comments where we should follow up if it
is still needed or not fixes #1157

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -39,6 +39,7 @@ EOF
 }
 
 CONFIGURATION="Release"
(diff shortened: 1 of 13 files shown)


gregoryyoung:
Seems reasonable and no dependency now. That code will just dissappear with unbuffered

PR #1168, opened 2017-01-08
During Scavenge operation count the number of times that it fails 

riccardone:
related to #964


hayley-jean:
Thanks for the PR @riccardone

Your PR pointed out that the final status of the scavenge is incorrect, however I don't think it's necessary to count the number of chunk failures during the scavenge. Instead, we should show a warning if there are errors during the scavenge both in the logs and the admin UI, it will be easy enough to look back through the logs to find the culprits and also see the errors that happened in those chunks.

I've created a github issue related to this here:
https://github.com/EventStore/EventStore/issues/1214

PR #1188, opened 2017-01-24
Add option for disabling insecure TCP

pgermishuys:


pgermishuys:
I have pushed up the necessary changes according to the review.
- If the user chooses to disable insecure tcp connections and have not set to use the internal ssl, the node will fail to start
- If the user chooses to disable insecure tcp connections but have not setup the external or internal tcp connections, the node will fail to start.

PR #1197, opened 2017-01-31
Remove debug message

pgermishuys:


gregoryyoung:
We probably want this as a log for debugging purposes on a db.

pgermishuys:
👍  and done @gregoryyoung

PR #1200, opened 2017-02-06
Exposes the index commit position

jageall:
Exposes index last commit position to the client api

### DIFF
--- a/src/EventStore.ClientAPI.Embedded/EmbeddedResponders.cs
+++ b/src/EventStore.ClientAPI.Embedded/EmbeddedResponders.cs
@@ -338,7 +338,8 @@ namespace EventStore.ClientAPI.Embedded
                     response.Events.ConvertToClientResolvedIndexEvents(),
                     response.NextEventNumber,
                     response.LastEventNumber,
-                    response.IsEndOfStream);
+                    response.IsEndOfStream,
+                    response.TfLastCommitPosition);
 
             }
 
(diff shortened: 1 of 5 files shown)


gregoryyoung:
Not a breaking change.

@jageall I notice you named it index commit position where is this derived from?

jageall:
@gregoryyoung mainly to call out where I think it is coming from and get some comments on whether I am understanding what it is correctly :)

so it is read from [here](https://github.com/EventStore/EventStore/blob/release-v4.0.0/src/EventStore.Core/Services/Storage/StorageReaderWorker.cs#L180) which after following it through is the commit position from the index file i believe, rather than the commit position of any of the events (although it could be)

gregoryyoung:
I'm not sure this is what you are looking for but instead wanting the event level P/C. https://github.com/EventStore/EventStore/blob/release-v4.0.0/src/EventStore.Core/Services/Storage/ReaderIndex/IndexCommitter.cs#L206 is where the index has actually read to I believe

jageall:
so i called it index commit position, because it is the commit position the index has read to. 

so ideally, what i want is the P/C for each event. but that is not actually available from the index. however, knowing where the index has got to gives me options to resolve whether i can switch to my $all cache I believe.

as in if the index has read past the tail of my cache, i can safely start processing from the tail as any future commits will appear after that in the $all subscription. And i think that is correct because the commit position will be > the commit position the index has read to because the TfPos comparison logic sorts first by commit position and then by prepare.

PR #1211, opened 2017-02-13
Release the resources the test is holding onto

pgermishuys:
Ensure that the test does not hold on to resources that needs to be cleaned up


gregoryyoung:
:+1:

PR #1225, opened 2017-02-22
Log expired operations

pgermishuys:
It might be worthwhile logging out these expired operations. This should not have the potential of filling up the logs.


gregoryyoung:
:+1:

PR #1227, opened 2017-02-24
Fixing some issues with the Verify option in test client.

simoncowie:
Been using the test client recently and wanted to add some fixes we found were necessary.

### DIFF
--- a/src/EventStore.TestClient/Commands/DvuBasic/DvuBasicProcessor.cs
+++ b/src/EventStore.TestClient/Commands/DvuBasic/DvuBasicProcessor.cs
@@ -208,7 +208,7 @@ namespace EventStore.TestClient.Commands.DvuBasic
 
             var failed = 0;
 
-            var rnd = new Random();
+            var rnd = new Random(writerIdx);
 
             var streamIdx = -1;
             var head = -1;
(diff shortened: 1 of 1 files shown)


src/EventStore.TestClient/Commands/DvuBasic/DvuBasicProcessor.cs
```diff
@@ -208,7 +208,7 @@ private void Write(Status status, int writerIdx, CommandProcessorContext context
 
             var failed = 0;
 
-            var rnd = new Random();
+            var rnd = new Random(writerIdx);
```
simoncowie: The writers were getting a lot of version conflicts as random without seed would spit out the same values

src/EventStore.TestClient/Commands/DvuBasic/DvuBasicProcessor.cs
```diff
@@ -296,6 +293,7 @@ private void Write(Status status, int writerIdx, CommandProcessorContext context
                                         wrongExpectedVersion, streamsDeleted, failed, requests);
             status.FinilizeStatus(writerIdx, failed != sent);
             connection.Close();
+            finish.Set();
```
simoncowie: With `finish.Set()` called in the packageHandler callback, the last writer to finish would sometimes miss calling the `status.FinilizeStatus` method before results were tallied

gregoryyoung:
This looks reasonable.

pgermishuys:
Thank you for the PR @simoncowie. The wercker build is failing fetching sources, asked the guys from `wercker` to look into that.

PR #1235, opened 2017-03-07
Cleanup for TCP Client Dispatcher

hayley-jean:
Remove `V1ClientDispatcher` and revert `TcpService`, `TcpCommand` and `ConnectionManager` to previous versions.
Add versioned registry of wrappers/unwrappers and pass the client version in when wrapping/unwrapping messages.
Add wrappers and unwrappers for tcp messages that need to be downgraded for version 1 clients.
Remove unnecessary proto messages.

Send an identify message when a client tries to connect that contains the client's version and connection name.
If identify on client fails, the client will try to reconnect. 
_Note: This IdentifyClient behaviour will need to be done on 3.9.4 as well if we want 4.0.0 clients to be able to communicate with 3.9.4 nodes_


src/EventStore.Core.Tests/Services/Transport/Tcp/TcpClientDispatcherTests.cs
```diff
@@ -0,0 +1,224 @@
...
+                Guid.NewGuid().ToString(), TcpServiceType.External, new ClientTcpDispatcher(),
+                InMemoryBus.CreateTest(), dummyConnection, InMemoryBus.CreateTest(), new InternalAuthenticationProvider(
+                                new Core.Helpers.IODispatcher(InMemoryBus.CreateTest(), new NoopEnvelope()), new StubPasswordHashAlgorithm(), 1),
+                TimeSpan.FromSeconds(10), TimeSpan.FromSeconds(10), (man, err) => { });
+        }
+
+        [Test]
+        public void when_unwrapping_message_that_does_not_have_version1_unwrapper_should_use_version2_unwrapper()
+        {
+            var dto = new TcpClientMessageDto.DeleteStream("test-stream", 2, true, false);
```
pgermishuys: We probably want to change this "2" here and use `ExpectedVersion.Any` for readability.

src/EventStore.Core.Tests/Services/Transport/Tcp/TcpClientDispatcherTests.cs
```diff
@@ -0,0 +1,224 @@
...
+        [Test]
+        public void when_wrapping_message_that_does_not_have_version1_wrapper_should_use_version2_wrapper()
+        {
+            var msg = new ClientMessage.DeleteStream(Guid.NewGuid(), Guid.NewGuid(), _envelope, true, "test-stream", 2, false, SystemAccount.Principal);
+            var package = _dispatcher.WrapMessage(msg, _version);
+            Assert.IsNotNull(package, "Package");
+            Assert.AreEqual(TcpCommand.DeleteStream, package.Value.Command);
+
+            var dto = package.Value.Data.Deserialize<TcpClientMessageDto.DeleteStream>();
+            Assert.IsNotNull(dto, "DTO");
```
pgermishuys: The descriptions for the failure cases should be consistent with the other failure cases in this PR. For example, "DTO is null". e.g. [here](https://github.com/EventStore/EventStore/pull/1235/commits/5b302e7b0edd704d99b7cfc21bab422767d5c5ee#diff-5ce4b7d447bba7a216eeaacad3c7e46aR77)

PR #1250, opened 2017-03-20
Made fromCategories available from projections.

### LINKED ISSUE #1244: Add `fromCategories` helper.
Hey there.

I'm aware this is somewhat of a small issue, but it would be quite helpful to have a `fromCategories([])` helper in the user-defined JavaScript projections. Where I work, we keep events for each instance of a resource separate, and we wish to output these all to a single stream that we then process using persistent consumers. Right now, I'm having to manually use `fromStreams(['$ce-brands', '$ce-manufacturers'])`. I'm very willing to create a PR with the code if you're open to the idea.

Thank you.

michielboekhoff:
This issue fixes #1244 by adding a function analogous to `fromStreams` for categories. This is almost a verbatim copy of `fromStreams`.

### DIFF
--- a/src/EventStore.Projections.Core/Prelude/1Prelude.js
+++ b/src/EventStore.Projections.Core/Prelude/1Prelude.js
@@ -210,6 +210,18 @@ function scope($on, $notify) {
         };
     }
 
+    function fromCategories(categories) {
+        var arr = Array.isArray(categories) ? categories : arguments;
+        for (var i = 0; i < arr.length; i++)
+            eventProcessor.fromCategory(arr[i]);
+
+        return {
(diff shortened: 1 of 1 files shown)


pgermishuys:
Thanks for the PR @michielboekhoff. Will have it reviewed. This is unlikely to make it into v4.0.0 final but will certainly be reviewed and decided upon for a minor release.

michielboekhoff:
Thanks for the quick response @pgermishuys. The work-around that exists currently is acceptable, just not optimal, so a minor release after v4.0.0 sounds great.

pgermishuys:
@michielboekhoff mind reopening this on the `release-v4.0.1` branch?

PR #1265, opened 2017-03-29
Add functionality to parse Dictionaries via Rags

pgermishuys:


etc/EventStore.Rags/Yaml.cs
```diff
@@ -61,6 +62,10 @@ public static IEnumerable<OptionSource> FromFile(string fileName, string section
                 {
                     options.Add(OptionSource.String("Config File", yamlElement.Key.ToString(), yamlElement.Value.ToString()));
                 }
+                else if (yamlMappingNode != null && !string.IsNullOrEmpty(sectionName))
```
pgermishuys: We are basically saying here that we will not parse map nodes unless we are parsing a section. This should be fine in most cases unless we start introducing dictionaries as parameters inside of Event Store.

pgermishuys: We are basically saying here that we will not parse map nodes unless we are parsing a section. This should be fine in most cases unless we start introducing dictionaries as parameters inside of Event Store.

PR #1268, opened 2017-04-02
Added "Warning" to the ScavengeResult enum; Fixes 1214

### LINKED ISSUE #1214: Add severity levels for a scavenge
Currently a scavenge will report that it was completed successfully even if some chunks were skipped due to errors. This masks the fact that the scavenge ran into issues, and can be misleading.

In these cases, we should set the scavenge result to `Warning` and show this both in the logs and in the UI.

riccardone:
Closes #1214 

@hayley-jean this PR log and return warnings if there has been errors during scavenge operation. The UI changes can be done in a separate PR. Is that correct?


src/EventStore.Core/Services/Storage/StorageScavenger.cs
```diff
@@ -96,8 +96,11 @@ private void Scavenge(ClientMessage.ScavengeDatabase message)
 
                     var scavenger = new TFChunkScavenger(_db, _ioDispatcher, _tableIndex, _readIndex, scavengeId,
                                                                                  _nodeEndpoint, unsafeIgnoreHardDeletes: _unsafeIgnoreHardDeletes);
-                    spaceSaved = scavenger.Scavenge(_alwaysKeepScavenged, _mergeChunks);
-                    result = ClientMessage.ScavengeDatabase.ScavengeResult.Success;
+                    int failedCount;
```
riccardone: @hayley-jean here is where the StorageScavenger received the number of failure and set the Warning status

riccardone:
@hayley-jean 
"...count how many chunks failed and return that to the StorageScavenger"
This is what this PR is doing. I count the number of failure and I return it to the StorageScavenger. To achieve this goal, I split the original Scavenge procedure in small parts and I also set using an enum a warning message as you suggested in one of the previous comments.

pgermishuys:
@riccardone thanks! We will manually merge this into 4.0.2. 👍

PR #1279, opened 2017-04-18
Handle all messages in queue

pgermishuys:
There are cases when the consumer can throw and in those cases the proceeeding messages in the batch are not handled


gregoryyoung:
@hayley-jean :+1:

PR #1280, opened 2017-04-20
Revert to legacy queued handler.

pgermishuys:


Narvalex:
Can someone tell me please what is the matter with the QueueHandlerUsingMpsc?

Narvalex:
I thought that the @Scooletz queue was better than the legacy one... What happen?

gregoryyoung:
It seems to run into an occasional issue but only on AWS linux, thread getting stuck in enqueue. We root caused an issue on AWS linux to that. Therefore moving back to the old one for now. The overall performance difference is extremely minimal.

Narvalex:
Thanks Greg. But why not just go like this
#if MONO
return DefaultHandler
#else
return MinimalPerformanceBoostHandler
#endif

Scooletz:
@gregoryyoung So it's enqueuing thread that is freezing? Is it MONO only?

gregoryyoung:
likely will. There is not a dramatic performance increase. My bigger worry is there is a bug in other environments possibly including windows and we have not run into it yet. As I said this would be reviewed.

@Scooletz not mono only. AWS linux only. We have not seen the issue in ubuntu etc. It may exist in OSX as well (have not tested enough)

Scooletz:
I'd say so. Revert them all. 
As you bumped up .NET there's `Volatile` accessible now which would greatly simplify this code (no `IntPtr` magic included)

gregoryyoung:
@Scooletz I can also send over threadstacks (managed + gdbg) showing the problem. Basically processing locks (in this case on rebuilding an index). CPU pegs up threadstacks keep coming back withenqueue on the top. It being there is a while(true) it seems like a reasonable candidate for an access pattern problem that loops forever. We have a reproduction now, a reasonable test would be to try limit that to some large number (say 50k) and see if the issue still occurs.

a perflog shows the enqueue methods using CPU.

Here is a thread stack:

"Finalizer"
"Threadpool worker"
"MainQueue"  at EventStore.Core.Bus.MPSCMessageQueue.Enqueue (EventStore.Core.Messaging.Message) <0x00124>
  at EventStore.Core.Bus.QueuedHandlerAutoResetWithMpsc.Publish (EventStore.Core.Messaging.Message) <0x0002b>
  at EventStore.Core.Services.Transport.Http.HttpService.Handle (EventStore.Core.Messages.HttpMessage/PurgeTimedOutRequests) <0x000cc>
  at EventStore.Core.Bus.MessageHandler`1<T_REF>.TryHandle (EventStore.Core.Messaging.Message) <0x000af>
  at EventStore.Core.Bus.InMemoryBus.Publish (EventStore.Core.Messaging.Message) <0x000db>
  at EventStore.Core.Services.VNode.VNodeFSMHandling`1/<Do>c__AnonStorey1<TMessage_REF>.<>m__0 (EventStore.Core.Data.VNodeState,EventStore.Core.Messaging.Message) <0x00058>
  at EventStore.Core.Services.VNode.VNodeFSM.Handle (EventStore.Core.Messaging.Message) <0x00133>
  at EventStore.Core.Services.VNode.ClusterVNodeController.EventStore.Core.Bus.IHandle<EventStore.Core.Messaging.Message>.Handle (EventStore.Core.Messaging.Message) <0x0002b>
  at EventStore.Core.Bus.QueuedHandlerAutoResetWithMpsc.ReadFromQueue (object) <0x00243>
  at System.Threading.ThreadHelper.ThreadStart_Context (object) <0x000d1>
  at System.Threading.ExecutionContext.RunInternal (System.Threading.ExecutionContext,System.Threading.ContextCallback,object,bool) <0x001c6>
  at System.Threading.ExecutionContext.Run (System.Threading.ExecutionContext,System.Threading.ContextCallback,object,bool) <0x00020>
  at System.Threading.ExecutionContext.Run (System.Threading.ExecutionContext,System.Threading.ContextCallback,object) <0x00059>
  at System.Threading.ThreadHelper.ThreadStart (object) <0x0004c>
  at (wrapper runtime-invoke) <Module>.runtime_invoke_void__this___object (object,intptr,intptr,intptr) <0x000ed>

PR #1286, opened 2017-04-24
Scavenge away LinkTo events that point to events that no longer exist

### LINKED ISSUE #1270: Scavenge LinkTos
If the original event of a linkTo no longer exists, we should scavenge it away.

hayley-jean:
Fixes #1270


gregoryyoung:
:+1: this will make quite a few dbs shrink a bunch on their next scavenge.

Issue #1231, opened 2017-03-01
Http on root includes listening port in redirect behind proxy


btrepp:
The webUI seems to include the port it is listening on in the redirect. Even if the Host: header doesn't include a port. It also doesn't respect the protocol, but it does grab the host.
Here is the verbose output of `curl -v https://eventstore.mycompany.org`. Sitting behind IIS/ARR as a reverse proxy

```
> GET / HTTP/1.1
> Host: eventstore.mycompany.org
> User-Agent: curl/7.49.1
> Accept: */*
>
* STATE: DO => DO_DONE handle 0x600057330; line 1659 (connection #0)
* STATE: DO_DONE => WAITPERFORM handle 0x600057330; line 1786 (connection #0)
* STATE: WAITPERFORM => PERFORM handle 0x600057330; line 1796 (connection #0)
* HTTP 1.1 or later with persistent connection, pipelining supported
< HTTP/1.1 302 Found
< Content-Length: 5
< Content-Type: text/plain; charset=utf-8
< Location: http://eventstore.mycompany.org:2113/web/index.html
* Server Microsoft-IIS/8.5 is not blacklisted
< Server: Microsoft-IIS/8.5
< Access-Control-Allow-Methods: GET, OPTIONS
< Access-Control-Allow-Headers: Content-Type, X-Requested-With, X-Forwarded-Host, X-PINGOTHER, Authorization, ES-LongPoll, ES-ExpectedVersion, ES-EventId, ES-EventType, ES-RequiresMaster, ES-HardDelete, ES-ResolveLinkTo
< Access-Control-Allow-Origin: *
< Access-Control-Expose-Headers: Location, ES-Position, ES-CurrentVersion
< X-Powered-By: ARR/3.0
< Date: Wed, 01 Mar 2017 23:46:03 GMT
<
```

I would expect the redirect to be. `Location: https://eventstore.mycompany.org/web/index.html` (which I will add appears to work perfectly when you go to that location manually in a browser.)

I can see the issue here
https://github.com/EventStore/EventStore/blob/115819a16af0f2f83ecc613151152a7b75a6c061/src/EventStore.Transport.Http/EntityManagement/HttpEntity.cs#L45
Looks like GES only attempts to look at the `X-Forwarded-Port` header for rebuilding urls, defaulting to the listening port. This information is inside the Host header eg

` Host = uri-host [ ":" port ] ; Section 2.7.1`
from https://tools.ietf.org/html/rfc7230#section-5.4

[...]

PR #1293, opened 2017-05-01
Fix reverse proxy issues

shaan1337:
Fixes reverse proxy issues (#1231) by rewriting the "Location" header as an absolute URL before sending out the response if the "Location" base matches the Requested URL base.

This should work by default under any reverse proxy if ES is mapped under the root directory. However, if the reverse proxy is mapping ES under a subfolder e.g. /es/, then the X-Forwarded-Prefix header must be set. Since URLs are now relative, there is now no need for X-Forwarded-Proto, X-Forwarded-Host and X-Forwarded-Port but it has been still kept in the code.


shaan1337:
Added a more robust check. Will send a second PR.

PR #1294, opened 2017-05-02
Fix reverse proxy issues

shaan1337:
Fixes reverse proxy issues (#1231) by rewriting the "Location" header as an absolute URL before sending out the response if the "Location" base matches the original Requested URL base.

This should work by default under any reverse proxy if ES is mapped under the root directory. However, if the reverse proxy is mapping ES under a subfolder e.g. /es/, then the X-Forwarded-Prefix header must be set. Since URLs are now relative, there is now no need for X-Forwarded-Proto, X-Forwarded-Host and X-Forwarded-Port but it has been still kept in the code.

### DIFF
--- a/src/EventStore.Core.Tests/Services/Transport/Http/build_requested_url_should.cs
+++ b/src/EventStore.Core.Tests/Services/Transport/Http/build_requested_url_should.cs
@@ -52,19 +52,19 @@ namespace EventStore.Core.Tests.Services.Transport.Http
         }
 
         [Test]
-        public void with_port_forward_header_only_port_is_changed()
+        public void with_port_forward_header_only_nothing_is_changed()
         {
             var headers = new NameValueCollection { { "X-Forwarded-Port", "4321" } };
             var requestedUri =
                 HttpEntity.BuildRequestedUrl(inputUri, headers, null, 0);
 
-            Assert.AreEqual(new Uri("http://www.example.com:4321/path/?key=value#anchor"), requestedUri);
+            Assert.AreEqual(new Uri("http://www.example.com:1234/path/?key=value#anchor"), requestedUri);
         }
 
         [Test]
         public void non_integer_port_forward_header_is_ignored()
         {
-            var headers = new NameValueCollection { { "X-Forwarded-Port", "abc" } };
+            var headers = new NameValueCollection { {"X-Forwarded-Proto", "http"}, {"X-Forwarded-Host","www.example.com"}, { "X-Forwarded-Port", "abc" }};
             var requestedUri =
                 HttpEntity.BuildRequestedUrl(inputUri, headers, null, 0);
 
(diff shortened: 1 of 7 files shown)


pgermishuys:
Hey @shaan1337, thank you for the PR, would you mind squashing your commits into a logical commit to keep the history clean?

Will have someone review during the weekend or on monday, sorry for the delay in getting back to you.

Thank you once again.

shaan1337:
No problem, thanks for the feedback @pgermishuys, commits have been squashed.

src/EventStore.Core.Tests/Services/Transport/Http/build_requested_url_should.cs
```diff
@@ -52,43 +52,43 @@ public void with_advertise_ip_and_http_port_set_both_host_and_port_is_changed()
         }
 
         [Test]
-        public void with_port_forward_header_only_port_is_changed()
+        public void with_port_forward_header_only_nothing_is_changed()
```
pgermishuys: Any particular reason you changed the description of these tests to "nothing_is_changed"? They did change though?

src/EventStore.Core.Tests/Services/Transport/Http/build_requested_url_should.cs
```diff
@@ -52,43 +52,43 @@ public void with_advertise_ip_and_http_port_set_both_host_and_port_is_changed()
...
+            var headers = new NameValueCollection { {"X-Forwarded-Proto", "http"}, {"X-Forwarded-Host","www.example.com"}, { "X-Forwarded-Port", "abc" }};
             var requestedUri =
                 HttpEntity.BuildRequestedUrl(inputUri, headers, null, 0);
 
             Assert.AreEqual(inputUri, requestedUri);
         }
 
         [Test]
-        public void with_proto_forward_header_only_scheme_is_changed()
+        public void with_proto_forward_header_only_nothing_is_changed()
```
pgermishuys: See comment above about "nothing_is_changed".
Here the scheme was changed from `http` to `https`.

pgermishuys:
Hey @shaan1337. Thanks again for getting these changes in so quickly. Mind reviewing the notes I made in the review? Did I potentially miss something?

src/EventStore.Core.Tests/Services/Transport/Http/build_requested_url_should.cs
shaan1337: Yes, there's a small change which is quite difficult to notice :) The port in the Assert.AreEqual was 4321 and I've changed it to 1234.

I've modified the behaviour of BuildRequestedUrl to change the base path only when the 3 headers: X-Forwarded-Scheme, X-Forwarded-Host and X-Forwarded-Port are all present. This is to ensure that the correct URL is being rebuilt. I saw some cases where reverse proxies are adding the Host header but not the Port or scheme which might not work properly. In any case, the 3 headers are actually no longer required since URLs will be rewritten as relative URLs. But the functionality is still supported in case someone wants it.

src/EventStore.Core.Tests/Services/Transport/Http/build_requested_url_should.cs
shaan1337: Here too, I've changed the scheme in Assert.AreEqual to "http://" instead of "https://"

src/EventStore.Core.Tests/Services/Transport/Http/build_requested_url_should.cs
pgermishuys: I see now, I should've looked this over properly. I do apologise.

PR #1296, opened 2017-05-05
Added gzip & deflate support for Accept-Encoding HTTP header

### LINKED ISSUE #965: Support gzip encoding
When using the Atom feed, with JSON content type, it would be great if EventStore natively supported gzip encoding (triggered by Accept-Encoding:gzip header). 

In my scenario, if I get all events from a store that has about 3M events, that currently translates to roughly 5GB of data (1.5KB/event). When doing it from localhost that's bearable, but if I were to get events over any kind of network it is pretty much required to put a proxy in front of GES that can do gzip encoding.

shaan1337:
Fixes #965 Also related to #1082.

Did a basic performance test with a 50MB event:
On my machine, increase in response time varies from 0.5s,99.9% compression (repeating a single character) to 2.1s,25% compression (random data), so compression takes around 10ms to 40ms per MB

### DIFF
--- a/src/EventStore.BufferManagement/BufferManager.cs
+++ b/src/EventStore.BufferManagement/BufferManager.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
+using EventStore.Common.Log;
 
 namespace EventStore.BufferManagement
 {
@@ -35,6 +36,7 @@ namespace EventStore.BufferManagement
 
         private readonly List<byte[]> _segments;
         private readonly object _creatingNewSegmentLock = new object();
+        private static readonly ILogger Log = LogManager.GetLoggerFor<BufferManager>();
 
         /// <summary>
         /// Gets the default buffer manager
@@ -161,7 +163,7 @@ namespace EventStore.BufferManagement
                     _buffers.Push(chunk);
                 }
 
-                Console.WriteLine("Segments count: {0}, buffers count: {1}, should be when full: {2}",
+                Log.Debug("Segments count: {0}, buffers count: {1}, should be when full: {2}",
                                   _segments.Count,
                                   _buffers.Count,
                                   _segments.Count * _segmentChunks);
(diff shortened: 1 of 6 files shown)


gregoryyoung:
I am just curious how you tested a 50mb event as that is larger than the largest allowed event? Were you testing the compressed version or?

shaan1337:
Well spotted. I changed the limits temporarily to do the test. There were two of them - a 4mb when posting in atom controller and a 16mb on event data

gregoryyoung:
:+1:

gregoryyoung:
https://github.com/EventStore/EventStore/pull/1296/files#diff-b093d7187d5fd93d11fd7be1c0dfbc6fR460 might benefit from reusing memory as opposed to new memorystreams each time. This is especially true without being able to set an initial size on the stream thus there will be many stream growth operations on larger data sizes.

shaan1337:
Thanks for the tip! I'll try to see if it's possible not to use streams at all. Gzip seems to be a simple algorithm based on what I've read, there must be some simple/fast implementation somewhere.

shaan1337:
After some research, seems that GzipStream from System.Io.Compression is still the preferred way to go. I didn't find any better implementation. I was wrong to say Gzip was simple :P The core idea is simple but lots of optimizations go into its implementation.

For the stream growth operations you mentioned, I'm not sure how memory can be reused, because I thought that an HttpEntityManager object was created for each HTTP request. Can you please clarify how it can be done?

Otherwise, we could also set the initial buffer size of the output MemoryStream to half (or a quarter) of the initial uncompressed response length . This way we're sure buffer will get resized at most 1 or 2 times(it gets doubled when allocating more memory). This would come with some memory tradeoff but since Gzip's compression ratio doesn't go below 20% on average not much memory would be wasted. I've tested this with a 50M string without much improvement. Log(50 million) to base 2 is around 25, I think it's too small to be noticeable.

gregoryyoung:
See: https://github.com/EventStore/EventStore/tree/release-v4.0.2/src/EventStore.BufferManagement

shaan1337:
Thanks 👍 will make use of the buffer manager

pgermishuys:
Hey @shaan1337, thanks once again!
Would you be so kind as to squash your PR into a logical set of commit(s) to keep the history clean?

Thank you so much!

We will likely review on monday unless someone picks this up over the weekend.

shaan1337:
@pgermishuys Commits squashed

PR #1302, opened 2017-05-12
Fixes checkpoint bug

### LINKED ISSUE #1301: Bad persistent subscription checkpoint resulting in skipped events
There's a scenario for a persistent subscription that can result in a checkpoint ahead of ack'd events:

- A client (or clients) is connected and processing events.
- The client is in the middle of processing an event when it gets disconnected.
- Internal to ES this results in all the outstanding events being moved to the retry queue.
- A checkpoint happens which will incorrectly be at the last sent event number due to the outstanding events set being empty.
- When the event store is restarted or fails over the retry queue is lost.
- The event store restarts and reads the bad checkpoint, skipping over the retried events.

See PersistentSubscription.TryMarkCheckpoint for where this bug is located.

We've probably been having bad checkpoints for a while, but this finally caught us this week when an Event Store node was restarted during another issue. In most cases without a restart the retry queue won't be lost so the bad checkpoint has no adverse impact.

The bug is around using int.MinValue as both a comparison and a sentinel value. Sorry for the lack of PR but I've not the time.

shaan1337:
Fixes #1301 
Wrong sentinel value was being used as described in issue.

From what I understood, correct behaviour is:
If there's at least one message in either _streamBuffer or _outstandingMessages:
"lowest" will contain the smallest event number among all messages
Otherwise:
"lowest" will contain the event number of "_lastKnownMessage"

### DIFF
--- a/src/EventStore.Core.Tests/Services/PersistentSubscription/OutstandingMessageCacheTests.cs
+++ b/src/EventStore.Core.Tests/Services/PersistentSubscription/OutstandingMessageCacheTests.cs
@@ -54,7 +54,7 @@ namespace EventStore.Core.Tests.Services.PersistentSubscription
             cache.StartMessage(new OutstandingMessage(id, null, Helper.BuildFakeEvent(id, "type", "name", 1), 0), DateTime.Now);
             cache.Remove(id);
             Assert.AreEqual(0, cache.Count);
-            Assert.AreEqual(long.MinValue, cache.GetLowestPosition());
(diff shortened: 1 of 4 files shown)


pgermishuys:
You are on a roll! Thanks for another PR @shaan1337!
I didn't see any test changes... this will need some reviewing.

shaan1337:
You're welcome :) Thanks for the feedback, I'll try to add in a test.

hayley-jean:
Looks good @shaan1337

PR #1317, opened 2017-05-29
Using null-conditional operators in several places. Corrects several …

### LINKED ISSUE #1315: Unsafe event raising could cause null reference exceptions
Hi.
This is a very low level issue but since I have been recently working on the code base for a PR I noticed there are a few places where raising events is done like:

`if (xxxx != null) xxxx(...)`

which is known to cause issues because subscribers can unsubscribe in between the null check and the invocation.

The solution is even easier now with the conditional operator:

`xxxx?.Invoke(...)`

I have spent a few hours replacing several if (yyy != null) yyy.F(), including the ones that cause the issue above.

I can submit the full PR if the team would take it, which would be a step towards using C# 6 features in your code, helping with readability, and ultimately, maintenability.

These are some of the places I have detected:

1. https://github.com/EventStore/EventStore/blob/release-v4.0.2/src/EventStore.ClientAPI/EventStoreCatchUpSubscription.cs#L250
1. https://github.com/EventStore/EventStore/blob/release-v4.0.2/src/EventStore.ClientAPI/EventStorePersistentSubscriptionBase.cs#L249-L252
1. https://github.com/EventStore/EventStore/blob/56c5680f51325cf0e99846da72ae313402c63c2d/src/EventStore.Transport.Http/AsyncStreamCopier.cs#L85-L86
1. https://github.com/EventStore/EventStore/blob/56c5680f51325cf0e99846da72ae313402c63c2d/src/EventStore.Transport.Http/AsyncStreamCopier.cs#L85-L86

This is not a full list. There are a few more.

rodolfograve:
Fixes #1315 
…instances of unsafe event handler invocation

This is intended to be a proper refactoring, i.e. no functional changes are introduced, except for those places where the invocation of a delegate was done in an unsafe way.

### DIFF
--- a/src/EventStore.ClusterNode/Program.cs
+++ b/src/EventStore.ClusterNode/Program.cs
@@ -30,7 +30,8 @@ namespace EventStore.ClusterNode
 
         public static void Main(string[] args)
         {
-            Console.CancelKeyPress += delegate {
+            Console.CancelKeyPress += delegate
+            {
                 Environment.Exit((int)ExitCode.Success);
             };
             var p = new Program();
@@ -51,13 +52,16 @@ namespace EventStore.ClusterNode
         {
             base.PreInit(options);
 
-            if (options.Db.StartsWith("~") && !options.Force){
+            if (options.Db.StartsWith("~") && !options.Force)
+            {
                 throw new ApplicationInitializationException("The given database path starts with a '~'. We don't expand '~'. You can use --force to override this error.");
             }
-            if (options.Log.StartsWith("~") && !options.Force){
+            if (options.Log.StartsWith("~") && !options.Force)
+            {
                 throw new ApplicationInitializationException("The given log path starts with a '~'. We don't expand '~'. You can use --force to override this error.");
             }
-            if (options.GossipSeed.Length > 1 && options.ClusterSize == 1){
+            if (options.GossipSeed.Length > 1 && options.ClusterSize == 1)
+            {
                 throw new ApplicationInitializationException("The given ClusterSize is set to 1 but GossipSeeds are multiple. We will never be able to sync up with this configuration.");
             }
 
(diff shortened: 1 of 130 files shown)


src/EventStore.ClusterNode/Program.cs
```diff
@@ -30,7 +30,8 @@ public class Program : ProgramBase<ClusterNodeOptions>
 
         public static void Main(string[] args)
```
pgermishuys: The only thing that changed in this file is the formatting and is likely confusing for git history. Would you mind terrible reverting this entire file?

src/EventStore.Core.Tests/Helpers/TestFixtureWithExistingEvents.cs
```diff
@@ -256,23 +256,23 @@ void IHandle<ClientMessage.ReadStreamEventsBackward>.Handle(ClientMessage.ReadSt
...
-                                    message.CorrelationId,
-                                    message.EventStreamId,
-                                    new EventLinkPair[0],
-                                    ReadStreamResult.Success,
-                                    nextEventNumber: -1,
-                                    lastEventNumber: list.Safe().Last().EventNumber,
-                                    isEndOfStream: true,// NOTE AN: don't know how to correctly determine this here
-                                    lastCommitPosition: _lastPosition));
-*/
+                    /*
```
pgermishuys: Let's just remove this entire commented out section of code.

src/EventStore.Core.Tests/Helpers/TestFixtureWithExistingEvents.cs
```diff
@@ -318,23 +318,23 @@ public void Handle(ClientMessage.ReadStreamEventsForward message)
...
```
pgermishuys: Remove the entire section of commented out code.

src/EventStore.Core.Tests/Http/HttpBehaviorSpecification.cs
```diff
@@ -31,7 +31,7 @@ public abstract class HttpBehaviorSpecification : SpecificationWithDirectoryPerT
         protected string _lastResponseBody;
         protected byte[] _lastResponseBytes;
         protected JsonException _lastJsonException;
-//MONOCHECK Does this work now?
```
pgermishuys: This file suffered the same fate as the other one mentioned above about being just reformatted. Would you mind terribly to just revert this particular file?

src/EventStore.ClusterNode/Program.cs
rodolfograve: There is one instance of where I applied the null-conditional operator (line 122). Do you still want me to revert all the formatting changes?

To clarify, I didn't go through the whole file changing all spaces and same-line brackets... I just hit Ctrl K + D which applies the formatting.

pgermishuys:
Both the Windows and Linux CIs are failing. I have attempted to go through the changes and see if I couldn't spot anything and made notes about those I did.

Since the PR is so large, it's difficult to see at a glance if any behaviour has changed, which is causing the tests to fail. Will try and have a closer look later.

Thank you once again for taking the time to go through the codebase and attempt to modernise it. Much appreciated @rodolfograve

PR #1329, opened 2017-06-04
Updating return types in doc comments

### LINKED ISSUE #1053: Codedoc wrong
https://github.com/EventStore/EventStore/blob/oss-v3.9.1/src/EventStore.ClientAPI/IEventStoreConnection.cs#L501-L502

Says it returns a value, but it doesn't.

jasonmitchell:
Improving consistency of return types in doc comments in `IEventStoreConnection` to fix #1053


gregoryyoung:
:+1:

Issue #1363, opened 2017-07-18
Memory leak and possible stackoverflowexception in ReadEventsInternalAsync


arnodenuijl:
The interaction between **ReadEventsInternalAsync** and **ReadEventsCallbackAsync** in **EventStoreCatchUpSubscription.cs** can cause memory pressure and can ultimately result in a stackoverflowexception.  (https://github.com/EventStore/EventStore/blob/release-v4.0.2/src/EventStore.ClientAPI/EventStoreCatchUpSubscription.cs#L541 & https://github.com/EventStore/EventStore/blob/release-v4.0.2/src/EventStore.ClientAPI/EventStoreCatchUpSubscription.cs#L548)

The two methods call each other which results in a big stack that takes up a decent amount of memory and, if continued long enough will result in a stackoverflowexception. See https://groups.google.com/d/msg/event-store/FrQEHpqCPhs/cR3zsB1_BQAJ for an example of a growing stack.

Easy way to reproduce is to just subscribe to a very large stream and wait. 
```
var es = EventStoreConnectionFactory.CreateConnection();
 es.SubscribeToStreamFrom("$ce-qbox", 0, CatchUpSubscriptionSettings.Default, (s, e) =>
 {

 });
```

I think there is an easy fix by replacing the mutually recursive function call with a while loop. I will make a pull request for it to show it.

gregoryyoung:
👍

arnodenuijl:
anything I can do for this item? I created a pull request with the changes. I've been running the client from my fork and it works great. Since we have streams with millions of events, it saves a huge amount of memory. 
I saw the pull request was out of date so i clicked the 'Update branch' button, but i'm not sure that is the way to go. Should i revert the merge and rebase and force push my changes so it's in one commit?!

pgermishuys:
No worries! This will get merged.

PR #1365, opened 2017-07-18
#1363 Fix

arnodenuijl:
ReadEventsInternalAsync and ReadEventsCallbackAsync called each other recursively. This resulted in stackoverflowexception when subscribing to a very large stream and a lot of memory pressure.

Main fix is to create a do..while in ReadEventsInternalAsync to facilitate continuing with the next slice if not done. This in stead of the recursive call.
To make this work ReadEventsCallbackAsync must return a bool indicating if we are done (shouldstop is true OR ProcessEventsAsync indicates that we are done)

As a side effect a couple of parameters could be dropped because they were only passed between the functions but not really used.

### DIFF
--- a/src/EventStore.ClientAPI/EventStoreCatchUpSubscription.cs
+++ b/src/EventStore.ClientAPI/EventStoreCatchUpSubscription.cs
@@ -536,33 +536,31 @@ namespace EventStore.ClientAPI
         /// <returns></returns>
         protected override Task ReadEventsTillAsync(IEventStoreConnection connection, bool resolveLinkTos, UserCredentials userCredentials,
             long? lastCommitPosition, long? lastEventNumber) =>
-            ReadEventsInternalAsync(connection, resolveLinkTos, userCredentials, lastCommitPosition, lastEventNumber);
-
+            ReadEventsInternalAsync(connection, resolveLinkTos, userCredentials, lastEventNumber);
+        
         private async Task ReadEventsInternalAsync(IEventStoreConnection connection, bool resolveLinkTos,
-                       UserCredentials userCredentials, long? lastCommitPosition, long? lastEventNumber)
+                       UserCredentials userCredentials, long? lastEventNumber)
         {
-            var slice = await connection.ReadStreamEventsForwardAsync(StreamId, _nextReadEventNumber, ReadBatchSize, resolveLinkTos, userCredentials).ConfigureAwait(false);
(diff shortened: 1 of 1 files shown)


src/EventStore.ClientAPI/EventStoreCatchUpSubscription.cs
```diff
@@ -536,34 +536,34 @@ public class EventStoreStreamCatchUpSubscription : EventStoreCatchUpSubscription
...
+			    var slice = await connection
+				    .ReadStreamEventsForwardAsync(StreamId, _nextReadEventNumber, ReadBatchSize, resolveLinkTos, userCredentials)
+				    .ConfigureAwait(false);
+				shouldStopOrDone = await ReadEventsCallbackAsync(slice, lastEventNumber).ConfigureAwait(false);
+		    } while (!shouldStopOrDone);
+	    }
+
+	    private async Task<bool> ReadEventsCallbackAsync(StreamEventsSlice slice, long? lastEventNumber)
+	    {
+		    bool shouldStopOrDone = ShouldStop || await ProcessEventsAsync(lastEventNumber, slice).ConfigureAwait(false);
```
arnodenuijl: Here is one small 'functional' change. If at this point ShouldStop is true we don't call ProcessEventsAsync. Before it would first call ProcessEventsAsync and then check ShouldStop.
I think this is better because if you indicated that you want to stop, there is no need to process anymore events.

pgermishuys:
Hi @arnodenuijl,
The formatting changes makes it difficult to spot the changes :)

pgermishuys:
@arnodenuijl would it be possible to not have the formatting changes included? Sorry if it's a little picky, it's just easier to see what's changed in the future when someone else comes along?

arnodenuijl:
No problem. I looked at the changes and fixed some tabs vs spaces. But I'm afraid this is as good as I can get it. There are some indentation differences that are caused by the do..while loop and the changes in the 'if' structure. So most of the formatting changes that are left are intentional. 

I think the best way to review the change is to look at the version before and the version after and reason about both behaviours.  Before it was a mutual recursive function, now it is a do..while loop.  It does the same, but in a different way. But that change in behaviour doesn't show well in a text diff. 

If you have any suggestions how to further clarify the changes i'm happy to give it a shot.

pgermishuys:
@arnodenuijl, no worries! thank you so much for the effort!

PR #1388, opened 2017-08-16
Revert "Scavenge away LinkTo events that point to events that no longer exist"

hayley-jean:
Reverts EventStore/EventStore#1286

This can cause streams to become unreadable, especially when "holes" are left in linkTo streams when some source events are deleted, but others aren't.
Requires more work and testing.


megakid:
@hayley-jean What were the issues with this? Can we contribute to a fix? Or is there any further progress on this?

hayley-jean:
Hi @megakid,

We haven't done any further work on this yet.

The issues this PR caused was mainly due to events being removed in the middle of a stream.
If you had two streams, A and B, being linked into stream C in an interleaving manner, you could end up with a resulting stream like this:

```
--- Stream C
0: 0@A
1: 0@B
2: 1@A
3: 1@B
4: 2@A
5: 2@B
```
If you were to delete and then scavenge away stream A, Events 0, 2, and 4 would be removed from stream C:
```
--- Stream C
0: ---
1: 0@B
2: ---
3: 1@B
4: ---
5: 2@B
```
After this, if you were to read stream C, you would get events 1, 3, 5. This results in a corrupt stream error, because Event Store expects events to follow on from each other sequentially.

To fix this, we would need to prevent the scavenge from removing events in the middle of a stream. Perhaps by checking whether a previous event exists in the stream. In the above case, we would basically want to remove event 0, but keep events 2 and 4 because there's still events before and after them.

megakid:
Thanks for the reply @hayley-jean,

If scavenging was prevented from removing events in the middle of streams, this would significantly reduce the benefits of this PR when applied to streams such as $ce-XXX and other cross stream projection linkTo streams?

How embedded is the assumption that stream versions are strictly sequential (1, 2, 3...) vs logically contiguous (4, 10, 12...)? Examples added as not entirely sure my vocabulary is correct here.

Could we not remove that assumption to get more out of this?

megakid:
We have discussed this locally and we think there's still quite a lot of value in scavenging up to the first item in the linkTo stream (e.g. $ce-) - this might be quite common as if people apply the same MaxAge value to all streams being linked to (e.g. `mycategory-A` (maxAge 30d) + `mycategory-B` - (maxAge 30d) => `$ce-mycategory`) would mean that during a scavenge, even using the restrictions above, it would be able to reclaim all of the `$ce-mycategory` links up until 30 days ago. 

However, @lscpike had the thought that this could be best done outside of the scavenge process as it requires detailed knowledge of the projection system (and persistent subscription system + checkpointing etc).  If scavenging was done on a $ce- stream up to version 50 (following the aforementioned rules), what would happen to a persistent subscription on that stream that was currently < version 50.  

A possible solution is a background/out-of-process action that inspects each stream containing dead links and any associated persistent subscriptions (and any other projection paraphernalia that needs to be checked) then modify the stream's metadata truncation position.  This would then line up the stream to be scavenged to a safe point next time a scavenge is run.  We would also have to consider how new persistent subscriptions (starting from earliest version) would handle the truncated stream (vs currently they would just be given lots of resolvable links).

megakid:
Hi @hayley-jean - I've created a third-party tool that uses the HTTP endpoints to set the `$tb` metadata entry on a given linkTo stream such as `$ce-...` - I have only done basic testing on this locally but give you some idea where I'm aiming.  https://github.com/megakid/EventStoreArchiver/blob/master/DeadLinkCleaner/Program.cs

PR #1410, opened 2017-09-04
Race condition causing projections to be stuck sometimes on startup

shaan1337:


shaan1337:
**Steps to reproduce:**

Use the file below where StartCoreSteps.Run() has been delayed by 10 seconds:
[ProjectionCoreServiceCommandReader.cs.zip](https://github.com/EventStore/EventStore/files/1277069/ProjectionCoreServiceCommandReader.cs.zip)

Result: Projections get stuck at "Preparing" stage (it could technically speaking get stuck at other stages too):
screenshot from 2017-09-05 14-20-15

PR #1414, opened 2017-09-05
Make ProjectionsManager work with Hostname

dersia:
As of now ProjectionsManager and ProjectionsClient only work with IPs even though it is based on Http because it is based on the IPEndPoint-class.

To allow Dns (hostname) next to IP I replaced the IPEnpointExtensions by EndpointExtensions, which uses pattern-matching to create the HttpUrl from IP or Hostname.

For now except for ProjectionsManager and ProjectionsClient, all other uses of ToHttpUrl will still rely on IPEndPoint. Eventually this should be changed in the future, although there are places that resolve dns to multiple IP addresses to use randomly out of that cluster.


dersia:
hmm looks like compiler is not up-to-date for C# 7.0 pattern matching

pgermishuys:
@dersia thanks for the PR. Would you mind not using the pattern matching for now as this requires an upgrade of Mono to 5.0 which hasn't been done yet.

Thank you once again for the PR!

dersia:
Removed pattern matching

pgermishuys:
Thanks @dersia, seems like the rebase didn't work as the PR has couple of extra commits.

dersia:
@pgermishuys oh yeah you're right, sorry didn't see that. I'll update it

pgermishuys:
@dersia no problem. Thanks again for the work on the PR.

dersia:
updated

pgermishuys:
@dersia apologise for being a pain, but would you mind squashing the commits into either a single commit or a logical set of commits.

Apologies for being pedantic.

dersia:
done :)

PR #1417, opened 2017-09-09
Prevent perfcounters from stopping MonitoringService

niklasi:
On Windows you can get problems with corrupt peformance counters and
this can prevent the MonitoringService from saving stats.
Some calls to the perfcounters is wrapped in a try-catch with a retry,
but not all of them.

Fix: Handle the calls to the gc-stats perfcounters the same way as the
rest of the perfcounters.


pgermishuys:
Looks good to me, waiting for the tests to pass.

PR #1430, opened 2017-09-15
Ensure c++ compiler before building V8

mjgpy3:
I needed to run `sudo apt-get install g++` for `scripts/build-js1/build-js1-linux.sh` to successfully run.

This is perhaps a little debian-centric (using `apt-get`) I'm open to suggestions.

This helped: https://groups.google.com/forum/#!topic/v8-users/Tt56ucRG3Tk

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,7 @@ The list of precompiled projections libraries can be found in `src/libs/x64`
 ### Linux
 **Prerequisites**
 - git
+- a c++ compiler (e.g `sudo apt-get install g++`)
 
 The scripts attempts to do it's best to have no dependencies from the user's point of view. If you do encounter any issues, please don't hesitate to raise an issue.
 


gregoryyoung:
Are you making changes to js1? Usually its included as binaries so it doesn't need to be built

mjgpy3:
@gregoryyoung I tried to run just `./build.sh 4.0.0 debug` and it failed asking if I had built the projections.

mjgpy3:
@gregoryyoung I was not intending to make changes to js1

pgermishuys:
@mjgpy3 there's a check that is performed that ensures that there is a libjs that can be used based on your system. You mentioned previously you were on mint?

You could likely just just force the use of the ubuntu projections native lib (libjs1).

Would you mind trying the following?
```
./build.sh 4.0.0 debug ubuntu-14.04
```

pgermishuys:
Would like to investigate this more before we do anything else.

mjgpy3:
@pgermishuys that command worked fine (I had to use `ubuntu-16.04` since it's the parent of LM 18.2).

Thanks for the help!

PR #1434, opened 2017-09-21
Fix slave-to-master forwarding issues when a header has the wrong formatting

shaan1337:
If there's a wrongly or slightly wrongly formatted header, the request will be rejected by the slave when trying to forward a request (results in a 500 internal error)

**Resolution:** 
If **System.FormatException** is thrown, try to add the header without validating it.
The request will handled and validated by the master anyway, so it doesn't affect the end result.

**Test examples:**
curl -H "Accept: application/json;" -v localhost:2113/users/admin
(The semicolon at the end is invalid)

curl -H "Via: abcd" -v localhost:2113/users/admin
(Correct format: "Via: 1.1 abcd")

### DIFF
--- a/src/EventStore.Core/Services/HttpSendService.cs
+++ b/src/EventStore.Core/Services/HttpSendService.cs
@@ -182,25 +182,30 @@ namespace EventStore.Core.Services
             // Copy unrestricted headers (including cookies, if any)
             foreach (var headerKey in srcReq.Headers.AllKeys)
             {
-                switch (headerKey.ToLower())
-                {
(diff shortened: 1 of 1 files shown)


pgermishuys:
The bit of code where the protected headers are being set needs to be reviewed. Prior to the PR (https://github.com/EventStore/EventStore/pull/897) where we moved over to `HttpClient` we used `HttpWebRequest` which did not allow the protected headers to be set other than through the properties.

With the PR, we moved over to `HttpRequestMessage` which allows the headers to be set through the headers collection.

PR #1435, opened 2017-09-21
Allow changing of custom metadata properties in the client StreamMetadataBuilder

lscpike:
Currently you can't modify a custom property if already set. Also added ways to enumerate and remove custom properties.


gregoryyoung:
Seems reasonable

PR #1436, opened 2017-09-22
added traverse in scavenge to count the new chunk size in memory

### LINKED ISSUE #1399: Scavange is not resuming after node restart
Scavenge is quite a long operation and cluster nodes may incidentally shut down during it. When the restart happens scavenge needs to be started again and it starts from the beginning. So eventually on large databases it becomes almost impossible to scavenge the database entirely. 

It would make more sense if the scavenge would resume after node goes up.

mpielikis:
Faster rescavange and more accurate inmemory precalculation of chunk size. Fixes #1399

scavenge stops on shutting down the node
image

and when restarted

- current speed when dealing with "no chunks scavenged"
image

- optimized
image


mpielikis:
@pgermishuys could you comment on why this PR was closed?

pgermishuys:
PR wasn't explicitly closed by me but because the 4.0.3 branch was deleted. Would you mind opening the PR against the 4.0.4 branch please?

PR #1438, opened 2017-09-22
Fix bug when Replaying parked Messages with an empty park stream

shaan1337:
If the park stream is empty and we manually 'Replay all parked messages', the flag PersistentSubscriptionState.ReplayingParkedMessages in _state variable is wrongly updated.
This results in inability to replay messages (until EventStore is restarted)

**Reproduction steps:**
[persistentsub.zip](https://github.com/EventStore/EventStore/files/1324408/persistentsub.zip)

Run the above application, which simulates the following:
1. Create a persistent subscription
2. 'Replay all parked messages' from the web ui
3. Send an event, and NAK it with park action.
4. Try to 'Replay all parked messages' again from the web ui. No event is received by the subscription.


gregoryyoung:
:+1:

PR #1448, opened 2017-09-29
Prevent reads of events that have not been replicated

hayley-jean:
Only update the index once a commit has been replicated in a cluster

Introduces ReplicationCheckpoint to track the position of the last committed event.

Introduces IndexCommitterService to replace the IndexCommitter in StorageChaser.
This tracks pending transactions from the StorageChaser and CommitAcks from the cluster.
When enough commit acks have been received, the index and the replication checkpoint is updated.
If a CommitAck is received on a non-master node, the event is immediately considered replicated.
If a master node changes to a non-master, all pending commits are considered replicated.

Update TwoPhaseRequestManagers to listen for CommitReplicated messages rather than CommitAck messages.
This means that a write is only complete once the commit has been replicated and the index updated.

Update AllReader to only read up to the ReplicationCheckpoint

Also refactored tests using `TFChunkDbConfig` to use a constructor from a helper method to make future constructor updates easier.


hayley-jean:
@gregoryyoung, @pgermishuys Would you mind having a look at this and seeing if there's any potential issues or warnings with this approach?

PR #1449, opened 2017-10-02
added traverse in scavenge to count the new chunk size in memory

### LINKED ISSUE #1399: Scavange is not resuming after node restart

mpielikis:
Faster rescavange and more accurate inmemory precalculation of chunk size. Fixes #1399

scavenge stops on shutting down the node
image

and when restarted

- current speed when dealing with "no chunks scavenged"
image

- optimized
image

### DIFF
--- a/src/EventStore.Core/TransactionLog/Chunks/TFChunkScavenger.cs
+++ b/src/EventStore.Core/TransactionLog/Chunks/TFChunkScavenger.cs
@@ -174,45 +174,48 @@ namespace EventStore.Core.TransactionLog.Chunks
                 foreach (var oldChunk in oldChunks)
                 {
                     TraverseChunk(oldChunk,
-                                  prepare => { /* NOOP */ },
-                                  commit =>
+                                  (prepare, _) => { /* NOOP */ },
+                                  (commit, _) =>
                                   {
                                       if (commit.TransactionPosition >= chunkStartPos)
                                           commits.Add(commit.TransactionPosition, new CommitInfo(commit));
                                   },
-                                  system => { /* NOOP */ });
+                                  (system, _) => { /* NOOP */ });
                 }
 
-                var positionMapping = new List<PosMap>();
+                int newSize = 0;
+                int positionMapCount = 0;
+
                 foreach (var oldChunk in oldChunks)
                 {
                     TraverseChunk(oldChunk,
-                                  prepare =>
+                                  (prepare, len) =>
                                   {
                                       if (ShouldKeepPrepare(prepare, commits, chunkStartPos, chunkEndPos))
-                                          positionMapping.Add(WriteRecord(newChunk, prepare));
+                                      {
+                                          newSize += len + 8;
+                                          positionMapCount++;
+                                      }
                                   },
-                                  commit =>
+                                  (commit, len) =>
                                   {
                                       if (ShouldKeepCommit(commit, commits))
-                                          positionMapping.Add(WriteRecord(newChunk, commit));
+                                      {
+                                          newSize += len + 8;
+                                          positionMapCount++;
+                                      }
                                   },
-                                  // we always keep system log records for now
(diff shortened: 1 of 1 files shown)


src/EventStore.Core/TransactionLog/Chunks/TFChunkScavenger.cs
```diff
@@ -174,45 +174,48 @@ private bool ScavengeChunks(bool alwaysKeepScavenged, IList<TFChunk.TFChunk> old
...
                                   {
                                       if (commit.TransactionPosition >= chunkStartPos)
                                           commits.Add(commit.TransactionPosition, new CommitInfo(commit));
                                   },
-                                  system => { /* NOOP */ });
+                                  (system, _) => { /* NOOP */ });
                 }
 
-                var positionMapping = new List<PosMap>();
+                int newSize = 0;
```
shaan1337: Can you please change this to a `long` instead of an `int`?

src/EventStore.Core/TransactionLog/Chunks/TFChunkScavenger.cs
```diff
@@ -174,45 +174,48 @@ private bool ScavengeChunks(bool alwaysKeepScavenged, IList<TFChunk.TFChunk> old
...
                 foreach (var oldChunk in oldChunks)
                 {
                     TraverseChunk(oldChunk,
-                                  prepare =>
+                                  (prepare, len) =>
                                   {
                                       if (ShouldKeepPrepare(prepare, commits, chunkStartPos, chunkEndPos))
-                                          positionMapping.Add(WriteRecord(newChunk, prepare));
+                                      {
+                                          newSize += len + 8;
```
shaan1337: Please change this line to: `len + 2*sizeof(int)` for more clarity

src/EventStore.Core/TransactionLog/Chunks/TFChunkScavenger.cs
```diff
@@ -174,45 +174,48 @@ private bool ScavengeChunks(bool alwaysKeepScavenged, IList<TFChunk.TFChunk> old
...
+                                          positionMapCount++;
+                                      }
                                   },
-                                  commit =>
+                                  (commit, len) =>
                                   {
                                       if (ShouldKeepCommit(commit, commits))
-                                          positionMapping.Add(WriteRecord(newChunk, commit));
+                                      {
+                                          newSize += len + 8;
```
shaan1337: Please change this line to: `len + 2*sizeof(int)` for more clarity

src/EventStore.Core/TransactionLog/Chunks/TFChunkScavenger.cs
```diff
@@ -174,45 +174,48 @@ private bool ScavengeChunks(bool alwaysKeepScavenged, IList<TFChunk.TFChunk> old
...
+                                          positionMapCount++;
+                                      }
                                   },
-                                  // we always keep system log records for now
-                                  system => positionMapping.Add(WriteRecord(newChunk, system)));
+                                  (system, len) =>
+                                  {
+                                      newSize += len + 8;
```
shaan1337: Please change this line to: `len + 2*sizeof(int)` for more clarity

shaan1337:
Hello @mpielikis 
Sorry for getting back to you after so long.

Thanks very much for your PR! It looks good, can you please make the changes suggested above? We'll then merge it.

Thanks again!

shaan1337:
Thanks for the changes @mpielikis !
Just a last request, if that's not too much asking: can you please squash your commits into a single commit?

We try to keep the log clean as much as possible.

PR #1455, opened 2017-10-07
fixed system projection query update bug

abdomohamed:
Hey Team, 

I've create a PR for this [issue](https://github.com/EventStore/EventStore/issues/1451

### DIFF
--- a/src/EventStore.ClientAPI/Projections/ProjectionsClient.cs
+++ b/src/EventStore.ClientAPI/Projections/ProjectionsClient.cs
@@ -142,9 +142,10 @@ namespace EventStore.ClientAPI.Projections
             return SendGet(endPoint.ToHttpUrl(httpSchema, "/projection/{0}/query", name), userCredentials, HttpStatusCode.OK);
         }
 
-        public Task UpdateQuery(EndPoint endPoint, string name, string query, UserCredentials userCredentials = null, string httpSchema = EndpointExtensions.HTTP_SCHEMA)
+        public Task UpdateQuery(EndPoint endPoint, string name, string query, UserCredentials userCredentials = null, string httpSchema = EndpointExtensions.HTTP_SCHEMA, bool emitEnabled = true)
         {
-            return SendPut(endPoint.ToHttpUrl(httpSchema, "/projection/{0}/query?type=JS", name), query, userCredentials, HttpStatusCode.OK);
+            var emit = emitEnabled ? "yes" : "no";
+            return SendPut(endPoint.ToHttpUrl(httpSchema, "/projection/{0}/query?emit={1}", name, emit), query, userCredentials, HttpStatusCode.OK);
         }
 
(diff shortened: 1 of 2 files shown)


src/EventStore.ClientAPI/Projections/ProjectionsClient.cs
```diff
@@ -142,9 +142,10 @@ public Task<string> GetQuery(EndPoint endPoint, string name, UserCredentials use
             return SendGet(endPoint.ToHttpUrl(httpSchema, "/projection/{0}/query", name), userCredentials, HttpStatusCode.OK);
         }
 
-        public Task UpdateQuery(EndPoint endPoint, string name, string query, UserCredentials userCredentials = null, string httpSchema = EndpointExtensions.HTTP_SCHEMA)
+        public Task UpdateQuery(EndPoint endPoint, string name, string query, UserCredentials userCredentials = null, string httpSchema = EndpointExtensions.HTTP_SCHEMA, bool emitEnabled = true)
```
hayley-jean: Defaulting the emit enabled to true could have unintended consequences.
The expected behaviour is if it's not set, the value will not be changed.
However, in this case, there is no way to _not_ set emit enabled.

Perhaps it would be better to create an override method that accepts the emitEnabled parameter, and only if that override is used will it be sent as part of the projection update.

hayley-jean:
Hi @abdomohamed, thanks for the PR! Apologies for taking so long to review it.

There is just one change that we'd like to see, I've added it as a review.

src/EventStore.ClientAPI/Projections/ProjectionsClient.cs
abdomohamed: OK, so you wanna make this method accepts bool?, and if not set it will be execlude emit paramater from Url, is that right ?

abdomohamed: OK, so you wanna make this method accepts bool?, and if not set it will be execlude emit paramater from Url, is that right ?

hayley-jean: Sounds good 👍

abdomohamed: @hayley-jean applied PR comments. please have a look. Thanks.

abdomohamed:
@hayley-jean I ran tests localy and all good, It seems "continuous-integration/appveyor/pr" failed because of something related to  HttpAsyncServer couldn't start. Is that right ? How I can run this check again ?

hayley-jean:
Thanks @abdomohamed, I kicked off the tests again and they're passing.

Would you mind squashing your commits? We do our best to keep the history clean

Faq:
@hayley-jean merge :)

abdomohamed:
@hayley-jean I've created another PR, becuase this one has now a lot of commits, happened after mine so it will be difficult to squash now.

abdomohamed:
**New PR:** https://github.com/EventStore/EventStore/pull/1472

PR #1462, opened 2017-10-11
v8: Add build for MacOS X 10.13

jen20:


jen20:
This isn't needed, a symbolic link to 10.12 libraries should be fine.

PR #1465, opened 2017-10-12
Stop reading the order stream once the checkpoint has passed.

lscpike:
Looking at the Windows file caching I noticed one of our event store clusters was always loading the entire database into the file cache. This turned out to be a multi-stream projection causing the projection system to read an entire stream all the way backwards during initialization. In particular the "order" stream. This takes about 2 hours and blows out the file cache...

It looks like the **break** was attempting to break out the **switch** but was actually breaking out the **foreach**.

Test included.


hayley-jean:
Thanks! 👍

PR #1485, opened 2017-11-08
Allow the user to skip verification of index integrity on startup

shaan1337:
While loading midpoints from PTable index files during start up, the whole PTable file is read to also verify its integrity using an MD5 checksum. This can take a long time for very large databases/indexes.

This change allows the user to skip this verification by specifying a command line argument:
`EventStore.ClusterNode.exe --skip-index-verify`

Specifying that parameter will not read the whole file but will only read the midpoints by seeking directly to the midpoint locations in the file.


gregoryyoung:
It might be worth pairing this change with a lazy midpoint load which would prevent the scan entirely (but the reads would be slower until warmed up). In my testing just disabling the check had little to no effect on the overall time to load.

PR #1488, opened 2017-11-13
Ability to disable index merges during busy periods.

megakid:
This is mainly a discussion point but if agreed, this is almost ready to go.
A few weeks ago we had an outage during a peak load period due to an index merge that was occurring on all the nodes within that particular cluster.  We saw SLOW MSG BUS messages stacking up in the logs and had no way to prevent the merge from occurring or to stop the merge once in progress.  

We feel the probability of merges occurring during peak periods is naturally higher (in terms of when they are likely to occur) due to the fact that the index is under more pressure due to the increased amount of events being added.  We discussed this internally and decided that the biggest win would be to suspend any index merges until after the peak period has passed.

I have exposed an admin POST endpoint to enable and disable the on-the-fly index merging.  The proposed setup we have discussed is that index merging is to be disabled during peak hours (we see this as 4 - 5 hours) and then re-enabled afterwards.  The toggle back to enabled would likely trigger a full merge soon afterwards.

This is a first effort that this.  Comments, suggestions, queries all welcome.

### DIFF
--- a/src/EventStore.ClusterNode/ClusterNodeOptions.cs
+++ b/src/EventStore.ClusterNode/ClusterNodeOptions.cs
@@ -201,6 +201,8 @@ namespace EventStore.ClusterNode
 
         [ArgDescription(Opts.IndexCacheDepthDescr, Opts.DbGroup)]
         public int IndexCacheDepth { get; set; }
+        [ArgDescription(Opts.DisableIndexMergingDescr, Opts.DbGroup)]
+        public bool DisableIndexMerging { get; set; }
 
         [ArgDescription(Opts.GossipIntervalMsDescr, Opts.ClusterGroup)]
         public int GossipIntervalMs { get; set; }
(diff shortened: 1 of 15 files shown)


jen20:
This looks like a reasonable approach to relieving some disk pressure, especially on larger databases, at the expense of index lookup time - which may be an acceptable trade off. @gregoryyoung any thoughts here?

gregoryyoung:
How is this related?

https://github.com/EventStore/EventStore/pull/1488/files#diff-3765b6dad4f260935b2cde1764a5282eR82

megakid:
@gregoryyoung I ended up with 3 endpoints all auth'ing on the same stuff - can revert that if necessary.

megakid:
Also a potential enhancement, instead of exposing a simple `bool`.  I could configure index merging levels `int` e.g. only merge the first `n` levels, `0` would mean merging totally disabled (same as `false`).  This would enable some compromise by enabling some levels to be merged whilst not merging two 16 or 32Gb files - for example.  Too complex?  Thoughts?

gregoryyoung:
(Taking the role of devil's advocate)

Might it be better to make the merge run slower by intention? Right now there is no throttling on it except what the OS pushes back.

megakid:
We did discuss automatically determining when to disable index merging but wasn't sure what metric to use (and would probably vary wildly depending on hardware) - how would slowing down index merges look?  `Thread.Sleep()` ¯\\_(ツ)_/¯ ?

gregoryyoung:
Thread.Sleep could work (as silly as it sounds). Where is your bottleneck IO/CPU (I am guessing IO)

megakid:
Yes, I/O

hayley-jean:
Hi @megakid 

Thanks

PR #1501, opened 2017-11-29
Failing test for index merge and upgrade.

lscpike:
Only a failing test. Note this affects all versions from 3.9.4.

This corrupted our index by dropping most of the events in the old v1 PTable during an index merge. The index still validated. We didn't notice the issue until a restart of the cluster. On restart the usual read of users and projections failed and resulted in a very bad time.

### DIFF
--- a/src/EventStore.Core.Tests/EventStore.Core.Tests.csproj
+++ b/src/EventStore.Core.Tests/EventStore.Core.Tests.csproj
@@ -275,6 +275,7 @@
     <Compile Include="Index\FakeIndexReader.cs" />
     <Compile Include="Index\IndexV1\when_merging_ptables.cs" />
     <Compile Include="Index\IndexV2\table_index_hash_collision_when_upgrading_to_64bit.cs" />
+    <Compile Include="Index\IndexV2\table_index_when_merging_upgrading_to_64bit_if_entry_doesnt_exist_drops_entry_and_carries_on.cs" />
     <Compile Include="Index\MemTableTests.cs" />
     <Compile Include="Index\IndexV2\adding_four_items_to_empty_index_map_with_four_tables_per_level_causes_merge.cs" />
     <Compile Include="Index\IndexV2\adding_four_items_to_empty_index_map_with_two_tables_per_level_causes_double_merge.cs" />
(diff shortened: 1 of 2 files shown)


lscpike:
Bug is in this method: https://github.com/EventStore/EventStore/blob/980bb6d769ef18e9042fa08ab2f17305958e9271/src/EventStore.Core/Index/PTableConstruction.cs#L351

When all entries for a hash are scavenged an empty list is generated. The empty list causes an early termination of the scan.

shaan1337:
Hello @lscpike 

Thanks very much for the details and your PR.
I've tested it, there seems to be a small bug in the test which is not triggering the PTable merge and not having the intended consequences: `maxSizeForMemory` is set to 5 instead of 2 [here](https://github.com/EventStore/EventStore/pull/1501/files#diff-37f35786de6c95c30ffd3505d29d4742R39
)

Thus, the index entries from the first `TableIndex` you've constructed are lost and the 
second `TableIndex` you built contains only [these two index entries](https://github.com/EventStore/EventStore/pull/1501/files#diff-37f35786de6c95c30ffd3505d29d4742R56
) and the second test is failing

I've reviewed the code you pointed to [here](https://github.com/EventStore/EventStore/blob/980bb6d769ef18e9042fa08ab2f17305958e9271/src/EventStore.Core/Index/PTableConstruction.cs#L351)
`current` and `hash` are assigned independently of whether the item is being added to the list or not. So, after the first iteration, `hash` will be assigned to the first stream hash. Then after that it'll stay the same and for the remaining iterations it'll be compared against `current.Stream`. So, the code should not terminate early even if the list is empty. Please let me know if that clears out your doubts or if we're missing something here.

Can you please provide us with more details (and the logs if available) about the issue you encountered?

lscpike:
You're right that test didn't show the issue. I've updated it. 

I also think the loop in ReadUntilDifferentHash is not doing as expected too. It includes the first entry in the next hash group. Hence my test has to have 2 deleted entries in order to trigger the bug.

It also means the attempt to reorder collisions after the hash upgrade could be missing the first entry.

e.g. (pretending 4 to 8 bit hash upgrade)

hash before, hash after, version
A, AA, 0
B, BB, 0
B, BA, 0

Will end up AA, BB, BA instead of the correct AA, BA, BB. I'm aware the index is reverse ordered so translate accordingly.

shaan1337:
Thanks very much for the details @lscpike, that makes complete sense :+1: 

Enumerating the empty list in `MoveNext()` is returning an empty `IndexEntry` (stream: 0, version: 0, position: 0) This causes all the next index entries from that PTable to be skipped since it'll be always less than the other PTable's enumerator.

lscpike:
It’s more to do with the fact MoveNext returns false on an empty list.

shaan1337:
Hi @lscpike , I've submitted a fix here: #1504 
It now passes your failing test + the case where the first item from the next stream group is added to the previous stream group. Please have a look at it to see if we're on the same page.

Thanks again so much :+1:

Issue #1479, opened 2017-11-06
HTTP: Allow persistent subscription to be created with timeout of 0


hayley-jean:
Related to this issue in the UI repo: https://github.com/EventStore/EventStore.UI/issues/141

It appears that creating a persistent subscription with 0 set on the message timeout, the subscription will be created with the default of 10 seconds.
As 0 is the value set for disabling message timeouts, it is possible that this would be set to 0 on purpose, and therefore we shouldn't change it.

PR #1519, opened 2017-12-13
Defaulting MessageTimeoutMilliseconds to int.MaxValue and showing 0 in the UI

kavir-bheeroo:
## Issue
- #1479 

## Comments
- Sending 0 as timeout from the UI is meant to disable any timeouts but instead a default timeout value of 10 seconds was being set on every such request.
- To remedy this, now every request with a timeout value of 0 is defaulted to int.MaxValue (2147483647) and on encountering such value, the UI will display 0 on its side.
- This [line](https://github.com/kavir-bheeroo/EventStore/blob/release-v4.0.4/src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs#L812) affects several other scenarios so it needs to be confirmed whether the mapping from MaxValue to 0 can be done here.
- Any comments, queries and suggestions are most welcome in order to get this done in a better way.


hayley-jean:
Thanks for the PR @kavir-bheeroo !

We've had a bit of a discussion around this change, and it's become apparent that this will cause some backwards compatibility issues, especially with creating a subscription over the HTTP API.

We have often seen people create a subscription just by sending an empty object to the subscriptions url, and expecting the subscription to be created with all default settings. If we now accept `0` as not timing out rather than the 10 second default, we may break functionality for people creating subscriptions this way.

Perhaps a better setting for `DontTimeout` is `-1`, as this would need to be explicitly set, rather than being a default value?
We can still accept 0 from the .NET client, but use -1 from the admin UI, and update the .NET client in the next major version.

Any thoughts @shaan1337  or @gregoryyoung?

shaan1337:
Hi @kavir-bheeroo @hayley-jean 

I've done a few tests, this line seems to be causing the problem :
https://github.com/EventStore/EventStore/blob/release-v4.0.4/src/EventStore.Core/Services/Transport/Http/Controllers/PersistentSubscriptionController.cs#L264

From the UI, the message timeout value is being sent as a string in the json request. It seems that  `"0"` is wrongly being parsed by the `JsonCodec`. Other non zero string values (e.g. `"123"`) seem to work fine though

kavir-bheeroo:
Hi @shaan1337 

Indeed the problem is where you pointed out and the reason why the serializer is incorrectly parsing a "0" value is because the `JsonSerializerSettings` being used have `DefaultValueHandling = DefaultValueHandling.Ignore` - https://github.com/EventStore/EventStore/blob/release-v4.0.4/src/EventStore.Transport.Http/Codecs/JsonCodec.cs#L21.

As the default value for an `int` is 0, it is ignored and the value set in the constructor `MessageTimeoutMilliseconds = 10000;` will then remain - https://github.com/EventStore/EventStore/blob/release-v4.0.4/src/EventStore.Core/Services/Transport/Http/Controllers/PersistentSubscriptionController.cs#L728.

kavir-bheeroo:
In my opinion, the solution using -1 instead of 0 for no timeouts as suggested by @hayley-jean seems a good candidate to solve this problem without causing any regressions.

shaan1337:
Hi @kavir-bheeroo 
Thanks, you're right, `DefaultValueHandling.Ignore` seems to be causing that trouble!

The same issue also happens with other integer values like `Max Retry Count` when set to 0 - it's set back to the default value.

Another solution is that we could set the "DefaultValue" attribute as mentioned in the link you sent for all properties of `SubscriptionConfigData` to the same values set in the constructor. But we would still need to keep the constructor: https://stackoverflow.com/a/2329878
What do you think? @kavir-bheeroo  @hayley-jean 

I've checked through the code base and haven't seen other places where we've used a custom constructor when deserializing to an object, so this seems the only place affected.

hayley-jean:
Hi @kavir-bheeroo 

Thanks

PR #1527, opened 2017-12-20
Set MaxDiscoverAttempts for KeepReconnecting

### LINKED ISSUE #1478: Set MaxDiscoverAttempts when using KeepReconnecting
`MaxDiscoverAttempts` behaves the same way as `MaxReconnections`, but only when cluster discovery is being used.
It would be least surprising if setting `KeepReconnecting` to true also sets `MaxDiscoverAttempts` to a similar value

Currently, `MaxDiscoverAttempts` doesn't have a value that matches `KeepReconnecting` (-1), so this would need to be added.

accraze:
Currently, `MaxDiscoverAttempts` doesn't have a value that matches `KeepReconnecting` (-1)
This sets the values to match.

fixes #1478

### DIFF
--- a/src/EventStore.ClientAPI/ConnectionSettingsBuilder.cs
+++ b/src/EventStore.ClientAPI/ConnectionSettingsBuilder.cs
@@ -180,6 +180,7 @@ namespace EventStore.ClientAPI
         public ConnectionSettingsBuilder KeepReconnecting()
         {
             _maxReconnections = -1;
+            _maxDiscoverAttempts = -1;
             return this;
         }
 


hayley-jean:
Thanks for the PR @accraze!

The endpoint discoverer still needs to handle a `-1` value properly when reconnecting, as currently it will just return immediately. You can see where the value is used [here](https://github.com/EventStore/EventStore/blob/release-v4.0.4/src/EventStore.ClientAPI/Internal/ClusterDnsEndPointDiscoverer.cs#L53)

There may also be some checks in the client code that ensures the `MaxDiscoverAttempts` value is greater than or equal to `0` - these will need to change too.

hayley-jean:
Hi @accraze 

Thanks

Issue #1530, opened 2017-12-21
4.4.1 error in libMonoPosixHelper.so after upgrade, no login screen


pvcon13:
Been running 4.0.2 on ubuntu 16.0x.0x. Did a apt-get dist-upgrade, stopped eventstore;reboot.
System came up and started rebuilding indexes (according to service eventstore status and logs.
When index rebuilding was complete, could not get a login prompt on port 2113 anymore.
Command line Version reported was 4.4.1 now. Downgraded to 4.0.2. Restarted service.
It rebuilt indexes again and now I can log back in.
Error in logs for failed attempts Error while handling HTTP request <address:port>
libMonoPosixHelper.so

1) so bug in 4.4.1
2) how can you prevent it from rebuilding indexes on an orderly restart?

hayley-jean:
Hi @pvcon13 

Would you be able to share your logs with us? If you don't want to add them here, you can email them to hayley at geteventstore.com.

hayley-jean:
Hi @pvcon13 

The http error is caused by a missing file in the package.
As a workaround until we can update the packages, you can install mono on the Event Store node.

With regards to the index rebuild, I have not been able to reproduce this.
1. Was there any reason in the logs as to why the rebuild happened? Corruption etc?
2. Was there an index backup folder created?

pvcon13:
E-mail with procedures and log files sent earlier.
Thank you

pvcon13:
I included 1) how the system was originally built  2) how the unintended upgrade happened  3) how the system was downrev'ed.  I assume the dist-upgrade process had something to so with this. you have all the logs. now

pvcon13:
Question: was the index rebuild a result of the upgrade? If not, how do we reboot systems without triggering and re-indexing?

pvcon13:
Can I add a parameter to the startup config to NOT re-index unless its really corrupted?

pvcon13:
Where on a ubuntu system would the data files and thus the index backups be located?

jen20:
The index will only ever be written to disk if it's hit a million entries (or the number of `--max-memtable-size`) - but the default if installed from packages for both data and indicies is: `/var/lib/eventstore`

pvcon13:
ok, 19 chunk files total. one index folder (empty), no index backup. 8 bytes in each chaser, epoch, truncate and writer files. Nothing else there. What and where else should I look?

jen20:
[...]

PR #1531, opened 2017-12-22
Ensure libMonoPosixHelper is packaged

hayley-jean:
This package is required for gzip support and needs to be manually included in the package.

### DIFF
--- a/scripts/package-mono/package-linux.sh
+++ b/scripts/package-mono/package-linux.sh
@@ -67,6 +67,15 @@ else
     exit 1
 fi
 
+MONOPOSIXHELPER=/usr/lib/libMonoPosixHelper.so
+if [[ -f $MONOPOSIXHELPER ]] ; then
+    writeLog "Using libMonoPosixHelper: $MONOPOSIXHELPER"
+else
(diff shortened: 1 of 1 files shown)


gregoryyoung:
There are other things that need this as well. Why have we run into a problem just now over it?

hayley-jean:
Not a clue @gregoryyoung. There haven't been any changes to the packaging scripts since changing to mono 4.

This missing package is the cause of the http errors in the release candidate reported in #1530, which goes away if you turn off gzip encoding in your browser. The encoding's a new feature, so I guess it stands to reason that it's the first thing to actually fall over with this file's absence

gregoryyoung:
As example see:

https://github.com/EventStore/EventStore/blob/release-v4.0.4/src/EventStore.Core/TransactionLog/Unbuffered/NativeFile.cs#L8

Which I believe uses libposix, might it be a versioning difference of libposix?

hayley-jean:
That might be it, though this is just a helper that's missing. Perhaps that uses libposix proper and doesn't require the helper?

gregoryyoung:
Is there a wrapper in libposixhelper for the functions used? In general we should use them if they are available as opposed to libposix directly.

PR #1539, opened 2018-01-16
Optimize existAt calls on scavenged chunks using bloom filters for faster index merges

shaan1337:
PTable merges of index entries related to scavenged chunks are around 10x slower than merges of index entries related to non-scavenged chunks as shown below:

```
Merge with non-scavenged chunk:
[PID:30978:007 2018.01.08 09:50:48.759 TRACE PTable              ] PTables merge finished in 00:00:01.8044394 ([2048000, 2048000] entries merged into 4096000).
Merge with scavenged chunk:
[PID:32224:007 2018.01.08 10:13:22.025 TRACE PTable              ] PTables merge finished in 00:00:25.2660264 ([2048000, 2048000] entries merged into 4096000).
```
The slowness is due to the [existAt call](https://github.com/EventStore/EventStore/blob/release-v4.0.4/src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunkReadSide.cs#L212) that verifies if an index entry exists in the chunk file by doing a binary search look up on the Posmap (which also requires disk access)

The proposed solution is to add an option **--optimize-index-merge** that uses bloom filters to keep the posmap in memory for as many scavenged chunks as possible. This reduces the time for PTable merges and also decreases the disk pressure which is already high during PTable merges **(Full description of the solution in commit message)**

**Results (tested with an index rebuild):**

**Merge with scavenged chunk without optimization:**
 ```
PTables merge finished in 00:00:23.1940588 ([1000000, 1000000] entries merged into 2000000).
PTables merge finished in 00:00:20.1107664 ([1000000, 1000000] entries merged into 2000000).
PTables merge finished in 00:00:35.5952172 ([2000000, 2000000] entries merged into 4000000).
```
**Total: 78s**

**Merge with scavenged chunk with --optimize-index-merge**
 ```
PTables merge finished in 00:00:03.1470243 ([1000000, 1000000] entries merged into 2000000).
PTables merge finished in 00:00:03.1325550 ([1000000, 1000000] entries merged into 2000000).
PTables merge finished in 00:00:04.3924874 ([2000000, 2000000] entries merged into 4000000).
```
**Total: 10s (around 7-8x speedup)**

The above also includes the time taken to create and populate the bloom filters which is done on the fly when calls are made to existAt.

### DIFF
--- a/src/EventStore.ClusterNode/ClusterNodeOptions.cs
+++ b/src/EventStore.ClusterNode/ClusterNodeOptions.cs
@@ -210,6 +210,9 @@ namespace EventStore.ClusterNode
         [ArgDescription(Opts.IndexCacheDepthDescr, Opts.DbGroup)]
         public int IndexCacheDepth { get; set; }
 
+        [ArgDescription(Opts.OptimizeIndexMergeDescr, Opts.DbGroup)]
+        public bool OptimizeIndexMerge { get; set; }
+
         [ArgDescription(Opts.GossipIntervalMsDescr, Opts.ClusterGroup)]
         public int GossipIntervalMs { get; set; }
         [ArgDescription(Opts.GossipAllowedDifferenceMsDescr, Opts.ClusterGroup)]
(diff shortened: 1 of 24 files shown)


shaan1337:
Verified data integrity by doing index rebuilds with a few small `--max-mem-table-size` values (to generate many merges) with and without `--optimize-index-merge` and then comparing the md5 hashes of the final PTables match. Since it's an index rebuild, all index entries should be included in the PTables, since there are no deleted events.

PR #1557, opened 2018-02-04
Add PreferSlaveNode to Connection Builder

Salgat:
This adds the option to prefer Slave nodes in the ClientAPI connection builder (similar to PreferRandom). No breaking changes to the public API, simply adds a PreferSlaveNode() to the fluent syntax connection builder, similar to PreferRandomNode(). 

The use case is that my company has many services that UseMaster along with several services that are read-only and use PreferRandom. The issue is that best case, a third of our read only services are still loading down the master which already has a disproportionate number of the connections. We'd like to shift as much read-only load to the slaves when possible.

### DIFF
--- a/src/EventStore.ClientAPI/ClusterSettings.cs
+++ b/src/EventStore.ClientAPI/ClusterSettings.cs
@@ -43,7 +43,7 @@ namespace EventStore.ClientAPI
         /// <summary>
         /// Prefer a randomly selected node. 
         /// </summary>
-        public bool PreferRandomNode;
+        public NodePreference NodePreference;
 
         /// <summary>
         /// Used if connecting with gossip seeds.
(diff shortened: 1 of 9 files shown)


hayley-jean:
Thanks 👍

mikaelisik:
With PreferSlaveNode, how are the slaves ordered? If several clients connects using PreferSlaveNode, would they all get the same slave?

Salgat:
@mikaelisik Great catch mikaelisik, I'll put up another PR to randomize the slaves.

Salgat:
PR up at #1559 @hayley-jean

ben-foster-cko:
@Salgat were you planning to implement the same in the .NET Core client?

Salgat:
@ben-foster-cko I had assumed PRs were automatically brought over to the .net core client also. I really agree with @bartelink so that we can avoid a divergence in the two clients (I can't imagine maintaining duplicate clients till the end of time instead of multi-targeting). I'll look into putting up a PR for the other repo when I have time.

samholder:
Is it possible to set this preference from the connection string in config?

Salgat:
@samholder Unfortunately not. When you connect to a cluster the client retrieves a list of nodes and has to choose which to use (which is where this setting comes in to play); the connection string only tells the IP/Domain, Port, and connection protocol to use.

PR #1559, opened 2018-02-06
Fix PreferSlaveNode to randomize slave node order

Salgat:
This prevents preference of a specific slave. See PR #1557, thanks @mikaelisik!


hayley-jean:
Good catch, thanks!

PR #1561, opened 2018-02-13
Fix collection modified exceptions that can occur and fault a projection

hayley-jean:
Change the dictionary that tracks EOF's in the `MultiStreamEventReader` to `ConcurrentDictionary` because it can be accessed from multiple threads.


gregoryyoung:
:+1:

PR #1567, opened 2018-02-22
Remove Thread.Sleep from ClientAPI

nielssandmann:
This prevents a threadpool starvation described in #1051. Instead of blocking the thread when trying to enqueue more operations than MaxQueueSize the thread is released letting other operations to continue.

PR is a dublicate of PR 1566 but with fixed tests


hayley-jean:
Thanks @nielssandmann!

PR #1577, opened 2018-03-14
Parked stream persistent subscription fixes

shaan1337:
This commit fixes two issues in parked streams:

**1) When replaying events from parked streams, the parked event is returned in the "Original" part of the ResolvedEvent (instead of the actual original event)**
The change ensures that the actual original event is now returned in the "Original" part.
If the event is a linkTo, the "Event" part is also resolved according to _settings.ResolveLinkTos_

**2) When writing an event to parked streams, and the Original event is a linkTo, the linked event is written to the parked stream instead of the Original event (basically the Original event information is lost)**
The change ensures that we always write the Original event to the parked stream instead of the linked event

Basically the change makes sure that _"Replaying parked messages" makes the subscriptions receive events exactly the same way they are received normally in the "EventAppeared" handler_

**Configuration affected # 1:** Clients using the "Original" part of the event
This configuration was not working correctly previously when replayed from parked streams - the parked event was returned as the "Original" part of the _ResolvedEvent_

**Configuration affected # 2:** _Events with links_ + _DoNotResolveLinkTos_
This configuration was not working correctly previously when replayed from parked streams - the linked event was returned instead of the original event as the "Event" part of the _ResolvedEvent_

_**Issue with existing parked streams for configuration # 2:**_
Replaying messages from existing parked streams will continue to return the **Linked event** instead of the **Original event** until the stream is completely replayed and truncated (because the parked stream contains the linked event)

### DIFF
--- a/src/EventStore.Core/Services/PersistentSubscription/PersistentSubscription.cs
+++ b/src/EventStore.Core/Services/PersistentSubscription/PersistentSubscription.cs
@@ -2,12 +2,14 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Threading.Tasks;
 using EventStore.Common.Log;
 using EventStore.Common.Utils;
 using EventStore.Core.Data;
 using EventStore.Core.Messages;
 using EventStore.Core.Messaging;
 using EventStore.Core.Services.PersistentSubscription.ConsumerStrategy;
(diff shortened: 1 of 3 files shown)


gregoryyoung:
Can you rebase and I will merge...

shaan1337:
Rebased, thanks :+1:

PR #1578, opened 2018-03-15
KeepDiscovering option added for infinte nodes discovery attempts

### LINKED ISSUE #1478: Set MaxDiscoverAttempts when using KeepReconnecting

avish0694:
Fixes #1478
An alternative solution to issue #1478

### DIFF
--- a/src/EventStore.ClientAPI/DnsClusterSettingsBuilder.cs
+++ b/src/EventStore.ClientAPI/DnsClusterSettingsBuilder.cs
@@ -55,6 +55,17 @@ namespace EventStore.ClientAPI
         }
 
         /// <summary>
+        /// Allows infinite nodes discovery attempts.
+        /// </summary>
+        /// <returns></returns>
+
+        public DnsClusterSettingsBuilder KeepDiscovering()
+        {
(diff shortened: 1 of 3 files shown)


gregoryyoung:
As far as I know you could set it to 2 billion. Is 2 billion not enough? I am guessing at this point someone has probably woken up and noticed.

hayley-jean:
@gregoryyoung This is mostly so we have an equivalent to `KeepReconnecting` for a cluster.

We have seen a few people who expect `KeepReconnecting` to work the same for a cluster connection as it does for a single connection, but of course it doesn't. It's easier for us to tell them to just change to `KeepDiscovering` than to tell them to set `MaxDiscoverAttempts` or something like that, and this is a simple enough change to make.

PR #1582, opened 2018-03-29
Add queue timeout facility

bartelink:
This PR results from load and resilience testing I've been carrying out on a production HA cluster and cross checking with local load tests.

The local load tests involve running 2000 tps of read, append (with expected version based on read), read write to very success for an independent stream per transaction. (box can take 2500 most of the time; this is an intentionally relatively low load). The 2000 tps is a target; the runner will attempt up to 4000 requests in parallel to catch up if the 2000 tps are not achieved; in other words, I configure it such that we don't saturate the request queue, but we do saturate the operations queue. (I'm not in a position to furnish the load test harness right now; securing clearance re open sourcing of my wrappers, tests and load harness is in progress)

Each transaction runs on an independent `Task` within an x64 .NET 4.6.1 console app with a `SetMinThreads(7768, 768)` startup tweak (I have tested without any `SetMinThreads`, and with `(100, 100)`; both settings work, but are not in alignment with my broader requirements).

During testing, I intermittently
- [x] close sockets (to trigger instant reconnect)
- [x] Ctrl-C the local server (reconnect, inc a cold/warming up server)
- [x] do heavy compiles (to trigger overloads leading to threadpool saturation), let screensaver kick in

A common issue observed in this test scenario is that the ES connection does not afford one a way in which to exert backpressure on the system as a whole. The provided `SetQueueTimeoutTo` API on the config DSL is what I utilise to effect this. I've load tested this in some depth and it exhibits the desired characteristics in that it:
- [x] never hangs
- [x] deals with the outages
- [x] will throw if overloaded, giving a predictable load-shedding/timeout behavior
- [x] can recover from the point of being saturated.

The OOTB behavior without invoking the API is intended to be identical (the `TimeSpan.Zero` default value means the condition is never checked and the exception can never be thrown); I've verified that the normal latencies under significant load without overload situations are the pretty much identical (despite that I've replaces a `Queue` with a `ConcurrentQueue` given the fact that the queue is `Enqueue`d to from multiple threads).

### DIFF
--- a/src/EventStore.ClientAPI/ConnectionSettings.cs
+++ b/src/EventStore.ClientAPI/ConnectionSettings.cs
@@ -61,6 +61,10 @@ namespace EventStore.ClientAPI
         /// </summary>
         public readonly TimeSpan ReconnectionDelay;
         /// <summary>
+        /// The amount of time a request for an operation is permitted to be queued awaiting transmission to the server.
+        /// </summary>
+        public readonly TimeSpan QueueTimeout;
+        /// <summary>
         /// The amount of time before an operation is considered to have timed out.
         /// </summary>
         public readonly TimeSpan OperationTimeout;
@@ -139,6 +143,7 @@ namespace EventStore.ClientAPI
                                     int maxReconnections,
                                     bool requireMaster,
                                     TimeSpan reconnectionDelay,
+                                    TimeSpan queueTimeout,
                                     TimeSpan operationTimeout,
                                     TimeSpan operationTimeoutCheckPeriod,
                                     UserCredentials defaultUserCredentials,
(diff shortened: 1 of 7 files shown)


shaan1337:
Thanks very much for your PR @bartelink! This looks like a very useful addition to the Client API. The changes look good, will do some tests and merge it if everything's fine.

shaan1337:
Merging this, I've done a few tests and everything works as expected.
Thanks again for your contribution @bartelink!

bartelink:
Wait; where's the nitpicking ;) Thanks for the quick merge 👍

bartelink:
Thanks @Lougarou - I can confirm master as of this merge has also been battle tested too in a prod environment too (it entered prod shortly after the merge).

Issue #1598, opened 2018-04-14
SetClusterDns throwing exceptions internally


bartelink:
On master right now (but also on NuGet 4.1), when I do a DNS resolve computing the endpoints by hand based on supplying "localhost" using (simplified):

        let dnsEntries = System.Net.Dns.GetHostAddressesAsync("localhost").Result
        let validIps = seq { for e in dnsEntries do if e.AddressFamily = System.Net.Sockets.AddressFamily.InterNetwork then yield e }
        [| for a in validIps -> System.Net.IPEndPoint(a, 30778) |]

and feed the endpoints to `DiscoverClusterViaGossipSeeds().SetGossipSeedEndPoints(ips)`, I get a clean discovery against a local 4.1 OSS node started via

    & $env:ProgramData\chocolatey\bin\EventStore.ClusterNode.exe --gossip-on-single-node --discover-via-dns 0  --ext-http-port=30778 

However when I instead try to avoid having to duplicate what I understand to be client lib functionality via `DiscoverClusterViaDns().SetClusterDns("localhost")`, I see it first report a failure:-

```
[03:05:02 INF] Discovering attempt 1 failed with error: System.UriFormatException: Invalid URI: The hostname could not be parsed.
   at System.Uri.CreateThis(String uri, Boolean dontEscape, UriKind uriKind)
   at EventStore.ClientAPI.Transport.Http.HttpAsyncClient.Receive(String method, String url, UserCredentials userCredentials, Action`1 onSuccess, Action`1 onException, String hostHeader) in C:\code\EventStore\src\EventStore.ClientAPI\Transport.Http\HttpAsyncClient.cs:line 83
   at EventStore.ClientAPI.Transport.Http.HttpAsyncClient.Get(String url, UserCredentials userCredentials, Action`1 onSuccess, Action`1 onException, String hostHeader) in C:\code\EventStore\src\EventStore.ClientAPI\Transport.Http\HttpAsyncClient.cs:line 38
   at EventStore.ClientAPI.Internal.ClusterDnsEndPointDiscoverer.TryGetGossipFrom(GossipSeed endPoint) in C:\code\EventStore\src\EventStore.ClientAPI\Internal\ClusterDnsEndPointDiscoverer.cs:line 190
[...]

PR #1600, opened 2018-04-16
Handle IPv6 addresses from Dns.GetHostAddresses

bartelink:
As mentioned in #1598, passing `"localhost"` to `SetClusterDns` causes the client to internally throw exceptions (and triggers a delay in resolution due to the fact that each failed connection attempt is succeededby a `Sleep(500)`), which can be observed by wiring up the logging.

This PR resolves the root problem (needing to route around this part of the config DSL using `SetGossipSeedEndPoints` by independenly generating the list and then filtering for only IPv4 addresses) by letting `IPAddress.ToString()` render IPV6 addresses in the IETF-defined canonical rendition format: `[addr]:port`, which renders the IPv6 `::1` as `[::1]`, replacing the existing logic which would instead render as `scheme://::1:port/path`, which triggers the exceptions cited in #1598 internally).

I've tested this fix on a 3-node HA cluster defined via a DNS entry (works, before and after as there are only IPv4 addresses) and on my `localhost` (which has two entries:- IPv4 and IPv6).

Ideally someone who has thought about it more should eval before this gets blindly merged - perhaps this signifies something that might be more appropriately remediated at a different level.

### DIFF
--- a/src/EventStore.ClientAPI/Transport.Http/EndpointExtensions.cs
+++ b/src/EventStore.ClientAPI/Transport.Http/EndpointExtensions.cs
@@ -11,7 +11,7 @@ namespace EventStore.ClientAPI.Transport.Http
             if(endPoint is IPEndPoint)
             {
                 var ipEndPoint = endPoint as IPEndPoint;
-                return CreateHttpUrl(schema, ipEndPoint.Address.ToString(), ipEndPoint.Port, rawUrl != null ? rawUrl.TrimStart('/') : string.Empty);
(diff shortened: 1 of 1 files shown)


bartelink:
Force-pushed a rebase as mindlessly clicking Update Branch made a merge commit ;)
(🤔 I assume there was no need/value for me do such a rebase - let me know if you want me to force push it back to how it was in the first instance ...)

shaan1337:
Thanks very much for your PR and all the details @bartelink!
Your changes are backward compatible so it should not bring up any issue. I will discuss internally with the team about also integrating #1610 to filter out the non-ipv4 addresses.

PR #1601, opened 2018-04-16
Make UseDebugLogger write to Debugger from NuGet version of client lib OOTB

### LINKED ISSUE #1575: UseDebugLogger is not practically usable
[DebugLogger](https://github.com/EventStore/EventStore/blob/29e5d77540504d6fc5f00307b56b56ec139f5112/src/EventStore.ClientAPI/Common/Log/DebugLogger.cs#L9) uses `Debug.WriteLine`, which bears a `Conditional("DEBUG")`Attribute:-

            System.Diagnostics.Debug.WriteLine(Log("ERROR", format, args));

The net effect of this is that only a DEBUG build of the nuget package would actually ever do anything.

For me, the solution lies in
a) a change to Trace.WriteLine
b) a slight doc change to mention that it writes to the debugger output (atm it's technically not lying)

I'd be happy to send a PR to address this, but the question is what combo of naming/deprecation/breaking changes based on the following make sense:-
1) Fix/add a new impl TraceLogger/UseTraceLogger
2) remove/deprecate old impl
3) adjust docs to mention that it sends to `Trace.WriteLine` and hence will show up in debugger and via tools such as SysInternals DbgView
4) For `EventStore.Client.NetCore`, do I do the same PR or anticipate it percolating through once I put it here?

bartelink:
This fixes #1575 by changing the internal logger triggered by the invocation of the `.UseDebugLogger()` config DSL element to write via `Trace.WriteLine` instead of using `Debug.WriteLine`, in order to make the API do what I as a user of the NuGet package assumed when the intellisense xmldoc tells us:

    /// Configures the connection to output log messages to the listeners
    /// configured on <see cref="System.Diagnostics.Debug" />.

(The reason that it does not emit anything as it stands is because `Debug.WriteLine` is tagged `Conditional("DEBUG")`  in the BCL impl; thus any retail build of the NuGet package does not emit any messages to `OutputDebugString` (verified via SysInternals DbgView). Trace.WriteLine is, per the docs, the unconditional equivalent of `Debug.WriteLine`)

### DIFF
--- a/src/EventStore.ClientAPI/Common/Log/DebugLogger.cs
+++ b/src/EventStore.ClientAPI/Common/Log/DebugLogger.cs
@@ -6,32 +6,32 @@ namespace EventStore.ClientAPI.Common.Log
     {
         public void Error(string format, params object[] args)
         {
-            System.Diagnostics.Debug.WriteLine(Log("ERROR", format, args));
+            System.Diagnostics.Trace.WriteLine(Log("ERROR", format, args));
         }
 
         public void Error(Exception ex, string format, params object[] args)
         {
-            System.Diagnostics.Debug.WriteLine(Log("ERROR", ex, format, args));
+            System.Diagnostics.Trace.WriteLine(Log("ERROR", ex, format, args));
         }
 
         public void Debug(string format, params object[] args)
         {
-            System.Diagnostics.Debug.WriteLine(Log("DEBUG", format, args));
+            System.Diagnostics.Trace.WriteLine(Log("DEBUG", format, args));
         }
 
         public void Debug(Exception ex, string format, params object[] args)
         {
-            System.Diagnostics.Debug.WriteLine(Log("DEBUG", ex, format, args));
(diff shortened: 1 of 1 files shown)


shaan1337:
Hi @bartelink thanks for noticing this and submitting another PR! :-)

Ultimately we should rename this to `UseTraceLogger` in a next major release (to avoid breaking changes) Writing to TRACE will also write to the DEBUG listener so adding Debug Listeners will still work as advertised in the api docs.

The changes look good after testing except that the TRACE constant was not defined here: https://github.com/EventStore/EventStore/blob/release-v4.1.1/src/EventStore.ClientAPI/EventStore.ClientAPI.csproj#L33

I've added it in #1611

bartelink:
Nice catch @shaan1337, was not aware `Trace.WriteLine` even had a `ConditionalAttribute` on it 😎  
I do agree the name is a little misleading, but in practice I don't think it's really going to surprise people as it is, and that config DSL is pretty choc-a-bloc atm.

PR #1610, opened 2018-04-23
Remove non ipv4 addresses from dns entries

shaan1337:
Related to #1600 and #1598

### DIFF
--- a/src/EventStore.ClientAPI/Internal/ClusterDnsEndPointDiscoverer.cs
+++ b/src/EventStore.ClientAPI/Internal/ClusterDnsEndPointDiscoverer.cs
@@ -10,6 +10,7 @@ using EventStore.ClientAPI.Messages;
 using EventStore.ClientAPI.Transport.Http;
 using System.Linq;
 using HttpStatusCode = EventStore.ClientAPI.Transport.Http.HttpStatusCode;
+using System.Net.Sockets;
 
 namespace EventStore.ClientAPI.Internal
 {
@@ -125,7 +126,7 @@ namespace EventStore.ClientAPI.Internal
             IPAddress[] addresses;
             try
             {
-                addresses = Dns.GetHostAddresses(dns);
+                addresses = Dns.GetHostAddresses(dns).Where(x => x.AddressFamily == AddressFamily.InterNetwork).ToArray();
             }
             catch (Exception exc)
             {


bartelink:
Would be good to see a comment added to the code it this gets merged (🤔 do duplicates actually cause an issue at all?).

shaan1337:
Will do. Duplicates should not cause any issue, the discovery phase will just try to get the gossip info twice from a node

bartelink:
🤔 hadnt considered it would actually trigger extra traffic/work. From that perspective it's welcome. Obviously the question is whether dropping IPv6 addresses breaks people and/or closes doors without a clean equivalent mechanism (i.e. forcing people to do the discovery explicitly if using IPv6 may or may not make sense)

shaan1337:
@bartelink thanks, we're investigating about the potentially breaking changes.
In your tests you mentioned the following error from the manager node:
```
curl http://[::1]:30778/gossip?format=json
curl : Bad Request - Invalid Hostname
HTTP Error 400. The request hostname is invalid
```
Did your client work correctly after your fixes in #1600 ? (no other exception)

bartelink:
My tests worked locally (I do a non-gossip connect, a pre-seeded gossip connect, and the DNS gossip connect).

I didn't see exceptions from my load test (connecting to a QA cluster and my localhost) when using the DNS gossip connect. It is possible that it internally communicated and failed but did not emit a message to my logger.

If you were to push a -PreRelease NuGet, I'd be more than willing to give it a run (I'm migrating a codebase from 3.6 to 4.1 atm)

bartelink:
Hm ☹️  @shaan1337 I'm seeing connections to `"tcp://localhost:1113"` are broken for me locally on my test rig using against OSS 4.1 locally for both 4.1.1-rc1 and trunk. The chances are this is a local misconfig (fairly sure I ran all tests before submitting), but one other theory is that its something else in 411rc1 (wasnt there a new discovery mechanism - although that looked innocuous) - I'll root cause later and report back

gregoryyoung:
The client and server logs should say why the connection was dropped (whether it was a timeout etc).

bartelink:
Yes that's true. There was an exception, but it was hard to diagnose at the time because drunkvs doesnt surface `ITestOutputHelper` output (it was client side only: the new `OperationExpiredException` from https://github.com/EventStore/EventStore/pull/1582)

shaan1337:
@bartelink thanks for your tests
After analysis I have decided to close this PR. Filtering out ipv6 addresses will very probably not make any difference to the gossip mechanism and it's safer to avoid any breaking change (although I doubt it will break anything)

gregoryyoung:
To make the change safe can we add a address type to gossip that defaults to "ip4" then use that to control the parsing?

bartelink:
I wouldn't be averse to that happening.

My main desire in this space was to be able to remove bespoke DNS resolution code from apps. In prod rigs I'm aware of, IPv6 addresses are not presently in play (and I'm sure they are not in most environments).

However, having the resolution steps/failure modes be as symmetric as possible between localhost running in single node gossip config vs a HA cluster is definitely a good thing in general.

gregoryyoung:
The reason I bring it up is I see no reason why it should *not* support a mixed environment.Yes it might be weird to run a mixed environment for a long period of time.... What about transitioning from ipv4->ipv6 without downtime?

bartelink:
I can imagine that being a useful facility for people seeking to do a migration in some specific fashion. However I'd argue that these advanced requirements are covered to a degree by the pre-seeding API.

For me, the dream API is something that I don't have to debug/read the docs to discover that by default IPv6 addresses are filtered out and that this is the reason things are not working.

If the API forced one to pick 4, 6 or both (with the risk of suboptimal perf if both are in play) explicitly, that'd be preferable to something defaulting to 4.

IOW I personally intend to wrap this in a `GossipSeeded of  System.Net.IPEndPoint [] | GossipDns of string | GossipDnsCustomPort of string * int`; having to add a `GossipDns4Only of string | GossipDns6Only of string` set of options or making `GossipDns` have to be `of dnsQuery: string * maybeCustomPort: int option * AddressFamily` is not ideal.

What are the perf implications of having an IPv4 and IPv6 version of the same address in play at the same time? IOW is it important that this be surfaced given it makes things more complex and necessitates more documentation?

Atm I'm about to remove stuff using the seeded gossip APIs from our systems as it's just a longwinded version of the `GossipDns` scheme now that it can cope with IPv6 addresses coming up in the results.

gregoryyoung:
:+1: Got it.

PR #1622, opened 2018-05-05
Bycorrid

gregoryyoung:
This PR adds a new internal projection $by_correlation_id. This makes a stream per correlation id which can be useful in many scenarios. In the future this is the internal work required for building up the correlationid graph which will also use causationid to show the full graph.

### DIFF
--- a/src/EventStore.Projections.Core/Standard/ByCorrelationId.cs
+++ b/src/EventStore.Projections.Core/Standard/ByCorrelationId.cs
@@ -116,9 +116,17 @@ namespace EventStore.Projections.Core.Standard
             else
                 linkTarget = data.EventSequenceNumber + "@" + data.EventStreamId;
 
-            var linkMetadata = new ExtraMetaData(
-                new Dictionary<string,string> {{"$originalEventTimestamp", "\""+data.Timestamp.ToString("yyyy-MM-ddTHH:mm:ss.ffffffZ")+"\""}}
-            );
+            var metadataDict = new Dictionary<string,string>();
+            metadataDict.Add("$eventTimestamp", "\""+data.Timestamp.ToString("yyyy-MM-ddTHH:mm:ss.ffffffZ")+"\"");
+            if(data.EventType == SystemEventTypes.LinkTo){
+                JObject linkObj = new JObject();
+                linkObj.Add("eventId", data.EventId);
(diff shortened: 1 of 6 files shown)


src/EventStore.Projections.Core/Standard/ByCorrelationId.cs
```diff
@@ -0,0 +1,132 @@
...
+            emittedEvents = null;
+            newState = null;
+            if (data.EventStreamId != data.PositionStreamId)
+                return false;
+            if (!data.IsJson) 
+                return false;
+            var metadata = JObject.Parse(data.Metadata);
+            var indexedEventType = data.EventType;
+            
+            string correlationId = metadata["$correlationId"].Value<string>();
```
Salgat: Can we make the property key "$correlationId" configurable (similar to how category has a configurable option). Currently we use "correlationId" in our metadata.

gregoryyoung: It ha always been $correlationId (ued by projection etc a well internally)

Salgat: Correct, we currently use our own by_correlation_id projection (which is almost identical to this one), it would just be nice to be able to switch over to the built in one instead. If it's too difficult to make this configurable similar to category, I understand.

```
fromAll().when({
  $any:(function(s,e) {
    if (e.metadata === null) return;
    var correlationId = e.metadata.correlationId;
    if (!correlationId) return;
    var formattedCorrelationId = correlationId.replace(/-/g, "");
    linkTo('$process.event.CorrelationId-' + formattedCorrelationId, e);
  })
})
```

shaan1337:
@gregoryyoung can you please squash your commits? I've also pushed a few fixes. Will test again tomorrow and merge.

src/EventStore.Projections.Core/Standard/ByCorrelationId.cs
shaan1337: @Salgat please note that this has now been added.
the projection source now  accepts a json string to set a custom correlation id property, example: `{"correlationIdProperty":"$myCorrelationId"}`

ChrisChinchilla:
Ah sorry @shaan1337 @gregoryyoung this is for that new feature that will come.

PR #1625, opened 2018-05-08
Client API: Add public ctors to other return values of IEventStoreConnection

megakid:
e.g. RecordedEvent, ResolvedEvent to aid unit testing and standardize (WriteResult already has a public constructor).


bartelink:
Would prefer for these to not pollute the public API - can factory methods live in a helper class in an EventStore.Testing maybe, or as named static factory methods ? Also I question whether you should be testing at that level; personally I have wrapper layers, but express them more as higher level integration tests than as unit tests that would necessitate this sort of thing. Also note I think the active branch is now `master`.

bartelink:
I'm not (completely) objecting to them being creatable without private reflection, more with adding public ctors that make it harder for someone jabbing at an API with intellisense to understand a usage pattern. Based on that, I'm requesting that things be pushed slightly off to the side. I'd reiterate that most unit tests re things like this would live in ES as consumer driven contract type things, and integration tests your side or my side focusing on higher level assertions where possible. Obviously there's lots of grey in the middle; I'm definitely not saying "over my dead body", all unit testing backdoors must be managed by `InternalsVisibleTo` ;)

megakid:
Thanks for the comments, I agree it would be more flexible and avoid API pollution if we were to have a EventStore testing lib but that feels a bit over-engineered for this limited use case.

I did consider a public static factory `.Create(...)` with a private constructor but this just feels like we’re dividing the API rather than changing its intention (you want it to be clear to others that these public creation methods - constructors or otherwise - are not considered the main use-case).

megakid:
Also changed the PR base to `master` as suggested.

SeanFarrow:
I think like @bartelink I'd prefer a testing assembly.
Then to have factories to create the types.
We have an abstraction that models a stream--whether this be event store or some other queuing system, so we have abstractions to allow us to process results etc. This requires us to be able to create EventReadResults, so the capability to create clent messages via a test assembly would be good.

jsjlewis96:
Was a final decision made on this?

pgermishuys:
Hey @megakid, thanks for the PR. We opened an issue to discuss this and would like you to weigh in on that discussion [here](https://github.com/EventStore/EventStore/issues/1986).

At the moment we are not going to provide public ctors for these types and will close this PR.

PR #1629, opened 2018-05-18
Scavenge log cleanup

lscpike:
## Intro
This is the first in a series of PRs I will submit around scavenging. I'm breaking it down to avoid a massive PR and also get approval for each part.

1. Clean up of scavenge log. (This PR)
2. Add ability to stop a scavenge.
3. Automatically mark interrupted scavenges as stopped.
4. Add ability to start a scavenge from a chunk number/log position.
5. Remove the scavenge from the merge pass. This takes a long time and is mostly doing nothing.

## Motivation
- We have a large db that takes about 3 weeks to scavenge.
- A lot of the time we'll have to restart a node resulting in the scavenge stopping.
- Starting a new scavenge starts from the beginning meaning we don't ever reach the most recent chunks.
- If we have a period of high load and need to stop the scavenge then we currently have to restart the node.
- Most of the benefit of scavenging comes from the recent chunks. Our oldest chunks never reduce in size.

## Details
- This PR moves all scavenge log writing to a single class out and out of the actual scavenge code.
- The $scavengeinitialized event has been removed. As there is a max age on the stream it's a little pointless.
- The ScavengeDatabaseCompleted has been replaced with ScavengeDatabaseResponse. This is sent immediately instead of after the scavenge has completed. Nobody will be listening for 3 weeks!
- The response to the ScavengeDatabase message is now being returned on the HTTP interface instead of being ignored.
- The TCP interface has changed to match the last 2 points. This doesn't appear to be implemented in the .net client. The JVM client looks semi implemented but I'm not sure. I can revert this part for compatibility sake, but...

## Questions
- Why does the TCP scavenge interface exist? It seems surplus, unused and not fitting the pattern used for other admin functions. Can I delete it?
- Any feedback on my plans?

### DIFF
--- a/src/EventStore.Core/Services/Transport/Http/Controllers/AdminController.cs
+++ b/src/EventStore.Core/Services/Transport/Http/Controllers/AdminController.cs
@@ -55,11 +55,11 @@ namespace EventStore.Core.Services.Transport.Http.Controllers
                     var completed = message as ClientMessage.ScavengeDatabaseResponse;
                     switch (completed?.Result)
                     {
-                        case ClientMessage.ScavengeDatabaseResponse.ScavengeResponse.Started:
+                        case ClientMessage.ScavengeDatabaseResponse.ScavengeResult.Started:
                             return Configure.Ok(e.ResponseCodec.ContentType);
-                        case ClientMessage.ScavengeDatabaseResponse.ScavengeResponse.InProgress:
+                        case ClientMessage.ScavengeDatabaseResponse.ScavengeResult.InProgress:
                             return Configure.BadRequest();
-                        case ClientMessage.ScavengeDatabaseResponse.ScavengeResponse.Unauthorized:
+                        case ClientMessage.ScavengeDatabaseResponse.ScavengeResult.Unauthorized:
                             return Configure.Unauthorized();
                         default:
                             return Configure.InternalServerError();
--- a/src/EventStore.Core/Services/Transport/Http/Controllers/AdminController.cs
+++ b/src/EventStore.Core/Services/Transport/Http/Controllers/AdminController.cs
@@ -48,7 +48,7 @@ namespace EventStore.Core.Services.Transport.Http.Controllers
             var envelope = new SendToHttpEnvelope(_networkSendQueue, entity, (e, message) =>
                 {
                     var completed = message as ClientMessage.ScavengeDatabaseResponse;
-                    return e.ResponseCodec.To(completed?.ScavengeId);
+                    return e.ResponseCodec.To(new ScavengeResultDto(completed?.ScavengeId));
                 },
                 (e, message) =>
                 {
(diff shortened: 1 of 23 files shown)


src/EventStore.Core/Services/Transport/Http/Controllers/AdminController.cs
```diff
@@ -41,16 +43,31 @@ private void OnPostShutdown(HttpEntityManager entity, UriTemplateMatch match)
...
-            else
-            {
-                entity.ReplyStatus(HttpStatusCode.Unauthorized, "Unauthorized", LogReplyError);
-            }
+            Log.Info("Request scavenging because /admin/scavenge request has been received.");
+
+            var envelope = new SendToHttpEnvelope(_networkSendQueue, entity, (e, message) =>
+                {
+                    var completed = message as ClientMessage.ScavengeDatabaseResponse;
+                    return e.ResponseCodec.To(completed?.ScavengeId);
```
shaan1337: @lscpike hi, would you be able to return `completed` instead of `completed?.ScavengeId` here so that it's serialized as JSON/XML depending on the `Content-Type` specified. It's just a just a matter of consistency among all the controllers. thanks again for this great PR :)

Lougarou:
Thank you @lscpike for the contribution! The reason for the ScavengeIndexInitializedEvent was https://github.com/EventStore/EventStore/pull/817 however, after some discussion with my colleague we think that the solution might not be ideal. I am approving this PR as everything else is fine. Thank you for also structuring the commits independently we **might** have to revert the remove ScavengeIndexInitializedEvent commit later.
Kind regards,
Lokhesh

src/EventStore.Core/Services/Transport/Http/Controllers/AdminController.cs
lscpike: I'm only returning the scavenge id and don't want to expose the whole ScavengeDatabaseResponse. As XML this comes back like this:

<?xml version="1.0" encoding="utf-8"?><string>2b10b34a-1cfe-4d5c-b48f-41eb985983ed</string>

If that's not good enough I'll add a DTO to wrap it. Thoughts?

lscpike:
@Lougarou #817 explains why it's there and I can see the problem it solves. I'm happy to revert it now I know the reason for ScavengeIndexInitializedEvent. Would you like me to do that or did the discussion with your colleague lead to a better option?

Lougarou:
@lscpike It's ok no need to revert it. We will probably patch the UI instead.

src/EventStore.Core/Services/Transport/Http/Controllers/AdminController.cs
shaan1337: @lscpike thanks. yes, a DTO would be perfect. the only issue is with JSON - it just prints out the id with quotes which is not a valid json string.

lscpike: Done.

shaan1337: @lscpike thanks!

PR #1637, opened 2018-05-24
Faster scavenge

lscpike:
Based on the branch from PR #1636. Thanks to @megakid for the majority of this work.

The current scavenge implementation takes between 1 and 5 minutes per chunk on our database. This is partly because of the number of events per chunk, but also because the chunks are already scavenged (scavenged chunk scan is slower). We've done some optimization to speed it up with a trade off of more memory.

This PR has 3 main changes:
- The merge pass no longer scavenges. 
  - We're aware this is less-optimal for scavenging transactions, but that was always dependent on a merge occurring anyway. Additional passes can be performed manually if required.
  - At 4 minutes per chunk to scavenge vs ~5s to just merge, the merge process is just not worth running as is.
- The scavenge process only does a single pass of the chunk file.
  - It does this by copying the whole chunk into a managed array. This obviously has an impact on the manged heap. For most users they aren't going to worry about the required ~300MB of memory. Anybody who is scavenging almost certainly isn't.
  - We've also used this in-memory array to deduplicate the ShouldKeep call. This call is 90% of the time spent scavenging.
- Multi-threading
  - You can now specify up to 3 threads to perform the scavenge in parallel. Default is still 1.
  - This requires even more memory.
  - We've limited to 3 to keep the various ObjectPools happy.


src/EventStore.Core/Services/Transport/Http/Controllers/AdminController.cs
```diff
@@ -41,16 +44,86 @@ private void OnPostShutdown(HttpEntityManager entity, UriTemplateMatch match)
...
+                if (!int.TryParse(startFromChunkVariable, out startFromChunk) || startFromChunk < 0)
+                {
+                    SendBadRequest(entity, "startFromChunk must be a positive integer");
+                    return;
+                }
+            } 
+            
+            int threads = 1;
+
+            var threadsVariable = match.BoundVariables["threads"];
```
megakid: Should be `QueryParameters` instead of `BoundVariables`

src/EventStore.Core/TransactionLog/Chunks/TFChunkScavenger.cs
```diff
@@ -23,60 +23,133 @@ public class TFChunkScavenger
...
-            var totalSw = Stopwatch.StartNew();
-            var sw = Stopwatch.StartNew();
-            long spaceSaved = 0;
-
-            Log.Trace("SCAVENGING: started scavenging of DB. Chunks count at start: {0}. Options: alwaysKeepScavenged = {1}, mergeChunks = {2}",
-                      _db.Manager.ChunksCount, alwaysKeepScavenged, mergeChunks);
+        public string ScavengeId => _scavengerLog.ScavengeId;
 
-            for (long scavengePos = 0; scavengePos < _db.Config.ChaserCheckpoint.Read(); )
+        private IEnumerable<TFChunk.TFChunk> GetAllChunks(int startFromChunk)
```
megakid: private IEnumerable<TFChunk.TFChunk> GetAllChunks(int startFromChunk)
        {
            for (int chunkNum = startFromChunk; chunkNum < _db.Manager.ChunksCount; )
            {
                var chunk = _db.Manager.GetChunk(chunkNum);
                if (!chunk.IsReadOnly)
                {
                    yield break;
                }

                yield return chunk;
                
                chunkNum = chunk.ChunkHeader.ChunkEndNumber + 1;
            }
        }

src/EventStore.Core/Services/Transport/Http/Controllers/AdminController.cs
lscpike: Fixed with addition of threads to the template.

src/EventStore.Core/TransactionLog/Chunks/TFChunkScavenger.cs
lscpike: I'm not sure how this doesn't work the way it is. Can you provide a failing test or explain the problem?

megakid: Your starting `scavengePos`, since it uses a fixed `_db.Config.ChunkSize` assumes chunk sizes are constant in relation to the checkpoint position. They aren’t once scavenged.

lscpike:
Chunk positions are constant. A scavenged chunk has the same log position
always. It just abstracts away the storage position.

src/EventStore.Core/TransactionLog/Chunks/TFChunkScavenger.cs
megakid: Well whatever it is, it didn’t work on my test database.

megakid:
Yep makes sense, thanks.  Looks all good then

lscpike:
We realised the Parallel.ForEach isn't strict on the number of partitions created so it was reallocating the large buffers more than anticipated. We've switched to the ObjectPool to minimise the allocation/GC overheads.

shaan1337:
@lscpike @megakid Kindly also rebase these commits (including 6b18816a7cd55b3d5b45b322cf3c5072be74a1b3)

lscpike:
Rebased as requested.

lscpike:
Closed as the changes will be included with PR #1638.

PR #1632, opened 2018-05-22
Stop scavenge

lscpike:
This PR includes #1629 and will look a lot smaller once that is merged.

This allows operations to stop a running scavenge by calling DELETE /admin/scavenge/{scavengeId}. The scavenge log will show a completed/stopped event.

I moved the threading code to inside the TFChunkScavenger and changed it to a Task.

PR #1638, opened 2018-05-24
Scavenge index

lscpike:
This is based on PR #1637.

The index currently gets scavenged during a level merge. This can happen at any time and can impact performance of the node. It is also very inefficient as each level performs a new scavenge even though it may have just been scavenged for a lower level. If the database hasn't been scavenged since the last merge this is a complete waste of time.

This PR adds an explicit scavenge operation to the TableIndex. This is called at the end of a DB scavenge meaning there is definite benefit to doing the work. It also means it can be stopped during high load situations (thanks to PR #1632.)

### Notes
- The index scavenge blocks the awaiting tables queue. This shouldn't be a problem as the impact is no different to a normal merge.
- The scavenge code is based upon the merge code.
- There's a decent test suite in the PR.
- The merge still has a lot of the scavenge code in there for the v1 upgrade path. If v1 support is dropped in future this code gets a lot smaller.
- The activity is logged to the $scavenges stream to fit the general pattern.


lscpike:
Just to note this my last PR around scavenging.

gregoryyoung:
Can you squash the changes?

lscpike:
Sure. You want the whole lot as 1 commit? There's a lot of files touched
across the 7 prs.

Each PR is mergable/working if you do them in order. The diffs will
obviously get smaller each time.

gregoryyoung:
I meant just for this one (in case we later need to revert). Its far easier to have the commit history with one commit for the feature.

lscpike:
Makes sense! Want the same for the others; faster scavenge and scavenge log clean up? I need to rebase anyway.

lscpike:
We realized that the OptimizeIndexMerge flag results in a lot of memory sitting unused when the index scavenge is complete. As this optimisation now only applies when doing a scavenge it makes sense to free the memory at the end of the scavenge process.

I can squash in if you want, but figured it's easier to see what we've changed as is for now.

megakid:
Any movement on this or if not, timelines for progression?

src/EventStore.Core/TransactionLog/Chunks/TFChunkScavengerLogManager.cs
```diff
@@ -0,0 +1,159 @@
...
+                        return;
+                    }
+
+                    foreach (var ev in readResult.Events)
+                    {                        
+                        if (ev.ResolveResult == ReadEventResult.Success)
+                        {
+                            var dictionary = ev.Event.Data.ParseJson<Dictionary<string, object>>();
+
+                            var nodeId = dictionary["nodeEndpoint"].ToString();
```
riccardone: this like throw a "key not found" exception

riccardone:
@lscpike @megakid we like the stop operation that is implemented in your PR. Thanks for that. 
We are not sure about the changes that you have done with multi-threading. The scavenge operation as is today is a very simple operation and we would like to keep it simple in order to serve all the different scenarios and environments where our clients are running ES. 
Could it be possible keep as is or with minimal modification the current implementation and provide a separate implementation with the parallel/multithread execution?
Something to be driven by command line param (--parallel-scavenge) or http?

lscpike:
By default it is single threaded. The thread count is exposed as a http
option.

This feature is very important for us as a 3 week scavenge time is
unmanagable where a 4 day scavenge time is more reasonable (real world
example).

riccardone:
I get the fact that there is a thread param count. I was asking about the possibility to keep separate your new scavenge multi-threaded implementation from the existing and use your one on demand. That's the reason for the --parallel-scavenge suggestion.

lscpike:
The parallel implementation is only about 3 lines of code. Are you more concerned about the performance optimizations we've made?

Would it be possible to start merging the prior PRs to make the faster scavenge and index scavenge PRs easier to review? I'll rebase the others as each PR is merged to give a clean diff. We can then focus on getting the scavenge optimizations as low risk as possible for you.

gregoryyoung:
The parallel can work out to be slower was more the comment due to disk head thrashing vs sequential reading

lscpike:
The default thread count of 1 will result in sequential reads of the chunks. It's then up to the user not to use the hidden option (not in the UI) that might have a decrease in performance for their disks. That can be in the documentation if required.

lscpike:
We're running this in production as a custom build now. Thought you might be interested to see the scavenge log.

image

shaan1337:
@lscpike @megakid Kindly also rebase these commits (including 6b18816a7cd55b3d5b45b322cf3c5072be74a1b3)

lscpike:
Rebased as requested.

riccardone:
@lscpike I'm doing the final tests of this PR. 
I wonder if it could be possible to log the 'total space saved' as it was before the changes?
https://github.com/lscpike/EventStore/blob/feature/scavenge_index/src/EventStore.Core/TransactionLog/Chunks/TFChunkScavenger.cs#L204
example:
SCAVENGING: total time taken: 00:08:47.3769352, total space saved: 802301658

lscpike:
Done. 

Note this log line occurs before the index scavenge so only includes the space saved by the DB scavenge.

lscpike:
The stop/start from feature was merged in another PR. The index *scavenge*
can also be interrupted/stopped. The index *merge* is cannot be stopped. It
just no longer includes scavenging.

riccardone:
@lscpike in PTableConstruction can I ask why the new method is called Scavenged and not Scavenge?

lscpike:
Probably a typo. I'll fix in a bit.

lscpike:
Just went to fix it and realised it's a factory method. It's createing a "Scavenged" PTable. I'll leave as is unless you have a different opinion.

riccardone:
@lscpike I updated your PR with the latest from the upstream and solved a conflict to save your time. Please let me know if that it's ok for you and then I will proceed with the merge

riccardone:
@lscpike it's probably better if you rebase your PR (not me). Now fw is on 4.7.1 and there is only a simple conflict to fix on index_map

riccardone:
I'm happy to merge all from here, thanks.

ChrisChinchilla:
Docs noted @riccardone and @shaan1337

Issue #1603, opened 2018-04-19
Windows File Cache Usage


lscpike:
I've raised this before in PR #1454.

We've been running 3.9.4 with that patch for the last 6 months and performance has been great. We recently did the upgrade to 4.1 and immediately started having major issues:
- Quorum loss - appears to be delayed/sporadic gossip and the cluster doesn't recover until the original master has been restarted. 
- Constant SLOW MESSAGE BUS errors
- Projections falling behind.
- Lots of client disconnects.
- General fluctuating performance of writes.

As mentioned in the PR these issues all start occurring when "Free memory" has dropped towards 4GB.

We have a scheduled task running every 10 minutes that calls SetSystemFileCacheSize with parameters to free up 40GB of memory. This doesn't suffice:
- This appears to stop having any effect under certain unknown circumstances. 
- It does not cope when ES is doing a large number of reads, e.g. during start up, an index merge, scavenge or something external is reading the history. The cache is being filled very quickly.
- When it does get called and Windows decides to obey it can freeze the entire OS for 10s of seconds.

Our only option was to reapply that patch to 4.1 and deploy a custom build.

Given all of the above, would you be willing to merge that patch or another patch with a configuration option to turn it on?

megakid:
#1516 also potentially reported this behaviour

gregoryyoung:
Just off the top of my head adding SetSystemFileCacheSize calls at opportune places such as during scavenging etc should alleviate this. We have however seen that windows does not seem to always honor what has been set.

megakid:
I think this can be closed now #1639 is merged.

shaan1337:
Thanks @lscpike for this contribution!

PR #1639, opened 2018-05-24
Add option to reduce pressure on the system file cache

lscpike:
Second attempt at getting this change in. See issue #1603.

- A command line option to enable this patch. Off by default.
- The only interesting change is TFChunk.cs#377.
- Removes the RandomAccess hint when opening chunk files.
- This makes windows mark the files as standby pages instead of active pages so they are discarded in preference to process memory.

We've been running with this for 9 months. No obvious performance issues and solved all the ones we had related to low memory.


gregoryyoung:
Looks good :+!:

PR #1640, opened 2018-05-25
Fix latency in index committer service queue & off by one in _serviceShutdownsToExpect

### LINKED ISSUE #1634: Write performance: 4.0.3 vs 4.1.0 and influence of watching the stream in web UI.
We've noticed a significant slow-down in event write speed upon upgrade from EventStore version 4.0.3 to 4.1.0.

Testing across 3 desktop workstations and one server instance shows time increase between 30% and 500% (for writing fixed number of events).

Demo project is attached (executable JAR and Eclipse project; zipped separately). It is a simple Java / Spring Boot application using [esjc](https://github.com/msemys/esjc) as client library. It writes fixed number of events to one stream, then reads those events, then reads them again and re-emits them to another stream and finally re-reads them again and create links on yet another stream. Each of those steps is timed separately. All reads and writes/appends are synchronous.

Our testing results, along with description of each machine used in tests, are in attached Excel file. Those results represent run timings (minutes/seconds) per EventStore version, per run, per machine. Timing is done for 20000 events. There was also some additional timing across 3 more PCs, and some more with smaller number of events (1000 and 4000) on one of the PCs. All show similar pattern. (Those results are not included in the workbook). All testing is done with single EventStore node run locally (always on same machine as test jar) with all default settings.

Another funny thing is that writes tend to go much faster (around 5 times faster) if target stream is watched via the web UI (auto-refreshing page displaying stream's events). It literally goes faster if you look at it. This is mostly noticeable for EventStore versions 4.1.0 and above; much less (if at all) for 4.0.3. For 4.1.0+ speed (with stream open in web UI) approaches that of 4.0.3.

Tested with 4.1.1-hotfix1 as well with same results.

Note that attached project/JAR is configured to work with 4000 events in order to reduce overall execution time (still more than enough to illustrate the behavior on our hardware).

[jar.zip](https://github.com/EventStore/EventStore/files/2027494/jar.zip)
[project.zip](https://github.com/EventStore/EventStore/files/2027495/project.zip)
[Results-20000.xlsx](https://github.com/EventStore/EventStore/files/2027501/Results-20000.xlsx)

shaan1337:
Fixes #1634 (although, for some reason, the latency overhead is heavily amplified in the example application posted)
- Additional latency incurred by Thread.Sleep(1) in index committer service
- Fixes index commiter service stats
- Also fixes off by one in `_serviceShutdownsToExpect` which causes issues with tests

### DIFF
--- a/src/EventStore.Core/Services/Storage/IndexCommitterService.cs
+++ b/src/EventStore.Core/Services/Storage/IndexCommitterService.cs
@@ -41,12 +41,17 @@ namespace EventStore.Core.Services.Storage
 
         public string Name { get { return _queueStats.Name; } }
         private readonly QueueStatsCollector _queueStats = new QueueStatsCollector("Index Committer");
-
         private readonly ConcurrentQueue<StorageMessage.CommitAck> _replicatedQueue = new ConcurrentQueue<StorageMessage.CommitAck>();
         private readonly ConcurrentDictionary<long, PendingTransaction> _pendingTransactions =
                             new ConcurrentDictionary<long, PendingTransaction>();
 
         private readonly CommitAckLinkedList _commitAcks = new CommitAckLinkedList();
+#if MONO
+        private readonly AutoResetEvent _addMsgSignal = new AutoResetEvent(false);
+#else
+        private readonly ManualResetEventSlim _addMsgSignal = new ManualResetEventSlim();
+#endif
+        private TimeSpan _waitTimeoutMs = TimeSpan.FromMilliseconds(100);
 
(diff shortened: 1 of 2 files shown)


gregoryyoung:
why the auto/manual reset event clr/mono?

shaan1337:
@gregoryyoung histograms attached before/after the fix (tested on a windows vm):
[histograms.zip](https://github.com/EventStore/EventStore/files/2044692/histograms_fixed.zip)

Regarding the auto/manual reset event, I've followed the same structure used in StorageChaser:
https://github.com/EventStore/EventStore/blob/HEAD/src/EventStore.Core/Services/Storage/StorageChaser.cs#L29

Faq:
Any chance to review and merge?

PR #1642, opened 2018-05-26
Faster node initialization

megakid:
By adding parallelism to 2 stages of the node initialization process it is now considerably faster, especially for nodes with lots of chunks.  I have introduced a new option called InitializationThreads which controls the level of parallelism for both stages - it defaults to 1 and can be increased to 4 (as a maximum).
You can be the judge but I think the code is not overly complex and looks similar before.

**Chunk initialization during database open**

Currently, the 'Open' method loads each chunk in a sequential fashion and, once initialized, uses the `ChunkHeader` to establish which chunk to open next.  This is not I/O bound so prime for parallelism.  In the changes I propose, the ChunkHeaders are read first to establish the list of chunk filenames to initialize, this list is then farmed out to `Parallel.ForEach` to do the initialization work.  The `Manager.Add` method is already threadsafe and isn't reliant upon sequential Adds so can safely be used within the Parallel loop.

**IndexMap load**

Whilst v4 Indexes have solved the slow index cache rebuild, it is, again, an easily parallizable task (on a per index file basis). 
Currently, the `LoadPTables' method loads each line of the indexmap file in sequence and builds up the PTables accordingly.  By passing the indexmap lines to `Parallel.ForEach` again, we can initialize each PTable in parallel, adding them when ready within a lock.

**Performance**

By no means scientific but I have a 550Gb database with both v3 (5 files) and v4 (5 files) indexes.

I measure from .exe startup time to when the node becomes ready for use in a single node configuration:

````D:\es-bin\EventStore.ClusterNode.exe -Db D:\EventStore\data -Index D:\EventStore\data\newindex --log D:\es-bin\slc-build\logs  --int-ip 10.0.11.109 --ext-ip 10.0.101.109 --int-tcp-port=2111 --ext-tcp-port=2112 --int-http-port=2113 --ext-http-port=2114 --RunProjections=All -StartStandardProjections -SkipIndexVerify -SkipIndexScanOnReads -OptimizeIndexMerge -ReduceFileCachePressure --InitializationThreads=4````

    v3 index node start (4.1.0):
    	14:44:21.308
    	15:11:15.754 (**26mins 55 secs**)
    
    v3 index node start (this build `-InitializationThreads=4`):
    	14:19:40.227
    	14:33:20.626 (**13 mins 40 secs**)
    
    v4 index node start (4.1.0):
    	13:58:23.655
    	14:04:21.928 (**5 mins**)
    
    v4 index node start (this build `-InitializationThreads=4`):
    	14:15:23.558
    	14:16:08.083 (**45 secs**)

**TL;DR;**

Using 4 threads, this code speeds up a 550Gb database node startup - with latest index (v4) - by 85% compared to 4.1.0.

### DIFF
--- a/src/EventStore.ClusterNode/ClusterNodeOptions.cs
+++ b/src/EventStore.ClusterNode/ClusterNodeOptions.cs
@@ -234,6 +234,9 @@ namespace EventStore.ClusterNode
 
         [ArgDescription(Opts.ReduceFileCachePressureDescr, Opts.DbGroup)]
         public bool ReduceFileCachePressure { get; set; }
+        
+        [ArgDescription(Opts.InitializationThreadsDescr, Opts.DbGroup)]
+        public int InitializationThreads { get; set; }
 
         public ClusterNodeOptions()
         {
@@ -349,6 +352,7 @@ namespace EventStore.ClusterNode
 
             SkipIndexScanOnReads = Opts.SkipIndexScanOnReadsDefault;
             ReduceFileCachePressure = Opts.ReduceFileCachePressureDefault;
+            InitializationThreads = Opts.InitializationThreadsDefault;
 
             ConnectionPendingSendBytesThreshold = Opts.ConnectionPendingSendBytesThresholdDefault;
             ChunkInitialReaderCount = Opts.ChunkInitialReaderCountDefault;
--- a/src/EventStore.ClusterNode/Program.cs
+++ b/src/EventStore.ClusterNode/Program.cs
@@ -217,7 +217,8 @@ namespace EventStore.ClusterNode
                         .AdvertiseExternalSecureTCPPortAs(options.ExtSecureTcpPortAdvertiseAs)
                         .HavingReaderThreads(options.ReaderThreadsCount)
                         .WithConnectionPendingSendBytesThreshold(options.ConnectionPendingSendBytesThreshold)
-                        .WithChunkInitialReaderCount(options.ChunkInitialReaderCount);
+                        .WithChunkInitialReaderCount(options.ChunkInitialReaderCount)
+                        .WithInitializationThreads(options.InitializationThreads);
 
             if(options.GossipSeed.Length > 0)
                 builder.WithGossipSeeds(options.GossipSeed);
--- a/src/EventStore.Core/Cluster/Settings/ClusterVNodeSettings.cs
+++ b/src/EventStore.Core/Cluster/Settings/ClusterVNodeSettings.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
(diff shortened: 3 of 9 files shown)


gregoryyoung:
Are these measurements on a clean boot or after having been run already (eg are the files in cache)

megakid:
Good point, they weren't performed in the order I've listed them but I will rerun with clean caches

shaan1337:
@megakid thanks for your PR! if you're on linux, you can also just run `sudo echo 3 > /proc/sys/vm/drop_caches` to clean the caches without the need for a reboot.

megakid:
Re-ran the tests with a reboot in between each one, 550Gb dataset identical (except for a couple of projection/stats events added on each node startup).  This was on Windows Server 2016, 8 * 2.6Ghz, 16Gb RAM.

| Test | Build | Threads | Result (secs) | v3 % | v4 % |
| ------- | ------ | ------ | ---- | --- | --- | 
| `v3 index node start` | `4.1.0` | `N/A` | `1622`  | `100%` | `631%` | 
| `v3 index node start` | `vNext` | `4` | `840` | `52%` | `327%` |
| `v3 index node start` | `vNext` | `8` | `835` | `51%` | `325%` |
| `v4 index node start` | `4.1.0` | `N/A` | `257` | `16%` | `100%` |
| `v4 index node start` | `vNext` | `4` | `74` | `5%` | `29%` |
| `v4 index node start` | `vNext` | `8` | `46` | `3%` | `18%` |

shaan1337:
@megakid thank you, the results look pretty impressive! We'll do some internal tests before merging

shaan1337:
Just to confirm, are you running with SkipIndexVerify: true on the v4 index nodes? Thanks

megakid:
Correct, I am running `-SkipIndexVerify` on all tests.

gregoryyoung:
Also what kind of disk are you running on?

megakid:
This is running on a virtualized setup backed with a top tier Pure Storage flasharray (350k IOps - but I imagine it will be QoS'd over our virtual estate) - all of which are running our production application servers.  We don't yet have a physical box to test on with a dataset of this size but might have one by this afternoon.

If the data is cached somewhere (e.g. my first test without reboots), it makes even more sense to parallelize since it will be less IO bound than when coming from disk.

gregoryyoung:
parallelizing may also make sense for some setups (sounds surprising but true) its also a pathological case for other setups

megakid:
Any movement on this or if not, timelines for progression?

gregoryyoung:
I have started looking through it. It will take me a bit though as there are quite a few changes.

shaan1337:
@megakid can you please also rebase this PR against master? thanks!

megakid:
Rebase is located at #1756

PR #1643, opened 2018-05-28
Clean up persistent subscription logging

lscpike:
- Make everything consistent with punctuation and naming.
- Use the full subscription id everywhere.
- When retrying an event, log the stream id and event number.
- Reduce the logging of the persistent subscription checkpoints.

The main thing here is the last point. The logs currently have a lot of fairly useless entries for the persistent subscription checkpoints. These are of course in the ES database anyway so no loss removing them.


gregoryyoung:
:+1:

PR #1644, opened 2018-05-30
Fix bug in MakeUrl() where url is wrongly formed when query parameters are present

shaan1337:
The bug manifests itself when trying to create a projection in the "Query" tab.


gregoryyoung:
:+1:

PR #1647, opened 2018-06-02
* Add Fabric Service Support

hanxinimm:
* Add Fabric Service EnvironmentVariable Config On Deploy an existing executable to Service Fabric

<EnvironmentVariable Name="EVENTSTORE_EXT_IP" Value="[Fabric_NodeIPOrFQDN]"/>

[VariableName] Reference System EnvironmentVariable Or Service Fabric environment variables

PS:
https://docs.microsoft.com/en-us/azure/service-fabric/service-fabric-environment-variables-reference

### DIFF
--- a/etc/EventStore.Rags/EnvironmentVariables.cs
+++ b/etc/EventStore.Rags/EnvironmentVariables.cs
@@ -1,19 +1,29 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text.RegularExpressions;
 
 namespace EventStore.Rags
 {
     public static class EnvironmentVariables 
     {
+        static readonly Regex _regex = new Regex(@"^\[(\w+)]$");
+
         public static IEnumerable<OptionSource> Parse<TOptions>(Func<string, string> nameTranslator) where TOptions : class
         {
             return
-                (from property in typeof (TOptions).GetProperties()
-                    let environmentVariableName = nameTranslator(property.Name)
-                    let environmentVariableValue = Environment.GetEnvironmentVariable(environmentVariableName.ToUpper())
-                    where !String.IsNullOrEmpty(environmentVariableValue)
-                    select OptionSource.String("Environment Variable", property.Name, environmentVariableValue));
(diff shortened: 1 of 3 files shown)


jen20:
Hi @hanxinimm! Can you explain a bit of the context around how this should be used, and what the purpose is?

hanxinimm:
I'm inspired by this article. 
https://nick.barrett.org.nz/running-eventstore-on-service-fabric-v2-3cb086ad96e2

I’ve hosting EventStore as a guest executable in Service Fabric.
I use environment variables to configure Eventstore, like this
image

it's hard to set internal and external IP
default config not work well
Service Fabric has built-in environment variables set for each service instance.
Service Fabric has an environment variable called Fabric_NodeIPOrFQDN.
I want to do it by placing the parameter name in [] brackets
So that it can reference the values of other environment parameters

hanxinimm:
@jen20  do you have any plans to support? Many people try to use eventstore on service fabric~thanks

shaan1337:
hi @hanxinimm 
thanks for your PR! we're sorry for the very long delay, we were in a migration phase.
Do you have some guidelines for testing the above change please? (this will help speed up the approval process)

Would you mind reapplying your changes to a fresh `master` branch? After the migration, the project structure has changed and the `Rags` project is now located under `src/EventStore.Rags/`

shaan1337:
although `${}` will conflict with bash syntax but can easily be escaped by adding a `\` before `$` if exporting from the command line (this can be documented):

example: `export myvar="\${env:refvar}"`

hanxinimm:
@shaan1337  thank you,it looks great~ i will test it when you publish！

hanxinimm:
> @hanxinimm thank you for the feedback :) Can I proceed with the rebase? I'll then make the above changes and add you as co-author on the commit.

that's ok,if have something to do for me, tell me! goode night~

shaan1337:
@hanxinimm I've reopened a PR here with your commits rebased + the changes we discussed: https://github.com/EventStore/EventStore/pull/1813

PR #1651, opened 2018-06-04
Replaced NLog with Serilog

avish0694:
This PR replaces Nlog with Serilog in the hope to achieve better performance. I'm trying to make Serilog backward compatible with how Nlog currently works in EventStore. We will later provide structured logging through Serilog's infrastructure itself.
Things to do : 
- [x]  Modify Serilog RollingFile appender to support {Date} in folder name.
Serilog Rolling file sink does not allow dates to be part of  folder.

- [x] Padding 
In Nlog Thread and process ID are displayed like this : [PID:05349:025] and [PID:5349:25] in Serilog.

- [x] Make Serilog Sinks Async work with subloggers.


src/EventStore.Common/Log/SeriLogger.cs
```diff
@@ -0,0 +1,117 @@
...
+            .ForContext("className",_name,false)
+            .ForContext("loglevel","INFO",false)
+            .Information(exc, format, args);
+        }
+
+        public void DebugException(Exception exc, string format, params object[] args)
+        {
+            _logger
+            .ForContext("className",_name,false)
+            .ForContext("loglevel","DEBUG",false)
```
bartelink: you should add an enricher like:

``` let NLogLevelEnricher () =
        let map = function
            | Serilog.Events.LogEventLevel.Verbose ->   "TRACE"
            | Serilog.Events.LogEventLevel.Debug ->     "DEBUG"
            | Serilog.Events.LogEventLevel.Information -> "INFO"
            | Serilog.Events.LogEventLevel.Warning  ->  "WARN"
            | Serilog.Events.LogEventLevel.Error  ->    "ERROR"
            | Serilog.Events.LogEventLevel.Fatal  ->    "FATAL"
            | x -> invalidArg "level" (sprintf "Invalid value %O" x)
        let enrich (e : Events.LogEvent) = e.AddPropertyIfAbsent(Events.LogEventProperty("NLogLevel", Events.ScalarValue(map e.Level)))
        { new Serilog.Core.ILogEventEnricher with member __.Enrich(evt,_) = enrich evt }
```
and avoid this boilerplate

src/EventStore.Common/Log/SeriLogger.cs
```diff
@@ -0,0 +1,117 @@
...
+
+        public void TraceException(Exception exc, string format, params object[] args)
+        {
+            _logger
+            .ForContext("className",_name,false)
+            .ForContext("loglevel","TRACE",false)
+            .Verbose(exc, format, args);
+        }
+
+        public static void FlushLog(TimeSpan? maxTimeToWait = null)
```
bartelink: is this a final flush? might be good to add a `///` comment

src/EventStore.Common/Log/SeriLogger.cs
```diff
@@ -0,0 +1,117 @@
...
+namespace EventStore.Common.Log
+{
+    public class SeriLogger : ILogger
+    {
+        private readonly Serilog.ILogger _logger;
+        private string _name;
+
+        public SeriLogger(string name)
+        {
+           _logger = Serilog.Log.ForContext(name, null);
```
bartelink: should be `Serilog.Core.Constants.CnotextName or similar, name` - see impl of ForContext

src/EventStore.Common/Log/SeriLogger.cs
```diff
@@ -0,0 +1,117 @@
...
+            .ForContext("loglevel","FATAL",false)
+            .Fatal(exc, format, args);
+        }
+
+        public void ErrorException(Exception exc, string format, params object[] args)
+        {
+            _logger
+            .ForContext("className",_name,false)
+            .ForContext("loglevel","ERROR",false)
+            .Error(exc, format, args);
```
bartelink: beware that there is a special case for nlog where if there are no args then it will not try to format it. for Serilog, it will. see https://stackoverflow.com/questions/49592351/redirect-all-nlog-output-to-serilog-with-a-custom-target

bartelink:
See https://github.com/serilog/serilog-sinks-file/issues/40 re need for a general extension to make naming for rolling files pluggable in serilog - if we can work out a good proposal, I might do the PR there (I want the rolling suffices to be more consistent and this has long been a need for serilog)

bartelink:
> Modify Serilog Process Enricher to add padding.

🤔 Maybe just having a locla enricher which does both and maybe sequester that in a self-answered SO question - in general each enricher in serilog is orthogonal

src/EventStore.Common/Log/SeriLogger.cs
avish0694: Thanks @bartelink for pointing out. I will make another commit in this PR.

src/EventStore.Common/Log/SeriLogger.cs
avish0694: Done. Thanks.

src/EventStore.Common/Log/SeriLogger.cs
avish0694: SourceContext gives the full namespace but to keep backwards compatibility with the previous logging we are using the same class name variable used before.

snakefoot:
Alternative one could just update the 3 years old custom build of NLog with NLog ver. 4.5.6 and it will be faster than Serilog

avish0694:
@snakefoot Also one of the reasons we want to implemented Serilog is because it is a much more modern logging solution, allowing us to pipe logs to loggly , Azure EventHubs,  Elasticsearch and so on.

snakefoot:
Yes more effort has been put into the Serilog ElasticSearch-Sink. But the NLog Loggly Target should work just fine, along with all the other NLog targets

shaan1337:
Closing this PR. We have decided to go ahead with the NLog upgrade instead: #1678 
As per our internal benchmarks, NLog was a bit faster than Serilog with `keepFileOpen=true` but much slower without this option (not enabled by default)

PR #1653, opened 2018-06-08
Off by one bug in TFChunkDB causing pre-last chunk file not to be verified

shaan1337:
Can be reproduced with following steps:
- Create a database with 3 chunk files (chunk-000000.000000 - chunk-000002.000000)
- Stop EventStore
- Change 1 byte of event data in chunk-000001.000000
- Start EventStore
- chunk-000001.000000 is not verified during startup and database starts without errors


gregoryyoung:
I am surprised we have not seen this more in production as the 2nd to last chunk is far more likely to become corrupted than others. Perhaps we have seen it just not discovered it until it became n-2

PR #1655, opened 2018-06-15
fix typo

prolic:
Max Allowed Wries In Flight => Maxa Allowed Writes In Flight


pgermishuys:
@prolic thank you so much for the PR
Would it be possible to fix the spelling in the appropriate repo as the file you amended is the minified version which was built from the UI repository.

You can submit a PR over [here](https://github.com/EventStore/EventStore.UI/blob/master/src/js/modules/projections/views/projections.item.config.tpl.html#L71)

prolic:
Sure, no problem

PR #1669, opened 2018-07-04
new EventStore.Plugins library to run plugins

riccardone:
The new EventStore.Plugins library contains few contracts that can be implemented by an external Plugin to run special features like the new GeoReplicaPlugin
When used to develop a new plugins the new 'EventStore.Plugins' can be referenced as a nuget package
example using Package Manager: install-package EventStore.Plugins

Other minor changes:
- Fixed .gitignore
- There are some changes to test projects that are made by latest version of Visual Studio. It adds an identifier to any Test project

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 TestResult.xml
 *.nupkg
 /packages
+*packages
 /Logs
 /Data
 /Data-logs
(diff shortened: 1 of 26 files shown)


gregoryyoung:
One thing that is worth a discussion here is "how open" do we want this to be? Do we want users to actually be writing plugins? This sounds like a recipe for lots of support questions.

shaan1337:
@riccardone  can you please rebase against master?

riccardone:
replaced by https://github.com/EventStore/EventStore/pull/1760

Issue #1664, opened 2018-06-27
Multitargeting `EventStore.Client` package for both `netstandard2.0` and `net`


enricosada:
It's possibile to have a single nuget package for the eventstore client api, using multitargeting instead of the two separate  `EventStore.ClientAPI` and `EventStore.ClientAPI.NetCore`?

i volounteer to help with that, just need to discuss what do you prefer

That will fix:

- multi targeting projects who reference both netstandard (in a project reference) and .net in the main. Atm is really hard to do, because `EventStore.ClientAPI.NetCore` will be used in the p2p reference with netstandard, and `EventStore.ClientAPI` in the main project, so types will be bound to different assemblies

That will help:

- discoverability
- maintenance for users, using same package version inside the project
- maintenance for library authors who already implement multitargeting and depends on eventstore client for both netstandard and .net
- templates
- easier diff of api changes

May i ask why a separate repository for .net standard? The reason it's to not upgrade to latest sdk (old csproj) in https://github.com/EventStore/EventStore/ repo?

There are some options to minimize that friction:

- add new csproj in original repo, in parallel directory but referencing same sources
- generate `net` package as usual, and `dotnet pack` the new `netstandard`
- unify the packages at the end with a cli tool https://www.nuget.org/packages/dotnet-mergenupkg/

ref https://github.com/EventStore/ClientAPI.NetCore/issues/31

megakid:
This would be good.  

The `EventStore.ClientAPI` can target `netstandard2.0` and `net46` (and maybe `netcoreapp2.1` if we want to have some `#ifdef`s to use fancy new .NET core features).  I don't think I've seen any issues with the `EventStore.ClientAPI.NetCore` (or major differences in the code) so, as you say, it must be tooling that is the blocker for integration.

enricosada:
@gregoryyoung is that fine for you? i'll send a PR with new .net core sdk project in parallel to existing codebase, and merge packages at end of build script.
That allow to maintain current projects for `net4*`, with minimal maintenance, without go all-in with new sdk

gregoryyoung:
:+1:

ppumkin:
Has this been merged?

I am facing this issue now where I want to use `EmbeddedClient` which is 4.6.1 but I did all my Architecture using `EventStore.ClientApi.Core` and my Architecture code needs `IEventStoreConnection` which is in both packages but not castable.

[...]

PR #1680, opened 2018-07-24
Multitarget client nupkg

enricosada:
ref #1664

Add .net standard project for `EventStore.Client` nuget package, to have a single multi targeting package for `EventStore.Client`

Merge the codebase from https://github.com/EventStore/ClientAPI.NetCore/commit/6162cf3ab0cd0fb5f0331caf6cf77b43f7fcc518 one commit per merge, with info in the README , so each one can be rejected if needed

After the pack of the new `src/EventStore.ClientAPI.NetCore/EventStore.ClientAPI.csproj` project, that generated package will be merged with existing nupkg

Is done like that to minimize changes to codebase and build script.

i tried:

- from root: `build quick -Version 1.2.5` to build with assembly version 1.2.5
- from `scripts/nuget-clientapi` :  `.\package-clientapi.ps1 -Version 1.2.5-alpha` to generate in `packages` the nupkg with version 1.2.5-alpha who contains both net40 and netstandard2.0, without rebuild so maintaning the assembly version 1.2.5

TODO

- [x] port test project
- [ ] run test project against netcoreapp2.0
- [x] update build script to pack and merge nupkg
- [x] remove `Version` from csproj, pass from commandline


shaan1337:
hi @enricosada thanks very much for your PR! that looks like some promising work, we'll go through it as soon as possible.

ppumkin:
Amazing! Thanks for your hard word!

enricosada:
The test differ quite a bit, so i need an hint.

Atm all test use `MiniNode`, instead in `EventStore/ClientApi.NetCore` was using a different way to create the node and connection

How should we refactor this? can be done later?

ppumkin:
Hi @shaan1337 - Don't want to be pushy but have you had a chance to look at this? 👀🥇 
🍺 ?

bartelink:
@shaan1337 bump ;) I have a lib I get to single target and not have to do lots of complex build script bifurcation between `EventStore.ClientAPI.NetCore` vs `EventStore.ClientAPI` if we can get this over the line - is there anything can do to assist with unblocking the evaluation and hopefully merging of this ?

gregoryyoung:
I'd say this would remove a lot of friction/duplication and is worth looking at.

bboyle1234:
Thank you @enricosada

ppumkin:
bump :begging:

gregoryyoung:
I will make sure this is discussed further tomorrow. Expect a response within 24 hours or so.

bartelink:
*bump* @shaan1337  @gregoryyoung ?

ppumkin:
Just going to have to bit the bullet on regression. The initial choice was a small mistake and people may grumble a bit but its all for the best. .. Also waiting for this and also rather push here instead of forking since it will future proof NetStandard / Core applications

riccardone:
@enricosada @bartelink We are working on both .net Core client and server internally. More info will follow asap as I'm not directly involved in these projects.

jen20:
Hi all, I can add some colour to the work that is going on internally now that .NET Core is ready for production, though I am not currently prepared to commit to any timescales for anything here.

Currently @jageall has done some work to target the Client API at .NET Standard 1.6. Urgent patches excepted, the next release of the `EventStore.Client` package will work across .NET Framework 4.6.1, Mono 4.6 and .NET Core 1.0 and above as a result. The current Core package will be deprecated, but likely have a new version which includes the main client package as a dependency in order not to break people currently depending on it. This is a divergent approach to that taken in this pull request, though I am going to leave it to James to go through and work out how to reconcile the two.

**If you need to run the latest client on an earlier version of .NET or Mono, you will need to stick to the last release, or build it yourself - .NET 4.6.2 and Mono 4.6 will represent the earliest supported versions.**

As @bartelink says, I disagree that current state was a mistake. The alternative at the time would have been _no_ support for .NET Core as a client rather than the work now proposed now being done then. However, now that the landscape is clearer and there is less churn in the world of .NET Core, we need to move to more standard approaches.

Unfortunately it is not as straightforward as it sounds to "just port the client": the client internalises it's dependencies, but has internals visible to the rest of the system for test purposes. Consequently, many of the _server_ dependencies also require upgrading (which is part of the work done). As part of this, we are switching the project formats to use the newer SDK `csproj` format, and to pulling dependencies from a public MyGet feed instead of checking them into this repository directly. Some of this can be seen in #1703, some is not yet public. This is not a straightforward substitution either - our build system is complicated by the need to build native code (V8) across multiple platforms, and to statically link binaries for Linux.

Once the client is published, we intend to port the server to .NET Core also. Although not completely decided, it is likely that Mono and .NET Framework support for the server will be dropped for the next major release of Event Store, and that .NET Core will be the _only supported runtime_ across all of the operating systems we support for running the server, and likely also for the **embedded client**. This will allow us to aggressively adopt attractive new constructs in .NET Core such as `Span<T>` without having to give too much thought to compatibility with other runtimes.

Again, this is not a straightforward process: much of the code base has (mistakenly) used `Mono` as a proxy for "Not Windows" for several years now, and that all needs to be corrected. Our packaging will also change substantially, as .NET Core does not yet have the ability to statically link the runtime. There will likely be several release candidates before the next major version of Event Store ships, and everything here is subject to change based on further research or (constructive) discussion.

Finally, as part of porting the server to .NET Core it is likely that we will make it more modular. Not everyone uses projections, and certainly not everyone wants to build V8 in order to work on the managed core of Event Store or the client. It is likely that both projections and persistent subscriptions will be moved to a plug-in model rather than being part of the core system in the next major version as a result, and we will likely extract interfaces which allow other aspects of the system to working a similar model in future - we have a number of commercial-only projects under way which will also benefit from this: asynchronous geo-replication being one example.

Event Store is one of the most complex open-source .NET projects around thanks to the long history of support for running on Mono and Linux (most of our production clients have been using this for years). Because of the critical nature of a database, it is also not desirable to rush this work. Consequently it is going to take longer than some people think it should - we acknowledge that there is a desire to see this work done and that there are benefits to doing it, but this is not going to be a "quick fix".

You'll likely see incremental work towards this being merged into the codebase over the coming days, weeks and months and we will be creating a new label and milestone for .NET Core related work to help track this.

enricosada:
Hi @jen20 awesome news!

I'll close this PR so, i think is not needed if you convert all repo to new sdk and build the package there.
This PR was only to unblock the client, doing the minimal changes needed to address a real issue with packaging, hoping for a possibile merge.

Some feedback:

- **please do not remove the `net` target from the client package**. While is written in doc that  `net46` support `netstandard1.6`+, in reality there are lot of small issues consuming netstandard from `net461`. (see https://twitter.com/terrajobst/status/1031999730320986112  from .NET team lead ). Having a `net` target allow to use that an assembly built for .net directly, and works better also for older sdk who doesnt understandard netstandard
- **please add also a `netstandard2.0` target, in addition to `netstandard1.6`**. The 1.6 is useful just for .net core 1.0 pratically (and the 2.0 .net core runtime is LTS now, and support netstandard2.0 like `net472`). The important bit is **that .net standard 2.0 target remove lot of dependencies** (just check https://www.nuget.org/packages/Netstandard.library in netstandard1.6 vs 2.0 vs net461). All libraries packages should have a `netstandard2.0` target, and if all repo is converted to .net sdk, is really easy to add (that's a reccomended guidance from .net and asp.net team too). Having just `netstandard2.0` as target is fine for `>= net472` and `>= netcoreapp2.0`. Having just `netstandard1.6` is not enough if you want to target `>= net461` and `>= netcoreapp1.0` 
- **it's fine to not want to merge a PR or do that differently**. I tried asking in the related issue about how to make this PR before do it, but i did know there was a possibility to not have it merged because was unrequested and not a bugfix. I know priorities changes and there are maintenance strategies. It's fine. But please next time just add some comment about "we are not sure, we are evaluating how to do that internally", or "we want to do that differently, because long term maintanance", etc. There wasnt a real comment in 1 month, and that make me as contributor less confident about sending more PR (again, rejected PR is always fine, but communication helps long term trust)

riccardone:
"There wasnt a real comment in 1 month, and that make me as contributor less confident about sending more PR"
@enricosada Thanks for the effort with this PR. We will take on board your feedback and try to improve on this

PR #1688, opened 2018-08-01
Batch log read expired messages when a high number of reads are expiring

shaan1337:
Batch log read expired messages when a high number of reads are expiring to prevent excessive logging.

Some important notes:
- `StorageReaderWorker` was previously immutable and was used by multiple threads. To be able to keep the internal state `_lastExpireTime`, `_expiredBatchCount` and `_batchLoggingEnabled`, we now create one Storage Reader Worker/Storage Reader Bus object for each reader thread.
- A new message has been introduced: `BatchLogExpiredMessages` which is published every 2 seconds when a high number of expired messages is detected. This message extends `IQueueAffineMessage` to make sure that the message is sent to the correct storage reader thread.

### DIFF
--- a/src/EventStore.Core/Services/Storage/StorageReaderService.cs
+++ b/src/EventStore.Core/Services/Storage/StorageReaderService.cs
@@ -35,18 +35,23 @@ namespace EventStore.Core.Services.Storage
             _readIndex = readIndex;
             _threadCount = threadCount;
 
-            var readerWorker = new StorageReaderWorker(bus, readIndex, writerCheckpoint);
-            var storageReaderBus = new InMemoryBus("StorageReaderBus", watchSlowMsg: false);
-            storageReaderBus.Subscribe<ClientMessage.ReadEvent>(readerWorker);
-            storageReaderBus.Subscribe<ClientMessage.ReadStreamEventsBackward>(readerWorker);
-            storageReaderBus.Subscribe<ClientMessage.ReadStreamEventsForward>(readerWorker);
-            storageReaderBus.Subscribe<ClientMessage.ReadAllEventsForward>(readerWorker);
-            storageReaderBus.Subscribe<ClientMessage.ReadAllEventsBackward>(readerWorker);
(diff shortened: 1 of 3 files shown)


src/EventStore.Core/Services/Storage/StorageReaderService.cs
```diff
@@ -35,18 +35,23 @@ public StorageReaderService(IPublisher bus, ISubscriber subscriber, IReadIndex r
...
-            var readerWorker = new StorageReaderWorker(bus, readIndex, writerCheckpoint);
-            var storageReaderBus = new InMemoryBus("StorageReaderBus", watchSlowMsg: false);
-            storageReaderBus.Subscribe<ClientMessage.ReadEvent>(readerWorker);
-            storageReaderBus.Subscribe<ClientMessage.ReadStreamEventsBackward>(readerWorker);
-            storageReaderBus.Subscribe<ClientMessage.ReadStreamEventsForward>(readerWorker);
-            storageReaderBus.Subscribe<ClientMessage.ReadAllEventsForward>(readerWorker);
-            storageReaderBus.Subscribe<ClientMessage.ReadAllEventsBackward>(readerWorker);
-            storageReaderBus.Subscribe<StorageMessage.CheckStreamAccess>(readerWorker);
+            StorageReaderWorker[] readerWorkers  = new StorageReaderWorker[threadCount];
+            InMemoryBus[] storageReaderBuses = new InMemoryBus[threadCount];
```
riccardone: @shaan1337  Can you please explain the need of introducing an array of InMemoryBus here?

shaan1337: @riccardone thanks for pointing this to my attention, I'll have a look to see if we can reuse the same object but at first sight, it looks like it's saving some internal state

shaan1337: @riccardone we need an array of InMemoryBus here (one per thread) because the internal state of the StorageReaderWorker is changing after each message is published.

PR #1707, opened 2018-08-24
Add a new property CmdDefineConstants that's appended to DefineConstants in all csproj files under EventStore.sln

shaan1337:
- Add a new property CmdDefineConstants that's appended to DefineConstants
- Make sure TRACE is added to all release builds
- Make sure TRACE, DEBUG is added to all debug builds
- Change build.sh & build.ps1 to use CmdDefineConstants

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -222,7 +222,7 @@ function buildEventStore {
     patchVersionFiles
     patchVersionInfo
     rm -rf bin/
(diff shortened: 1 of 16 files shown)


shaan1337:
@jen20 @jageall  merging this into master to allow mono builds to work in debug mode. it will be superseded by #1710 later on but until then, this can help to have working DEBUG builds on wercker.

PR #1703, opened 2018-08-22
Updates the project format to the new sdk style format

jageall:

PR #1710, opened 2018-08-27
[WIP] Convert projects to .NET SDK Format

jen20:
This is intended to supersede #1703 and add in the changes necessary to get our current CI process going again. It may take in some other PRs over time too.


jen20:
cc @jageall

build.sh
```diff
@@ -222,7 +222,13 @@ function buildEventStore {
     patchVersionFiles
     patchVersionInfo
     rm -rf bin/
-    xbuild src/EventStore.sln /p:Platform="Any CPU" /p:DefineConstants="$DEFINES" /p:Configuration="$CONFIGURATION" || err
+    msbuild \
```
enricosada: you can use .net core sdk to build also `net461`.
so in the build script you can use `dotnet build`, `dotnet pack`, etc. who are easier to use (and same command line experience for users).

That also mean you dont need to use `msbuild` of mono for compiling, so you can use mono just as runtime.
It make it easier to upgrade (the .net sdk can be installed side by side, easier to upgrade and pin with a `global.json`)

src/EventStore.Rags.Tests/EventStore.Rags.Tests.csproj
```diff
@@ -0,0 +1,24 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net46</TargetFramework>
+    <Platform>x64</Platform>
+    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+    <OutputPath>..\..\bin\tests\</OutputPath>
```
enricosada: that make a lot harder to run `dotnt test` with multitargeting (useful later to test both `net461` and `netcoreapp`)

using `dontet test` is easier also from command line and build scripts, no need to use custom test runner

src/EventStore.Rags.Tests/EventStore.Rags.Tests.csproj
```diff
@@ -0,0 +1,24 @@
...
+    <DebugSymbols>true</DebugSymbols>
+    <DebugType>full</DebugType>
```
enricosada: new mono support `portable` too

src/EventStore.Projections.Core/EventStore.Projections.Core.csproj
```diff
@@ -1,399 +1,20 @@
...
-    <TargetFrameworkProfile />
-  </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
-    <DebugSymbols>true</DebugSymbols>
-    <DebugType>full</DebugType>
-    <Optimize>false</Optimize>
+    <TargetFramework>net46</TargetFramework>
+    <Platform>x64</Platform>
+    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <OutputPath>..\..\bin\intermediate\</OutputPath>
```
enricosada: doing like that make use of multitargeting a lot harder later.
new sdk use `bin/{Configuration}/{tfm}` to make it easier to do incremental build

build.sh
jen20: Hi @enricosada! Thanks for looking over this and providing feedback. We're actually using `dotnet msbuild` at the moment internally, but the build scripts are due for a rewrite once the rest of the cleanup work is done, so that hasn't been ported into the old ones yet (the change you are commenting on will not be in the eventual version of this pull request).

The only slightly weird thing appears to be that to build with Mono it is necessary to set `FrameworkPathOverrides` (or something similar, I forget the exact variable and am on a Windows box right now so don't have it in history).

Ultimately we want to fit into the "standard" .NET workflow as much as possible given the native dependencies.

src/EventStore.Rags.Tests/EventStore.Rags.Tests.csproj
jen20: This is mostly legacy from the current build scripts, and will be changed as per your comment.

src/EventStore.Rags.Tests/EventStore.Rags.Tests.csproj
jen20: Oh, interesting - we should likely switch to use that.

enricosada:
@jen20 sure! 👍  to the work

PR #1711, opened 2018-08-28
Upgrade nunit console to 3.8.0

shaan1337:


jen20:
With #1710 this won't actually be necessary, because the runner is pulled in as a NuGet package. It does seem that there are some other fixes pulled in here that might be worth porting over?

PR #1712, opened 2018-08-29
Added max count to PersistentSubscriptionConfig

Lougarou:
The $PersistentSubscriptionConfig stream can become huge. This is to limit the max count of the stream.

### DIFF
--- a/src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
+++ b/src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
@@ -693,7 +693,12 @@ namespace EventStore.Core.Services.PersistentSubscription
             Log.Debug("Saving persistent subscription configuration");
             var data = _config.GetSerializedForm();
             var ev = new Event(Guid.NewGuid(), "PersistentConfig1", true, data, new byte[0]);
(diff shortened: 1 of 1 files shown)


shaan1337:
@Lougarou can you please rebase against origin/master?

ChrisChinchilla:
@Lougarou and @shaan1337 I guess this needs some docs changes to mention the new setting too?

shaan1337:
@ChrisChinchilla this is an internal system setting, so no documentation required here. thanks!

gregoryyoung:
I think this could use a one sentence blurb about it in the config section. "by default only the last N configs will actually be saved this can be changed via ...". I am assuming if I change the metadata the subscription will not rewrite the maximum?

ChrisChinchilla:
@shaan1337 and @gregoryyoung So final opinions on if this needs public docs or not?

shaan1337:
@ChrisChinchilla it doesn't hurt to add more details so we can mention something briefly like Greg mentioned :)

ChrisChinchilla:
@shaan1337 Actually, looking at this, I think it gets picked up by the automated docs process for the codebase, as in fact, there are no docs covering this area explicitly I can see.

To clarify, when @gregoryyoung mentions 'config', this is config when you create a new stream?

shaan1337:
@ChrisChinchilla he's referring to the persistent subscription configuration (saved in the $PersistentSubscriptionConfig stream). Any time you add/remove/update a persistent subscription from the web ui, this stream is updated with a new event.

PR #1713, opened 2018-08-29
Chunk size

gregoryyoung:
Allows setting the size of TF chunks.


jen20:
This needs rebasing, squashing and retargeting at the current tip of master to be considered for merge.

gregoryyoung:
"We need to consider what happens in the case that someone starts a server with the chunk size set to (for example) 1GB while having 256MB chunk files in the data directory, and a whole host of other similar scenarios."

This scenario is already covered and can be seen in the source. The size is only used at the time of creation and the header is used in the lookup of data later. Thus the locations from the headers alleviate this issue and specify the range a chunk (even scavenged) covers.

gregoryyoung:
Squashing is no problem.
We have no good way of "updating documentation with the new option" without me at least knowing what version it may end up in.

"We need to consider what happens in the case that someone starts a server with the chunk size set to (for example) 1GB while having 256MB chunk files in the data directory, and a whole host of other similar scenarios."

The size applies to new chunks. Chunks define their own size/range in their header. There can be chunks of different sizes in the same db.

ChrisChinchilla:
I have a PR waiting on the docs repo to merge when this is merged, and by the looks of things the code docs were already done, but the functionality didn't exist, so nothing for me to do here.

pgermishuys:
Hey @gregoryyoung, We are going to close this PR for now. We can keep the branch around to keep the work but this will require some rework.

gregoryyoung:
What precisely is the rework? IIRC this provides variable chunk size ... and works. There isn't much to that change. I can see pulling it out into a fresh PR though.

PR #1723, opened 2018-09-11
Fixing macOS build

mjaric:
Two issues are fixed in this pull request:
1. build scripts do not work with official mono 5.x package installations (osx 10.13)
2. old NLog dependency in lib folder do not work anymore with mono 5+ throwing ```System.ReadOnlySpan`1[System.Char]' cannot be converted to type 'System.String``` error once EventStore.ClusterNode.exe is started. So dependency is replaced with version 4.5.9

Before pulled into master, would be good if someone can test this in "older" linux environment.


jen20:
Hi @mjaric! Thanks for opening a pull request! We actually have work in progress to upgrade to a new version of Mono, on #1710, which takes a slightly different approach to this - and will be getting CI up for Mac OS as part of that. If you want to build in the meantime, it might be best to check out that branch until it gets merged? I'm going to close this one for now in favour of that work - I just confirmed it builds fine on the latest version of Mono from Homebrew. If you want to use the `dotnet` CLI to build, remember you'll need to set `FrameworkPathOverride` to the correct version.

mjaric:
@jen20, sorry, I didn't saw that branch. I will give it a try!

PR #1724, opened 2018-09-14
Minor corrections to text and brand name

ChrisChinchilla:
Just removing 'the' from Event Store to bring it inline with docs and branding elsewhere.


gregoryyoung:
seems legit

jen20:
Thanks @ChrisChinchilla!

PR #1709, opened 2018-08-27
Add appveyor.yml for CI builds

shaan1337:

PR #1725, opened 2018-09-15
Moved test results to TESTS tab in AppVeyor

idubnori:
#### Improvement of CI test results visibility
- Test results show in real-time on `TESTS`
- `CONSOLE` log reduces (60k+ rows to 0.9k, all to see)
- The existing detailed test logs are stored in `ARTIFACTS`

#### Notes
- This PR is related to #1709

***
eventstore-tests-realtime-3

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,6 +10,9 @@ test_script:
        /framework:net-4.5 /timeout=60000 
        /out:TestResults.txt 
        --result=TestResults.xml;format=AppVeyor
+artifacts:
+  - path: '**\TestResults.*'
+    name: TestResults
 notifications:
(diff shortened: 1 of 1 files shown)


gregoryyoung:
This looks great! Probably it will be reviewed Monday.

jen20:
Hi @idubnori! Thanks for opening a pull request here - this is a nice change! However, we're in the process of switching over from a combination of Wercker (for Linux) and AppVeyor to use Azure DevOps Pipelines instead - you can see that work in #1710. Fortunately that is already configured with a similar feature - if we were sticking with AppVeyor we would definitely merge this though!

PR #1728, opened 2018-09-16
Giving the correct reason when the subscription drops due to overflow

### LINKED ISSUE #1726: Subscription queue overflow causes subscription to drop with the reason UserInitiated
There are in fact two issues:

- When real-time subscription queue gets full, the subscription drops but the reason is `UserInitiated`, which is incorrect
- When historical events are being processed and the queue is overflown, the reason `ProcessingQueueOverflow`, which is correct.

Solution: use `ProcessingQueueOverflow` reason in both cases for consistency

- Unlike the catch-up historical load, the real-time subscription has a fixed maximum queue size. This size is hard coded and the `MaxQueueSize` setting is not applied to it.

Solution: use the `MaxQueueSize` value from subscription settings for the real-time subscription as well.

alexeyzimarev:
Fixes #1726

### DIFF
--- a/src/EventStore.ClientAPI/ClientOperations/SubscriptionOperation.cs
+++ b/src/EventStore.ClientAPI/ClientOperations/SubscriptionOperation.cs
@@ -257,7 +257,10 @@ namespace EventStore.ClientAPI.ClientOperations
         private void ExecuteActionAsync(Func<Task> action)
         {
             _actionQueue.Enqueue(action);
-            if (_actionQueue.Count > _maxQueueSize) DropSubscription(SubscriptionDropReason.UserInitiated, new Exception("client buffer too big"));
+            
+            if (_actionQueue.Count > _maxQueueSize) 
+                DropSubscription(SubscriptionDropReason.ProcessingQueueOverflow, new Exception("client buffer too big"));
+            
             if (Interlocked.CompareExchange(ref _actionExecuting, 1, 0) == 0)
                 ThreadPool.QueueUserWorkItem(ExecuteActions);
         }


jen20:
Hi @alexeyzimarev! Thanks for opening a pull request. An initial look over this looks fine to me, we'll get it reviewed and merged soon - possibly via #1710 which is still ongoing.

src/EventStore.ClientAPI/ClientOperations/SubscriptionOperation.cs
```diff
@@ -257,7 +259,10 @@ protected void EventAppeared(TE e)
         private void ExecuteActionAsync(Func<Task> action)
         {
             _actionQueue.Enqueue(action);
-            if (_actionQueue.Count > _maxQueueSize) DropSubscription(SubscriptionDropReason.UserInitiated, new Exception("client buffer too big"));
+            
+            if (_actionQueue.Count > _maxQueueSize)
```
riccardone: This change make sense as you are redefining the exception in order to handle it properly. I don't understand all the other changes related to the maxQueueSize setting and the default value.

alexeyzimarev: Yes, I can. But I'd like to know what is the reason for the `_maxQueueSize` default value to be hardcoded instead of using the setting?

riccardone: That is an internal setting that has been decided to be set 2000 as an appropriate value. 

The setting MaxQueueSize is already used in the EventStoreNodeConnection class to limit the maximum number of outstanding items allowed in the queue

https://github.com/EventStore/EventStore/blob/7e98bd4206d5b0aec0cbc24d12002dfe195643fe/src/EventStore.ClientAPI/Internal/EventStoreNodeConnection.cs#L242

I like the improvement of changing the SubscriptionDropReason but not the other changes that extend the use of MaxQueueSize setting in the EventStoreConnectionLogicHandler class.

alexeyzimarev: Ok, sounds reasonable. I will remove other changes.

rcknight: Hit this bug myself in a production system today. Since @alexeyzimarev didn't get round to making these changes yet i've done so in #1747.

alexeyzimarev:
Will do today

alexeyzimarev:
@shaan1337 done

shaan1337:
@alexeyzimarev thanks very much!

PR #1751, opened 2018-11-04
Non Promotable Clone

### LINKED ISSUE #263: NonPromotable Clone
useful feature or use a shoveler?

riccardone:
There is a new IsPromotable config setting (default true)
When it is set to false, the node does not partecipate in the election process as a candidate and it does not allow the cluster to stay up if there are not enough (other) nodes to form a quorum
Added a new test for the ElectionService changes (more tests can be added... as usual)

To test this PR, run a cluster of 3 nodes. This cluster could be the current stable version.
Once that the cluster is up and running, run a node using this PR code and IsPromotable set to false.
Remember to keep the clustersize 3 and use the same gossip port of the other nodes in order to join the existing cluster.
You can verify that the node is recognize as NonPromotableClone and it does not partecipate as a candidate in the Election Process when you shut down the other nodes for testing.

Closes #263
Closes EventStore/EventStore.CommercialHA#36

### DIFF
--- a/src/EventStore.ClusterNode/ClusterNodeOptions.cs
+++ b/src/EventStore.ClusterNode/ClusterNodeOptions.cs
@@ -305,6 +305,9 @@ namespace EventStore.ClusterNode {
 		[ArgDescription(Opts.MaxAutoMergeIndexLevelDescr, Opts.DbGroup)]
 		public int MaxAutoMergeIndexLevel { get; set; }
 
+		[ArgDescription(Opts.IsPromotableDescr, Opts.ClusterGroup)]
+		public bool IsPromotable { get; set; }
+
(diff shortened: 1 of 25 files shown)


ChrisChinchilla:
This is now in progress

shaan1337:
@riccardone there are some conflicts to resolve here too (I guess related to argument passing)

jcelgin:
Will this make the cut for 5.0?  It is worth noting that [the docs](https://eventstore.org/docs/server/node-roles/index.html#non-promotable-clone) suggest this is already a thing...

jen20:
Hi @jcelgin - 5.0.0 is already out, so no - but it should make 5.1.0 shortly, once it’s had some more thorough testing.

PR #1758, opened 2018-11-08
Multitarget client with net46 and netstandard2.0

shaan1337:
Thanks to @jageall for helping out with this. Suggestions are welcome.

The following is based on the compatibility matrix available here: https://docs.microsoft.com/en-us/dotnet/standard/net-standard#net-implementation-support

.NET core 2.0 projects (on Windows/Linux/Mac) and existing .NET Framework 4.6.1+ projects (on Windows) should be able to reference the `netstandard2.0` ClientAPI library according to the compatibility matrix.

.NET Framework 4.6.1+ projects (on Linux/Mac with mono 4.6 as runtime) should be able to reference the `net46` version of the ClientAPI (only mono 5.4+ is compatible with `netstandard2.0` according to the compatibility matrix). This will maintain backward compatibility for mono 4.6 clients.

.NET Framework 4.6.1+ projects (on Linux/Mac with mono 5.4+ as runtime) should be able to reference the `netstandard2.0` ClientAPI library according to the compatibility matrix.

### DIFF
--- a/src/EventStore.ClientAPI/EventStore.ClientAPI.csproj
+++ b/src/EventStore.ClientAPI/EventStore.ClientAPI.csproj
@@ -22,6 +22,5 @@
       <PackageReference Include="System.Net.Http" Version="4.3.4" />
       <PackageReference Include="System.Net.Requests" Version="4.3.0" />
       <PackageReference Include="System.Net.Security" Version="4.3.2" />
-      <PackageReference Include="System.Data.SqlClient" Version="4.5.1" />
   </ItemGroup>
 </Project>
--- a/src/EventStore.ClientAPI/EventStore.ClientAPI.csproj
+++ b/src/EventStore.ClientAPI/EventStore.ClientAPI.csproj
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>net47</TargetFramework>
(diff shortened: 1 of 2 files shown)


src/EventStore.ClientAPI/EventStore.ClientAPI.csproj
```diff
@@ -11,11 +11,17 @@
...
       <PackageReference Include="protobuf-net" Version="2.4.0" />
   </ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net46' ">
+    <Reference Include="System.Net.Http" />
+  </ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
+      <PackageReference Include="System.Net.Http" Version="4.3.4" />
+      <PackageReference Include="System.Net.Requests" Version="4.3.0" />
+      <PackageReference Include="System.Net.Security" Version="4.3.2" />
+      <PackageReference Include="System.Data.SqlClient" Version="4.5.1" />
```
jageall: if we change the documentation comment on the EventStoreConnection that refers to SqlConnection using <seealso /> then we can remove our dependency on the SqlClient,.

shaan1337: @jageall fixed thanks!

shaan1337:
Tested with a sample basic project targeting `net46` and `netcoreapp2.0` on Ubuntu. Also tested the `net46` binary with mono 4.6.2 runtime. Worked fine.

Similar tests done on windows by @riccardone.

PR #1760, opened 2018-11-08
Plugins

riccardone:
EventStore.Plugins library with few contracts to be implemented by an external Plugin to run special features like the new GeoReplicaPlugin
When used to develop a new plugins the new 'EventStore.Plugins' can be referenced as a nuget package
example using Package Manager: install-package EventStore.Plugins


ChrisChinchilla:
@riccardone Does this need any docs changes?

riccardone:
@ChrisChinchilla not for now

PR #1773, opened 2018-11-20
StreamEventReader Fix

Lougarou:
Fixes StreamEventReader bug when end of stream is met when reading from a stream with deleted events($MaxAge/$MaxCount) using read and check positions instead of a single variable. Also added unit tests.

Reproduction Steps:
1. Create a continuous projection from one stream
2. Stream has maxage or maxcount
3. Append events to stream
4. Stop projection and wait for events to get appended and events to get deleted
5. Start projection. 
Without fix no exception is thrown, with fix Event Store should throw: "Event number x was expected in the stream y, but event number z was received. This may happen if events have been deleted from the beginning of your stream, please reset your projection."

Might be related to https://github.com/EventStore/EventStore/issues/1468

### DIFF
--- a/src/EventStore.Projections.Core.Tests/Services/event_reader/stream_reader/when_handling_streams_with_deleted_events_and_reader_starting_after_event_zero.cs
+++ b/src/EventStore.Projections.Core.Tests/Services/event_reader/stream_reader/when_handling_streams_with_deleted_events_and_reader_starting_after_event_zero.cs
@@ -35,11 +35,12 @@ namespace EventStore.Projections.Core.Tests.Services.event_reader.stream_reader
             _edp.Resume();
         }
 
-        private void HandleEvents(long[] eventNumbers){
+        private void HandleEvents(long[] eventNumbers, int nextEventNumber = -1){
             string eventType = "event_type";
             List<ResolvedEvent> events = new List<ResolvedEvent>();
 
             foreach(long eventNumber in eventNumbers){
+                
                 events.Add(
                     ResolvedEvent.ForUnresolvedEvent(
                         new EventRecord(
@@ -51,8 +52,7 @@ namespace EventStore.Projections.Core.Tests.Services.event_reader.stream_reader
                 );
             }
 
-            var correlationId = _consumer.HandledMessages.OfType<ClientMessage.ReadStreamEventsForward>().Last().CorrelationId;
-
+            var correlationId = _edp.PendingRequestCorrelationId;
             long start, end;
             if(eventNumbers.Length > 0){
                 start = eventNumbers[0];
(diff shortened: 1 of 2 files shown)


megakid:
Don't quite follow, can you explain more?

From that:

> 1. Create a continuous projection from one stream

Ok

> 2. Stream has a maxage or maxcount

I assume this could also apply to `$tb` (truncate before) metadata.

> 3. Append events to stream

Is this to start the projection/ensure the projection has some events?  Does this rely on the projection checkpointing atleast once?

> 4. Stop projection and wait for events to get appended and events to get deleted

Does this rely on the fact that the projection hasn't checkpointed ahead of where we're deleting to?
e.g.  Projection checkpoints at v10, we append 10 events and have a maxcount of 5.  Meaning we only keep v15 -> v20.

Does this happen only after a scavenge or will it occur if just soft deleted?

> 5. Start projection.

What happens after the projection is reset? -> checkpoints are cleared so where does it begin from?  v0 or v15 (using our example above).

Lougarou:
> Don't quite follow, can you explain more?
> 
> From that:
> 
> > 1. Create a continuous projection from one stream
> 
> Ok
> 
> > 1. Stream has a maxage or maxcount
> 
> I assume this could also apply to `$tb` (truncate before) metadata.
> 
> > 1. Append events to stream
> 
>> Is this to start the projection/ensure the projection has some events? Does this rely on the projection checkpointing atleast once?

Yes it's not at the start, it has to checkpoint at least once
> 
> > 1. Stop projection and wait for events to get appended and events to get deleted
> 
> Does this rely on the fact that the projection hasn't checkpointed ahead of where we're deleting to?
> e.g. Projection checkpoints at v10, we append 10 events and have a maxcount of 5. Meaning we only keep v15 -> v20.
>
Normally if let's say the checkpoint is at v10 but v10 has already been deleted. Instead of just resuming the projection it should throw an exception and force a reset projection(this pr). We plan to add a flag for that though like "ensure strict event ordering" to let the users choose this behaviour or the previous one.
>> Does this happen only after a scavenge or will it occur if just soft deleted? 

Yes it happens with soft deletion
> 
> > 1. Start projection.
> 
>> What happens after the projection is reset? -> checkpoints are cleared so where does it begin from? v0 or v15 (using our example above).

v15

Lougarou:
@riccardone This problem cannot be solved by changes you have mentioned or the PR https://github.com/EventStore/EventStore/pull/1774

src/EventStore.Projections.Core.Tests/Services/event_reader/stream_reader/when_handling_streams_with_deleted_events_and_reader_starting_after_event_zero.cs
```diff
@@ -96,6 +96,18 @@ public void should_not_allow_first_event_to_be_greater_than_sequence_number()
 
             Assert.AreEqual(1, HandledMessages.OfType<ReaderSubscriptionMessage.Faulted>().Count());            
         }
+        [Test]
+        public void should_not_allow_first_event_to_be_greater_than_sequence_number_when_first_read_eof()
```
Lougarou: @riccardone

pgermishuys:
Hi @Lougarou would you mind having a look at this again and rebase if possible?

jen20:
This is now sufficiently stale that we'll need to reapply the changes anyway @Lougarou - I'll close out this PR for now.

PR #1774, opened 2018-11-21
StreamEventReader Fix (no InvalidOperationException)

riccardone:
Fix scenario when $maxAge/$MaxCount are set on the original stream and events are written while the projection is stopped. 
To manually reproduce:
1. stop the projection
2. write events on the original stream
3. start the projection
4. write other events on the original stream
5 (original code) the projection goes in Fault state + thrown an InvalidOperationException 
5 (with this fix) the projection keep working and log the error 

Fixes https://github.com/EventStore/EventStore/issues/1468

This PR introduce a new option: FaultOutOfOrderProjections (default true)
FaultOutOfOrderProjections=true: the projection that process an event with a not expected number will go in Fault state and a reset is required
FaultOutOfOrderProjections=false: the projection that process an event with a not expected number will keep going working and the error * will be logged

*Event number {0} was expected in the stream {1}, but event number {2} was received. This may happen if events have been deleted from the beginning of your stream, please reset your projection.

### DIFF
--- a/src/EventStore.ClusterNode/ClusterNodeOptions.cs
+++ b/src/EventStore.ClusterNode/ClusterNodeOptions.cs
@@ -152,8 +152,8 @@ namespace EventStore.ClusterNode
         public int WorkerThreads { get; set; }
         [ArgDescription(Opts.ProjectionsQueryExpiryDescr, Opts.ProjectionsGroup)]
         public int ProjectionsQueryExpiry { get; set; }
-        [ArgDescription(Opts.FailOutoforderProjectionsDescr, Opts.ProjectionsGroup)]
-        public bool FailOutoforderProjections { get; set; }
+        [ArgDescription(Opts.FaultOutOfOrderProjectionsDescr, Opts.ProjectionsGroup)]
+        public bool FaultOutOfOrderProjections { get; set; }
 
         [ArgDescription(Opts.IntHttpPrefixesDescr, Opts.InterfacesGroup)]
         public string[] IntHttpPrefixes { get; set; }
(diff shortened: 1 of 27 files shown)


src/EventStore.Projections.Core/Services/Processing/EventReaderCoreService.cs
```diff
@@ -262,6 +262,13 @@ public void Handle(ReaderSubscriptionMessage.Faulted message)
             if (!_eventReaderSubscriptions.TryGetValue(message.CorrelationId, out projectionId))
                 return; // unsubscribed
 
+            if (message.Reason.Contains("was expected in the stream"))
+            {
+                // Log without fault the projection
+                _logger.Debug(message.Reason);
```
jageall: Is debug an appropriate level for this log message? If a customer is not expecting the new behaviour it won't necessarily show up in the logs. Suggest minimum level of Trace

riccardone: set to Trace now

src/EventStore.ClusterNode/ClusterNodeOptions.cs
```diff
@@ -152,6 +152,8 @@ public class ClusterNodeOptions : IOptions
         public int WorkerThreads { get; set; }
         [ArgDescription(Opts.ProjectionsQueryExpiryDescr, Opts.ProjectionsGroup)]
         public int ProjectionsQueryExpiry { get; set; }
+        [ArgDescription(Opts.FailOutoforderProjectionsDescr, Opts.ProjectionsGroup)]
+        public bool FailOutoforderProjections { get; set; }
```
jageall: Nit: to keep the terminology the same as projections, I think this should be FaultOutOfOrderProjections, also the casing is inconsistent

src/EventStore.Core/Util/Opts.cs
```diff
@@ -130,6 +130,9 @@ public static class Opts
         public const string ProjectionThreadsDescr = "The number of threads to use for projections.";
         public const int    ProjectionThreadsDefault = 3;
 
+        public const string FailOutoforderProjectionsDescr = "Fail the projection if the Event number that was expected in the stream differs from what is received. This may happen if events have been deleted or expired";
```
jageall: Again, use fault not fail

src/EventStore.ClusterNode/ClusterNodeOptions.cs
riccardone: done

src/EventStore.Core/Util/Opts.cs
riccardone: done

PR #1795, opened 2018-12-13
Correct ExpectedVersion values

ChrisChinchilla:
This was also reported here - https://github.com/EventStore/EventStore/issues/1441

But I was reviewing something for docs and found the same inconsistency, if it is indeed an error, it seemed quick to fix, but maybe it's intentional.


prolic:
same as here? https://github.com/EventStore/ClientAPI.NetCore/pull/34/

hayley-jean:
ExpectedVersion.EmptyStream has been marked obsolete in PR #1974 and will be removed in the next version.

PR #1821, opened 2019-01-07
Best effort server side filtering of $all stream

Yspadadden:
This pull request add a best effort server side filtering capability to the EventStore when reading the `$all` stream. It should solve the issue: #718 . 

The changes are based on the OSS 4.1.1-hotfix1 so they currently do not merge cleanly to master. 

## Problem to solve

Reading from the `$all` stream is by far the fasted way for getting events out of the EventStore. But this has the issue of returning all events, even ones that the client is not interested in. This includes system events, but also user events this particular client is not interested in. 

Using EventStore projections to create a "view" of events a particular client is interested is an option, but it can be very slow. In our tests, it is usually faster to read the `$all` stream and live with the overhead than to read a projection stream. Also, using such a projection means we increase the `$all` stream size by a factor of 2, because each event now has a corresponding link event in the projection stream. This means that the projection approach is an all or nothing solution. As clients that still read the `$all` stream now become even slower. 

## Proposed solution

The proposed solution is to add "best effort" server side filtering to the `$all` stream. Best effort, because we do not want to break the EventStore protocol. See below. 

The proposed change adds an optional field to the `ReadAllEvents` message which contains the list of desired event types. This change is backwards compatible. Clients that do not know about this field, will not send it, and, since it is optional, the message is still valid. 

## Why "Best Effort"

It would be simple to implement a hard filter on the server side. But this would lead to an unbounded execution time for reading a single batch of events. In the extreme case, where there are no valid events in the store at all, the EventStore would have to read the whole database before returning an empty batch of events. This will most likely lead to a timeout on client side. 

In addition, returning an empty batch is interpreted as end-of-stream, by some parts of the code. To preserve backwards compatibility, we need to return at least one event, if we are not at the end of stream. 

## Implementation

The main change is a new parameter for all `ReadAllEvents...` methods. The new parameter is an instance of a new `StringFilter` class. This class is responsible for deciding, if a given string is in the whitelist or not. 

The performance of the `StringFilter` class is important to prevent reduced performance of the `$all` stream. For this reason, the `StringFilter` implements different strategies for filtering, depending on the size of the whitelist. For example, if the whitelist is empty and therefore all events should be returned, the code becomes a static `return true`. 

We also increase the window of events we consider for returning a single batch. Because we might be filtering a lot of events, if we only ever read `batchSize` events from the all stream, we would usually return empty or nearly empty batches. But, we should at least try to return full batches of events, if possible, while keeping a not-unbounded execution time. 

## Known issues

* Code does not merge cleanly. 
* Only implemented for `ReadAllEventsForward`. Needs to be implemented for `..Backward` as well.


ahjohannessen:
Why not also consider eventId filter as well? E.g. using `*` in ["customer-*”, “product-*”, “some_singleton”] to get all customers, all products and some singleton stream.
Seems more useful to me, in general.

Yspadadden:
@ahjohannessen Is the EventId filter a component that I could use or is this more of a concept? Adding the "*" as wildcard to the filter should be possible. I will have a look at it next week. 

As for the other comment. So you would suggest leaving the Message `ReadAllEvents` as is and introduce a new Message `ReadAllEventsFiltered` or something. This might include fields like `allowedEventTypes` and probably all of the fields of `ReadAllEvents` but might also include something like `maxSearchWindow` which would indicate how many events we should look at before giving up and returning a non-full batch of events. Also, this message might have the semantic to be able to return a completely empty batch of events to indicate no matching events where found in the window. 

I like this idea. But I will need some time to implement this. Hopefully I'll get it done next week.

ahjohannessen:
>@ahjohannessen Is the EventId filter a component that I could use or is this more of a concept? Adding the "*" as wildcard to the filter should be possible. I will have a look at it next week.

The filtering could be a bit more sophisticated wrt. inclusion, exclusion and system events, as well. But in order to keep things simple, one could do the majority of filtering with inclusion on server side and handle the minor cases of exclusion on client side.

>ReadAllEventsFiltered or something

Yes, something like that. You would be able to specify clearer semantics wrt. filtering and results.

ahjohannessen:
@Yspadadden As far as I can tell, you have not introduced ability to filter `streamId/eventId`, right?

Yspadadden:
@ahjohannessen No I didn't. Sorry, I seem to have misunderstood you. Filtering by eventId didn't make sense to me. After all they are usually UUIDs, so what would a filter for a UUID look like or achieve? 

Is there an ability by the clients to set their own eventIds? Is that something that is done? Then maybe it would make sense. 

Now, filtering by streamId would make more sense. Most straight forward way would be to add `repeated string allowed_stream_ids = 8;` to the message. Same semantics with regexes or plain strings. The question would then be how to combine this filter with the event_type filter. Simplest way here would be an event has to match both the event_type filter and the stream_id filter to be returned. Would that be a workable solution? 

Another option would be to have some sort of query language. But this is really outside my expertise. 

So, should I add stream_id and event_id filters as outlined above?

ahjohannessen:
@Yspadadden When I referred to `EventId` in an earlier comment, I really meant `StreamId` :)

ahjohannessen:
>Simplest way here would be an event has to match both the event_type filter and the stream_id filter to be returned. Would that be a workable solution?

Would it not make more sense to match when any of them match? Doing exclusion on the client side for things that should match both?

Yspadadden:
Of course`OR`ing is also an option. The rest of the implementation outline would be ok? So an additional field etc.

ahjohannessen:
@Yspadadden I think so, however I have no say in what is ok, just sprinkling ideas as I think this could be useful. Someone from the team should rather provide feedback :)

Yspadadden:
It is an interesting option. @shaan1337 any opinions?

daMupfel:
Could we have feedback for this pull request? Would be very much appreciated :).

ChrisChinchilla:
Hey @Yspadadden thanks for this! I'm trying to get docs updated to reflect this new feature, and I must admit, I'm a little lost in the code, maybe you can just help me out and it will be much quicker 😆.

So, is this for the HTTP API and the .NET API, or just one of them?

In which case, how will the endpoint / method look when a developer uses it? How do they specify what they want to filter by etc?

Thanks for your help, then I'll get it written up.

Yspadadden:
@ChrisChinchilla Let me attempt to summarize. 

The change adds a command to the protobuf client-server protocol. It is intended to be used by a protobuf client like the .net client. However, we did not actually implement the .net client changes as we don't use it. The server changes are currently in production in our environment and therefore we have some confidence that they work. But we are a Java shop and therefore cannot test the .net client with the same thoroughness. 

What we did is clone the ReadAllEvents command as ReadAllEventsFiltered: 

```
message ReadAllEventsFiltered {
    required int64 commit_position = 1;
    required int64 prepare_position = 2;
    required int32 max_count = 3;
    optional int32 max_search_window = 4;
    required bool resolve_link_tos = 5;
    required bool require_master = 6;
    repeated string allowed_event_types = 7;
}
```

It has the same arguments as the ReadAllEvents command but has an additional string paramter `allowed_event_types` which can be repeated and may also contain regexes. So one way to specify the filter is to add a list of all event types you want to read. Another is to add one or more regexes that should match the event types you want to read. You can also mix and match. Each item in the list is interpreted accordingly. 

Now the arguments `max_count` and `max_search_window` might need some explanation. `max_count` is the same as for the standard read command. The server will return this many events at maximum. It might return fewer, if the stream is at is end or contains less than `max_count` events. 

When reading the `$all` stream with filtering we would normally return less than `max_count` events. Say f.e. you want 5 events max and only events of type `$UserCreated`. The whole store contains one `$UserCreated` event, then 5 other events, then another 4 `$UserCreated` events. If we just read `max_count` events from the stream, filter out all the ones that don't match and then return to the client we return a list containing exactly one `$UserCreated` event. We do not fill the window the client wants. 

To fix this we introduced the `max_search_window` parameter. This is the maximum number of events the server will look at while attempting to fill the window of `max_count` events to return to the client. The server will stop when either the window is full, or the `max_search_window` has been reached. 

In the example above we could pass `20` as the `max_search_window` parameter. Then the server would read events until it either has filled the client requested window or reaches 20 events. In this case it would read the first `$UserCreated` event, then read and discard the 4 other events, then read the next 4 `$UserCreated` events. At this point it would stop, as it now has fulfilled the client request of 5 events. 

What these parameters should be in real life depends heavily on the sparseness of the data the client wants. We found that using `4096` (the maximum the server will allow) for `max_count` and `20 * 4096` for `max_search_window` gives us nice performance while still staying in the 30 second timeout window of the client. 

## Java client

We did implement these changes in our preferred Java client. See [this diff](https://github.com/BertschiAG/esjc/commit/16d5d4a5306485ca98de668d5c2d3c438239ba62) for the implementation details. Basically we just need to send the correct protobuf message to the server and parse the result accordingly. The basic classes for using the protobuf are auto generated.

ChrisChinchilla:
Thanks @Yspadadden amazing explanation! I attempted a draft here - https://github.com/EventStore/documentation/pull/352

Which abstracts quite a lot of what you said. I also realise that a lot of the client API docs need some updating, this is in progress, but in the meantime, let me know if I've missed anything important.

Yspadadden:
@ChrisChinchilla Thanks for the write-up. It contains every relevant information, I think, but is much shorter than my writeup 👍 . I just left one small comment.

ChrisChinchilla:
Great thanks @Yspadadden now I'll just wait for this to be merged, and then merge the docs :)

jen20:
Hi @Yspadadden! Thanks for opening this PR, it's functionality we intend to integrate. Since there's been some divergence since this was opened, @mat-mcloughlin is working to rebase on top of the latest release in order that it can be merged - we'll ensure that you are credited with the work here, even though it's likely that we'll move this into a new pull request. We'll update here as this progresses.

mat-mcloughlin:
New Pull request:

https://github.com/EventStore/EventStore/pull/1972

jageall:
Closing this PR as the changes have been moved into this one #1972 with some additions to support read backwards and subscriptions

PR #1842, opened 2019-01-28
Adds a maximum index level for automatically merging ptables

jageall:
- Adds a startup option to set the maximum auto merge level
    which defaults to int.MaxValue to keep the current behavior
    of automerging all levels (after this, the list would break
    anyway)
  - Changes the indexmap to store the maximum automerge level
    reducing the automerge level is supported, but increasing
    it requires an index rebuild and will throw an exception
  - Adds an endpoint to trigger a manual merge via an http post
  - Manual merge will merge all the current ptables at the maximum
    automerge level

Signed-off-by: James Geall <jageall@gmail.com>


src/EventStore.Core/Index/IndexMap.cs
```diff
@@ -394,7 +415,40 @@ public void SaveToFile(string filename)
             }
         }
 
+		//TODO: this shouldn't really be called manual table, come up with a better name
```
jen20: Do we need to resolve this before merge?

src/EventStore.Core/Index/IndexMap.cs
```diff
@@ -409,23 +463,64 @@ public void SaveToFile(string filename)
             Ensure.Nonnegative(commitCheckpoint, "commitCheckpoint");
 
             var tables = CopyFrom(_map);
-            AddTableToTables(tables, 0, tableToAdd);
+            AddTableToTables(tables, 0, tableToAdd); // only when is automatic merge
```
jen20: This comment doesn't necessarily seem to be reflected in the code?

src/EventStore.Core/Index/TableIndex.cs
```diff
@@ -222,25 +229,77 @@ public void AddEntries(long commitPos, IList<IndexKey> entries)
...
+                TryProcessAwaitingTables(commitPos, prepareCheckpoint);
+            }
+        }
+     
+        public Task MergeIndexes()
+        {
+            TryManualMerge();
+            return Task.CompletedTask;
+
+            //return Task.Run(() =>
```
jen20: This can probably be nuked?

src/EventStore.Core/Index/TableIndex.cs
```diff
@@ -222,25 +229,77 @@ public void AddEntries(long commitPos, IList<IndexKey> entries)
...
+	        lock (_awaitingTablesLock)
+	        {
+		        
+		        var (maxLevel, highest) = _indexMap.GetTableForManualMerge();
+		        if (highest == null) return; //no work to do
+
+				//These values are actually ignored later as manual merge will never change the checkpoint as no
+				//new entries are added, but they can be helpful to see when the manual merge was called
+				//because of the way the "queue" currently works (LIFO) it should always be the same
+			    var prepare = _indexMap.PrepareCheckpoint;
```
jen20: Indentation here looks weird given the return above?

src/EventStore.Core/Index/IndexMap.cs
jageall: place holder comment left over from when I was figuring out how to break it up. will delete

src/EventStore.Core/Index/TableIndex.cs
jageall: probably a lack of .editorconfig and my editor liking tabs if there isn't one. will fix

src/EventStore.Core/Index/IndexMap.cs
jageall: no, i changed the behavior so the comment became irrelevant, know it does get a table for manual merge. will delete

src/EventStore.Core/Index/TableIndex.cs
jageall: There is an outstanding item over surfacing this in the UI, but that doesn't need to happen to merge this. I suspect it will get nuked when I surface it in the admin page

ChrisChinchilla:
@jageall and @jen20 Does this need docs changes?

jen20:
@ChrisChinchilla It will want a section about it - probably best to collaborate with @jageall on how to explain this feature.

ChrisChinchilla:
@jageall Can you give me a quick summary of the effect of the changes? Looking through code I also think I'll add some more code documentation might be useful for the autogenerated code docs.

PR #1854, opened 2019-02-08
Fixes issues with protobuf lock contention

jageall:
- Retries 12 times. This will add a maximum of 1 min to startup if
 there is heavy contention
 - Uses a static field initializer, this will help make sure that
 for most applications this code will be run very early in it's
 lifecycle, probably before any multithreading happens
 - if we can't initialize inside 1 minute, there is probably something
 very broken in the client app so we will always throw the error when
 calling create
 - this issue should completely go away if we internalize protobuf


jen20:
Waiting on CI for this, and then will merge.

PR #1855, opened 2019-02-08
Change post scavenge merge process to ignore startFromChunk

lscpike:
Whilst doing a lot of scavenging of large databases we've come across the following; if you stop a scavenge and then restart it (using startFromChunk), the chunks scavenged the first time will not be merged. This is a pain as it requires the whole database to be scavenged in one to allow the merge of chunks at the start. 

Therefore, I've raised this small PR to ignore the startFromChunk for the merge phase of a scavenge.

### DIFF
--- a/src/EventStore.Core.Tests/TransactionLog/Scavenging/Helpers/FakeTFScavengerLog.cs
+++ b/src/EventStore.Core.Tests/TransactionLog/Scavenging/Helpers/FakeTFScavengerLog.cs
@@ -22,6 +22,7 @@ namespace EventStore.Core.Tests.TransactionLog.Scavenging.Helpers
         public event EventHandler<EventArgs> CompletedCallback;
 
         public IList<ScavengedLog> Scavenged { get; } = new List<ScavengedLog>(); 
+        public IList<ScavengedLog> Merged { get; } = new List<ScavengedLog>(); 
         public IList<IndexScavengedLog> ScavengedIndices { get; } = new List<IndexScavengedLog>(); 
 
         public void ScavengeStarted()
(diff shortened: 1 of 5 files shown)


shaan1337:
@lscpike thanks very much for your PR. I've not looked at it in detail yet, just a question, I'm probably missing something:
What's the reason the merge phase cannot handle merging of chunks from a starting chunk number when the scavenge has been interrupted and restarted?

shaan1337:
@lscpike I've done the following test:
1. Create 60 chunk files mostly filled with a stream called `a`.
2. Delete stream `a`
3. Start scavenge with `startFromChunk=20`
4. Interrupt the scavenge when reaching `chunk 30`. So `chunks 20-30` have been scavenged but not merged
5. Start another scavenge with `startFromChunk=20`
6. This resulted in `chunks 20-60` being merged into a single chunk: `chunk-000020.000002`, so chunks 20-30 have effectively been merged.

I'm probably missing something when you're saying: `the chunks scavenged the first time will not be merged`. Can you please clarify?

By any chance, do you have some stats of the time penalty involved with the change for a large database? It doesn't look like it'll be too much since only the chunk header is inspected to calculate the size during merging. I'm thinking about a scenario where for example, we have 30k chunks and startFromChunk = 25000. We would need to inspect 25k chunks during the merge phase.

shaan1337:
Just an idea that came up: we could alternatively add a `POST` parameter, `onlyMerge` which would skip the scavenge phase.

lscpike:
So an example:
- Same 60 chunk db with single deleted stream.
- Start scavenge from 0 (as would be expected behaviour)
- Stop scavenge at 20 (maybe you've got a high load period where you need solid performance, maybe the node got restarted for a server patch, etc.)
- Start scavenge from where you stopped it (chunk 20)
- Scavenge finishes.
- You now have chunks 20 - 60 merged into 1 chunk, but 0 - 19 are still tiny individual files.

Currently you'd need to restart the scavenge from 0 to merge 0 - 19. This would be fine for this trivial example, but a real database with billions of streams and a per chunk scavenge time of 1 minute this isn't very feasible.

An onlyMerge option would work, but requires the user to figure out they need to do so. This PR means that it will always attempt to merge tiny chunk files without user intervention. There is no downside to always attempting to merge and no reason a user would not want small chunks merging. If chunks 0-19 have already been merged then it's very quick to decide not to do anything (just a file size check.)

lscpike:
To answer your specific question on 25k chunks. Yes we would need to do that work, but the minutes? it would take are irrelevant compared to days it takes to scavenge a database that size.

shaan1337:
@lscpike thanks very much for all the clarifications, looks good :+1:

PR #1861, opened 2019-02-12
Add warn exception to logger interface

avish0694:


src/EventStore.ClientAPI/Common/Log/ConsoleLogger.cs
```diff
@@ -51,6 +51,17 @@ public void Debug(Exception ex, string format, params object[] args)
             Console.WriteLine(Log("DEBUG", ex, format, args));
         }
 
+        /// <summary>
+        /// Writes a Warn message to the logger
+        /// </summary>
+        /// <param name="ex">A thrown exception.</param>
+        /// <param name="format">Format string for the log message.</param>
+        /// <param name="args">Arguments to be inserted into the format string.</param>
+        public void Warn(Exception ex, string format, params object[] args)
```
shaan1337: This function can be removed since EventStore.ClientAPI.ILogger does not expose warn functions:
`public void Warn(Exception ex, string format, params object[] args)`

PR #1866, opened 2019-02-14
Add and enforce code style via .editorconfig

jen20:
A common complaint with an expanded team has been that much of the code differs in formatting and style, and that adopting a "states rights" model has not served us well. This commit attempts to rectify that by committing a .editorconfig file, and reformatting all the code to use this.

This is an opportune moment to do this, as no new code has gone in since the release of major version 5.0.0.

All major editors used in the team have support for .editorconfig in one form or another:

- Rider via the EditorConfig plugin (installed by default)
- Visual Studio (2017 and above) has native support - older versions have a  plugin available in the marketplace.
- VS Code via the EditorConfig for VS Code plugin
- Sublime Text has an extension available - editorconfig-sublime
- vim has a plugin available - editorconfig/editorconfig-vim.

There is a valid discussion to be had about whether this diff is too large - let's have that and decide on a course of action (which may involve modified `editorconfig` settings to what I have here as a starting point).


jen20:
I just rebased this and will merge pending CI.

jen20:
Curiously the build failed because of hitting the (configured) timeout in Azure Pipelines - I've added a second commit that increases it from 20 to 30 minutes to try to stop this kind of false negative.

PR #1882, opened 2019-03-05
Ignore IsJson property in $by_correlation_id projection

### LINKED ISSUE #1875: Problem with $by_correlation_id projection
Hi,

I found the problem with $by_correlation_id in ES version 5, when using this library to append events through HTTP API (https://groups.google.com/forum/#!topic/event-store/1_e1yZfeWP0). The library sends json with escaped quotes in data and metadata, so the projection is not working.

Is it problem of the library or EventStore? The both formats are valid jsons I think.

Thanks,
Jakub Koci

shaan1337:
Fixes #1875

**Issue**
Relying on `IsJson` to determine if the event's metadata is JSON is not appropriate. For e.g. this PR shows one example where a JSON string is not marked as `IsJson=True`: https://github.com/EventStore/EventStore/pull/1881

**Resolution**
This PR simply removes the check. The code will then try to parse the metadata as JSON and continue if successful.

**Upgrading EventStore with this PR may require a reset of the $by_correlation_id projection since the projection may fault with "An event emitted in recovery for stream <$bc-stream> differs from the originally emitted event"**

### DIFF
--- a/src/EventStore.Projections.Core/Standard/ByCorrelationId.cs
+++ b/src/EventStore.Projections.Core/Standard/ByCorrelationId.cs
@@ -79,8 +79,6 @@ namespace EventStore.Projections.Core.Standard {
 			newState = null;
 			if (data.EventStreamId != data.PositionStreamId)
 				return false;
-			if (!data.IsJson)
-				return false;
 
 			JObject metadata = null;
 


Narvalex:
This change is the reason for issue #1938 .
It breaks when the following projection is enabled and running:
``` js
fromAll()
    .when({ 
        $any: (s, e) => {
            if (e.streamId.substring(0, 1) === '$')
                return;

            linkTo('all', e);
       }
})
```
I checked it twice.

PR #1887, opened 2019-03-07
Use absolute path as mutex name when starting up node

### LINKED ISSUE #1665: Couldn't acquire exclusive lock on DB at '.\ESData' on Windows on a second running instance
The issue is not about clustering. Is more about multi tenancy, like mentioned here #907. Maybe this is also related to #40 .
Steps to reproduce:
1. Unzip the latest database executables (version 4.1.1 hotfix) on two diferent directories, like
 - Instance 1: C:\Databases\FolderOne\
 - Instance 2: C:\Databases\FolderTwo\
2. Run the first instance from the cmd passing the following params, [like in the docs](https://eventstore.org/docs/server/index.html): `EventStore.ClusterNode.Exe --db .\ESData`
3. Run the second instance from cmd the same way as before  `C:\Databases\FolderTwo\EventStore.ClusterNode.Exe --db .\ESData`

shaan1337:
Fixes #1665

### DIFF
--- a/src/EventStore.ClusterNode/Program.cs
+++ b/src/EventStore.ClusterNode/Program.cs
@@ -84,7 +84,11 @@ namespace EventStore.ClusterNode {
 			var dbPath = opts.Db;
 
 			if (!opts.MemDb) {
-				_dbLock = new ExclusiveDbLock(dbPath);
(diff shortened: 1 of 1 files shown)


Narvalex:
Thanks @shaan1337 !

PR #1902, opened 2019-03-27
use the 5min default if no override was specified

mcollins4551:
Currently we're always overriding the default 5 minute expiration with 0 minutes if options.ProjectionsQueryExpiry has not been set.  

This was causing transient projections is our environment to expire before they were finished, even after less than a minute.  Since the projection was getting disposed of too early, it was causing the projection to fail and we were seeing InvalidOperationException: Disposed errors in the log from the V8ProjectionStateHandler


shaan1337:
@mcollins4551 hello, thanks very much for your PR!

The expected behaviour should be that the projections query expiry time should be set to the default (5 mins) if it has not been specified in the config or on the command-line.

Due to a missing assignment, the value would be set to 0 mins. I've opened another PR here: #1921 with the fix and added you as co-author.

PR #1906, opened 2019-04-04
stop throwing unhandleable exceptions

dunkymole:
Per the title, there was an unhandlable exception being thrown from within the catch-up phase of EventStoreCatchUpSubscription which isnt awaitable to consumers via StartAsync or as invoked upon reconnection. Furthermore this PR adds a backstop observation of async exceptions in order that any unforseen exceptions on that path also terminate the subscription.


dunkymole:
OK, I was a bit hasty with that initial commit and I broke the tests. The tests now pass again and the case where we have an exception during the catch-up phase *after a re-connection* is now covered and guaranteed to not throw unhandleable exceptions. 
Sadly we are now in a position where the tests are happy by virtue of an internal method raising an async-exception which the final API consumers dont have access to so I consider this still broken.

dunkymole:
Closing in search of another approach. I feel like we are being drawn down the wrong path by the needs of the tests.

PR #1907, opened 2019-04-04
remove uncatchable exception

dunkymole:
EventStoreCatchUpSubscription threw an exception in order to error out the result of StartAsync - this API is internal and in all cases except the tests, the result is thrown away meaning that the exception which was thrown went unobserved. By removing the expectation on that exception from the tests while preserving all the valid expectations around errors being notified through the API, its possible to remove the problematic throw.


ylorph:
Apologies for taking so long to get back to this.
Closing this pull request as it's the legacy API & no longer in this repository

PR #1914, opened 2019-04-21
Don't MarkForDeletion a chunk until all references to it have been swapped

lscpike:
Don't MarkForDeletion a chunk until all references to it have been swapped out. This avoids it still being accessible for reads resulting in FileBeingDeletedException.

I'm 99% we've been hitting this whilst scavenging. The scenario is:
- Previously scavenged and merged multiple chunks.
- Scavenging multiple threads.
- 1 thread finishes re-scavenging a merged chunk (chunk "#1-5") and swaps in in the DB (TfChunkManager).
- TfChunkManager swaps Chunk index 1 for the new chunk and deletes the old chunk.
- Note indices 2-5 still point at the deleted chunk.
- Another thread is scavenging a different but close proximity chunk.
- The second thread needs to check for stream deletion of a stream so reads the metadata which happens to be in chunk "#2".
- BOOM! FileBeingDeletedException

This is obviously a problem for any read operation on the chunk, not just the scavenge process.

The fix is simple; don't MarkForDeletion until all references in the DB are swapped.

### DIFF
--- a/src/EventStore.Core/TransactionLog/Chunks/TFChunkManager.cs
+++ b/src/EventStore.Core/TransactionLog/Chunks/TFChunkManager.cs
@@ -234,17 +234,26 @@ namespace EventStore.Core.TransactionLog.Chunks {
 				}
 			}
 
-			TFChunk.TFChunk lastRemovedChunk = null;
+			TFChunk.TFChunk previousRemovedChunk = null;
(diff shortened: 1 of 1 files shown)


shaan1337:
@lscpike looks good. thanks for your contribution! we have seen at least one other case where the bug occurred.

PR #1918, opened 2019-04-25
Add regular file flushes whilst writing out scavenged chunk

lscpike:
After our scavenge performance improvements we've been experiencing intermittent commit timeouts and general spikes in write time during a scavenge. This is caused by the large file write direct from memory at the end of a chunk scavenge or chunks merge. 

The file is written quickly and then flushed/closed resulting in up to 256MB being put on the disk write queue. We can see in our monitoring that the disk transfer latency spikes resulting in the slow writes for new events.

This PR adds a regular flush to the scavenge and merge process so that the individual flushes are smaller. Looking at stats when testing locally this results in avg write size of ~60KB instead of 1MB. 

I've run a custom build with this change (plus #1914)  and run a full scavenge on one of our large databases. There was still occasional impact, but there were no commit timeouts and write times were much more stable.

I played with adding a Thread.Sleep after each flush to throttle the write even more. This gave good results locally, but added many seconds to the chunk write. I left it out as I'm not sure it's necessary.

I'm sure there's a similar issue with the index scavenge and merge processes. I haven't touched them here.

### DIFF
--- a/src/EventStore.Core/TransactionLog/Chunks/TFChunkScavenger.cs
+++ b/src/EventStore.Core/TransactionLog/Chunks/TFChunkScavenger.cs
@@ -30,6 +30,7 @@ namespace EventStore.Core.TransactionLog.Chunks {
 		private readonly int _threads;
 		private const int MaxRetryCount = 5;
 		internal const int MaxThreadCount = 4;
+		private const int FlushPageInterval = 32; // max 65536 pages to write resulting in 2048 flushes per chunk
 
 		public TFChunkScavenger(TFChunkDb db, ITFChunkScavengerLog scavengerLog, ITableIndex tableIndex,
 			IReadIndex readIndex, long? maxChunkDataSize = null,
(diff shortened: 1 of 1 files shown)


gregoryyoung:
This brings up some interesting questions of *where* this kind of thing should happen. As example its perfectly legitimate that they could end up on different disks. My guess is that overall more knowledge is required to implement this in a reasonable way.

lscpike:
Are you referring to raid set ups? On windows and hardware raid it appears the disk queues are per file system partition rather than physical disk. It could be misleading me though.

gregoryyoung:
No the possibility of delays from a source code perspective. There are a few such operations that could possibly want a delay/priority added. Scavenge, Index Merge, Checkpoints, Version Upgrade etc. 

It is also that we don't really know enough such as if these things are actually on shared/remote disks etc and what might possibly be conflicting in terms of iops. In other words we might want to centralize the decision so we know what the index is doing when we are deciding what a chunk should do. Does that make more sense? There has been some work towards this already ... think IO scheduling.

lscpike:
I see what you mean. There's definitely value in that. I've already seen a commit timeout from an index dump to disk at the same time as a scavenge write.

In the meantime this change is very useful though!

PR #1920, opened 2019-04-29
Fix assertion regarding # of midpoints cached

shaan1337:
If this line is hit: https://github.com/EventStore/EventStore/blob/master/src/EventStore.Core/Index/PTableConstruction.cs#L514, no midpoints will be cached to the PTable file.

This will make the following assertion fail:
```
if (_version >= PTableVersions.IndexV4 && _count > 0 && _midpointsCached < 2)
```

*Resolution*
Check if there is at least one midpoint cached before applying the `_midpointsCached < 2` assertion.


Lougarou:
The fix (https://github.com/EventStore/EventStore/pull/1920) prevents the server from crashing however we still have not found the root cause

PR #1923, opened 2019-04-30
Adds benchmark project for microbenchmarking

jageall:
- Adds benchmark for message dispatching


jageall:
Closing this as we are moving where we do benchmarks with the next version

PR #1937, opened 2019-05-29
DRY ClusterDnsEndPointDiscoverer creation

bartelink:
This provides a minor tidyup which makes the code more legible wrt how the node preference is applied when using the `EventStoreConnection.Create(ConnectionSettings,ClusterSettings,string)` overload (which bit me, resulting in the need for https://github.com/jet/equinox/pull/135)

### DIFF
--- a/src/EventStore.ClientAPI/EventStoreConnection.cs
+++ b/src/EventStore.ClientAPI/EventStoreConnection.cs
@@ -110,19 +110,7 @@ namespace EventStore.ClientAPI {
 					var clusterSettings = new ClusterSettings(uri.Host, connectionSettings.MaxDiscoverAttempts,
 						uri.Port,
 						connectionSettings.GossipTimeout, connectionSettings.NodePreference);
-					Ensure.NotNull(connectionSettings, "connectionSettings");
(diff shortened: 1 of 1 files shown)


bartelink:
And an aside... while the linked issue is on me, it does highlight a potential for confusion which can equally bite others:

- the `ConnectionSettings` DSL has a `PreferSlaveNode` and a `PreferRandomNode`, as do the two call chains (seeded and not) in the `ClusterSettings` builder
- when you call the `EventStoreConnection.Create(ConnectionSettings, Uri, ...)` overload, the `ConnectionSettings` ones apply
- when you call the `EventStoreConnection.Create(ConnectionSettings, ClusterSettings, ...)` overload, the `ConnectionSettings` ones get dropped

Unfortunately I can't think of a good way to resolve the forces, aside from making an even messier config DSL out of it. (I'd personally settle for removing `PreferSlaveNode` and a `PreferRandomNode` from the `ClusterSettings` builder, but that's a pretty breaking change and is also confusing/incorrect in terms of the semantics)

shaan1337:
@bartelink thank you!

PR #1940, opened 2019-06-05
Running server on .Net Core

### LINKED ISSUE #1650: .NET Core 2.1
I don't want to do a PR for this as there is clearly *a lot* of considerations to be discussed and debated.  I spent a few hours yesterday getting a working .NET Core 2.1 build - I used Jetbrains Rider 2018.1.2 but looks to be buildable from VS2017 as well.  

https://github.com/BlizardPuppet/EventStore/tree/netcore21

Some of my initial changes/observations/questions:

* The big API gap was `UrlTemplate` - Copied it verbatim from .net framework source with very small edits (e.g. delocalization of Exception messages) - probably needs a clean up 
* Converted all EventStore.sln csprojs to the new VS2017 format - this mean dropping < VS2017 support
* Using package references (nuget) instead of checked-in lib dlls (upgrades necessary for .NET Core support include `protobuf-net` 2.0.X -> 2.2.X, `Newtonsoft.Json` 6.X -> 11.X, both didn't require any code changes) 
* Some minor .csproj tweaks to ensure it works nicely for multitargetting .NET Core + net46.
* Fixed up `FlushSafe` to work across .NET Core on Windows and Full Framework builds - it seems `FileStream` now exposes the `SafeFileHandle` instead of having to use reflection to find it.
* Added new 'dotnet' "visual studio version" to build.ps1 script inc auto-detect.  We can probably drop < VS2017 support for the C# code but unsure about the js1 build - personally, I have never built this.
* dotnet version smoke tested inc js1 usage in projections.  Joined a cross-build cluster (2 nodes `dotnet`, 1 node `net46` - manual smoke testing passed)
* Unit tests need looking at, not sure why they are not runnable using standard IDE - I think aim should be to get them running under `dotnet test`
* What do we need to do about `Mono` support?  Can we drop it (eventually) if this path is chosen? 
* haven't tested anything else other than build.cmd quick and build from Rider (windows)
* Need to look at building + running under Linux

riccardone:
This PR run EventStore as is today (v5.0.1) on .Net Core 2.2

I ported UriTemplate and UriTemplateMatch into a new local project called EventStore.UriTemplate to let the current Http service layer running 
Test projects are still running on .netframework to avoid rewriting all tests (nunit run on fw)

Automation fails and need to be reconfigured/simplified as it relies on mono with dedicated builds

Closes #1650


martinothamar:
Hey, any ETA on this being merged?

jageall:
Unfortunately, UriTemplate and UriTemplateMatch have been copied from reference source and are not licensed by microsoft anywhere that i can find that allows straight copying the code into a codebase in such a manner.

We do plan to move over to .net core in the next major release with initial release candidates planned for the end of November this year.

riccardone:
@jageall the .net Framework is distributed under MIT license. This license allow to copy and change the code as long as the original license ref is maintained. I've just added the license file. 

As long as the code is OSS there are no problems on doing in that way. That porting is necessary to run the EventStore OSS code on .net core without rewriting the whole http section. 
That is basically why licenses like MIT and GPL has been made for. 

I will also open a PR on the .net standard repo with but I suggest to keep this bunch of classes in a local project to avoid problems with API changes made by others.

riccardone:
https://github.com/microsoft/dotnet/blob/master/LICENSE

Issue #1938, opened 2019-05-31
$by_correlation_id projection faulted after upgrade to Event Store 5.0.1 


Narvalex:
Description:
The '$by_correlation_id' projection faults after upgrading from v5.0.0 to v5.0.1, even after following the upgrade instruction found in the official blog (we stopped the projections in version 5.0.0 before the rolling upgrade).

1. Follow the upgrade instructions from version 5.0.0 to 5.0.1 found in the blog.
2. After some time (it happened al least 3 times to us) the projection will fail, like when not following the update instructions.

Expected behavior
The projection should not fail, even after reseting the projection $by_correlation_id

Actual behavior
The projection fails, and we can not recover the $bc-corrId streams with, for instance, reseting the projection.

Screenshot:
![image](https://user-images.githubusercontent.com/11648387/58730567-8761cc80-83ba-11e9-8241-d13a4b1d2da5.png)

EventStore details:
 - EventStore server version: 5.0.1

 - Operating system:
Windows 10 pro, v 1809

Additional context
The database is small, like 11.000 events.

Narvalex:
We are willing to provide the full database to reproduce the error, as this happens consistently every time we try to upgrade from v5.0.0 to v5.0.1

Narvalex:
We have just tried to upgrade **without** following the instructions (stop the projections before upgrading) and now works, aparently... wich is wierd.

Narvalex:
> We have just tried to upgrade **without** following the instructions (stop the projections before upgrading) and now works, aparently... wich is wierd.

Update: It did not work every time. We are rolling back to version 5.0.0 until this issue is fixed.

Narvalex:
I was watching carefully the stream $projections-$by_correlation_id-checkpoint, and **only** after watching it was updated I shutdown the server and proceed to do the rolling upgrade on our single node DB. And It worked! I will close this issue for now, but I will reopen if any issues arise again. Thanks @shaan1337 for your careful observation

Narvalex:
It did happen again. Bellow the screenshot
![image](https://user-images.githubusercontent.com/11648387/58840520-ebdb9080-8633-11e9-8fbd-9907837537b2.png)

Narvalex:
> It did happen again. Bellow the screenshot
This time, I did followed the instructions. I appeared to work. But as soon as I restarted the node, it happened again 😟

shaan1337:
@Narvalex hello, thanks for all the details, we're trying to reproduce this issue.

[...]

PR #1952, opened 2019-06-26
Fix By Correlation Id projection on link event appeared bug

Narvalex:
This is a bug that only happens when link events are written to the database. This is related to issue #1938 .

**Prerequisites:**

- Event Store version 5.0.1 
- All system projections enabled
- The following continuous projection enabled:
``` js
fromAll()
    .when({ 
        $any: (s, e) => {
            if (e.streamId.substring(0, 1) === '$')
                return;

            linkTo('all', e);
       }
})
```

**Symtomps:**
When appending events with correlation id, the stream `$bc-[correlationId]` appears with all the correlated events as expected.
After restarting the node we can see the stream `$bc-[correlationId]` is empty, and or not making progress.

**Root cause:**
The system projection appears to be handling the link events in disruptive way. Before the pull request #1882 link events where ignored. After that merge, link events are not ignored at all.

**Resolution:**
Ignore link events.


pgermishuys:
Hey @Narvalex!
Thanks for taking the time to root cause this aswell as opening a PR.
Just letting you know that we have the issue on our V6 backlog. There will likely be a delay in getting to accepting and merging this PR.

Thank you once again!

Narvalex:
Thanks for your kind words @pgermishuys ! Just letting you know that we have a custom build of Event Store with this pull request merge in production for several month now without any issues. Thanks you all for your hard work in this great database

Narvalex:
Can someone help me with the failed build on Windows please? I cant understand why...

YoEight:
Hi @Narvalex, don't worry about that windows build right now, it's a false negative. With your patch, did you stop having the issue mentioned in #1938 ?

I'm asking because I think #2447 should fix it but it's possible I missed something.

Narvalex:
The #2447 prevents the `$by_correlationId` projection to get in a faulted state, but does not fix the issue, as described in the #1938 thread

PR #1959, opened 2019-07-08
Setting to override http client & setting to enable tls based gossip comms

ChrisMcKee:
Adds the ability to tell GES to use load seeds over HTTPS (as settings and via extension method) 
`SetGossipSeedEndPoints(seedOverTls: true, params IPEndPoint[] gossipSeeds)`

Adds the ability to pass in a custom httpclient or use the built in implementation and override the http client handler `HttpAsyncClient(TimeSpan timeout, HttpClientHandler clientHandler = null)`

The interface addition would duplicate one thats in `EventStore.Transport.Http` which wasn't intentional but from the dependencies I assume its correct to duplicate this rather than create a dependency.

ref: https://github.com/EventStore/ClientAPI.NetCore/pull/29

### DIFF
--- a/src/EventStore.ClientAPI/ConnectionSettings.cs
+++ b/src/EventStore.ClientAPI/ConnectionSettings.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using EventStore.ClientAPI.Common.Utils;
 using EventStore.ClientAPI.SystemData;
+using EventStore.ClientAPI.Transport.Http;
 
 namespace EventStore.ClientAPI {
 	/// <summary>
@@ -32,6 +33,11 @@ namespace EventStore.ClientAPI {
 		/// The <see cref="ILogger"/> that this connection will use.
 		/// </summary>
 		public readonly ILogger Log;
+
+		/// <summary>
+		/// Allows overriding the HTTPClient <see cref="IHttpClient"/>
+		/// </summary>
+		public IHttpClient CustomHttpClient { get; set; }
 
 		/// <summary>
 		/// Whether to use excessive logging of <see cref="EventStoreConnection"/> internal logic.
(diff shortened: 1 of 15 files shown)


gregoryyoung:
re dependency, probably best ...

ChrisMcKee:
I have other changes i'd like to make but I'll see how this goes; and I'm aware there are various changes pending around modules but I'm mostly interested in the client for the time being.

pgermishuys:
Hey @ChrisMcKee,
Thanks for the PR!
This introduces a breaking API change here https://github.com/EventStore/EventStore/pull/1959/files#diff-686da6a6903eb257efa7e72f5675cbdbR29.
One option might be to introduce an overload that accepts the TLS option or introduce a separate method that uses `Uri`.

ChrisMcKee:
This bit `public GossipSeedClusterSettingsBuilder SetGossipSeedEndPoints(bool tlsTerminatedEndpoints = false, params IPEndPoint[] gossipSeeds) {` ?

ChrisMcKee:
force-pushed as I had to rebase + merge the changes in

pgermishuys:
Hi @ChrisMcKee,
What I meant to say that given that an existing user has the following code
```
var settings = ConnectionSettings.Create()
  .KeepReconnecting()
  .KeepRetrying()
  .EnableVerboseLogging()
  .SetGossipSeedEndPoints(new IPEndPoint(IPAddress.Loopback, 1113));
```
Your change would essentially have the client upgrade and have their code no longer build.

ChrisMcKee:
@pgermishuys yeah sorry, I worked that out and lost the change I made rebasing on upstream

pgermishuys:
Hey @ChrisMcKee. Thanks for doing this. My apologies but would you mind splitting this work into 2 separate commits or PRs? Namely the HttpClient and then the gossiping over `https` or `http`.

Sorry for going back and forth on this.

ChrisChinchilla:
@pgermishuys If this was a breaking change, does this need docs updates?

PR #1964, opened 2019-07-12
Fix issue 1950

smartcodinghub:
As the Issue says, EmptyStream should be ZERO.
https://github.com/EventStore/EventStore/issues/1950


hayley-jean:
Thank you for the PR @smartcodinghub 

The ExpectedVersion.EmptyStream was used in a very early version of Event Store where a "StreamCreated" event had to be written to each stream to actually create it.

When this was removed, EmptyStream became the same as NoStream, and was kept for backwards compatibility. It has now been marked obsolete in PR #1974, and will be removed in the next version.

We're going to close this PR, but look forward to seeing you contribute in the future

PR #1972, opened 2019-08-06
[WIP] Implementing server side filtering of events and streams

mat-mcloughlin:
Extension upon the server side filtering referenced here https://github.com/EventStore/EventStore/pull/1821


mat-mcloughlin:
Just an update so we don't forget where we are

So far I've implemented
- `ReadAllEventsForwardFilteredAsync`
- `ReadAllEventsBackwardsFilteredAsync`
- `SubscribeToAllFilteredAsync` (still requires testing)

and I've just got 
- `SubscribeToAllFilteredFrom`

left to do along with these general tasks

- [ ] Document methods correctly
- [ ] Investigate the HTTP client and any changes that may need to be made there
- [ ] Benchmarking as it's touched some of the main read/subscribe methods.
- [ ] Check interaction between old server/new client and vice versa

I think it's also worth examining the API as there's a lot going on in the new filtering methods that can easily be confusing. So here are some examples.

### EventFilter class

```
// Create event filter.
var eventFilter = EventFilter.Create()
    .WithEventPrefixFilter("someevent")
    .WithStreamPrefixFilter("somestream")
    .WithEventFilter(new Regex(@"someregex"))
    .WithStreamFilter(new Regex(@"someregex"))
    .ExcludeSystemEvents()
    .Build();
```

The new event filter class allows users to build a filter using the above extension methods. This is then passed into the new filtering methods.

### ReadAllEventsForwardFilteredAsync & ReadAllEventsBackwardsFilteredAsync

```
// Read events from all stream.
var readEvents = await connection.ReadAllEventsForwardFilteredAsync(
    position: Position.Start, 
    maxCount: 4096, 
    resolveLinkTos: false,
    eventFilter: eventFilter,
    maxSearchWindow: 1000);
```

The last argument on this method is the max search window. This specifies how many events the read will search through before returning the slice. For example in the code above it will search through 10,000 events and even if the filtered list hasn't reached 4,096 it will return. However, it still won't return more than 4,096 events in any given slice.

### SubscribeToAllFilteredAsync

```
// Subscribe to all stream.
await connection.SubscribeToAllFilteredAsync(
    resolveLinkTos: false, 
    eventFilter: eventFilter, 
    eventAppeared: (s, e) => Task.CompletedTask,
    checkpointReached: (s, p) => Task.CompletedTask,
    sendCheckpointMessageCount: 1000);
```

In the case of subscribe to all we provide an additional delegate that gets called with the position. This is to deal with the fact that when a filter is applied the subscription may read through a lot of events before returning one and we don't want to lose the position if the service crashes. This method gives the user of the api the ability to record the checkpoint even if an event hasn't been found.

The send checkpointMessageCount argument defines how often this delegate will be called.

mat-mcloughlin:
Also need to implement the imbedded code too

pgermishuys:
Fixes #1985

mat-mcloughlin:
Closed due to being too large so split into smaller PR's. Replaced with #2009 and #2011

PR #1977, opened 2019-08-20
Use ExpectedVersion.NoStream instead of ExpectedVersion.EmptyStream

avish0694:
Since `ExpectedVersion.EmptyStream` is now obsolete, it gives warning when compiling the project. 
This pull request replaces `ExpectedVersion.EmptyStream` to `ExpectedVersion.NoStream` everywhere in the project.

### DIFF
--- a/src/EventStore.ClientAPI/ExpectedVersion.cs
+++ b/src/EventStore.ClientAPI/ExpectedVersion.cs
@@ -29,7 +29,7 @@ namespace EventStore.ClientAPI {
 		/// <summary>
 		/// The stream should exist but be empty when writing. If it does not exist or is not empty treat that as a concurrency problem.
 		/// </summary>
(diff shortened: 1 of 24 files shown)


src/EventStore.ClientAPI/ExpectedVersion.cs
```diff
@@ -29,7 +29,6 @@ public static class ExpectedVersion {
 		/// <summary>
 		/// The stream should exist but be empty when writing. If it does not exist or is not empty treat that as a concurrency problem.
 		/// </summary>
-		[Obsolete("ExpectedVersion.EmptyStream has been deprecated. Use ExpectedVersion.NoStream instead")]
```
Narvalex: Why the Obsolete decorator is being remove?

avish0694: It was mistake. Fixed! Thanks! :)

PR #1993, opened 2019-09-09
[WIP] Fix Build on Linux w/ Rider

thefringeninja:
The build script works because it sets the FrameworkPathOverride varaible [here](https://github.com/EventStore/EventStore/blob/3b2237a088051e7ab9f40c21a377e67583852d58/build.sh#L113). You can set the MSBuild property in rider but that only affects building, not solution analysis.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -5,4 +5,29 @@
     -->
     <RestorePackagesPath>$(MSBuildThisFileDirectory)dependencies</RestorePackagesPath>
 </PropertyGroup>
+<PropertyGroup>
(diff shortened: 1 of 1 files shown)


bartelink:
Can https://stackoverflow.com/a/57384175/11635 not do the job for you ?

thefringeninja:
Doesn't seem to actually work, will have to investigate further.

thefringeninja:
Ok, after further investigation I see that the folder this file resides in is not even mounted inside the CI container. I will put the fix in #1996 instead.

PR #1996, opened 2019-09-11
Implement Kestrel Http Transport and Server

thefringeninja:
- [x] Compiles
- [x] Open issue on mono/mono re: netstandard problem mono/mono#16826
- [x] Figure out how to run tests against both implementations of `IHttpServer`


thefringeninja:
Unanswered questions:

Kestrel does not like to bind against both `localhost` and `127.0.0.1` unlike `HttpListener`. Is this necessary? Can we get rid of it entirely?

There are a couple of ways to run the same tests against both implementations of `IHttpServer`. The simplest thing to do is use an environment variable and run the test suite twice (or at least against `EventStore.Core.Tests` and `EventStore.Core.Projections.Tests`), but that might be too slow. Another solution is to pull the HTTP tests into a separate folder, make two sets of test projects with different MiniNode configurations, and include these files with a glob pattern in the project file. Thoughts?

PR #2012, opened 2019-09-26
GRPC Client and Server

thefringeninja:
- Migrates solution to dotnet core
- Replaces httplistener with Kestrel
- Implements GRPC transport for Streams, Persistent Connections, Projection Management, and User Management
- New ClientAPI acceptance tests for both the tcp client and the embedded client


ahjohannessen:
@thefringeninja I have been following your work, great stuff 👍 I have started working on gRPC client for Scala that uses a protobuf code generator called ScalaPB. It has a few issues with some of the naming of fields, especially cases with `Empty empty` and oneof’s like `AllOptions all_options` and similar, nothing I can fix by rename those fields, but it would be nice to use your definitions as is. I can outline what I have changed with field names in order to get ScalaPB codegen without naming conflicts below if you like?

ahjohannessen:
@thefringeninja I have a diff here: https://github.com/ahjohannessen/sec/commit/494e6ed4a5f4520b6fc75c4095c7c1016ff38e2d

The reason for changes in field names for `stream_options`, `filter_options` and `all_options` is because there are message types called `StreamOptions`, `FilterOptions` and `AllOptions`, respectively. ScalaPB generates code for `oneof` field names in a special way that conflicts with the already defined message types.

Reason I have changed names for `revision_options` and `count_options` is just for consistency.

The other change is with respect to not using the name `empty` in `Empty empty` usage inside `oneof` as that also causes conflicts in generated code, ScalaPB generates a type called `Empty` for `oneof`'s and that results in errors like `method isEmpty is defined twice`.

ahjohannessen:
@thefringeninja Btw, you could alternatively use `import "google/protobuf/empty.proto";` instead of those `Empty` definitions.

Furthermore, I think it would be nice if you could allow for a setup of a HTTP/2 endpoint without TLS, e.g. `... , o => o.Protocols = HttpProtocols.Http2);` because one can imagine people using something like Envoy or Consul Connect to secure service communication.

thefringeninja:
@ahjohannessen,

Have you considered wrapping what's generated by the java tooling instead? The scala one honestly sounds like it's busted. TBF the dotnet one has problems in this area too. Ultimately, I think I'll have to generate the client in all the supported languages to ensure there are no conflicts anywhere.

I am on the fence on reusing another proto. I did end up importing `google/protobuf/struct.proto` for projections, but I may end up copy&pasting it in the end.

Currently we are using a temporary dev cert so we can run this inside a container. I would love to get rid of this and support non TLS HTTP/2.0; however I am told that it is almost impossible to do from the official grpc clients in other languages. I will have to investigate further.

ahjohannessen:
>Have you considered wrapping what's generated by the java tooling instead? 

I base my Scala client on [fs2-grpc](https://github.com/fiadliel/fs2-grpc), a project that I am involved with too. fs2-grpc also uses [ScalaPB](https://scalapb.github.io) for code-gen, so plain java tooling is a no go and not really interesting.

I could consider only using the runtime parts of fs2-grpc, but the ScalaPB codegen generates code that is more in line with what I like. I am already heavily invested into ScalaPB with hundreds of proto messages for events that are stored in ES and for grpc services as well.

>The scala one honestly sounds like it's busted. 

I can get the Scala tooling to work just fine. It just generates code that conflicts with your namings, It is no problem to get it to work with a few minor adjustments as I pointed out earlier.

>TBF the dotnet one has problems in this area too. Ultimately, I think I'll have to generate the client in all the supported languages to ensure there are no conflicts anywhere.

Even if you guys provide a Java client, I would not use it. The one I build uses functional programming concepts such as cats-effect and fs2 streams and the stuff from java-grpc is just too far away from what I like.

>I am on the fence on reusing another proto. I did end up importing google/protobuf/struct.proto for projections, but I may end up copy&pasting it in the end.

I do not mind, just a friendly pointer :) However, that is not just another random proto, it is from google/protobuf/*.

>Currently we are using a temporary dev cert so we can run this inside a container. I would love to get rid of this and support non TLS HTTP/2.0; however I am told that it is almost impossible to do from the official grpc clients in other languages. I will have to investigate further.

I got the server on your grpc branch running with with non TLS HTTP/2.0 and connect to it java-grpc. I suppose the issue is dotnetcore on MacOS.

ahjohannessen:
@thefringeninja Btw, your work is awesome :) I already have a `subscribeToAll` example working in my Scala client using fs2-grpc 😎

ahjohannessen:
@thefringeninja Perhaps your `bytes id` could be represented as
```
// A 128-bit id, transferred as two 64-bit longs
message Uuid {
  sint64 least_significant_bits = 1;
  sint64 most_significant_bits  = 2;
}
```

It is bit unfortunate that clients for other platforms need to look at the C# implementation to see how `id` is encoded.

ahjohannessen:
@thefringeninja Just a heads up, I got the maintainer of ScalaPB fix the code generator with respect to naming conflicts of `oneof`. So do not worry about how you name your proto defs :)

ahjohannessen:
@thefringeninja I wonder, why not represent `End` for `Position` and `StreamRevison` as you do with `Start`, i.e. as `End` instead of magic values like `-1L` and (`-1L`, `-1L`), respectively?

ahjohannessen:
@thefringeninja Can you consider making `ExpectedStreamRevision`, `Empty` and `Position` top level message definitions.

For example:

`DeleteReq.Options.ExpectedStreamRevision.NoStream(DeleteReq.Empty())`, `TombstoneReq.Options.ExpectedStreamRevision.NoStream(TombstoneReq.Empty())`
`AppendReq.Options.ExpectedStreamRevision.NoStream(AppendReq.Empty())`

could be:

`com.eventstore.client.streams.ExpectedStreamRevision.NoStream(com.eventstore.client.streams.Empty())` or `com.eventstore.client.ExpectedStreamRevision.NoStream(com.eventstore.client.Empty())`.

and:

`ReadReq.Options.Position`, 
`AppendResp.Position`,
`DeleteResp.Position`,
`TombstoneResp.Position`

could be:

`com.eventstore.client.streams.Position` or `com.eventstore.client.Position`

src/EventStore.Rags.Tests/Properties/AssemblyInfo.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
+using NUnit.Framework;
```
pgermishuys: Don't think this is needed.

thefringeninja:
@hayley-jean I don't think so. We need fixes contained in `Microsoft.AspNetCore.TestHost@3.1.0-preview3.19555.2` and that targets `.NETCoreApp 3.1`. That is expected to ship in December so I think we're ok here.

I will update the PR description with all the changes.

PR #2013, opened 2019-10-07
Added histogram measurement around resolving LinkTos

Lougarou:
Added histogram to get breakdown on read times when ResolveLinkTos is true vs normal reads. This will allow to compare the performance impacts of resolving linkTos.

Sample usage: Histogram can be called with curl http://localhost:2113/histogram/reader-linkto -u admin:changeit

Sample run resolving event from by_category stream vs reading from stream directly: 
reader-linkto
[Mean    =       22.645, StdDeviation   =        8.924]
[Max     =       51.053, Total count    =          168]
[Buckets =           20, SubBuckets     =         2048]
reader-streamrange
[Mean    =        0.396, StdDeviation   =        3.039]
[Max     =       37.356, Total count    =          396]
[Buckets =           20, SubBuckets     =         2048]


jageall:
This has been superseded by the work @condron is doing on profiling read times

PR #2016, opened 2019-10-10
Make ResolvedIndexedEvent's event field optional.

YoEight:
This change reflects the reality that `ResolvedIndexedEvent` 's event field **can** be optional in some situations.

For example if you read `$streams` stream and one of the listed streams got deleted, the associated `ResolvedIndexedEvent` resulting from a `ReadStreamEventsCompleted` will only have its `link` defined.

The old behaviour was causing wrong protobuf generation if you use the official `protoc` program. In the example I mentioned, protobuf code generated by `protoc` will throw `MessageNotInitialized` because the `ResolveIndexedEvent` message didn't comply with the protobuf declaration file.

note: I don't understand the edit at the end of the file, I edited the code directly on Github.

### DIFF
--- a/src/Protos/ClientAPI/ClientMessageDtos.proto
+++ b/src/Protos/ClientAPI/ClientMessageDtos.proto
@@ -35,7 +35,7 @@ message EventRecord {
 }
 
 message ResolvedIndexedEvent {
-	required EventRecord event = 1;
+	optional EventRecord event = 1;
 	optional EventRecord link = 2;
 }
 


jen20:
Hi @YoEight! Thanks for the pull request here. As part of #2012 (for Event Store V6) we are switching to a gRPC-backed client protocol, and thus to protocol buffers 3 (rather than the current v2) where everything is optional. I'm not 100% sure of the semantics around changing a required field into an optional field, but I suspect it is not backwards compatible and thus probably not something we'd like to do at this stage. We don't yet have a clear deprecation plan for the current Client API, but we'll put out a post on the mailing list when that comes about.

Hopefully the gRPC-backed protocol makes it easier for your own Haskell client, as [grpc-haskell](https://github.com/grpc/grpc-haskell) will end up implementing a lot of the complex bits around connection management, and subscriptions become much easier to deal with. If you'd like to talk more about that migration path, please feel free to drop us an email!

ahjohannessen:
@YoEight I wonder how I should deal with this scenario in my Scala gRPC client. For now I have choosen to filter it as I do not see the use case to get a link to something that is no longer there. Was it required for you to surface the `ResolvedIndexedEvent` with only a link, if so then why? In scala I do something like this:

```scala
  def mkEvent[F[_]: ErrorM](re: ReadResp.ReadEvent): F[Option[Event]] = {

    val eventOptF: F[Option[EventRecord]] = re.event.traverse(mkEventRecord[F])
    val linkOptF: F[Option[EventRecord]]  = re.link.traverse(mkEventRecord[F])

    // Alternative that selects option of link as EventRecord if event is None.
//    (eventOptF, linkOptF).mapN { (eOpt, lOpt) =>
//      eOpt
//        .map(er => lOpt.fold[Event](er)(li => ResolvedEvent(er, li)))
//        .getOrElse(lOpt)
//    }

    eventOptF >>= { eOpt =>
      linkOptF.map { lOpt =>
        eOpt.map(er => lOpt.fold[Event](er)(li => ResolvedEvent(er, li)))
      }
    }
  }
```

and using fs2 I do something like this:
```scala
val mkEvents: Stream[F, ReadResp] => Stream[F, Event] =
  s => s.evalMap(rr => rr.event.map(mkEvent[F]).getOrElse(none[Event].pure[F])).unNone
```
The last part, `unNone`, filters `None`. This is then used like this:

```scala
 client.read(request, auth(creds)).through(mkEvents)
```

YoEight:
@ahjohannessen In my case, the issue was in my Rust client where the `protobuf` library refused to parse `ResolvedIndexedEvent` because of the field `event` is tagged as required. My Haskell client didn't have that issue because the protobuf lib I use isn't as zealous as its Rust's counterpart. The workaround was easy in Rust anyway. I switched to `optional` in the protobuf file.  I let the user decide what they want to do with an unresolved link.

ahjohannessen:
Ok, I see. I am not sure about the value of making event and link on `ResolvedEvent` `Option`. Perhaps I should rather introduce a special case `UnresolvedLink(linkRecord)` instead.

YoEight:
There is a use for just having only a link. If you are only interested in where that link comes from without paying the read from disk, as a link is just `event_number@stream_name` with `$>` type. I think your suggestion is good. Like we said in Haskell, don't implement a sum type as a product type :smile:

ahjohannessen:
>There is a use for just having only a link. If you are only interested in where that link comes from without paying the read from disk

Is that not just setting `resolveLinkTos = false` and read from, e.g., `$streams`?

PR #2024, opened 2019-10-17
Changes to Plugin system

shaan1337:
- Add support for subsystem plugins
- Always start subsystems - not only on the master node
- Stop subsystems on shutdown
- Add plugin subdirectories to the catalog


shaan1337:
Closing in favor of #2122

PR #2036, opened 2019-10-25
Require master on all writes and deletes from projections

hayley-jean:
Projections only run on the master node, and writes published from projections should only be processed if they originate from the master node.
Setting `RequireMaster` to true for all writes and deletes from projections enforces this, causing writes to fail if the node is not in the master state when processing them.

### Current issue

It's possible for a deposed master to publish writes for emitted streams while it is shutting down the projections subsystem.
This is fine if the node is still in the `Master` state, as then it simply writes the emitted event to the transaction log, which will likely be truncated soon.

But when a master is deposed, it initially returns to the `PreReplica` state to subscribe to the new master.
In this state, it will forward writes to the new master node.

This means that the deposed master can forward writes based off its own data and state to the new master.
Usually this is harmless, but it can sometimes result in an `ExpectedVersion` error on the new master, or the projection faulting with `An event emitted in recover differs from a previously emitted event.`

### DIFF
--- a/src/EventStore.Projections.Core/Services/Processing/RequestResponseQueueForwarder.cs
+++ b/src/EventStore.Projections.Core/Services/Processing/RequestResponseQueueForwarder.cs
@@ -31,14 +31,14 @@ namespace EventStore.Projections.Core.Services.Processing {
 		public void Handle(ClientMessage.WriteEvents msg) {
 			_externalRequestQueue.Publish(
 				new ClientMessage.WriteEvents(
-					msg.InternalCorrId, msg.CorrelationId, new PublishToWrapEnvelop(_inputQueue, msg.Envelope), false,
+					msg.InternalCorrId, msg.CorrelationId, new PublishToWrapEnvelop(_inputQueue, msg.Envelope), true,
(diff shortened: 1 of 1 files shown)


hayley-jean:
### Steps to reproduce the forwarded write

You need to be able to simulate a node going offline and then reconnecting quickly.
In my tests I did the following:

1. Alter the resign master feature in v6-master to:
    1. Not go into the `ResigningMaster` state, but stay as `Master`
    2. Change the handler for `MasterIsResigning` to update the gossip and set the current master as dead.
2. Add a log in the `ClusterVNodeController` to log out forwarded writes for the projection output stream - `outstream-*`.
3. Start the cluster and lower the priority of the master node.
4. Create a **stopped** projection with the following definition:

```
        fromAll()
        .when({
            $any: function(s,e) {
                linkTo("outstream-" + e.eventType, e);
            }
        })
```

5. Configure the projection to checkpoint after 400,000 events and 1000000 mb. Set the max concurrent writes to 20.
6. Start the projection.
7. Start writing events of different event types to fan out the projection.
8. Send a resignation to the master node. This should cut it off from the rest of the nodes, forcing an election and a truncation.
9. Monitor the logs for the log message added in step 2.

PR #2050, opened 2019-11-12
Improved liveness determination for nodes

pgermishuys:
There are 2 places inside of Event Store where we publish the
VNodeConnectionLost message. MasterReplicationService and
ReplicaService.

The ReplicaService's responsibility is to subscribe to Master amongst
other things.

When the ReplicaService changes state, it will disconnect and reconnect
to master. It terminates the connection cleanly. As a result, we publish
a `VNodeConnectionLost` which the `GossipService` uses as the truth that
the node is dead, which in fact it's not.

The suggested change here is not publish this message if the
connection was terminate cleanly. This will avoid false positives for
nodes being marked as dead.

The Master Replication Service's responsbility is to ensure that it has
a quorum of replicas to replicate it's log to.

When it checks the connection statuses of the list of subscriptions it
has, it publishes a `VNodeConnectionLost` will mark a node as dead which
it might not be as it might have been a replica that disconnected
cleanly.

The suggested change here is similar to above with the ReplicaService.
We want to publish if the connection was not terminated succesfully.
We want to publish this message still as it will result in a much faster
result than waiting for a gossip to fail to determine if a node is alive
or not.

When Elections are done, we can safely update our view of the cluster as
we know who master is. We now do this in the GossipService.


pgermishuys:
First round of testing with the changes.
**Cluster Size:** 5
**Number of Flooders:** 8
**Size of Messages:** 4096 bytes
**Master Resignation Interval:** "00:00:30"

**Results:**
- 71 Successful Master Resignations.
- Total of 524393 messages written

PR #2058, opened 2019-11-19
Properly scale up object pool based on number of reader threads

### LINKED ISSUE #2044: PTable WorkItem ObjectPool is not scaled up according to number of reader threads
**Describe the bug**
This is a bug similar to the one this PR fixes:
https://github.com/EventStore/EventStore/pull/1737/files

Basically, the PTable work item object pool is not scaled up according to the number of reader threads:
https://github.com/EventStore/EventStore/blob/master/src/EventStore.Core/Index/PTable.cs#L105

Like it's done here for the `readerPool`:
https://github.com/EventStore/EventStore/blob/master/src/EventStore.Core/ClusterVNode.cs#L204

Instead the default parameter `maxReaders` is used which is equal to `ESConsts.PTableMaxReaderCount`, currently 17:
https://github.com/EventStore/EventStore/blob/master/src/EventStore.Core/Index/PTable.cs#L84

If the number of reader threads is set to a value greater than `ESConsts.PTableMaxReaderCount` (which is currently 17) and there is a high volume of reads (making use of all reader threads), the object pool limit will be reached and it will throw `ObjectPoolMaxLimitReachedException`.

It appears there's also a similar issue with the TFChunk constructor which uses the `ESConsts.TFChunkMaxReaderCount` constant as `maxReaderCount` parameter in several places.

**EventStore details**
 - EventStore server version:
5.0.3

 - Operating system:
All

 - EventStore client version (if applicable):
N/A

shaan1337:
Fixes #2044

*Reproduction steps:*

1. Apply following diff on the `master` branch to slow down returning the work item to make issue more easily reproducible:
```
diff --git a/src/EventStore.Core/Index/PTable.cs b/src/EventStore.Core/Index/PTable.cs
index 26ee7ff..afa8b69 100644
--- a/src/EventStore.Core/Index/PTable.cs
+++ b/src/EventStore.Core/Index/PTable.cs
@@ -708,6 +708,7 @@ namespace EventStore.Core.Index {
                }
 
                private void ReturnWorkItem(WorkItem workItem) {
+                       Thread.Sleep(10);
                        _workItems.Return(workItem);
                }
 
diff --git a/src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs b/src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
index 3f49f6c..70e3420 100644
--- a/src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
+++ b/src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
@@ -1035,6 +1035,7 @@ namespace EventStore.Core.TransactionLog.Chunks.TFChunk {
                }
 
                private void ReturnReaderWorkItem(ReaderWorkItem item) {
+                       Thread.Sleep(10);
                        if (item.IsMemory) {
                                _memStreams.Enqueue(item);
                                if (_isCached == 0 || _selfdestructin54321)
```

2. Start eventstore with following parameters:
```
--db ../db --log ../log --structured-log=False --run-projections=All --reader-threads-count 26 --max-memtable-size 1000 --cached-chunks 0 --chunks-cache-size 0
```

3. Start test client and write a few events:
```
mono bin/Release/EventStore.TestClient/net471/EventStore.TestClient.exe
>> WRFL 1 3000
>> WRLT 1 1 1 1 stream
```
`WRFL` 3000 events will ensure that we create a PTable. Check your index directory to make sure it's been created.
`WRLT 1 1 1 1 stream` will ensure that we have a stream called `stream` because we're going to read events from it later.

4. Start the following project with `dotnet run`:
[multireader.zip](https://github.com/EventStore/EventStore/files/3868851/multireader.zip)

It will launch 30 threads to spam reads.

5. When the 30 threads have been launched and the reads start, the following errors should be visible in the logs. You should see two types of errors:
```
System.Exception: Unable to acquire reader work item. Max internal streams limit reached.
  at EventStore.Core.TransactionLog.Chunks.TFChunk.TFChunk.GetReaderWorkItem () [0x00068] in <437e25a373a84d3594ba358923fe3cb4>:0 
```
and
```
System.Exception: Unable to acquire work item.
  at EventStore.Core.Index.PTable.GetWorkItem () [0x00016] in <437e25a373a84d3594ba358923fe3cb4>:0 
```

6. Test EventStore with the branch `fix-max-object-pool-size` with the same steps, the errors do not occur.

### DIFF
--- a/src/EventStore.Core.Tests/Constants.cs
+++ b/src/EventStore.Core.Tests/Constants.cs
@@ -1,0 +1,24 @@
+using EventStore.Core.Settings;
+using EventStore.Core.TransactionLog.Chunks;
+using EventStore.Core.Util;
+
+namespace EventStore.Core.Tests {
+	public class Constants {
+		public const int PTableInitialReaderCount = ESConsts.PTableInitialReaderCount;
+
+		public const int PTableMaxReaderCountDefault = 	1 /* StorageWriter */
+		                                                + 1 /* StorageChaser */
+		                                                + 1 /* Projections */
+		                                                + TFChunkScavenger.MaxThreadCount /* Scavenging (1 per thread) */
+		                                                + 1 /* Subscription LinkTos resolving */
+		                                                + Opts.ReaderThreadsCountDefault
+		                                                + 5 /* just in case reserve :) */;
+
+		public const int TFChunkInitialReaderCountDefault = Opts.ChunkInitialReaderCountDefault;
+
+		public const int TFChunkMaxReaderCountDefault = PTableMaxReaderCountDefault
+		                                                + 2 /* for caching/uncaching, populating midpoints */
+		                                                + 1 /* for epoch manager usage of elections/replica service */
+		                                                + 1 /* for epoch manager usage of master replication service */;
+	}
+}
--- a/src/EventStore.Core.Tests/Helpers/MiniClusterNode.cs
+++ b/src/EventStore.Core.Tests/Helpers/MiniClusterNode.cs
@@ -23,8 +23,6 @@ using EventStore.Core.TransactionLog.FileNamingStrategy;
 using EventStore.Core.Services.Transport.Http.Controllers;
 using EventStore.Core.Util;
 using EventStore.Core.Data;
-using EventStore.Core.Settings;
-using EventStore.Core.Tests.TransactionLog;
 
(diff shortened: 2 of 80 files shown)


shaan1337:
Closing and reopening against v6-master

PR #2068, opened 2019-11-25
v6 miscellaneous fixes

shaan1337:
- Fix CI artifact names
- Enable TLS 1.3 protocol
- Fix build warnings. This required also removing Obsolete attribute on internal/external http service for now


jen20:
IMO we should disable everything except TLS 1.3 by default, and allow opting in via configuration to earlier versions.

pgermishuys:
Closing in favor of https://github.com/EventStore/EventStore/pull/2109

PR #2075, opened 2019-11-27
Rework FlushSafe for DotNet Core

pgermishuys:
The field that we used to get is no longer available as `_handle`. This is currently throwing an exception and resulting in `FlushSafe = f => f.Flush(flushToDisk: true);`


condron:
there is a SafeFileHandle now directly available so this would be the most direct replacement.

```
FlushSafe = f => {
	if.Flush(flushToDisk: false);
	if (!FlushFileBuffers(f.SafeFileHandle)){
              throw new Exception($"FlushFileBuffers failed with err: {Marshal.GetLastWin32Error()}");
         };
};
```

PR #2079, opened 2019-11-27
client api acceptance tests with the v5 client

thefringeninja:


thefringeninja:
Only one test doesn't pass right now: `connect_to_persistent_subscription.that_does_not_exist_throws`. But this is because of what is a bug IMO in v5 of the client api - If you await a task you should never ever see an `AggregateException`.

PR #2083, opened 2019-11-28
Use Certificate Options for Kestrel HTTPS

thefringeninja:


src/EventStore.ClusterNode/Program.cs
```diff
@@ -124,15 +124,21 @@ public class Program : ProgramBase<ClusterNodeOptions> {
 			_host = new WebHostBuilder()
 				.UseKestrel(o => {
 					o.Listen(opts.IntIp, opts.IntHttpPort);
-					o.Listen(opts.ExtIp, opts.ExtHttpPort,
-						listenOptions => listenOptions.UseHttps());
+					o.Listen(opts.ExtIp, opts.ExtHttpPort, listenOptions => {
+						if (_node.Certificate == null) {
+							listenOptions.UseHttps();
```
jen20: What certificate does this give us?

thefringeninja: If you don't supply one it will use the default kestrel one.

PR #2085, opened 2019-11-29
Log Kestrel Startup Messages

thefringeninja:


thefringeninja:
There's only one problem - all http requests are getting logged to the console which we definitely don't want.

pgermishuys:
Work has been re-assigned.

PR #2086, opened 2019-12-02
Fix Issue Where Non Dotnet Clients Get Error

thefringeninja:


thefringeninja:
See aspnet/Extensions#2732 and aspnet/AspNetCore#11050 for info on the startup class.

PR #2087, opened 2019-12-02
Dockerized Build

thefringeninja:
PR builds coming from forks will not initiate a github actions build if no ci.yml exists in the target branch. If this works this will replace #2070.


Dockerfile
```diff
@@ -1,18 +1,60 @@
...
+    chmod +x /build/test.sh
+CMD ["/build/test.sh"]
+
+FROM build as publish
+ARG RUNTIME=linux-x64
+
+RUN dotnet publish --configuration=Release --no-build --runtime=${RUNTIME} --self-contained \
+     --framework=netcoreapp3.1 --output /publish /p:PublishTrimmed=true EventStore.ClusterNode
+
+FROM mcr.microsoft.com/dotnet/core/runtime-deps:3.1-${CONTAINER_RUNTIME} AS runtime
```
pgermishuys: We are publishing a self contained application, we shouldn't need the 3.1 runtime.

PR #2090, opened 2019-12-02
Limit Append to User Configurable Value

thefringeninja:
not to exceed 16MB

### DIFF
--- a/src/EventStore.Core/Services/Transport/Grpc/Streams.Append.cs
+++ b/src/EventStore.Core/Services/Transport/Grpc/Streams.Append.cs
@@ -1,6 +1,5 @@
 using System;
 using System.Collections.Generic;
-using System.Security.Principal;
 using System.Threading.Tasks;
 using EventStore.Core.Data;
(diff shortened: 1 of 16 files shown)


src/EventStore.Core/Services/Transport/Grpc/Streams.Append.cs
```diff
@@ -38,19 +37,27 @@ partial class Streams {
...
 			while (await requestStream.MoveNext()) {
 				if (requestStream.Current.ContentCase != AppendReq.ContentOneofCase.ProposedMessage)
 					throw new InvalidOperationException();
 
 				var proposedMessage = requestStream.Current.ProposedMessage;
+				var data = proposedMessage.Data.ToByteArray();
+				size += data.Length;
+
+				if (size > _maxAppendSize) {
+					throw RpcExceptions.MaxAppendSizeExceeded();
```
ahjohannessen: Perhaps adding `_maxAppendSize` and `size` to exception is useful for clients?

PR #2102, opened 2019-12-03
Protobuf Contract Improvements

thefringeninja:
- [x] make Position.End explicit on reads
- [x] make StreamRevision.End explicit on reads
- [x] support both types of UUID

### DIFF
--- a/src/EventStore.Grpc.Common/Uuid.cs
+++ b/src/EventStore.Grpc.Common/Uuid.cs
@@ -1,0 +1,114 @@
+using System;
+
+namespace EventStore.Grpc {
+#if EVENTSTORE_GRPC_PUBLIC
+	public
+#else
+	internal
+#endif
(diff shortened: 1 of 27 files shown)


ahjohannessen:
@pgermishuys Could you please take time to review so this can get merged? :)

src/EventStore.Grpc.Common/Uuid.cs
```diff
@@ -0,0 +1,114 @@
...
+			? Guid.Parse(value)
+			: throw new ArgumentNullException(nameof(value))) {
+		}
+
+		private Uuid(long msb, long lsb) {
+			_msb = msb;
+			_lsb = lsb;
+		}
+
+		internal PersistentSubscriptions.UUID ToPersistentSubscriptionsDto() =>
```
pgermishuys: Is there any reason why there are persistent subscriptions related stuff in the UUID type?

Issue #1917, opened 2019-04-25
Slave restart can lead to master re-elections (most common in 5 node cluster)


shaan1337:
Thanks to @megakid for reporting this.

Restarting a slave node can lead to master re-elections.
This can lead to some instability since the previous master (and maybe some slaves too) very likely needs to truncate if writes were ongoing since uncommitted data is written immediately to chunk files causing a few more elections when these truncated nodes are back up.

*Reproduction steps:*
Here's an instance of how it can occur:
1. Node X (a slave) is restarted
2. Node X triggers elections process
3. Node X is selected as the leader of the current elections
4. Node X waits for a quorum number of `PREPARE_OK` messages from other nodes before choosing the best master candidate. It does not receive a `PREPARE_OK` from the original master.
5. At this point in the code https://github.com/EventStore/EventStore/blob/master/src/EventStore.Core/Services/ElectionsService.cs#L338 , Node X does not have the `_lastElectedMaster` value since it was restarted. Based on the current algorithm, a different master will be elected.

So far this has been reported only on 5-node clusters and a reproduction on 3-node clusters has not been successful.

*Potential solution 1:*
 If a node has been selected as leader of elections and does not have a `_lastElectedMaster` value check in gossip if there is at least one slave/master nodes. If yes, let the elections time out since we know that these nodes will have a `_lastElectedMaster` value.

*Potential solution 2:*
If a node has been selected as leader of elections and does not have a `_lastElectedMaster` value, let the elections time out X times where X = 1 or 2. This will lead to a few more elections (particularly when all nodes are being started for first time) but could make the cluster more stable for single node restarts.

Ticket no: 1314

pgermishuys:
This particular issue requires further testing and investigation as there are 2 issues. One where the TCP being lost caused elections to take place which is what #2103 attempts to remedy and another where 2 nodes accepted a new master proposal even though they both had knowledge of the previous master being elected and that master still being alive.

The other 2 nodes outright rejected the proposal as both of them knew that the previously elected master node is still alive.

gregoryyoung:
"This can lead to some instability since the previous master (and maybe some slaves too)"

Can you provide an example of a slave needing to truncate in this case? This should not be anything resembling a normal case...

pgermishuys:
Having a had a look at this closer, the above case is already covered whereby the nodes in the cluster should reject the master proposal from the restarting slave if the current master of the nodes that need to accept the proposal is still alive.

[...]

PR #2103, opened 2019-12-03
Use gossip to determine node liveness after tcp connection drops

pgermishuys:
When elections occur and a new master node is elected, we drop the TCP Connection and reconnect to the new master. We however mark the previous node as dead during this time which isn't entirely accurate.

This work introduces a double check on the liveness of a node by issuing a new `GetGossip` (The name could be better) when the subscription TCP Connection is dropped (`VNodeConnectionLost`) instead of assuming the node is dead immediately.

We then mark the node as **dead** if the `GetGossip` fails or mark the node as **alive** if the `GetGossip` succeeds.

This will provide a little more stability for example during elections, specifically in cases where a master resignation has been performed and a new master gets elected.

Ref: #1917

### DIFF
--- a/src/EventStore.Core/Services/Gossip/GossipServiceBase.cs
+++ b/src/EventStore.Core/Services/Gossip/GossipServiceBase.cs
@@ -22,8 +22,11 @@ namespace EventStore.Core.Services.Gossip {
 		IHandle<GossipMessage.GossipSendFailed>,
 		IHandle<SystemMessage.VNodeConnectionLost>,
 		IHandle<SystemMessage.VNodeConnectionEstablished>,
+		IHandle<GossipMessage.GetGossipReceived>,
+		IHandle<GossipMessage.GetGossipFailed>,
 		IHandle<ElectionMessage.ElectionsDone> {
 		private static readonly TimeSpan DnsRetryTimeout = TimeSpan.FromMilliseconds(1000);
+		private static readonly TimeSpan GossipTimeout = TimeSpan.FromMilliseconds(1000);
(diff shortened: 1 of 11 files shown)


src/EventStore.Core/Services/Gossip/GossipServiceBase.cs
```diff
@@ -94,31 +104,34 @@ public abstract class GossipServiceBase : IHandle<SystemMessage.SystemInit>,
...
 				_bus.Publish(new HttpMessage.SendOverHttp(node.InternalHttpEndPoint,
-					new GossipMessage.SendGossip(_cluster, NodeInfo.InternalHttp), DateTime.Now.Add(GossipInterval)));
+					new GossipMessage.SendGossip(_cluster, NodeInfo.InternalHttp),
+					_timeProvider.LocalTime.Add(GossipInterval)));
 			}
 
-			var interval = message.GossipRound < 20 ? GossipStartupInterval : GossipInterval;
-			var gossipRound = Math.Min(2000000000, node == null ? message.GossipRound : message.GossipRound + 1);
+			var interval = message.GossipRound < GossipRoundStartupThreshold ? GossipStartupInterval : GossipInterval;
+			var gossipRound = Math.Min(int.MaxValue, node == null ? message.GossipRound : message.GossipRound + 1);
```
shaan1337:
```suggestion
			var gossipRound = Math.Min(int.MaxValue - 1, node == null ? message.GossipRound : message.GossipRound + 1);
```

nit pick: this should be int.MaxValue - 1, otherwise if `message.GossipRound` has a value of `int.maxValue`, `message.GossipRound + 1` will overflow.

PR #2120, opened 2019-12-06
Set the default to false for WriteStatsToDb

pgermishuys:
For Event Store V6, we will no longer by default write stats to the database. Writing stats to the database can be enabled by setting `WriteStatsToDb` to `true`.


mat-mcloughlin:
Build potentially timing out due to the following test? https://github.com/EventStore/EventStore/blob/master/src/EventStore.Grpc.Tests/Streams/write_stream_meta_security.cs#L77

PR #2124, opened 2019-12-09
Handle GRPC Client Cancellations Better

thefringeninja:
If the grpc client cancels an in-flight request, a message can appear in the log e.g.
```
[08399,11,11:13:48.595] Error when executing service method 'Read'.
EXCEPTION OCCURRED
The CancellationTokenSource has been disposed.
```
We don't want this especially if people have monitoring around errors in the logs.


danielmarbach:
Needs a rebase

PR #2134, opened 2019-12-11
Add Development Mode

pgermishuys:
Add `Development Mode`. (e.g. `--dev`)
- When the development mode is turned on. 
   - Use the SSL certificate included in the package for HTTPS.
   - The `MemDb` flag will be set to `true` only if the flag hasn't been specified (i.e. Default)
- When the development mode is turned off
   - An SSL Certificate is **required**
- Moves the `dev-ca` directory into the `/src` directory. This avoids any pathing related issues when building for the docker build specifically.

### Notes
When the development mode is set, a warning will tell the user that the development mode is not intended for production use.

```
========================================================================================================
DEVELOPMENT MODE IS ON. THIS MODE IS *NOT* INTENDED FOR PRODUCTION USE.
WHEN IN DEVELOPMENT MODE EVENT STORE WILL
 - NOT WRITE ANY DATA TO DISK.
 - USE A SELF SIGNED CERTIFICATE.
========================================================================================================

```

In addition to the above, when logging out the effective configuration, we will note that the `MemDb` has been set by the `Development Mode`.

```
MODIFIED OPTIONS:

        DEV:                      True (Command Line)
        MEM DB:                   True (Set by 'Development Mode' mode)
        CERTIFICATE FILE:         C:\source\EventStore\bin\Debug\EventStore.ClusterNode\netcoreapp3.1\dev-ca\server1.pfx (Set by 'Development Mode' mode)
```


ahjohannessen:
grpc would work in development mode when doing this:
```
if (!opts.Dev) {
  listenOptions.UseHttps(_node.Certificate);
} else {
  listenOptions.Protocols = HttpProtocols.Http2;
}
```

jen20:
I suspect we might want a sterner warning indicating that data is being written only to memory and will not persist across restarts?

Narvalex:
Is the SSL requirement only applicable to the http api or does this affect the native tcp api? Does this affect the Event Store Web UI?

src/EventStore.ClusterNode/Program.cs
```diff
@@ -90,6 +113,16 @@ public class Program : ProgramBase<ClusterNodeOptions> {
 		protected override void Create(ClusterNodeOptions opts) {
 			var dbPath = opts.Db;
 
+			if (opts.Dev) {
+				Log.Warn(
+					"\n========================================================================================================\n" +
+					"DEVELOPMENT IS ON. THIS MODE IS *NOT* INTENDED FOR PRODUCTION.\n" +
+					"WHEN IN DEVELOPMENT MODE EVENT STORE WILL\n" +
+					" - WRITE ANY DATA TO DISK.\n" +
```
shaan1337: Should this be: "NOT WRITE ANY DATA TO DISK."?

src/EventStore.ClusterNode/Program.cs
```diff
@@ -90,6 +113,16 @@ public class Program : ProgramBase<ClusterNodeOptions> {
 		protected override void Create(ClusterNodeOptions opts) {
 			var dbPath = opts.Db;
 
+			if (opts.Dev) {
+				Log.Warn(
+					"\n========================================================================================================\n" +
+					"DEVELOPMENT IS ON. THIS MODE IS *NOT* INTENDED FOR PRODUCTION.\n" +
```
shaan1337:
```suggestion
					"DEVELOPMENT MODE IS ON. THIS MODE IS *NOT* INTENDED FOR PRODUCTION.\n" +
```

shaan1337:
```suggestion
					"DEVELOPMENT MODE IS ON. THIS MODE IS *NOT* INTENDED FOR PRODUCTION USE.\n" +
```

pgermishuys:
Hey @Narvalex, apologies for only replying to you now.
With the preview release, the external HTTP service will be HTTPS only (This means that the Web UI as well as all the external HTTP traffic will be HTTPs only).

The External TCP interface will by default still work without TLS unless you set the `ExtSecureTcpPort` at which point it will use the same certificate as the one being used by the external HTTP service.

Does this answer your question?

PR #2145, opened 2019-12-13
Support Checkpoint Interval on Filtered Subscriptions

thefringeninja:


pgermishuys:
This requires a rebase @thefringeninja

hayley-jean:
We've previously discussed making use of the subscriptions API rather than issuing reads for subscriptions on the GRPC client.
Perhaps this is a good opportunity to make this change, either in this PR or in a separate one to be merged before this one?

That way we might not need to change the read result for the other operations to include the checkpoint position, and can eliminate the wait on subscriptions.

PR #2153, opened 2019-12-16
Operations GRPC Implementation for Scavenging

### LINKED ISSUE #2151: Add Operations Proto Contract
The following operational operations need to be in a proto:

- Scavenge
- Subsystem Restarts

Furthermore, they should go into a separate project / nuget package e.g. `EventStore.Grpc.Operations.Client`

thefringeninja:
Part of #2151

### DIFF
--- a/src/EventStore.Core/ClusterVNodeStartup.cs
+++ b/src/EventStore.Core/ClusterVNodeStartup.cs
@@ -27,6 +27,7 @@ namespace EventStore.Core {
 
 		private static readonly PathString StreamsSegment = "/event_store.grpc.streams.Streams";
 		private static readonly PathString UsersSegment = "/event_store.grpc.users.Users";
+		private static readonly PathString OperationsSegment = "/event_store.grpc.operations.Operations";
 
 		private readonly ISubsystem[] _subsystems;
 		private readonly IQueuedHandler _mainQueue;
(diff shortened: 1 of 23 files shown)


src/EventStore.Core/Services/Storage/StorageScavenger.cs
```diff
@@ -89,7 +89,7 @@ public class StorageScavenger :
 					} else {
 						message.Envelope.ReplyWith(new ClientMessage.ScavengeDatabaseResponse(message.CorrelationId,
 							ClientMessage.ScavengeDatabaseResponse.ScavengeResult.InvalidScavengeId,
-							_currentScavenge?.ScavengeId));
```
hayley-jean: Is there any particular reason why `currentScavenge.ScavengId` was changed to the `message.ScavengeId`?

thefringeninja: Yes, I would like to report back which scavengeid was invalid.

hayley-jean: Judging by the code around this, it looks like the original purpose was to return the currently running scavenge Id so if you sent the wrong one, you'd be able to get the correct one and be able to stop it in a subsequent call

thefringeninja: ok, makes sense. fixed now

ahjohannessen:
Should there not be something like:

```
message GetStatusReq {
 	Options options = 1;
 	message Options {
 		string scavenge_id = 1;
 	}
 }

rpc GetStatus (GetStatusReq) returns (ScavengeResp);
```

src/EventStore.Core/ClusterVNodeStartup.cs
```diff
@@ -94,7 +95,10 @@ public class ClusterVNodeStartup : IStartup, IHandle<SystemMessage.SystemReady>,
 								endpoint.MapGrpcService<Users>()))
 						.UseWhen(context => context.Request.Path.StartsWithSegments(StreamsSegment),
 							inner => inner.UseRouting().UseEndpoints(endpoint =>
-								endpoint.MapGrpcService<Streams>())),
+								endpoint.MapGrpcService<Streams>()))
+				.UseWhen(context => context.Request.Path.StartsWithSegments(OperationsSegment),  // TODO JPB figure out how to delete this sadness
```
ahjohannessen: You guys should seriously consider using a different port for grpc stuff.

PR #2154, opened 2019-12-17
Consistent ConfigureAwait(false)

danielmarbach:


danielmarbach:
Also raised a separate PR that brings in an analyzer in case you want to capture them early. Note that the analyzer currently doesn't cover async enumerable and async disposable

PR #2157, opened 2019-12-17
Remove Task.Run from IO bound path

danielmarbach:
I'm reasonably certain the `Task.Run` remove here are not necessary because the underlying path is IO-bound anyway and there is no need to explicitely offload to the worker thread pool

I deliberately left the `StreamSubscription` and `PersistenceSubscription` `Task.Run` in the constructor for now because haven't figured out yet how the GRPC Read behaves and the async enumerable in `StreamSubscription` could in theory yield synchronously


danielmarbach:
Seems tests are failing because master is failing

jen20:
I've rebased this onto the current `master` - hopefully that should resolve the failing tests.

PR #2158, opened 2019-12-17
Consistently go with TaskCreationOptions.RunContinuationsAsynchronously

danielmarbach:
plus Get rid of TaskCompletionSourceFactory

Generally, the code base is full of TCS usage and the creation options are not consistently applied. This PR goes for the safer default of adding `TaskCreationOptions.RunContinuationsAsynchronously` consistently. It was not possible for me to evaluate if the usage of the TCS was deliberately not using the creation options in some places because I don't have the context of this project. So in some of the core infrastructure having this option present might slightly increase the thread pool pressure because the continuations are forced to run on the worker pool but that is in most of the cases anyway what you want. 

// cc @jen20


jageall:
TaskCompletionSourceFactory was added because of .net452, we still need to target that and it was removed by mistake from the preview (because there  is no TCO.RCA option in 452)

Additionally, i would like to split this PR up between client and server because server side requires far more review, client side is much more straightforward

pgermishuys:
Hey @danielmarbach. Thanks for taking the time and effort you put into this PR. As per @jageall's suggestion, we are splitting up the PR into 2 separate PRs for Client and Server.

The server PR is here: https://github.com/EventStore/EventStore/pull/2202.

PR #2160, opened 2019-12-17
Add exponential backoff + jitter to projection writes

### LINKED ISSUE #2159: Provide the ability for projections to backoff exponentially
The current back off when projections are writing is linear. We could employ a different potentially better backoff strategy which will ease the load on Event Store. Possible options might include exponential.

shaan1337:
Fixes #2159

This PR adds exponential backoff + jitter to projection writes. The jitter strategy used, called "Full Jitter", is described in this article: https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

## Rationale
Under heavy load, commit timeouts can become very common. With the current code, projections attempt writes for 5 times with an interval of `currentAttempt` seconds between each retry before they are marked as faulted and manual intervention is required to re-enable the projections. This is not desirable in production.

The PR attempts to fix two things:
- Increase the retry count so that it takes longer before the projection is marked as faulted. 5 retries corresponds to only 10 seconds between the first attempt and the last attempt and is too low since there may be peak periods having a duration much longer than that in production. The PR attempts to increase this time period to an average time of 10 mins before the projection is marked as faulted.

- Add exponential backoff + jitter so that the load is spread out to avoid projection writes hammering the server with write retries.

## Fix description
- Add backoff to projection emitted event writes
- Add backoff to projection checkpoint writes
- Avoid rewriting to projection checkpoint metadata stream when retrying first checkpoint write

## Backoff algorithm description
The fix will simply backoff projections for the following time period between retries:
`random.uniform(1, min(2^(attempt-1), 256))` seconds

Note: we delay by at least one second between retries.
For the first attempt, there is no delay.

## Total time before being marked as faulted
To calculate average amount of time it would take before a projection is marked as faulted (if max attempts = 12):

```
TOTAL_TIME = 0 /* 1st attempt */+ rand(1,2) + rand(1,4) + rand(1,8) + rand(1,16) + ... +  rand(1, 128) + rand(1, 256) + ... + rand(1, 256)  /* random between 1 to 256 for 9th to 12th attempt */
```

Since rand(a,b) is uniformly distributed, the mean value of rand(a,b) is (a + b) / 2.
Thus, average total time is (by linearity of expectation):
```
E[TOTAL_TIME] = 0 + 1.5 + 2.5 + 4.5 + 8.5 + ... + 128.5 + 128.5 + 128.5 + 128.5 = 644.5 seconds ~= 10.7 mins.
```

## Test environment
Following tests were carried out in a docker container:
1. Create 200 projections called projection-$i that read from stream `test` and emit to `test-$i`:
```
fromStream('test').when({'$any':function(s,e){emit('test-$i', 'type', e.bodyRaw );}})
```
2. Stop all the projections
3. Write around 100000 events to stream `test`
4. Limit write iops to 10 (by modifying cgroup property: `blkio.throttle.write_iops_device`)
5. Start all projections

With the fix, the projections manage to progress consistently (it managed to progress up to around 50,000 events and I stopped it there).

With the current code, all the 200 projections fault after a few minutes.

With the current code but simply increasing the max retry count to a very high value (tested with 40), the projections manage to progress but very very slowly compared to with the exponential backoff mechanism since there is still much contention during retries.

### DIFF
--- a/src/EventStore.Projections.Core/Services/Processing/CoreProjectionCheckpointWriter.cs
+++ b/src/EventStore.Projections.Core/Services/Processing/CoreProjectionCheckpointWriter.cs
@@ -23,7 +23,9 @@ namespace EventStore.Projections.Core.Services.Processing {
 		private Event _checkpointEventToBePublished;
 		private CheckpointTag _requestedCheckpointPosition;
 		private IEnvelope _envelope;
-		private const int MaxNumberOfRetries = 5;
+		private const int MaxNumberOfRetries = 12;
+		private const int MinAttemptWarnThreshold = 5;
+		private Random _random = new Random();
 
 		public CoreProjectionCheckpointWriter(
 			string projectionCheckpointStreamId, IODispatcher ioDispatcher, ProjectionVersion projectionVersion,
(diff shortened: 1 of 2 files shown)


jageall:
Yeah, there are a few things that have been scattered around and retries would be good to pull together later. Changes look good. We should investigate something similar for reads

PR #2171, opened 2019-12-20
Fix hanging during shutdown if node shut down when not yet ready

### LINKED ISSUE #2168: EventStore hangs on shutdown / when going offline for truncation [6.0.0-preview1]
**Describe the bug**
The EventStore process seems to hang when shutting down at this line when running in a cluster:
```
Exit reason: "Shutdown and exit from process was requested."
```
Pressing Ctrl-C exits the application normally.

Similar behaviour observed when a node goes offline for truncation. This could be problematic during automatic truncations since the node will not come back online.

The issue does not seem to happen on single nodes.

**To Reproduce**
1. Start a 3 node cluster
2. On one of the nodes, go to the web ui, go to "Admin" -> "Shutdown server"

**Expected behavior**
The process should exit

**Actual behavior**
Process hangs

**EventStore details**
 - EventStore server version: 6.0.0-preview1

 - Operating system: Ubuntu 16.04

shaan1337:
Fixes #2168

In the current code, `StartAndWaitUntilReady()` waits for the `SystemReady` message to be received before marking the task as complete. The `SystemReady` message is received only after a node has been able to join a cluster as a slave or master. Waiting for this message is useful for tests to determine that the node has successfully joined a cluster but when running a real node, there are some situations where you want to shut down the node but the node was not yet ready, for example:
- When a node is trying to join a cluster but it needs to go offline for truncation
- When a node has not even joined a cluster but you still want to stop it from the UI

In the above scenarios, the node will hang here: https://github.com/EventStore/EventStore/blob/master/src/EventStore.Core/ProgramBase.cs#L92 since the task completion source in `StartAndWaitUntilReady()` was never set. Even if you trigger a shutdown, it will hang there.

*Reproduction steps*
1. Start one node out of a 3 node cluster
2. Try to shut down the node from the UI
3. It will hang after shutting down

*Resolution*
The fix simply adds a boolean parameter called `waitUntilReady` to the start function and calls it with false when a real node is starting up.
Start and stop functions have also been renamed for consistency with async-await pattern.

### DIFF
--- a/src/EventStore.ClientAPIAcceptanceTests/EventStoreClientAPIFixture.cs
+++ b/src/EventStore.ClientAPIAcceptanceTests/EventStoreClientAPIFixture.cs
@@ -34,7 +34,7 @@ namespace EventStore.ClientAPI.Tests {
 		}
 
 		public async Task InitializeAsync() {
-			await _node.StartAndWaitUntilReady();
+			await _node.StartAsync(true);
 			await Connections[true].ConnectAsync();
 			await Connections[false].ConnectAsync();
 		}
--- a/src/EventStore.ClientAPIAcceptanceTests/EventStoreClientAPIFixture.cs
+++ b/src/EventStore.ClientAPIAcceptanceTests/EventStoreClientAPIFixture.cs
@@ -42,7 +42,7 @@ namespace EventStore.ClientAPI.Tests {
 		public Task DisposeAsync() {
 			Connections[true].Dispose();
 			Connections[false].Dispose();
-			return _node.Stop().WithTimeout();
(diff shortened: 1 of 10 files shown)


shaan1337:
Steps to trigger offline truncation for testing:

1. Start a 3 node cluster: nodes 1,2,3. Let's assume node 1 is master.
2. Start a WRFL (write flood) on the master node with 200000 events
3. Stop nodes 2 & 3.
4. Stop node 1.
5. Stop the write flood
6. Start nodes 2 & 3 and wait for cluster to stabilize
7. Start node 1. It will go offline for truncation.

PR #2181, opened 2019-12-27
Replay *some* parked messages

megakid:
Currently there is the ability to replay all parked messages in a particular persistent subscription.  Unfortunately, the way it does this is to load *all* of those parked messages into memory and push them onto a replay buffer in preparation for dispatch ahead of the on-going stream subscription.

The issue with this is that when a parked message stream gets beyond a certain size, it becomes impossible to replay.  We have had numerous examples of this in our production environment whereby the parked streams are gathering messages overnight and by the time a support person was aware, it was too late to replay them; upon replay, the persistent subscription subsystem crashes and a master failover is necessary.  Unfortunately, at this point, the truncate `$metadata` has already been written to the parked stream but without the parked messages being dispatched meaning when we failover, we lose the entire parked message stream.  We have a custom tool to fix this but it is far from ideal.

This PR enables an optional `stopAt` version to be passed into the replay command so chunked parked message replays are possible.  We envision that we will use it for times when we have more than, for example, 5,000 or so parked messages outstanding so we can reduce the instant load & memory pressure on the persistent subscription subsystem.

### DIFF
--- a/src/EventStore.Core.Tests/Http/PersistentSubscription/parked.cs
+++ b/src/EventStore.Core.Tests/Http/PersistentSubscription/parked.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Concurrent;
 using System.Net;
 using System.Text.RegularExpressions;
 using EventStore.Core.Tests.Http.Users.users;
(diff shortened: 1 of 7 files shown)


megakid:
Looks like the tests are failing here are also failing on `master` (see #2182)

PR #2186, opened 2019-12-31
Stop persisting command and response messages

### LINKED ISSUE #2135: Projection Manager Response Reader doesn't stop after master changeover
The `ProjectionManagerResponseReader` has no stop functionality.

If the projections subsystem starts on a node, the response reader will always be running.
It will keep reading the projection manager stream and publishing messages on the projection bus even if a master changeover occurs and projections stop on the node.
This seems harmless at the moment, as the other components aren't running so these messages aren't acted upon.

**Reproduction**:

1. Start a cluster with projections running.
2. Resign the master node such that a new node takes over.
3. Observe the `$prepared`, `$starting`, and `$started` message logs on both the old and new master nodes.

### LINKED ISSUE #2166: Remove support for distributed projections
Projections were originally designed to distribute load to slave nodes.
However, they can only reasonably run on a master node given they are write and read heavy.

To support the original distributed design, projection operations are written to and read from control and manager streams.
This adds extra overhead to running projections, as well as many more opportunities for failure. If one of the readers or writers stops working, the projections affected by that can get stuck in specific states or become unresponsive.
It also makes the projection code much more complex.

Simplifying this design would result in faster, stabler projections, and potentially resolve a number of issues we see around projections getting stuck in certain states.

hayley-jean:
Fixes #2135, fixes #2166

### Remove response readers and command writers

Remove the response readers and command writers that were persisting projection commands to streams. These have been replaced by simply forwarding the messages between the manager and the core queues, and back.

- Use `ProjectionManagerMessageDispatcher` to send messages from manager to core queues.
- Set up forwarding in `ProjectionWorkerNode` to forward responses straight back to the manager node.
- Delete core and manager response readers.
- Remove response writers.
- Remove `ReaderReady` and Starting Management Messages.
- Delete persisted commands and responses.
- Remove unused messages.
- Fix some core coordinator tests.
- Remove unused projection stream names.

As part of this PR, the projections now also wait for the IODispatchers to drain before coming to a stop. This is to prevent outstanding writes from causing WrongExpectedVersions on restarts of the subsystem.

The Checkpoint manager is also stopped when the projections enter a Suspended state.

### Start Standard Projections

There was also a bug with starting standard projections that this PR fixes.

To reproduce you can do the following :
1. Start Event Store with `startStandardProjections` enabled and `memDb` disabled.
2. Open the UI and view the projections.
3. Restart Event Store.
4. Try and disable the system projections.
5. An error is thrown in the server when trying to update the system projection's version.

This happens because the subsystem is trying to force the projections to start when they are manually disabled. This does not look like it was intended behaviour.

When `startStandardProjections` is enabled, the subsystem listens for the `Stopped` messages of the system projections, and issues an `Enable` command when it receives them.

On the first run, the system projections will write `Stopped` messages when they have been successfully created.  
However, on the next run, the system projections will not write those `Stopped` messages, meaning that trying to stop the projection will inadvertently trigger an `Enable` command.

This PR listens for the `Started` messages of system projections as well to determine if the system projections are running, and to prevent sending `Enable` commands when manually stopping the projections.


src/EventStore.Core/Helpers/IODispatcher.cs
```diff
@@ -8,9 +8,12 @@
 using EventStore.Core.Services.AwakeReaderService;
 using EventStore.Core.Services.TimerService;
 using System.Collections.Generic;
+using System.Linq;
+using EventStore.Common.Utils;
 using EventStore.Core.Util;
 
 namespace EventStore.Core.Helpers {
+	
 	public sealed class IODispatcher : IHandle<IODispatcherDelayedMessage> {
```
pgermishuys: We could create a new `RequestTrackingIODispatcher` or have a switch that toggles tracking requests as it currently stands we are only interested in tracking requests for a single scenario and we likely don't want the overhead of tracking requests for everyone using the `IODispatcher`.

hayley-jean: Given that we are using the IODispatcher in a number of different places through the projections and it's not abstracted, I've changed it to use a switch to toggle tracking the requests.

However, I think it's worth revisiting this to pull out a separate dispatcher for projections at some stage

src/EventStore.Projections.Core/Messages/ProjectionSubsystemMessage.cs
```diff
@@ -2,7 +2,7 @@
 using EventStore.Core.Messaging;
 
 namespace EventStore.Projections.Core.Messages {
-	public class ProjectionSubsystemMessage {
+	public static class ProjectionSubsystemMessage {
```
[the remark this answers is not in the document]
hayley-jean: Just to align it with other `-Message` classes and to prevent Rider from trying to instantiate it every time I typed `ProjectionSubsystemMessage`

PR #2191, opened 2020-01-02
Use Internal Subscriptions Model in GRPC Subscriptions

thefringeninja:
Currently the GRPC subscriptions model uses reads internally even when the subscription is caught up. This may result in lowered performance and/or an unacceptable delay to receive a message in a subscription.

### DIFF
--- a/src/EventStore.Core/Services/Transport/Grpc/Enumerators.AllSubscription.cs
+++ b/src/EventStore.Core/Services/Transport/Grpc/Enumerators.AllSubscription.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.Security.Principal;
 using System.Threading;
 using System.Threading.Tasks;
+using EventStore.Common.Utils;
 using EventStore.Core.Bus;
 using EventStore.Core.Data;
 using EventStore.Core.Messages;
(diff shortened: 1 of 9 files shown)


src/EventStore.Core/Services/Transport/Grpc/Enumerators.AllSubscription.cs
```diff
@@ -41,93 +50,344 @@ public class AllSubscription : IAsyncEnumerator<ResolvedEvent> {
...
+				var result = await _inner.MoveNextAsync().ConfigureAwait(false);
+				if (result) {
+					return true;
 				}
 
-				if (_buffer.TryDequeue(out var current)) {
-					_current = current;
-					return true;
+				if (!_catchUpRestarted) {
+					await _inner.DisposeAsync().ConfigureAwait(false);
```
pgermishuys: It doesn't appear that we ever move to a live subscription here unless a cancellation has been requested. Are we missing some logic where we decide to switch over to the live subscription based on `IsEndOfStream` from the `ReadAllEventsForwardCompleted`.

thefringeninja:
See #2206

PR #2194, opened 2020-01-02
Add Client Operations Tests

condron:
Add tests for client operations directly against the vNode using the `ClientMessages` and without using any of the client connections.

This will allow testing and confirmation of behavior in the vNode with conflating it with behaviors in the various clients. This will also simply debugging greatly.
replaced by #2259


condron:
replaced by #2259

PR #2205, opened 2020-01-07
Fix Off by One Errors in Subscriptions

thefringeninja:
When sending a position / revision in a subscription it should be exclusive of. However, currently you can receive a message you did not want.


thefringeninja:
See #2206 instead

PR #2206, opened 2020-01-07
Fix Off by One Errors in Subscriptions

thefringeninja:
When sending a position / revision in a subscription it should be exclusive of. However, currently you can receive a message you did not want.

### DIFF
--- a/src/EventStore.Client.Tests/Streams/subscribe_to_all_filtered_live.cs
+++ b/src/EventStore.Client.Tests/Streams/subscribe_to_all_filtered_live.cs
@@ -6,13 +6,16 @@ using System.Threading;
 using System.Threading.Tasks;
 using EventStore.Core.Services;
 using Xunit;
+using Xunit.Abstractions;
 
 namespace EventStore.Client.Streams {
-	public class subscribe_to_all_from_filtered : IClassFixture<subscribe_to_all_from_filtered.Fixture> {
+	public class subscribe_to_all_filtered_live : IAsyncLifetime, IDisposable {
 		private readonly Fixture _fixture;
+		private readonly IDisposable _loggingContext;
 
-		public subscribe_to_all_from_filtered(Fixture fixture) {
-			_fixture = fixture;
+		public subscribe_to_all_filtered_live(ITestOutputHelper outputHelper) {
+			_fixture = new Fixture();
+			_loggingContext = LoggingHelper.Capture(outputHelper);
 		}
 
 		[Fact]
@@ -22,7 +25,7 @@ namespace EventStore.Client.Streams {
 			var result = new List<ResolvedEvent>();
 			var source = new TaskCompletionSource<bool>();
 
-			_fixture.Client.SubscribeToAll(EventAppeared,
+			await _fixture.Client.SubscribeToAllAsync(Position.End, EventAppeared,
 				false, filter: new StreamFilter(new RegularFilterExpression(new Regex($"^{streamPrefix}"))));
 
 			foreach (var e in events) {
@@ -51,8 +54,9 @@ namespace EventStore.Client.Streams {
 			var result = new List<ResolvedEvent>();
 			var source = new TaskCompletionSource<bool>();
 
-			_fixture.Client.SubscribeToAll(EventAppeared,
-				false, filter: new StreamFilter(new PrefixFilterExpression(streamPrefix)));
+			await _fixture.Client.SubscribeToAllAsync(Position.End, EventAppeared, false,
+				filter: new StreamFilter(new PrefixFilterExpression(streamPrefix)));
+
 
 			foreach (var e in events) {
 				await _fixture.Client.AppendToStreamAsync($"{streamPrefix}_{Guid.NewGuid():n}",
@@ -84,8 +88,8 @@ namespace EventStore.Client.Streams {
 			var result = new List<ResolvedEvent>();
 			var source = new TaskCompletionSource<bool>();
 
-			_fixture.Client.SubscribeToAll(EventAppeared,
-				false, filter: new EventTypeFilter(new RegularFilterExpression(new Regex($"^{eventTypePrefix}"))));
+			await _fixture.Client.SubscribeToAllAsync(Position.End, EventAppeared, false,
+				filter: new EventTypeFilter(new RegularFilterExpression(new Regex($"^{eventTypePrefix}"))));
 
 			foreach (var e in events) {
 				await _fixture.Client.AppendToStreamAsync($"{streamPrefix}_{Guid.NewGuid():n}",
@@ -117,8 +121,8 @@ namespace EventStore.Client.Streams {
 			var result = new List<ResolvedEvent>();
 			var source = new TaskCompletionSource<bool>();
 
-			_fixture.Client.SubscribeToAll(EventAppeared,
-				false, filter: new EventTypeFilter(new PrefixFilterExpression(eventTypePrefix)));
+			await _fixture.Client.SubscribeToAllAsync(Position.End, EventAppeared, false,
+				filter: new EventTypeFilter(new PrefixFilterExpression(eventTypePrefix)));
 
 			foreach (var e in events) {
 				await _fixture.Client.AppendToStreamAsync($"{streamPrefix}_{Guid.NewGuid():n}",
@@ -148,5 +152,10 @@ namespace EventStore.Client.Streams {
 			protected override Task When() =>
 				Client.AppendToStreamAsync(FilteredOutStream, AnyStreamRevision.NoStream, CreateTestEvents(10));
 		}
+
+		public Task InitializeAsync() => _fixture.InitializeAsync();
+
+		public Task DisposeAsync() => _fixture.DisposeAsync();
+		public void Dispose() => _loggingContext.Dispose();
 	}
 }
--- a/src/EventStore.Client/EventStore.Client.csproj
+++ b/src/EventStore.Client/EventStore.Client.csproj
@@ -28,9 +28,10 @@
 			<PrivateAssets>all</PrivateAssets>
 			<IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
 		</PackageReference>
+		<PackageReference Include="Microsoft.Extensions.Logging" Version="3.1.0" />
+		<PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
 		<PackageReference Include="System.Linq.Async" Version="4.0.0" />
 		<PackageReference Include="System.Text.Json" Version="4.7.0" />
-		<PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
 	</ItemGroup>
 	<ItemGroup>
 		<Protobuf Include="../Protos/Grpc/*.proto" Exclude="../Protos/Grpc/operations.proto" Access="internal" GrpcServices="Client" ProtoRoot="../Protos/Grpc" />
--- a/src/EventStore.Client/EventStoreGrpcClient.Read.cs
+++ b/src/EventStore.Client/EventStoreGrpcClient.Read.cs
@@ -32,8 +32,10 @@ namespace EventStore.Client {
 			},
 			operationOptions,
 			userCredentials,
-			cancellationToken);
-		
+			cancellationToken)
+			.Where(x => x.Item1 == SubscriptionConfirmation.None)
+			.Select(x => x.Item2);
+
 		/// <summary>
 		/// Asynchronously reads all events.
 		/// </summary>
@@ -79,7 +81,7 @@ namespace EventStore.Client {
 
 			var operationOptions = _settings.OperationOptions.Clone();
 			configureOperationOptions(operationOptions);
-			
+
 			return ReadAllAsync(direction, position, maxCount, operationOptions, resolveLinkTos, filter, userCredentials,
 				cancellationToken);
 		}
@@ -106,8 +108,10 @@ namespace EventStore.Client {
 			},
 			operationOptions,
 			userCredentials,
-			cancellationToken);
-			
+			cancellationToken)
+			.Where(x => x.Item1 == SubscriptionConfirmation.None)
+			.Select(x => x.Item2);
+
 		/// <summary>
 		/// Asynchronously reads all events from a stream.
 		/// </summary>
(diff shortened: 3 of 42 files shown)


src/EventStore.Client.Tests/Streams/AssertEx.cs
```diff
@@ -0,0 +1,39 @@
+using System;
```
jageall: Consider changing the check to just validate the ids for readability

src/EventStore.Client.Tests/Streams/EventDataComparer.cs
```diff
@@ -1,31 +1,34 @@
+using System;
```
jageall: Need to create an issue to replace this with something that gives better error messages

src/EventStore.Client.Tests/Streams/subscribe_to_all_filtered_live.cs
```diff
@@ -84,8 +91,10 @@ public class subscribe_to_all_from_filtered : IClassFixture<subscribe_to_all_fro
 			var result = new List<ResolvedEvent>();
 			var source = new TaskCompletionSource<bool>();
 
-			_fixture.Client.SubscribeToAll(EventAppeared,
-				false, filter: new EventTypeFilter(new RegularFilterExpression(new Regex($"^{eventTypePrefix}"))));
+			_fixture.Client.SubscribeToAll(Position.End, EventAppeared, false,
+				filter: new EventTypeFilter(new RegularFilterExpression(new Regex($"^{eventTypePrefix}"))));
+
+			await Task.Delay(100);
```
jageall: Consider raising an issue so that the client gets notified when the subscription has actually been setup server side

src/EventStore.Client/Logging/LogProvider.cs
```diff
@@ -0,0 +1,27 @@
+using System;
```
jageall: Review further how we want logging to work in the client

src/EventStore.Core/Services/Transport/Grpc/Enumerators.cs
```diff
@@ -1,4 +1,5 @@
 namespace EventStore.Core.Services.Transport.Grpc {
 	internal static partial class Enumerators {
+		private const int MaxLiveEventBufferCount = 16;
```
jageall: Does this need to be configurable?

src/EventStore.Core/Services/Transport/Grpc/Streams.Read.cs
```diff
@@ -140,6 +140,8 @@ partial class Streams {
 					_ => throw new InvalidOperationException()
 				};
 
+			context.CancellationToken.Register(() => enumerator.DisposeAsync());
```
jageall: Does this need to also dispose the registration returned here?

thefringeninja: yes, fixed in latest commit.

src/EventStore.Client.Tests/Streams/subscribe_to_all_filtered_live.cs
thefringeninja: We can; this would require a change to the proto.

src/EventStore.Core/Services/Transport/Grpc/Enumerators.cs
thefringeninja: I'm going to say no, except for possibly testing purposes

src/EventStore.Core/Services/Transport/Grpc/Enumerators.AllSubscription.cs
```diff
@@ -40,95 +51,412 @@ public class AllSubscription : IAsyncEnumerator<ResolvedEvent> {
...
+
+					var (commitPosition, preparePosition) = _nextPosition.ToInt64();
+
+					Log.Trace(
+						"Catch-up subscription {subscriptionId} to $all: reading next page starting from {nextPosition}.",
+						_subscriptionId, _nextPosition);
+
+					_bus.Publish(new ClientMessage.ReadAllEventsForward(
+						correlationId, correlationId, new CallbackEnvelope(OnMessage), commitPosition, preparePosition,
+						32, _resolveLinks, false, default, _user));
```
hayley-jean: `MaxCount` should be configurable on the subscription.
`RequireMaster` should probably also be configured on the client, though that could be handled as a separate PR

src/EventStore.Core/Services/Transport/Grpc/Enumerators.AllSubscriptionFiltered.cs
```diff
@@ -39,104 +49,445 @@ public class AllSubscriptionFiltered : IAsyncEnumerator<ResolvedEvent> {
...
+
+					var (commitPosition, preparePosition) = _nextPosition.ToInt64();
+
+					Log.Trace(
+						"Catch-up subscription {subscriptionId} to $all:{eventFilter} reading next page starting from {nextPosition}.",
+						_subscriptionId, _eventFilter, _nextPosition);
+
+					_bus.Publish(new ClientMessage.FilteredReadAllEventsForward(
+						correlationId, correlationId, new CallbackEnvelope(OnMessage), commitPosition, preparePosition,
+						32, _resolveLinks, false, 32, default, _eventFilter, _user));
```
hayley-jean: See AllSubscription, `MaxCount` should be configurable.
`MaxSearchWindow` should possibly also be configurable.

src/EventStore.Core/Services/Transport/Grpc/Enumerators.AllSubscriptionFiltered.cs
```diff
@@ -39,104 +49,445 @@ public class AllSubscriptionFiltered : IAsyncEnumerator<ResolvedEvent> {
...
+								RpcExceptions.UnknownMessage<ClientMessage.FilteredReadAllEventsForwardCompleted>(
+									message));
+							return;
+						}
+
+						switch (completed.Result) {
+							case FilteredReadAllResult.Success:
+								foreach (var @event in completed.Events) {
+									Log.Trace(
+										"Catch-up subscription {subscriptionId} to $all received event {position}.",
```
hayley-jean: This log should include the filter to match the previous logs

src/EventStore.Core/Services/Transport/Grpc/Enumerators.StreamSubscription.cs
```diff
@@ -42,105 +53,447 @@ public class StreamSubscription : IAsyncEnumerator<ResolvedEvent> {
...
-				if (readIndex == null) throw new ArgumentNullException(nameof(readIndex));
+				if (readIndex == null) {
+					throw new ArgumentNullException(nameof(readIndex));
+				}
 
+				_subscriptionId = Guid.NewGuid();
 				_bus = bus;
 				_streamName = streamName;
-				_nextRevision = startRevision;
+				_startRevision = startRevision == StreamRevision.End ? StreamRevision.Start : startRevision;
```
hayley-jean: When creating a subscription using `StreamRevision.End` it starts from the beginning instead of immediately creating a live subscription.

src/EventStore.Core/Services/Transport/Grpc/Enumerators.StreamSubscription.cs
```diff
@@ -42,105 +53,447 @@ public class StreamSubscription : IAsyncEnumerator<ResolvedEvent> {
...
+						_subscriptionId, _streamName, _nextRevision);
 
-				void OnMessage(Message message) {
-					if (message is ClientMessage.NotHandled notHandled &&
-					    RpcExceptions.TryHandleNotHandled(notHandled, out var ex)) {
-						readNextSource.TrySetException(ex);
-						return;
+					_bus.Publish(new ClientMessage.ReadStreamEventsForward(
+						correlationId, correlationId, new CallbackEnvelope(OnMessage), _streamName,
+						_nextRevision.ToInt64(), 32, _resolveLinks, false, default, _user));
```
hayley-jean: See previous comments about `maxCount`

src/EventStore.Core/Services/Transport/Grpc/Enumerators.StreamSubscription.cs
```diff
@@ -42,105 +53,447 @@ public class StreamSubscription : IAsyncEnumerator<ResolvedEvent> {
...
+									}
+
+									_buffer.Enqueue(@event);
+								}
+
+								_nextRevision = StreamRevision.FromInt64(completed.NextEventNumber);
+								readNextSource.TrySetResult(completed.IsEndOfStream);
+								return;
+							case ReadStreamResult.NoStream:
+								readNextSource.TrySetException(RpcExceptions.StreamNotFound(_streamName));
```
hayley-jean: Should we not start a subscription if there is no stream?

src/EventStore.Core/Services/Transport/Grpc/Enumerators.cs
hayley-jean: This should be configurable, and 16 is likely too low as a default.
If I remember correctly, the default in the existing catchup subscriptions is in the hundreds

src/EventStore.Client/EventStore.Client.csproj
```diff
@@ -18,6 +18,7 @@
 		<RootNamespace>EventStore.Client</RootNamespace>
 	</PropertyGroup>
 	<ItemGroup>
+		<PackageReference Include="CompareNETObjects" Version="4.65.0" />
```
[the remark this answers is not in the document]
thefringeninja: No!

src/EventStore.Core/Services/Transport/Grpc/Enumerators.AllSubscription.cs
```diff
@@ -40,95 +51,412 @@ public class AllSubscription : IAsyncEnumerator<ResolvedEvent> {
...
-				await readNextSource.Task.ConfigureAwait(false);
+				return await _inner.MoveNextAsync().ConfigureAwait(false);
+			}
 
-				if (_disposedTokenSource.IsCancellationRequested) {
-					return false;
+			public ValueTask DisposeAsync() => _inner.DisposeAsync();
+
+			private async ValueTask OnLiveSubscriptionDropped(Position caughtUpPosition) {
+				Log.Trace(
```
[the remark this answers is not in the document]
thefringeninja: Yes, this is strange behavior as I would assume that Information was the minimum allowed in the console logger.

src/EventStore.Core/Services/Transport/Grpc/Enumerators.AllSubscription.cs
```diff
@@ -40,95 +51,412 @@ public class AllSubscription : IAsyncEnumerator<ResolvedEvent> {
...
+										Fail(RpcExceptions.UnknownError(dropped.Reason));
+										return;
+								}
+							case ClientMessage.StreamEventAppeared appeared:
+								var fromPosition = Position.FromInt64(appeared.Event.OriginalPosition.Value.CommitPosition,
+									appeared.Event.OriginalPosition.Value.PreparePosition);
+
+								if (!_subscriptionConfirmed.Task.IsCompleted) {
+									_subscriptionConfirmed.TrySetResult(fromPosition);
+									ReadHistoricalEvents(fromPosition);
```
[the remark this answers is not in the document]
thefringeninja: In fact i tried that first but the tests wouldn't pass...

src/EventStore.Core/Services/Transport/Grpc/Enumerators.StreamSubscription.cs
thefringeninja: This is the current behavior from the client api

src/EventStore.Core/Services/Transport/Grpc/Enumerators.AllSubscription.cs
```diff
@@ -40,95 +51,412 @@ public class AllSubscription : IAsyncEnumerator<ResolvedEvent> {
...
+
+					void ReadHistoricalEvents(Position fromPosition) {
+						Log.Trace(
+							"Live subscription {subscriptionId} to $all loading any missed events starting at {fromPosition}",
+							subscriptionId, fromPosition);
+
+						var correlationId = Guid.NewGuid();
+						var (commitPosition, preparePosition) = fromPosition.ToInt64();
+						bus.Publish(new ClientMessage.ReadAllEventsBackward(correlationId, correlationId,
+							new CallbackEnvelope(OnHistoricalEventsMessage), commitPosition, preparePosition, 32,
```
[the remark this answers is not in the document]
thefringeninja: Made it a const for now

src/EventStore.Core/Services/Transport/Grpc/Enumerators.StreamSubscription.cs
thefringeninja: Fixed this by immediately choosing a live subscription if subscribing from End

src/EventStore.Core/Services/Transport/Grpc/Enumerators.cs
thefringeninja: I am reluctant to make this configurable as subscriptions now live on the server, setting this too high could take down a node.

hayley-jean: Fair enough

src/EventStore.Client/EventStoreGrpcClient.Read.cs
```diff
@@ -181,19 +185,27 @@ public partial class EventStoreClient {
...
 				.ReadAllAsync(cancellationToken)
-				.Select(ConvertToResolvedEvent)
+				.Select(ConvertToItem)
 				.WithCancellation(cancellationToken)
 				.ConfigureAwait(false)) {
 				yield return e;
 			}
 
-			ResolvedEvent ConvertToResolvedEvent(ReadResp response) =>
+			(SubscriptionConfirmation, ResolvedEvent) ConvertToItem(ReadResp response) => response.ContentCase switch {
```
jageall: We might want to do some bench marks around this. I believe it is just destructuring anonymous tuples that has performance implications, so I don't think we are affected by it here, but we should check in the RC time frame

src/EventStore.Client/RecordedEvent.cs
```diff
@@ -1,3 +1,5 @@
+using System.Threading.Tasks;
```
jageall: Why do we need this namespace imported?

PR #2210, opened 2020-01-08
Allow Clients to Specify Which UUID Type They Want

thefringeninja:


src/EventStore.Grpc.Common/Uuid.cs
```diff
@@ -1,5 +1,6 @@
 using System;
 
+
```
jen20:
```suggestion
```

src/EventStore.Core/Services/Transport/Grpc/PersistentSubscriptions.Read.cs
```diff
@@ -86,8 +87,13 @@ public partial class PersistentSubscriptions {
 				if (e == null) return null;
 				var position = Position.FromInt64(commitPosition ?? e.LogPosition, e.TransactionPosition);
 				return new ReadResp.Types.ReadEvent.Types.RecordedEvent {
-					Id = new UUID {
-						String = e.EventId.ToString()
+					Id = uuidOptionsCase switch {
```
jen20: Do we want to default here if this is not set in the message?

thefringeninja: actually yeah

PR #2214, opened 2020-01-10
Fix race condition in TCP connections which allowed messages to be dispatched after connection closed

shaan1337:
Fixes: https://github.com/EventStore/EventStore/issues/1392
May also fix other bugs in both server or client.

## Bug
There is a race condition in TCP connections that allows data/messages to be dispatched even after the TCP connection has been closed.

In the above issue: #1392, the following was happening:
- [ClientMessage.ConnectToPersistentSubscription message](https://github.com/EventStore/EventStore/blob/master/src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs#L407) was being received after [TcpMessage.ConnectionClosed message](https://github.com/EventStore/EventStore/blob/master/src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs#L397)
- When the `ConnectionClosed` message is received, the persistent subscription service will attempt to remove any subscription that matches this connection ID, but it does not exist yet.
- When the `ConnectToPersistentSubscription` message is received, the persistent subscription service will create a connection, although the socket has already been closed.
- This will result in a stale connection and the persistent subscription will silently attempt to dispatch events to this connection. These events will never be acknowledged and the persistent subscription will keep retrying (possibly on a different connection depending on chosen strategy) until it eventually it is dispatched or parked. If the strategy is `Round Robin` and there are a good number of stale connections, it may appear as if the projection is no longer processing any event, or if the chosen strategy is `DispatchToSingle` it will never be able to dispatch any event once it hits the stale connection.
- Proof of concept:  [persistentsub_quick_disconnect.zip](https://github.com/EventStore/EventStore/files/4045569/persistentsub_quick_disconnect.zip)

Launch a single node, then launch the PoC with `dotnet run`. Number of connections will keep increasing:
`curl -s -v -k https://localhost:2113/subscriptions/stream/group/info -H "accept:application/json" -u admin:changeit|jq -r '.connections | length'`

## Race condition
- It is possible that while sending data, [`CloseInternal` is called](https://github.com/EventStore/EventStore/blob/master/src/EventStore.Transport.Tcp/TcpConnection.cs#L175). This may be called from a threadpool thread:
`OnSendAsyncCompleted` -> `ProcessSend` -> `CloseInternal`
- `TryDequeueReceivedData` may also be called from a threadpool thread:
`OnReceiveAsyncCompleted` -> `ProcessReceive` -> `TryDequeueReceivedData`
- Thus it's possible that `callback(this, data);` in `TryDequeueReceivedData` is called after `CloseInternal` is called since sending/receiving occur in parallel

There are other possibilities (not race conditions) where the same issue occurs:
`ProcessReceive` -> `StartReceive` -> !firedAsync -> `ProcessReceive` -> `CloseInternal`. After `StartReceive` returns, it will call `TryDequeueReceivedData` although the connection has already been closed.

## Fix description
For `TcpConnection` classes, the fix simply adds a flag: `dispatchingData` which can have 3 different states:
- `0` - not dispatching any data
- `1` - data is being dispatched
- `2` - final state

When data is being dispatched, transition from `0` to `1` and back to `0` once the callback is done.
When the connection is closed, spin until no data is being dispatched (dispatchingData=0), then transition to the final state (dispatchingData=2). No data is dispatched after this.

For `TcpConnectionSsl` classes, the fix is similar to above but uses the existing `_receiveHandling` lock and transitions to a value of `2` when closing the connection.

## Regression tests
Regression tests have been added to:
- `TcpConnection.cs` (server)
- `TcpConnection.cs` (client)
- `TcpConnectionSsl.cs` (server)
- `TcpConnectionSsl.cs` (client)

All four of them fail the regression test (intermittently) when the fix is reverted. It appears that it fails more often when CPU affinity is set to 2 processors (for reviewing/testing purposes):
```
taskset -c 0,1 dotnet test -v normal -c Release src/EventStore.ClientAPI.Tests/EventStore.ClientAPI.Tests.csproj --filter FullyQualifiedName~EventStore.ClientAPI.Tests.Services.Transport.Tcp.TcpConnectionSslTests
```

Note: Unused TcpConnectionLockless classes have also been deleted.


src/EventStore.Transport.Tcp/TcpConnectionSsl.cs
```diff
@@ -91,7 +91,7 @@ public class TcpConnectionSsl : TcpConnectionBase, ITcpConnection {
 
 		private readonly object _streamLock = new object();
 		private bool _isSending;
-		private int _receiveHandling;
+		private int _receiveHandling; //states: 0 - not receiving data, 1 - receiving/dispatching data, 2 - final state, no data received/dispatched after reaching this state
```
jen20: Could this not be an enumeration instead?

PR #2218, opened 2020-01-13
Respect Max Count When Reading From Everything

thefringeninja:


src/EventStore.Core/Services/Transport/Grpc/Enumerators.ReadAllBackwardsFiltered.cs
```diff
@@ -93,7 +96,7 @@ public class ReadAllBackwardsFiltered : IAsyncEnumerator<ResolvedEvent> {
 
 				_bus.Publish(new ClientMessage.FilteredReadAllEventsBackward(
 					correlationId, correlationId, new CallbackEnvelope(OnMessage),
-					commitPosition, preparePosition, Math.Min(_maxCount, 32),
+					commitPosition, preparePosition, 32,
```
jen20: Same question as above.

src/EventStore.Core/Services/Transport/Grpc/Enumerators.ReadAllForwards.cs
```diff
@@ -76,7 +78,7 @@ public class ReadAllForwards : IAsyncEnumerator<ResolvedEvent> {
 
 				_bus.Publish(new ClientMessage.ReadAllEventsForward(
 					correlationId, correlationId, new CallbackEnvelope(OnMessage),
-					commitPosition, preparePosition, Math.Min(_maxCount, 32),
+					commitPosition, preparePosition, 32,
```
jen20: Same as above.

src/EventStore.Core/Services/Transport/Grpc/Enumerators.ReadAllBackwards.cs
```diff
@@ -76,7 +79,7 @@ public class ReadAllBackwards : IAsyncEnumerator<ResolvedEvent> {
 
 				_bus.Publish(new ClientMessage.ReadAllEventsBackward(
 					correlationId, correlationId, new CallbackEnvelope(OnMessage),
-					commitPosition, preparePosition, Math.Min(_maxCount, 32),
+					commitPosition, preparePosition, 32,
```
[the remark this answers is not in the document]
thefringeninja: IIRC I put that in there because I had confused max count with page size. To me it only makes sense if someone is trying to read the all stream with a page size of 1, which they can't do now

PR #2200, opened 2020-01-04
Commit Tracker

condron:
Centralizes commit status in a single deterministic service and allowing for selecting different required commit levels before returning requests.

- `MasterWrite` Write completed on Master Node
- `ClusterWrite` Write completed a quorum of nodes in the cluster
-  `MasterIndexed` Indexed (aka readable) on Master Node

n.b. explicit `TransactionCommit` operations require `MasterIndex`

Rationalizes Request Manager via the creation of a common `RequestManagerBase` linked to the `CommitTrackingService`

Decouples `StorageChaser` `IndexCommiterService` and RequestManagers on preparation for refactoring the `IndexCommitter`.

WIP Open Items
- [x] Authorization error in `OneTimeSetup` in Http.Transport.Authorization` test fixture
- [x] WriteEvents error in `EventStore.Core.Tests.Integration.specification_with_cluster` `OneTimeSetup`
- [ ] Address root cause on Projections test failures
- [ ] Expose command line settings for `CommitLevel`
- [X] Additional tests for commit levels with 3 node clusters
- [ ] Additional tests for commit levels with 5 node clusters
- [ ] Patch oss client v5 with #2192

PR #2219, opened 2020-01-14
Replication Tracking

### LINKED ISSUE #2037: Reading events that have not been replicated
If a client tries to read events that have been written, but not replicated, the read returns an empty result.

If we are able to determine that there is an event that is waiting to be acked by the other nodes in the cluster, we should rather delay until the write has been acked or timed out before returning to the client.

condron:
- Centralizes replication status in a single deterministic service on master.
- Enforce index writes are committed after cluster replication on all nodes.
- Normalizes request handling in Request Managers 
- Add tests for request managers
- Adds Tests for replication in 1, 3, and 5 node clusters
- Decouples `StorageChaser` `IndexCommitterService` and `RequestManagers` in preparation for refactoring the `IndexCommitter`.
- Resolves #2037 
- Resolves potential for "Stale Reads" on replica nodes
- Replaces #2200 

Open items 
- [ ] Race conditions in Projection Tests
- [ ] Grpc Projection status return types


jen20:
Closed in favour of #2224.

PR #2232, opened 2020-01-16
Expose Read Direction in Client API

pgermishuys:
Fixes https://github.com/EventStore/EventStore/issues/2231

Remove ReadDirection specific methods from the Client API


jen20:
How about `Direction` instead of `ReadDirection` (just for brevity)?

Issue #2234, opened 2020-01-17
Review test for rapid Projection Subsystem restart


condron:
The test `when_projection_subsystem_restarted_twice` is unstable in the CI builds.

While a useful developer test, rapid and repeated restarts of the Projection subsystem is not a system feature or goal.

The test has been marked `Explicit` as a work around to avoid false negatives in the CI test runs in #2235 

Requested resolution is any one of 
- Confirm the `Explicit` designation 
- Fix the test to be stable in CI
- Remove the test as unwarranted

PR #2235, opened 2020-01-17
Test stability fixes

condron:
Projection subsystem test changes per #2234 are out of scope, workaround added, decision required

Replication Tracking Test instability fixed via adding deterministic checking on published status

Read request tests fixed by adding deterministic `IndexCheckpoint` evaluation on replica nodes before attempting to read

Fixing Instability due to projections race conditions #2236 is out of scope for this PR, workarounds added.

Instability due to GRPC issues is out of scope #2237
Instability due to V8 issues are out of scope #2240 workaround added


src/EventStore.Client.Tests.Common/EventStoreGrpcFixture.cs
```diff
@@ -74,6 +74,7 @@ protected static EventData CreateTestEvent(int index, string type)
 			Client?.Dispose();
 		}
 
+
```
jen20:
```suggestion
```

PR #2245, opened 2020-01-22
Use int64 for Time Consistently

### LINKED ISSUE #2242: users.proto time encoding
The `last_updated` field below could be a `int64` representing ticks since unix epoch as it is done for `created` metadata field on streams:

```proto
message DetailsResp {
	UserDetails user_details = 1;
	message UserDetails {
		string login_name = 1;
		string full_name = 2;
		repeated string groups = 3;
		string last_updated = 4;
		bool disabled = 5;
	}
}
```

thefringeninja:
Fixes #2242

### DIFF
--- a/src/Protos/Grpc/users.proto
+++ b/src/Protos/Grpc/users.proto
@@ -86,8 +86,14 @@ message DetailsResp {
 		string login_name = 1;
 		string full_name = 2;
 		repeated string groups = 3;
-		string last_updated = 4;
+		DateTime last_updated = 4;
 		bool disabled = 5;
+
+		message DateTime {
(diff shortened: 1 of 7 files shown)


src/Protos/Grpc/users.proto
```diff
@@ -86,8 +86,14 @@ message DetailsResp {
 		string login_name = 1;
 		string full_name = 2;
 		repeated string groups = 3;
-		string last_updated = 4;
+		DateTime last_updated = 4;
 		bool disabled = 5;
+
+		message DateTime {
+			int64 value = 1;
```
ahjohannessen: perhaps `ticks_since_epoch` instead of `value` ?

PR #2252, opened 2020-01-23
Replication Tracking Fix

### LINKED ISSUE #2244: [Preview 2] All writes to a cluster fail after a node loses connection
If a non-master node loses connection to the master node, all following writes to the cluster will time out on the clients.
The data is still written and present after a restart, however the new events cannot be read and will not be published to subscriptions.

After getting into this state, the cluster cannot recover until the master node is resigned or restarted.
Bringing the dead node back online does not fix the issue

It appears that this is because the replication checkpoint is not accurately updated, and does not prune dead nodes.

**Reproduction steps:**

1. Start a cluster of 3 nodes
2. Kill one of the slaves
3. Issue a write to one of the nodes in the cluster. This write will succeed.
4. Issue another write to the cluster, this write and any other writes will fail.

condron:
Resolves #2244


condron:
squashed the additional file name commit down, no changes but I do need an updated review @hayley-jean

PR #2253, opened 2020-01-23
Allow Persistent Subscription to be created with Timeout of 0 from HTTP

### LINKED ISSUE #1479: HTTP: Allow persistent subscription to be created with timeout of 0

arwinneil:
- Change `DefaultValueHandling` to `Include` so the serializer does not ignore `MessageTimeoutMilliseconds` when set to 0

- Added `ToMessageTimeout()` & `ToCheckPointAfterTimeout()` for respective handling of conversion from `int` to `TimeSpan` when set to 0

Resolves #1479

### DIFF
--- a/src/EventStore.Transport.Http/Codecs/JsonCodec.cs
+++ b/src/EventStore.Transport.Http/Codecs/JsonCodec.cs
@@ -16,7 +16,7 @@ namespace EventStore.Transport.Http.Codecs {
 			ContractResolver = new CamelCasePropertyNamesContractResolver(),
 			DateParseHandling = DateParseHandling.None,
 			NullValueHandling = NullValueHandling.Ignore,
-			DefaultValueHandling = DefaultValueHandling.Ignore,
+			DefaultValueHandling = DefaultValueHandling.Include,
 			MissingMemberHandling = MissingMemberHandling.Ignore,
 			TypeNameHandling = TypeNameHandling.None,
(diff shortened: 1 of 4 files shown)


src/EventStore.Transport.Http/Codecs/JsonCodec.cs
```diff
@@ -16,7 +16,7 @@ public class JsonCodec : ICodec {
 			ContractResolver = new CamelCasePropertyNamesContractResolver(),
 			DateParseHandling = DateParseHandling.None,
 			NullValueHandling = NullValueHandling.Ignore,
-			DefaultValueHandling = DefaultValueHandling.Ignore,
+			DefaultValueHandling = DefaultValueHandling.Include,
```
jageall: This change will alter the behaviour for pretty much anything that uses json over http.. are we sure this is a good idea @pgermishuys @hayley-jean ?

hayley-jean: Judging by where this is used, this should be fine and is likely what we actually want to happen in the first place

PR #2257, opened 2020-01-29
Allow projections to subscribe from the end of the transaction log

### LINKED ISSUE #1911: Processing stream not from beginning
I know this is not the intended use, but I'm wondering if it's possible to create projection that will process events on stream from now on only. Without processing older ones. Or even to specify point in time from which events will be processed.

Surely events could be filtered by projection, but the projection would still have to process the whole stream before being ready.

hayley-jean:
Fixes #1911
Related UI change https://github.com/EventStore/EventStore.UI/pull/234

Add a `subscribeFromEnd` option to projection creation.
This option has been added to the HTTP, TCP and gRPC API's.

When this is set, the projection will not start an `EventReader`, but will rather immediately subscribe to the `HeadingEventReader`.
This subscription will start from whatever position the `HeadingEventReader` is at at the time of the projection's start.

It will also cause the projection to write a checkpoint on the first event it processes.
At the moment, it will need to receive an event that passes the filter in order for it to write the checkpoint.

Once a checkpoint has been written, restarting the projection will continue it from the checkpointed position rather than from the end of the stream.
If the projection is reset, it will start again from the end of the transaction log.

**Note:**
The proto for the gRPC projections client has changed.

### DIFF
--- a/src/EventStore.Client.Projections.Tests/Projections/create.cs
+++ b/src/EventStore.Client.Projections.Tests/Projections/create.cs
@@ -15,11 +15,11 @@ namespace EventStore.Client.Projections {
 				"fromAll().when({$init: function (state, ev) {return {};}});", TestCredentials.Root);
 		}
 
-		[Theory, InlineData(true), InlineData(false)]
-		public async Task continuous(bool trackEmittedStreams) {
+		[Theory, InlineData(true, true), InlineData(false, false)]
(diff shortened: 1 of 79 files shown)


hayley-jean:
After discussion with @jageall, we are going to delay this feature until after the release of v6 as we would like to take the time to do it properly. This is not a breaking change, so can be added afterwards.

PR #2258, opened 2020-01-29
Add Host Statistics

### LINKED ISSUE #2208: Use native methods for collecting CPU and Memory statistics on Unixes
We need to find a suitable replacement for the `sys.cpu` (System CPU %) cross platform.
Once https://github.com/EventStore/EventStore/pull/2204 is merged, the `sys.cpu` will only be available via Windows as it still uses Performance Counters there.

pgermishuys:
Fixes #2208 

Add [HostStats.NET](https://www.nuget.org/packages/HostStats.NET).

- Introduce CPU Load Averages for Linux and OSX
- Replace Individual Free Memory calls with calls to HostStat.NET

### Note:
- There is a difference between Windows and other operating systems for fetching CPU related information. Windows gets CPU Usage whereas the other OSes gets Load Averages

### DIFF
--- a/src/EventStore.Core/Services/Monitoring/SystemStatsHelper.cs
+++ b/src/EventStore.Core/Services/Monitoring/SystemStatsHelper.cs
@@ -128,9 +128,9 @@ namespace EventStore.Core.Services.Monitoring {
 					case OsFlavor.Linux:
 					case OsFlavor.MacOS:
 						var loadAverages = _hostStat.GetLoadAverages();
-						stats["sys-cpu-1m-loadaverage"] = loadAverages.Average1m;
-						stats["sys-cpu-5m-loadaverage"] = loadAverages.Average5m;
-						stats["sys-cpu-15m-loadaverage"] = loadAverages.Average15m;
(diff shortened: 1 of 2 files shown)


src/EventStore.Core/Services/Monitoring/SystemStatsHelper.cs
```diff
@@ -119,7 +121,22 @@ public class SystemStatsHelper : IDisposable {
...
 
-				stats["sys-cpu"] = GetTotalCpuUsage();
+				switch (OS.OsFlavor) {
+					case OsFlavor.Windows:
+						stats["sys-cpu"] = _perfCounter.GetTotalCpuUsage();
+						break;
+					case OsFlavor.Linux:
+					case OsFlavor.MacOS:
+						var loadAverages = _hostStat.GetLoadAverages();
+						stats["sys-cpu-1m-loadaverage"] = loadAverages.Average1m;
```
jen20: I'd remove "cpu" from these I think, and make them `sys-loadavg-1m` etc

PR #2269, opened 2020-02-04
Introduce ability to set deadlines (timeouts) in the gRPC Client

pgermishuys:
Introduces the ability to specify a deadline on the gRPC operations.

In addition to the ability to set deadlines, we introduced the notion of operation options which can be configured per operation.

**Note**
- There are no tests for the subscriptions client operations. They only set the timeout to never timeout.

### DIFF
--- a/src/EventStore.Client/EventStoreGrpcClient.Read.cs
+++ b/src/EventStore.Client/EventStoreGrpcClient.Read.cs
@@ -5,7 +5,6 @@ using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
 using EventStore.Client.Streams;
-using Google.Protobuf.WellKnownTypes;
 using Grpc.Core;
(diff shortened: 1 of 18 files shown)


src/EventStore.Client/EventStoreGrpcClient.Read.cs
```diff
@@ -16,15 +15,17 @@ public partial class EventStoreClient {
 		/// <param name="direction">The direction in which to read. <see cref="Direction"/></param>
 		/// <param name="position">The position to start reading from.</param>
 		/// <param name="maxCount">The maximum count to read.</param>
+		/// <param name="operationOptions"><see cref="EventStoreClientOperationOptions" /> to perform the operation with.</param>
```
hayley-jean: Given that these methods are now private, should the documentation be moved to the public ones?

PR #2272, opened 2020-02-05
Add option to enable or disable ATOM Controller

Lougarou:
PR adds option to Enable/Disable Atom Controller and shows a deprecation message if Atom controller interface is enabled. Atom Controller is enabled by default, for now.
Flag is --enable-atom-pub-over-http

Breaks UI
This breaks stream browser page and every url /streams/* i.e. viewing streams, events and adding new events. Admin UI is partly broken, scavenges are not shown although scavenge button will still trigger scavenging.

Related to issue https://github.com/EventStore/EventStore/issues/2268
Enable external tcp interface moved to another pr

### DIFF
--- a/src/EventStore.ClusterNode/ClusterNodeOptions.cs
+++ b/src/EventStore.ClusterNode/ClusterNodeOptions.cs
@@ -313,7 +313,10 @@ namespace EventStore.ClusterNode {
 		
 		[ArgDescription(Opts.DevDescr, Opts.AppGroup)]
 		public bool Dev { get; set; }
-
+		
+		[ArgDescription(Opts.EnableHTTPInterfaceDescr, Opts.InterfacesGroup)]
+		public bool EnableHTTPInterface { get; set; }
+		
 		public ClusterNodeOptions() {
 			Config = "";
 			Help = Opts.ShowHelpDefault;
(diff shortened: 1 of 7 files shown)


src/EventStore.ClusterNode/Program.cs
```diff
@@ -110,6 +110,10 @@ public class Program : ProgramBase<ClusterNodeOptions> {
 					"========================================================================================================\n");
 			}
 
+			if (opts.EnableAtomPubOverHTTP) {
+				Log.Warn("\n DEPRECATION WARNING: AtomPub over HTTP Interface has been deprecated in version 6. It is recommended to use GRPC instead.\n");
```
hayley-jean: Is it correct to call it version 6 here? Should this not be 20.02?

Lougarou: @hayley-jean thanks changed to 20.02

jageall:
Lets add an issue to the RC to fix view parked messages @mat-mcloughlin , we can do the same as we did for scavenge and route it through the PersistentSubscriptions controller instead of the atom controller. Then once @thefringeninja does his review, this can get merged

jageall:
Lets add an issue to the RC to fix view parked messages @mat-mcloughlin , we can do the same as we did for scavenge and route it through the PersistentSubscriptions controller instead of the atom controller. Then once @thefringeninja does his review, this can get merged

Lougarou:
@jageall this PR does not reroute reading scavenge streams to admin controller

PR #2279, opened 2020-02-11
Add flag to enable external TCP interface

Lougarou:
Flag to enable external tcp interface. External TCP Interface is disabled by default.
--enable-external-tcp
Related to issue https://github.com/EventStore/EventStore/issues/2268


Lougarou:
todo: don't let people set external tcp related settings when this is false

Lougarou:
TODO: Need to fix failing tests
TODO: Add deprecation message

PR #2293, opened 2020-02-17
Fetch All Tags

thefringeninja:
minver is not selecting the correct docker tag as git tags are not being fetched


thefringeninja:
```
MinVer: Using { Commit: 018de99, Tag: 'oss-v6.0.0-preview2', Version: 6.0.0-preview2, Height: 23 }.
MinVer: Calculated version 6.0.0-preview2.23.
```
this is what I'd expect to see

PR #2295, opened 2020-02-17
Replace IsJson w/ Content-Type

### LINKED ISSUE #2294: Replace IsJson with Content-Type in Client
Currently we have an `isJson` flag in `EventData` to indicate what 'type' exists in the `Data` field. This is really inflexible. Replacing this with a field like `Content-Type` would allow for much greater flexibility down the line - schema validation comes to mind.

For now, until we update the underlying storage format, we can only support two `Content-Type`:

- `application/json`
- `application/octet-stream`

thefringeninja:
Fixes #2294


src/EventStore.Client/EventData.cs
```diff
@@ -1,26 +1,41 @@
...
 		public readonly byte[] Data;
 		public readonly Uuid EventId;
 		public readonly byte[] Metadata;
 		public readonly string Type;
-		public readonly bool IsJson;
+		public readonly string ContentType;
 
-		public EventData(Uuid eventId, string type, byte[] data, byte[] metadata = default, bool isJson = true) {
+		public EventData(Uuid eventId, string type, byte[] data, byte[] metadata = default,
+			string contentType = "application/json") {
```
pgermishuys:
```suggestion
			string contentType = Constants.Metadata.ContentTypes.ApplicationJson) {
```

PR #2306, opened 2020-02-21
Internal gossip and elections over gRPC

### LINKED ISSUE #2290: Upgrade elections and gossip to run over gRPC

hayley-jean:
Fixes #2290 

- Add proto definition for internal gossip and elections
- Add a Cluster gRPC service and client
- Add a cluster client cache to allow channel reuse
- Remove ElectController and Post for the GossipController
- Remove the internal http service
- Forward external http requests over the external interface, not the internal one
- Increase the log level for gRPC errors from `Warning` to `Fatal`

**Note:** This PR also removes the internal http service.
None of the http controllers will be available over internal http after this PR is merged.


src/Protos/Grpc/cluster.proto
```diff
@@ -0,0 +1,138 @@
+﻿syntax = "proto3";
+package event_store.cluster;
+
+import "shared.proto";
```
thefringeninja: Some of this is snake_case and some of it is pascalCase

ahjohannessen:
Why no ability to subscribe to gossip info with best node info as grpc stream instead of req/res?

Seems like clients have to do a lot of logic determining best nodes where this perhaps rather could be provided by grpc streams with fresh and continually updated info.

pgermishuys:
Hi @ahjohannessen. This PR specifically addresses internal communications over gRPC.

src/EventStore.Core/Services/GrpcSendService.cs
```diff
@@ -0,0 +1,64 @@
...
+				case ElectionMessage.LeaderIsResigning resigning:
+					_eventStoreClientCache.Get(message.DestinationEndpoint)
+						.SendLeaderIsResigning(resigning, message.DestinationEndpoint, message.LiveUntil);
+					break;
+				case ElectionMessage.LeaderIsResigningOk resigningOk:
+					_eventStoreClientCache.Get(message.DestinationEndpoint)
+						.SendLeaderIsResigningOk(resigningOk, message.DestinationEndpoint, message.LiveUntil);
+					break;
+				default:
+					throw new NotImplementedException($"Message of type {message.GetType().Name} cannot be handled.");
```
jageall: Is this logging the right type? the switch is on message.Message where as GetType is called on message

src/EventStore.Core/Services/Transport/Grpc/Cluster.Elections.cs
```diff
@@ -0,0 +1,98 @@
...
+using System.Threading.Tasks;
+using EventStore.Client;
+using EventStore.Client.Shared;
+using EventStore.Cluster;
+using EventStore.Core.Messages;
+using Grpc.Core;
+
+namespace EventStore.Core.Services.Transport.Grpc {
+	partial class Cluster {
+		public override Task<Empty> ViewChange(ViewChangeRequest request, ServerCallContext context) {
```
jageall: We can cache the completed task in this case as a static singleton which will reduce allocations

src/Protos/Grpc/cluster.proto
jageall: I believe snake case is preferred for interop @thefringeninja ?

src/Protos/Grpc/cluster.proto
```diff
@@ -0,0 +1,138 @@
...
+
+service Cluster {
```
jageall: It might be better to split this service into Gossip {read, update} and Election{viewchange, ...}  that will eliminate Cluster.Cluster which is a pet peeve.

ahjohannessen:
@pgermishuys Ok Pieter, thanks for the heads up :)

src/EventStore.Common/Log/logconfig.json
```diff
@@ -4,7 +4,7 @@
 	  "Default": "Debug",
 	  "System": "Warning",
 	  "Microsoft": "Warning",
-	  "Grpc": "Warning"
+	  "Grpc": "Fatal"
```
thefringeninja: What's the reason for this?

PR #2308, opened 2020-02-24
EventData byte[] -> ReadOnlyMemory<byte>

martinothamar:
Changed: Replace `byte[]` with `ReadOnlyMemory<byte>` to reduce allocations

Since `EventData` holds byte arrays we're having limited options of keeping allocations down. Trying to dig into what it would take to use `ReadOnlyMemory<byte>` instead of byte arrays.

Issue: #2309


jen20:
Hi @martinothamar! This is definitely something we're interested in exploring, however more likely than not that will be for the new gRPC client rather than the older one which is about to be deprecated. We have a secondary issue that we have to support .NET Framework users who cannot upgrade yet also, but that is likely to be via a second client ultimately since the .NET gRPC bits are already core-only.

thefringeninja:
Hi @martinothamar can you rebase your PR and force push please? Azure has not kept around the build results.

PR #2331, opened 2020-03-03
Add certificate generation script

pgermishuys:
- Provide the ability to specify the SANS when using the terraform certificate generation.
- Add option (`CertificatePrivateKeyFile`) to provide the ability to set a private key when loading up a certificate.
- Use the certificate and key pair without the password in development mode.

#### Usage
```
> terraform init
> terraform plan
> terraform plan
var.dns_names
  Enter a value: ["eventstore.localhost.com", "localhost"]

var.ip_addresses
  Enter a value: ["127.0.0.1", "10.7.14.150"]
```


jen20:
Personally I do not like relying on OpenSSL instead of Terraform here. It's commonly not available on Windows, and it is more difficult to verify that things are generated correctly.

jen20:
If we decide to go with this approach instead of TF though, we should ensure that `shellcheck(1)` reports no potential issues with it (there are a few right now - notably SC2034, SC2164 and SC2059).

src/EventStore.ClusterNode/ClusterVNodeHostedService.cs
```diff
@@ -44,17 +44,17 @@ internal class ClusterVNodeHostedService : EventStoreHostedService<ClusterNodeOp
 					x.Source = "Set by 'Development Mode' mode";
 				}
 
-				if (x.Name == nameof(ClusterNodeOptions.CertificatePassword)
+				if (x.Name == nameof(ClusterNodeOptions.CertificateFile)
 				    && x.Source == "<DEFAULT>"
 				    && developmentMode) {
-					x.Value = SystemUsers.DefaultAdminPassword;
+					x.Value = Path.Combine(Locations.DevCertificateDirectory, "server1.pem");
```
jen20: I suspect we do not want a default here?

src/EventStore.ClusterNode/ClusterVNodeHostedService.cs
```diff
@@ -44,17 +44,17 @@ internal class ClusterVNodeHostedService : EventStoreHostedService<ClusterNodeOp
...
 					x.Source = "Set by 'Development Mode' mode";
 				}
-
-				if (x.Name == nameof(ClusterNodeOptions.CertificateFile)
+				
+				if (x.Name == nameof(ClusterNodeOptions.CertificatePrivateKeyFile)
 				    && x.Source == "<DEFAULT>"
 				    && developmentMode) {
-					x.Value = Path.Combine(Locations.DevCertificateDirectory, "server1.pfx");
+					x.Value = Path.Combine(Locations.DevCertificateDirectory, "server1.key");
```
jen20: Same here re: defaults

src/EventStore.ClusterNode/ClusterVNodeHostedService.cs
jen20: Actually I misread this, for development mode this should be fine.

src/EventStore.ClusterNode/ClusterVNodeHostedService.cs
jen20: Likewise for this.

src/EventStore.Core/VNodeBuilder.cs
```diff
@@ -677,13 +679,31 @@ public abstract class VNodeBuilder {
...
+			string password) {
 
-			_certificate = cert;
+			if (string.IsNullOrEmpty(privateKeyPath)) {
+				_certificate = new X509Certificate2(certificatePath, password);
+				return this;
+			}
+			
+			var privateKey = Convert.FromBase64String(string.Join(string.Empty, File.ReadAllLines(privateKeyPath)
+			.Skip(1)
```
jen20: Formatting here looks a little screwy.

PR #2332, opened 2020-03-03
Thread Through Cancellation Token on Appends and Deletes

### LINKED ISSUE #2277: Thread gRPC deadlines through to the server to avoid unnecessary work being done on the server
With the introduction of the ability to set [deadlines](https://github.com/EventStore/EventStore/pull/2269) in the gRPC Client. We can leverage this deadline on the server by threading it through to our read and write operations as we already have the ability to shed load on the server.

Here is where we set the timeouts for reads https://github.com/EventStore/EventStore/blob/master/src/EventStore.Core/Messages/ClientMessage.cs#L92 and the writes https://github.com/EventStore/EventStore/blob/master/src/EventStore.Core/Services/RequestManager/Managers/WriteEvents.cs#L65 on the server.

If we don't thread this through on the server we might have the following scenario occur.
The default and hardcoded timeout for reads on the server is **10 seconds**. If the client were to set a timeout/deadline of **2 seconds**, the client could send a number of operations that will actually have the server do work where the client has already abandoned the request(s).

thefringeninja:
Changed: Use cooperative cancellation instead of timeouts to cancel appends and deletes

Fixes #2277

### DIFF
--- a/src/EventStore.Core/ClusterVNode.cs
+++ b/src/EventStore.Core/ClusterVNode.cs
@@ -502,7 +502,8 @@ namespace EventStore.Core {
 			var requestManagement = new RequestManagementService(
 				_mainQueue,
 				vNodeSettings.PrepareTimeout,
-				vNodeSettings.CommitTimeout);
+				vNodeSettings.CommitTimeout,
+				vNodeSettings.WriteTimeout);
 
 			_mainBus.Subscribe<SystemMessage.SystemInit>(requestManagement);
 			_mainBus.Subscribe<SystemMessage.StateChangeMessage>(requestManagement);
--- a/src/EventStore.Core/ClusterVNode.cs
+++ b/src/EventStore.Core/ClusterVNode.cs
@@ -338,9 +338,7 @@ namespace EventStore.Core {
 			var httpPipe = new HttpMessagePipe();
 			var httpSendService = new HttpSendService(httpPipe, forwardRequests: true);
 			_mainBus.Subscribe<SystemMessage.StateChangeMessage>(httpSendService);
-			SubscribeWorkers(bus => {
-				bus.Subscribe<HttpMessage.HttpSend>(httpSendService);
-			});
+			SubscribeWorkers(bus => bus.Subscribe<HttpMessage.HttpSend>(httpSendService));
 
 			var grpcSendService = new GrpcSendService(_eventStoreClusterClientCache);
 			_mainBus.Subscribe(new WideningHandler<GrpcMessage.SendOverGrpc, Message>(_workersHandler));
@@ -463,7 +461,7 @@ namespace EventStore.Core {
 			var histogramController = new HistogramController();
 			var statController = new StatController(monitoringQueue, _workersHandler);
 			var atomController = new AtomController(_mainQueue, _workersHandler,
-				vNodeSettings.DisableHTTPCaching);
+				vNodeSettings.DisableHTTPCaching, vNodeSettings.WriteTimeout);
 			var gossipController = new GossipController(_mainQueue, _workersHandler, vNodeSettings.GossipTimeout,
 				_httpMessageHandler);
 			var persistentSubscriptionController =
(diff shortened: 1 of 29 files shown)


src/EventStore.Core/Messages/StorageMessage.cs
```diff
@@ -37,16 +34,18 @@ public class WritePrepares : Message, IPreconditionedWriteMessage, IFlushableMes
 
 			public string EventStreamId { get; private set; }
 			public long ExpectedVersion { get; private set; }
+			public CancellationToken CancellationToken { get; }
 			public readonly Event[] Events;
 
 			public readonly DateTime LiveUntil;
```
pgermishuys: `LiveUntil` is no longer being used apart from `ToString`. We can choose not to affect the TCP side and only introduce the cancellation support on the gRPC side.

src/EventStore.Core/Services/Transport/Tcp/ClientTcpDispatcher.cs
```diff
@@ -8,10 +8,15 @@
 using EventStore.Core.Messages;
 using EventStore.Core.Messaging;
 using EventStore.Core.Util;
+using Microsoft.AspNetCore.SignalR;
 
 namespace EventStore.Core.Services.Transport.Tcp {
 	public class ClientTcpDispatcher : ClientWriteTcpDispatcher {
-		public ClientTcpDispatcher() {
+		public ClientTcpDispatcher(int commitTimeoutMs = Opts.CommitTimeoutMsDefault)
```
pgermishuys: We could choose not to have a default parameter here but to update the tests and revert the change in `Opts.cs`.

src/EventStore.Core/Services/RequestManager/Managers/WriteEvents.cs
```diff
@@ -1,26 +1,28 @@
 ﻿using System;
+using System.Threading;
 using EventStore.Core.Bus;
 using EventStore.Core.Data;
 using EventStore.Core.Messages;
 using EventStore.Core.Messaging;
+using EventStore.Core.Services.Storage.ReaderIndex;
```
pgermishuys: This can safely be removed.

src/EventStore.Core/Util/Opts.cs
```diff
@@ -170,7 +170,7 @@ public static class Opts {
 		public static readonly int PrepareTimeoutMsDefault = 2000; // 2 seconds
 
 		public const string CommitTimeoutMsDescr = "Commit timeout (in milliseconds).";
-		public static readonly int CommitTimeoutMsDefault = 2000; // 2 seconds
+		public const int CommitTimeoutMsDefault = 2000; // 2 seconds
```
pgermishuys: See comment about just setting this in the affected tests.

src/EventStore.Core/ClusterVNode.cs
```diff
@@ -647,7 +648,8 @@ public class ClusterVNode :
 					_internalAuthenticationProvider, AuthorizationGateway,
 					gossipInfo, !vNodeSettings.DisableInternalTls, true,
 					Certificate == null ? null : new X509Certificate2Collection(Certificate),
-					vNodeSettings.IntTcpHeartbeatTimeout, vNodeSettings.ExtTcpHeartbeatInterval);
+					vNodeSettings.IntTcpHeartbeatTimeout, vNodeSettings.ExtTcpHeartbeatInterval,
+					vNodeSettings.CommitTimeout);
```
pgermishuys:
```suggestion
					vNodeSettings.WriteTimeout);
```

src/EventStore.Core/Services/RequestManager/RequestManagementService.cs
```diff
@@ -52,20 +53,22 @@ public class RequestManagementService :
...
 			_commitTimeout = commitTimeout;
+			_writeTimeout = writeTimeout;
 			_commitSource = new CommitSource();
 		}
 		
 		public void Handle(ClientMessage.WriteEvents message) {
 			var manager = new WriteEvents(
 								_bus,
-								_commitTimeout,
+								_writeTimeout,
```
pgermishuys:
```suggestion
								_commitTimeout,
```
WriteTimeout should not need to be propagated further than where the cancellation token was created in each of the transports.

PR #2333, opened 2020-03-04
Remove TLS target host configuration

shaan1337:
This PR builds on top of #2267 to also remove the TLSTargetHost configuration option.
After this change, the only supported configuration for TLS will be with IP addresses in SANs.


shaan1337:
Moved to #2267

PR #2335, opened 2020-03-04
Restrict trust to specified root certificates and add mutual authentication to internal HTTP interface used for gossip/elections

### LINKED ISSUE #2321: [Preview] Mutual TLS authentication for gossip
Gossip is now done over gRPC on HTTPs, however only the remote party's certificate is validated but not the client.

We can also send client certificates over the gRPC channel to also have mutual authentication for gossip.

shaan1337:
Added: A mandatory configuration parameter named: `TrustedRootCertificatesPath`. The certificate store will be expanded with the root certificates in this path before certificate validation. For server certificate validation, trust is restricted to system certificates + the specified root certificates but for client certificate validation, trust is restricted only to the specified root certificates. Before this change, all root certificates installed on the system were trusted.
Added: Client certificate validation to the internal HTTP interface which is used for gossip/elections

## Detailed changes
The PR makes the following changes:
- Add a mandatory `TrustedRootCertificatesPath` configuration parameter. The certificate store will be expanded with those root certificates before validation. Thus, root certificates no longer need to be installed on the system.
- Centralize certificate validation in ClusterVNode using delegates and remove validateServer/validateClient parameters.
- For server certificate validation, trust is restricted to system certificates + the specified root certificates
- For client certificate validation, trust is restricted to the specified root certificates
- Add client certificate validation to the internal HTTP interface which is used for gossip/elections.
Fixes #2321
It can be tested with following curl command:
```
curl -k -v https://127.0.0.1:<IntHttpPort>/ --cert /path/to/certificate.crt --key /path/to/certificatekey.key
```
Without the client certificate file/key provided to curl, the TLS connection will be terminated but before the change it would get through.
- Remove installation of root certificates from CI since they no longer need to be installed on the system
- In development mode, set the TrustedRootCertificatesPath to Locations.DevCertificateDirectory and no longer disable internal/external TLS
- Clean up and fix TLS tests
- Regenerate untrusted test certificate with a valid IP/DNS SAN (127.0.0.1/localhost) to make sure the certificate validation fails only due to chain errors in tests


src/EventStore.ClusterNode/Program.cs
```diff
@@ -46,7 +47,10 @@ await CreateHostBuilder(hostedService, args)
 				.ConfigureWebHostDefaults(builder =>
 					builder.UseKestrel(server => {
 							server.Listen(hostedService.Options.IntIp, hostedService.Options.IntHttpPort,
-								listenOptions => listenOptions.UseHttps(hostedService.Node.Certificate));
+								listenOptions => listenOptions.UseHttps(new HttpsConnectionAdapterOptions {
```
jageall: Where is the client certificate validation done? How do we limit which CA's are valid?

shaan1337: We currently accept all CA certificates installed on the system (the same applies to the replication channel). I'll try to add a `CACertificatesPath` configuration option which will be mandatory if we apply the "secure by default" principle.

shaan1337:
The PR still needs some work, please do not merge yet.

shaan1337:
Ok, it's ready for review.

jen20:
My only concern here is whether we should _expand_ the pool if `TrustedRootCertificatesPath` is set, or `replace` the pool contents with it. I'm tempted to suggest replacing it here, since if you provide root certs that is fairly indicative that you have a specific one in mind.

shaan1337:
@jen20 thanks for the feedback.

When doing client certificate validation, we're currently strictly restricting the pool to the specified root certificates here:
https://github.com/EventStore/EventStore/pull/2335/files#diff-083712360d7b6f16653edd5b6e90d9e5R849

We can also restrict it for server certificates if required.

The .NET framework does not directly allow you to restrict the set of root certificates, so we have to expand the certificate store first, do the chain validation, then finally (if we want) check if the root certificate of the provided certificate is in the set of supplied root certificates.

PR #2353, opened 2020-03-10
Fix recursive lock causing deadlock in TCP connections

### LINKED ISSUE #2318: EventStore 5.0.6 stops accepting connections after some time [Potential bug]
The issue was originally reported here by Zvjezdan Tomičević:
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/event-store/a1e0qAAg3vc/uawiRYYnGgAJ

Quoting from the google groups link:
```
Hi. We have updated our staging server from 5.0.5 to 5.0.6, and have started to experience strange behavior.
After a while, server rejects all new connections from .NET client. All write or read attempts end up in ConnectionClosed exception.
Restarting the apps does not help.
Normal operation resumes only after we restart server instance.

We are running on bare metal, WIN 2016 server, 250GBSSD, 64GB RAM.
Any pointers as to what can saturate and block server in this way? Our code has not changed, and it worked until last update.
BR
```

**To Reproduce**
No reproduction steps yet.

**Expected behavior**
Server should be able to accept new connections from clients.

**Actual behavior**
After some time, it appears that the server no longer accepts  connections from clients. Restarting the clients does not make any difference, so the issue must be server-side.

**EventStore details**
 - EventStore server version: 5.0.6

 - Operating system: Windows

**Additional context**
This could be a regression due to the multiple TCP connection/socket fixes applied in 5.0.6.
https://eventstore.com/blog/event-store-5.0.6-release/

shaan1337:
Fixes #2318

This fixes a regression introduced by commit 68cd7f0b979e95739d11e2138280a88e4dbd9467.

It is possible for `connection.Close()` to be called from the [callback](https://github.com/EventStore/EventStore/commit/68cd7f0b979e95739d11e2138280a88e4dbd9467#diff-a976f5fbfba0888ba9072863f48146aaR259) for example, when there is a TCP framing error (can be tested with `curl -v http://localhost:1113/` where 1113 is the external TCP port)
 
If this happens, it results into a recursive lock which causes a deadlock since the thread doing the callback keeps spinning [here](https://github.com/EventStore/EventStore/commit/68cd7f0b979e95739d11e2138280a88e4dbd9467#diff-a976f5fbfba0888ba9072863f48146aaR286), waiting for the callback to complete.

This fix reverts the commit 68cd7f0b979e95739d11e2138280a88e4dbd9467 and uses a simple lock statement instead which already supports recursive locking.

It would also have been possible to fix the issue by keeping the previous code and saving the callback thread ID and verifying if the closed is being done from the same thread but in this context there is no significant performance benefit compared to using a simple lock since contention will occur only when closing the TCP connection while data is being dispatched.

### DIFF
--- a/src/EventStore.Transport.Tcp/TcpConnection.cs
+++ b/src/EventStore.Transport.Tcp/TcpConnection.cs
@@ -86,8 +86,9 @@ namespace EventStore.Transport.Tcp {
 
 		private readonly object _receivingLock = new object();
 		private readonly object _sendLock = new object();
+		private readonly object _closeLock = new object();
 		private bool _isSending;
-		private volatile int _closed;
+		private volatile bool _isClosed;
 
 		private Action<ITcpConnection, IEnumerable<ArraySegment<byte>>> _receiveCallback;
 
@@ -185,7 +186,7 @@ namespace EventStore.Transport.Tcp {
 			} else {
 				NotifySendCompleted(socketArgs.Count);
 
-				if (_closed != 0)
+				if (_isClosed)
 					ReturnSendingSocketArgs();
 				else {
 					lock (_sendLock) {
@@ -296,7 +297,10 @@ namespace EventStore.Transport.Tcp {
 				data[i] = new ArraySegment<byte>(d.Buf.Array, d.Buf.Offset, d.DataLen);
 			}
 
-			callback(this, data);
+			lock (_closeLock) {
+				if(!_isClosed)
+					callback(this, data);
+			}
 
 			for (int i = 0, n = res.Count; i < n; ++i) {
 				BufferManager.CheckIn(res[i].Buf); // dispose buffers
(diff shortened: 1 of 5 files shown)


src/EventStore.Transport.Tcp/TcpConnection.cs
```diff
@@ -186,7 +186,7 @@ public class TcpConnection : TcpConnectionBase, ITcpConnection {
 			} else {
 				NotifySendCompleted(socketArgs.Count);
 
-				if (_closed != 0)
+				if (_isClosed)
```
hayley-jean: Do we not need to take a lock here?

shaan1337: @hayley-jean I think that stale data is acceptable here based on what the previous code was doing. If `isClosed` is true but we miss it here we will get the value of `true` it in the next `TrySend()->ProcessSend()` call. Or alternatively, we will call `ReturnSendingSocketArgs()` in the `catch (ObjectDisposedException)` of `TrySend()`

PR #2392, opened 2020-03-26
Add new route to view parked messages in PersistentSubscriptionController

Lougarou:
Added: New route in PersistentSubscriptionController to view parked messages /subscriptions/viewparkedmessages/{stream}/{group}

Fixes https://github.com/EventStore/EventStore/issues/2305
To be used from UI https://github.com/EventStore/EventStore.UI/pull/242

### DIFF
--- a/src/EventStore.Core/Services/Transport/Http/Controllers/PersistentSubscriptionController.cs
+++ b/src/EventStore.Core/Services/Transport/Http/Controllers/PersistentSubscriptionController.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using EventStore.Core.Bus;
 using EventStore.Core.Messages;
@@ -11,6 +12,9 @@ using ClientMessages = EventStore.Core.Messages.ClientMessage.PersistentSubscrip
 using EventStore.Core.Services.PersistentSubscription;
 using EventStore.Common.Utils;
 using EventStore.Plugins.Authorization;
+using EventStore.Core.Settings;
+using EventStore.Transport.Http.Atom;
+using Microsoft.Extensions.Primitives;
 using ILogger = Serilog.ILogger;
 
 namespace EventStore.Core.Services.Transport.Http.Controllers {
@@ -19,7 +23,7 @@ namespace EventStore.Core.Services.Transport.Http.Controllers {
 		private readonly IPublisher _networkSendQueue;
 		private const int DefaultNumberOfMessagesToGet = 1;
 		private static readonly ICodec[] DefaultCodecs = {Codec.Json, Codec.Xml};
-
+		public static readonly char[] ETagSeparatorArray = { ';' };
 		private static readonly ICodec[] AtomCodecs = {
 			Codec.CompetingXml,
 			Codec.CompetingJson,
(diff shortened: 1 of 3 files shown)


src/EventStore.Core/Services/Transport/Http/Controllers/PersistentSubscriptionController.cs
```diff
@@ -63,6 +65,116 @@ public class PersistentSubscriptionController : CommunicationController {
...
+				return true;
+		
+			if (string.Equals(onlyLeader, "True", StringComparison.OrdinalIgnoreCase) ||
+			    string.Equals(onlyMaster, "True", StringComparison.OrdinalIgnoreCase)) {
+				requireLeader = true;
+				return true;
+			}
+
+			return string.Equals(onlyLeader, "False", StringComparison.OrdinalIgnoreCase) ||
+			       string.Equals(onlyLeader, "False", StringComparison.OrdinalIgnoreCase);
```
pgermishuys, line 144: The following should likely be...

```suggestion
			       string.Equals(onlyMaster, "False", StringComparison.OrdinalIgnoreCase);
```

We can check the AtomController as well and either fix it in this PR or raise an issue to ensure that we don't have more of this logic lying around.

src/EventStore.Core/Services/Transport/Http/Controllers/PersistentSubscriptionController.cs
```diff
@@ -63,6 +65,116 @@ public class PersistentSubscriptionController : CommunicationController {
...
+			Register(service, "/subscriptions/viewparkedmessages/{stream}/{group}?embed={embed}", HttpMethod.Get,
+				ViewParkedMessages, Codec.NoCodecs, AtomCodecs, WithParameters(Operations.Subscriptions.Statistics));
+		}
+		private void ViewParkedMessages(HttpEntityManager http, UriTemplateMatch match) {
+			var stream = match.BoundVariables["stream"];
+			var groupname = match.BoundVariables["group"];
+			var evNum = match.BoundVariables["event"];
+			var cnt = match.BoundVariables["count"];
+
+			stream = "$persistentsubscription-" + stream + "::" + groupname+"-parked";
```
pgermishuys, line 80: We should have other places that build up this key and we should be able to centralize this logic. Just have a look if it's feasible to do this cleanup in this PR.

Lougarou: Thanks, added a BuildSubscriptionParkedStreamName which uses previous build function BuildSubscriptionGroupKey

src/EventStore.Core/Services/Transport/Http/Controllers/PersistentSubscriptionController.cs
Lougarou: Thanks! Looks like they were only in Admin and Atom controllers. Made the fixes.

src/EventStore.Core/Services/Transport/Http/Controllers/PersistentSubscriptionController.cs
```diff
@@ -63,6 +66,175 @@ public class PersistentSubscriptionController : CommunicationController {
 				WithParameters(Operations.Subscriptions.ProcessMessages), AckMessages);
 			RegisterUrlBased(service, "/subscriptions/{stream}/{subscription}/nack?ids={messageids}&action={action}",
 				HttpMethod.Post, WithParameters(Operations.Subscriptions.ProcessMessages), NackMessages);
+			//map view parked messages
+			Register(service, "/subscriptions/viewparkedmessages/{stream}/{group}/{event}/backward/{count}?embed={embed}", HttpMethod.Get,
+				ViewParkedMessagesBackward, Codec.NoCodecs, AtomCodecs, WithParameters(Operations.Subscriptions.Statistics));
+			Register(service, "/subscriptions/viewparkedmessages/{stream}/{group}?embed={embed}", HttpMethod.Get,
+				ViewParkedMessagesBackward, Codec.NoCodecs, AtomCodecs, WithParameters(Operations.Subscriptions.Statistics));
```
pgermishuys, line 73: We should be able to safely remove this endpoint now
```suggestion
```

pgermishuys:
@Lougarou would you mind resolving the merge conflicts and then this should be good to get merged.

hayley-jean:
It also might be a good idea to update the UI submodule as part of this PR, as the UI changes have been merged already.

PR #2395, opened 2020-03-26
Skip pull request linting when not targetting src directory.

YoEight:


.github/workflows/changelog.yml
```diff
@@ -10,8 +10,15 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+      - name: Check pull requests
+        uses: EventStore/Automations/pr-check@master
+        id: pr_check
+        with:
+          github-token: ${{ secrets.GH_PAT }}
```
thefringeninja: what if you used  `${{ secrets.GITHUB_TOKEN }}` ?

YoEight: For `pr-check` we could have. For `changelog-update` github action we need rights that what `GITHUB_TOKEN` has. At the end, I used `GH_PAT` solely to use the same token across the board.

PR #2407, opened 2020-04-01
Tag Bionic Container as ci for Integration Tests

thefringeninja:
After some discussion about EventStore/EventStore-Client-Dotnet#1, we agreed that we did not want to continue changing the docker image tag every time we resolve an issue in the main repository. The best course of action is to update a specific tag (`ci`) and run integration tests against that.

Partiality fixes EventStore/home#115


thefringeninja:
Part of EventStore/home#115

PR #2415, opened 2020-04-06
Add MaxTruncation safety feature

megakid:
This feature was built to avoid huge unexpected truncations due to misconfiguration.

Today I did some rejigging of our EventStore deployment mechanism, after testing on a development and UAT environment, I pushed my changes live.  We don't take deploying EventStore lightly but this change seemed relatively harmless but little did I know, a 1 character mistake ended up truncating our entire data set on 4 of our 5 nodes (approx 800GB of data in total).  Essentially I had typed an IP address wrong and the deployment process automatically creates a DNS entry to our gossip DNS entry.  That one (bad) mistake ended up causing a potentially catastrophic loss of data (outside of backups obviously).

Unfortunately upon deployment with this misconfiguration, there was another cluster (with a completely different dataset) running on this erroneous IP address.  On top of this, it also shared the same port setup (something we try to avoid but didn't achieve that here).  This meant that these two clusters quickly merged due to the gossip networks joining one another.  The cluster I was deploying then quickly logged the below and truncated their *entire* dataset:

```
Master [10.194.0.10:5412,{ebd4c7e7-efe1-4419-a6a0-624865099018}] subscribed us at 0 (0x0), which is less than our last epoch and LastCommitPosition 225919458447 (0x3499D95C8F) >= lastEpoch.EpochPosition 225919442304 (0x3499D91D80). That might be bad, especially if the LastCommitPosition is way beyond EpochPosition.
......
Truncate checkpoint is present. Truncate: 0 (0x0), Writer: 225919474285 (0x3499D99A6D), Chaser: 225919474285 (0x3499D99A6D), Epoch: 225919442304 (0x3499D91D80)
```

To be clear, this is absolutely not a bug but I feel it's a bit 'violent' for a database to completely wipe itself following such a "small" misconfiguration.

In this PR, I've added a `MaxTruncation` value which can be configured to a small value (say `20,000` to allow for "usual" truncations during failovers/re-elections etc) but doesn't allow huge truncations to occur without someone changing the config.  If you manually set the truncate.chk then you should also ensure that your configuration `MaxTruncation` value is set to `-1` or removed (defaults to `-1`).  It would be a huge weight off our minds to know that instead of wiping all data, it would go into an infinite restart loop (with clear logs as to why) if something like this occurs in the future.

We did manage to fully recover from this incident fairly easily by copying the intact data from our last remaining node across the other nodes, restarting the 2nd cluster it accidently joined (so it reset the cluster definition) and starting up the cluster.

### DIFF
--- a/src/EventStore.Core/Util/Opts.cs
+++ b/src/EventStore.Core/Util/Opts.cs
@@ -431,5 +431,10 @@ namespace EventStore.Core.Util {
 
 		public const string WriteStatsToDbDescr = "Set this option to write statistics to the database.";
 		public const bool WriteStatsToDbDefault = true;
+
+		public static readonly long MaxTruncationDefault = -1;
+		public const string MaxTruncationDescr =
+			"When truncate.chk is set, the database will be truncated on startup.  This is a safety check to ensure large amounts of data truncation does not happen accidentally.  This value should be set in the low 10,000s for allow for standard cluster recovery operations. -1 is no max (default).";
+
 	}
 }
--- a/src/EventStore.ClusterNode/ClusterNodeOptions.cs
+++ b/src/EventStore.ClusterNode/ClusterNodeOptions.cs
@@ -316,6 +316,9 @@ namespace EventStore.ClusterNode {
 		[ArgDescription(Opts.WriteStatsToDbDescr, Opts.DbGroup)]
 		public bool WriteStatsToDb { get; set; }
 
+		[ArgDescription(Opts.MaxTruncationDescr, Opts.DbGroup)]
+		public long MaxTruncation { get; set; }
+
 		public ClusterNodeOptions() {
 			Config = "";
 			Help = Opts.ShowHelpDefault;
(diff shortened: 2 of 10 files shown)


src/EventStore.Core/Util/Opts.cs
```diff
@@ -431,5 +431,10 @@ public static class Opts {
 
 		public const string WriteStatsToDbDescr = "Set this option to write statistics to the database.";
 		public const bool WriteStatsToDbDefault = true;
+
+		public static readonly long MaxTruncationDefault = -1;
```
hayley-jean, line 435:
```suggestion
		public static readonly long MaxTruncationDefault = 256 * 1024 * 1024;
```
A single chunk is probably a sensible default here, and could prevent others from only learning about this setting after they've run into this problem and had their data truncated

pgermishuys:
Closed in favor of https://github.com/EventStore/EventStore/pull/2436. @megakid we wanted to get this PR through as soon as possible. Thank you for the contribution!

PR #2422, opened 2020-04-08
Always honor NodePreference.Leader in ClusterDnsEndPointDiscoverer

### LINKED ISSUE #2416: NodePreference.Leader not always honored in ClusterDnsEndPointDiscoverer
This has not been tested but it seems that `NodePreference.Leader` will not be honored here since a case is missing in the switch statement:
https://github.com/EventStore/EventStore/blob/0a53f8cfcca84a5981a836f833c91509037876ec/src/EventStore.ClientAPI/Internal/ClusterDnsEndPointDiscoverer.cs#L221

EDIT: After discussing with @pgermishuys , it appears that we never took `NodePreference.Leader` into account even in v5 or before although it's set as the default here which is a bit confusing:
https://github.com/EventStore/EventStore/blob/0a53f8cfcca84a5981a836f833c91509037876ec/src/EventStore.ClientAPI/GossipSeedClusterSettingsBuilder.cs#L14

Thus this affects all versions.

IMHO, clients should be able to force a connection to the leader node (for example if doing only writes) even if it's done by default or with a `PreferLeaderNode()` API call.

EDIT 2:
After discussion with @jageall it seems that the original intention was to depend on the order of this enum:
https://github.com/EventStore/EventStore/blob/0a53f8cfcca84a5981a836f833c91509037876ec/src/EventStore.ClientAPI/Messages/ClusterMessages.cs#L68

Since we're sorting the State in descending order here:
https://github.com/EventStore/EventStore/blob/0a53f8cfcca84a5981a836f833c91509037876ec/src/EventStore.ClientAPI/Internal/ClusterDnsEndPointDiscoverer.cs#L215
```
			var nodes = members.Where(x => x.IsAlive)
				.Where(x => !notAllowedStates.Contains(x.State))
				.OrderByDescending(x => x.State)
				.ToArray();
```

However, we have recently added new states in this enum: `ReadOnlyLeaderless` , `PreReadOnlyReplica` and `ReadOnlyReplica` which means that those states will be given preference over `Leader` which looks like a potential bug.

shaan1337:
Fixed: NodePreference.Leader is not always honored in ClusterDnsEndPointDiscoverer

Fixes #2416
*Reproduction steps*
- Connect to a cluster having 2 nodes and a read-only replica with default settings and verbose logging on using the TCP client.
- The client will connect to the ReadOnlyReplica instead of the Leader:
```
Discovering: found best choice [127.0.0.1:0,127.0.0.1:3103] (ReadOnlyReplica)
```

- Explicitly sort nodes giving preference to VNodeState.Leader when NodePreference.Leader is specified to make the code clearer and avoid future bugs.
- Reorder the VNodeState enum and add a comment mentioning that the order in the VNodeState enum is important

### DIFF
--- a/src/EventStore.ClientAPI/Messages/ClusterMessages.cs
+++ b/src/EventStore.ClientAPI/Messages/ClusterMessages.cs
@@ -65,24 +65,26 @@ namespace EventStore.ClientAPI.Messages {
 			}
 		}
 
+		//The order specified in this enum is important.
+		//It defines the default sorting order (descending, ignoring a few specific states) used in ClusterDnsEndPointDiscoverer when choosing which node the client should connect to.
 		public enum VNodeState {
-			Initializing,
-			Unknown,
-			PreReplica,
-			CatchingUp,
-			Clone,
-			Slave,
-			Follower,
-			PreMaster,
-			PreLeader,
-			Master,
-			Leader,
-			Manager,
-			ShuttingDown,
-			Shutdown,
-			ReadOnlyLeaderless,
-			PreReadOnlyReplica,
-			ReadOnlyReplica
+			Initializing = 1,
+			ReadOnlyLeaderless = 2,
+			Unknown = 3,
+			PreReadOnlyReplica = 4,
+			PreReplica = 5,
+			CatchingUp = 6,
+			Clone = 7,
+			ReadOnlyReplica = 8,
+			Slave = 9,
+			Follower = 10,
+			PreMaster = 11,
+			PreLeader = 12,
+			Master = 13,
+			Leader = 14,
+			Manager = 15,
+			ShuttingDown = 16,
+			Shutdown = 17
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


src/EventStore.ClientAPI/Messages/ClusterMessages.cs
```diff
@@ -65,24 +65,26 @@ public class MemberInfoDto {
...
-			PreLeader,
-			Master,
-			Leader,
-			Manager,
-			ShuttingDown,
-			Shutdown,
-			ReadOnlyLeaderless,
-			PreReadOnlyReplica,
-			ReadOnlyReplica
+			Initializing = 1,
```
jageall, line 71: It would be good to put Unknown back in at 0 as that is the default value for the enum

shaan1337: If we set `Unknown` to 0, it will also imply that clients will favor nodes in `Initializing` and `ReadOnlyLeaderLess` states over nodes in the `Unknown` state (both `Unknown` and `ReadOnlyLeaderLess` accept reads). Is it ok or should we set `Initializing` to 0 instead?

PR #2436, opened 2020-04-16
Add MaxTruncation safety feature

pgermishuys:
Added: Max Truncation Safety Feature to avoid large unexpected truncations due to misconfiguration

Original PR by @megakid here https://github.com/EventStore/EventStore/pull/2415

This feature was built to avoid huge unexpected truncations due to misconfiguration.

Today I did some rejigging of our EventStore deployment mechanism, after testing on a development and UAT environment, I pushed my changes live.  We don't take deploying EventStore lightly but this change seemed relatively harmless but little did I know, a 1 character mistake ended up truncating our entire data set on 4 of our 5 nodes (approx 800GB of data in total).  Essentially I had typed an IP address wrong and the deployment process automatically creates a DNS entry to our gossip DNS entry.  That one (bad) mistake ended up causing a potentially catastrophic loss of data (outside of backups obviously).

Unfortunately upon deployment with this misconfiguration, there was another cluster (with a completely different dataset) running on this erroneous IP address.  On top of this, it also shared the same port setup (something we try to avoid but didn't achieve that here).  This meant that these two clusters quickly merged due to the gossip networks joining one another.  The cluster I was deploying then quickly logged the below and truncated their *entire* dataset:

```
Master [10.194.0.10:5412,{ebd4c7e7-efe1-4419-a6a0-624865099018}] subscribed us at 0 (0x0), which is less than our last epoch and LastCommitPosition 225919458447 (0x3499D95C8F) >= lastEpoch.EpochPosition 225919442304 (0x3499D91D80). That might be bad, especially if the LastCommitPosition is way beyond EpochPosition.
......
Truncate checkpoint is present. Truncate: 0 (0x0), Writer: 225919474285 (0x3499D99A6D), Chaser: 225919474285 (0x3499D99A6D), Epoch: 225919442304 (0x3499D91D80)
```

To be clear, this is absolutely not a bug but I feel it's a bit 'violent' for a database to completely wipe itself following such a "small" misconfiguration.

In this PR, I've added a `MaxTruncation` value which can be configured to a small value (say `20,000` to allow for "usual" truncations during failovers/re-elections etc) but doesn't allow huge truncations to occur without someone changing the config.  If you manually set the truncate.chk then you should also ensure that your configuration `MaxTruncation` value is set to `-1` or removed (defaults to `-1`).  It would be a huge weight off our minds to know that instead of wiping all data, it would go into an infinite restart loop (with clear logs as to why) if something like this occurs in the future.

We did manage to fully recover from this incident fairly easily by copying the intact data from our last remaining node across the other nodes, restarting the 2nd cluster it accidently joined (so it reset the cluster definition) and starting up the cluster.

### DIFF
--- a/src/EventStore.Core/Cluster/Settings/ClusterVNodeSettings.cs
+++ b/src/EventStore.Core/Cluster/Settings/ClusterVNodeSettings.cs
@@ -85,6 +85,7 @@ namespace EventStore.Core.Cluster.Settings {
 		public readonly bool ReduceFileCachePressure;
 		public readonly int InitializationThreads;
 		public readonly int MaxAutoMergeIndexLevel;
+		public readonly long MaxTruncation;
 
 		public readonly bool GossipOnSingleNode;
 		public readonly bool FaultOutOfOrderProjections;
@@ -160,6 +161,7 @@ namespace EventStore.Core.Cluster.Settings {
 			int maxAutoMergeIndexLevel = 1000,
 			bool disableFirstLevelHttpAuthorization = false,
 			bool logFailedAuthenticationAttempts = false,
+			long maxTruncation = -1,
 			bool readOnlyReplica = false,
 			int maxAppendSize = 1024 * 1024,
 			bool unsafeAllowSurplusNodes = false,
@@ -272,6 +274,7 @@ namespace EventStore.Core.Cluster.Settings {
 			MaxAppendSize = maxAppendSize;
 			PTableMaxReaderCount = ptableMaxReaderCount;
 			UnsafeAllowSurplusNodes = unsafeAllowSurplusNodes;
+			MaxTruncation = maxTruncation;
 		}
 
 		public override string ToString() =>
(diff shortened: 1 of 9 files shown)


src/EventStore.Core/Util/Opts.cs
```diff
@@ -434,6 +434,10 @@ public static class Opts {
 		public const string WriteStatsToDbDescr = "Set this option to write statistics to the database.";
 		public const bool WriteStatsToDbDefault = false;
 		
+		public const string MaxTruncationDescr =
+			"When truncate.chk is set, the database will be truncated on startup. This is a safety check to ensure large amounts of data truncation does not happen accidentally. This value should be set in the low 10,000s for allow for standard cluster recovery operations. -1 is no max (default).";
```
megakid, line 438: Note to self: update this Descr with new default value.

src/EventStore.Core/Cluster/Settings/ClusterVNodeSettings.cs
```diff
@@ -160,6 +161,7 @@ public class ClusterVNodeSettings {
 			int maxAutoMergeIndexLevel = 1000,
 			bool disableFirstLevelHttpAuthorization = false,
 			bool logFailedAuthenticationAttempts = false,
+			long maxTruncation = -1,
```
megakid, line 164: Should this be the default value too? (256 * 1024 * 1024)

src/EventStore.Core/TransactionLog/Chunks/TFChunkDbConfig.cs
```diff
@@ -38,7 +39,8 @@ public class TFChunkDbConfig {
 			bool unbuffered = false,
 			bool writethrough = false,
 			bool optimizeReadSideCache = false,
-			bool reduceFileCachePressure = false) {
+			bool reduceFileCachePressure = false,
+			long maxTruncation = -1) {
```
megakid, line 43: Should this be the new default too? (256 * 1024 * 1024)

Issue #2114, opened 2019-12-05
Review Configuration Options


pgermishuys:
At the moment, we have a substantial list of options. We should review our current configuration options and normalize.

```
Application Options
   --help (-Help)                              Show help.
   --version (-Version)                        Show version.
   --log (-Log)                                Path where to keep log files.
   --config (-Config)                          Configuration files.
   --defines (-Defines)                        Run-time conditionals.
   --what-if (-WhatIf)                         Print effective configuration to console and then exit.
   --start-standard-projections                Start the built in system projections.
   --disable-http-caching                      Disable HTTP caching.
   --force (-Force)                            Force the Event Store to run in possibly harmful environments such as with Boehm GC.
   --stats-period-sec                          The number of seconds between statistics gathers.
   --worker-threads                            The number of threads to use for pool of worker services.
   --enable-histograms                         Enables the tracking of various histograms in the backend, typically only used for debugging etc
   --log-http-requests                         Log Http Requests and Responses before processing them.
   --log-failed-authentication-attempts        Log the failed authentication attempts.
   --skip-index-scan-on-reads                  Skip Index Scan on Reads. This skips the index scan which was used to stop reading duplicates.
   --max-append-size                           The maximum size of appends, in bytes. May not exceed 16MB.
   --dev (-Dev)                                Enable Development Mode for Event Store.

   Authentication/Authorization Options
   --authentication-type                       The type of authentication to use.
   --authentication-config                     Path to the configuration file for authentication configuration (if applicable).
   --disable-first-level-http-authorization    Disables first level authorization checks on all HTTP endpoints. This option can be enabled for backwards compatibility with EventStore 5.0.1 or earlier.

   Certificate Options (from file)
   --certificate-file                          The path to a PKCS #12 (.p12/.pfx) or an X.509 (.pem, .crt, .cer, .der) certificate file.
   --certificate-private-key-file              The path to the certificate private key file (.key) if an X.509 (.pem, .crt, .cer, .der) certificate file is provided.
   --certificate-password                      The password to the certificate if a PKCS #12 (.p12/.pfx) certificate file is provided.

   Certificate Options (from store)
   --certificate-store-location                The certificate store location name.
   --certificate-store-name                    The certificate store name.
   --certificate-subject-name                  The certificate subject name.
   --certificate-thumbprint                    The certificate fingerprint/thumbprint.

   Cluster Options
   --cluster-size                              The number of nodes in the cluster.
   --node-priority                             The node priority used during leader election
[...]

PR #2442, opened 2020-04-21
Remove Force option

pgermishuys:
Removed: Force option

Related to reviewing options in #2114

This option was used initially used to be able to force Event Store to run on unsupported versions of the Mono runtime or an unsupported GC such as boehm. This is no longer the case since Event Store has moved to .NET Core. The only other place this option is used is to force the use of a directory that starts with a `~` which we don't expand.

Do we want to keep this option around for future use or remove it?

- We need to update the command line documentation to remove this [here](https://eventstore.com/docs/server/command-line-arguments/index.html) if this PR does get accepted. (cc: @ChrisChinchilla)

### DIFF
--- a/src/EventStore.ClusterNode/ClusterVNodeHostedService.cs
+++ b/src/EventStore.ClusterNode/ClusterVNodeHostedService.cs
@@ -80,14 +80,14 @@ namespace EventStore.ClusterNode {
 		protected override void PreInit(ClusterNodeOptions options) {
 			base.PreInit(options);
 
-			if (options.Db.StartsWith("~") && !options.Force) {
+			if (options.Db.StartsWith("~")) {
 				throw new ApplicationInitializationException(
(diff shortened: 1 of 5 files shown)


src/EventStore.ClusterNode/ClusterVNodeHostedService.cs
```diff
@@ -77,12 +77,12 @@ internal class ClusterVNodeHostedService : EventStoreHostedService<ClusterNodeOp
 		protected override void PreInit(ClusterNodeOptions options) {
 			base.PreInit(options);
 
-			if (options.Db.StartsWith("~") && !options.Force) {
+			if (options.Db.StartsWith("~")) {
```
jageall, line 80: the error message is now wrong, but i would rather just ditch support for this entirely.  What is the reason for it being supported in the first place?

PR #2449, opened 2020-04-27
Support JWT for TCP

thefringeninja:
Added: TCP jwt authorization on the wire

Fixes EventStore/home#131

### DIFF
--- a/src/EventStore.Core/Services/Transport/Tcp/TcpPackage.cs
+++ b/src/EventStore.Core/Services/Transport/Tcp/TcpPackage.cs
@@ -1,5 +1,5 @@
 using System;
-using System.Text;
+using System.Collections.Generic;
 using EventStore.Common.Utils;
 
 namespace EventStore.Core.Services.Transport.Tcp {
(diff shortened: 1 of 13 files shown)


src/EventStore.Core/Services/Transport/Tcp/TcpPackage.cs
```diff
@@ -40,14 +53,25 @@ public struct TcpPackage {
 			string login = null;
 			string pass = null;
 			if ((flags & TcpFlags.Authenticated) != 0) {
-				var loginLen = data.Array[data.Offset + AuthOffset];
-				if (AuthOffset + 1 + loginLen + 1 >= data.Count)
-					throw new Exception("Login length is too big, it does not fit into TcpPackage.");
+				if (!BitConverter.IsLittleEndian) {
```
hayley-jean, line 56:
```suggestion
				if (BitConverter.IsLittleEndian) {
```

thefringeninja: I think we don't need to check this at all actually

Issue #2213, opened 2020-01-09
Multi-Master scenario due to node using outdated information about last elected master


pgermishuys:
There is a possibility whereby a multi-master scenario will arise due to nodes using out of date information (from gossip) during elections to propose or accept a node as a master candidate.

The gossip carries with it the Epoch Number which is used to determine whether a master candidate is a legitimate master candidate. In the following example, a master proposal with a higher epoch number could result in a node accepting a new master candidate even if its current master is still alive.

The above means that if a node gets a master proposal before it received a gossip from it's master after an election, the master proposal is accepted.

### Example:
1. 5 node cluster (**Node 1** through to **Node 5**)
2. **Node 1** is elected master
3. An election is started because **Node 2** is restarted
4. **Node 2** is the leader of the elections
5. **Node 2** proposes **Node 3** as the master as according to it, it's the best master candidate (it might have not received a gossip from the current master node)
6. **Node 2** _accepts_ it's proposal
7. **Node 1** _rejects_ the proposal as it's the **master** and it has the most up to date information about itself (Epoch Number included) and it's still alive
8. **Node 3** _accepts_ the proposal because even though it has a previously elected master, it hasn't received a gossip from the master (**Node1**) with the updated **Epoch Number** whereas the proposal has a higher Epoch Number than the current master (**Node1**) at this point
9. **Node 4** _rejects_ the proposal as it received a gossip from the current master (**Node1**)
10. **Node 5** _accepts_ the proposal because it also hasn't received a gossip from the current master (**Node1**)

The above results in multiple masters existing in the same cluster without a network partition having occurred.

Once this happens, a series of elections will take place.

PR #2454, opened 2020-04-30
Write leader instance ID in epoch record and pass on epoch's leader's ID and gossip info to leader of elections

shaan1337:
Changed: Write leader's instance ID in epoch record. Pass on the epoch record's leader's instance id and each node's gossip information during elections to the leader of elections to determine more accurately if the previous leader is still alive when choosing the best leader candidate.

Fixes: #2213

### Current algorithm
The current election algorithm works like this when choosing the best candidate:
- Each node keeps in memory the last elected leader for the elections they've participated in
- When a node becomes the leader of the elections, it chooses the best leader candidate as follows:
  - If the previously elected leader it has stored in memory appears to be alive (according to the node's gossip info), immediately propose it as the leader
  - Otherwise, choose the best candidate (by selecting the node that has the latest data) among the quorum number of nodes that have sent a PrepareOk message.

There are a few problems with this approach:
### Problem 1
It is not necessary that the in-memory previously elected leader is accurate. For instance, consider this scenario in a 3 node cluster:
 - Nodes 1,2,3 elect node 1 as the leader
 - A network partition occurs which isolates node 3. Node 3's in memory previously elected leader is node 1.
 - Node 2 becomes the leader for some reason. (e.g node 1 has resigned, node 2 is restarted, etc.)
 - The partition is re-established and node 3 triggers elections.
 - Node 3 becomes the leader of the elections. It still thinks that node 1 is the leader and will propose it immediately although node 2 is the new leader.

### Problem 2
The leader of the elections also uses only it's own gossip information to determine if the previous leader is alive or not. It may be possible that the gossip information is outdated and the node thinks that the leader is dead but it is not.

### Problem 3
Restarting a node loses the in memory previously elected leader. If this node becomes leader of the elections, it may not propose the previous leader.

### New algorithm
The new algorithm works as follows:
 - Instead of keeping the last elected leader's ID in memory, we store the leader's ID in the epoch record that the leader writes as soon as it is elected. The epoch record is part of the transaction log and is replicated to each node. Each node also passes this information to the leader of the Elections in the `PrepareOk` message.
 - Instead of only using the leader of the elections's gossip information, each node sends its latest gossip information in the `PrepareOk` message.
 - The leader of the elections chooses the best candidate leader as follows:
   - Start by finding the candidate that has the latest data among the quorum number of nodes that are participating in the elections. Let's call this candidate `BestCandidate`
   - Obtain the leader's instance ID (`PreviousLeader`) from the epoch record in the `PrepareOk` message sent by `BestCandidate`.
   - Next try to determine if `PreviousLeader` is alive by either seeing if we have directly received a `PrepareOk` message from it or if at least one node among the quorum thinks it is alive (using the gossip info sent in the `PrepareOk`).
   - If it appears to be alive and is not resigning, then we propose `PreviousLeader` as best candidate
   - Otherwise we propose `BestCandidate` as the best candidate.

### Guarantees
- Using the above algorithm, we guarantee that we are electing a leader that has the latest data while with the previous algorithm the in-memory value may be stale. Since data needs to be replicated to a quorum number of nodes to be considered committed, we know that if the epoch has been replicated, the chosen `BestCandidate` will have it independently of which subset quorum of nodes participate in the elections.
See this comment for more details: https://github.com/EventStore/EventStore/pull/2454/files#diff-65f89e62d3f64576ec45b0860d37ffc8R437
- There are two possibilities when the previous leader is alive but we do not elect it:
  - The previous leader writes an epoch record but it does not have enough time to be replicated to a quorum number of nodes and new elections occur. This is an uncommitted write anyway. In this case, the previous leader will need to truncate when it joins the new leader. We still guarantee that the `BestCandidate` node has the latest replicated data.
  - All the quorum number of nodes participating in the elections think that the previous leader is dead while it's still actually alive. In this scenario, we will not be able to obtain a quorum number of Accepts anyway and the elections would fail, so by proposing the `BestCandidate` instead, we're still better than nothing.
  - On the other hand if at least one node says that the previous leader is alive, we may still have a quorum number of Accepts if we choose this node and the other floor(N/2) nodes that have not sent their PrepareOks for the elections. So we still give the previous leader a chance to be elected.

### Stress tests
Stress tests have been carried out with https://github.com/EventStore/XstreamTester with random cluster sizes between 3 and 7 nodes. On this PR, out of around 5640 elections, 7 resulted in truncation which is around 0.1%. The reason for truncation is due to one of the two reasons stated above. There have also been no truncations of committed records.

Tests have also been carried out on V6 `master` (ca67b9a096b0717244ee7e2b0c773d3e0ebf42a3) and in this case, the number of truncations is higher: 14/4996 elections ~= (0.3%). This has probably improved a bit compared to V5 due to #2386. There was also a case where epoch with committed records was truncated.

Tests on 5.0.8 also show a higher number of truncations and a higher number of truncation of epoch with committed records.

| Branch | Number of unique elections tested | No. of elections resulting in Truncations | No. of elections resulting in Truncations of epoch with committed records |
| --- | --- | --- | --- |
| This PR | 5640 | 7 (0.1%) | 0 |
| v6-master (ca67b9a) | 4996 | 14 (0.3%) | 1 |
| v 5.0.8 (6b871e5) | 5290 | 17 (0.3%) | 4 |


shaan1337:
Tests currently failing due to #2505

src/EventStore.Core/Services/ElectionsService.cs
```diff
@@ -71,6 +71,7 @@ public class ElectionsService : IHandle<SystemMessage.BecomeShuttingDown>,
 		private LeaderCandidate _leaderProposal;
 		private Guid? _leader;
 		private Guid? _lastElectedLeader;
+		public Guid TestElectedLeaderId = Guid.Empty; //this property is used for test visibility purposes only
```
hayley-jean, line 74:
```suggestion
		public Guid ElectedLeaderId => _leader?.InstanceId ?? Guid.Empty;
```

It might be better to expose this properly rather than relying on setting a separate property just for testing.

shaan1337: I've pushed a new commit to improve this part: cd3760cf4a3a6cad73949914ed562e0b119b7181

thanks!

jageall: Yeah, I agree. There isn't much harm in exposing it properly as it is readonly.

PR #2463, opened 2020-05-06
Extend proto to define append response types

pgermishuys:
Changed: Appends will now return a oneof response types which currently is either a `Success` or a `WrongExpectedVersion`.

In an effort to reduce rpc exceptions being thrown from the server, we are extending the protocol to return responses which can be converted into errors or exceptions on the client side.

The `WrongExpectedVersion` error is described as the following:
**current_revision** is either an **actual revision** (as a `ulong`) _or_ `no_stream` (which is the default)
**expected_revision** is either an **actual revision** (as a `ulong`) _or_ `stream_exists` _or_ `any`.

Currently a write can fail for `Any` expected version of some of the events of in a batch has been written.

### DIFF
--- a/src/EventStore.Core/Services/Transport/Grpc/Streams.Append.cs
+++ b/src/EventStore.Core/Services/Transport/Grpc/Streams.Append.cs
@@ -103,21 +103,21 @@ namespace EventStore.Core.Services.Transport.Grpc {
 					return;
 				}
 
+				var response = new AppendResp();
 				switch (completed.Result) {
 					case OperationResult.Success:
-						var response = new AppendResp();
-
+						response.Success = new AppendResp.Types.Success();
 						if (completed.LastEventNumber == -1) {
-							response.NoStream = new Empty();
+							response.Success.NoStream = new Empty();
 						} else {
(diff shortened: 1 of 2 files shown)


src/EventStore.Core/Services/Transport/Grpc/Streams.Append.cs
```diff
@@ -103,21 +103,21 @@ partial class Streams {
...
 
+				var response = new AppendResp();
 				switch (completed.Result) {
 					case OperationResult.Success:
-						var response = new AppendResp();
-
+						response.Success = new AppendResp.Types.Success();
 						if (completed.LastEventNumber == -1) {
-							response.NoStream = new Empty();
+							response.Success.NoStream = new Empty();
```
jageall, line 111: This is a bit weird (but shouldn't be addressed in this PR). If we successfully appended to an empty stream we should still be able to return CurrentRevision for the next write? @thefringeninja do you know why we don't?

thefringeninja: I do not know

PR #2476, opened 2020-05-11
Prevent the same index and database path from being specified

avish0694:
The index system deletes anything that's not in the index map on startup. If the same index/db path is specified, the index system will delete all chunk files. To remedy this, we do a check to prevent the same index and database path from being used when EventStore starts up.

### DIFF
--- a/src/EventStore.ClusterNode/ClusterVNodeHostedService.cs
+++ b/src/EventStore.ClusterNode/ClusterVNodeHostedService.cs
@@ -83,6 +83,13 @@ namespace EventStore.ClusterNode {
 					"The given database path starts with a '~'. Event Store does not expand '~'.");
 			}
 
+			string absolutePathIndex = Path.GetFullPath(options.Index);
+			string absolutePathDb = Path.GetFullPath(options.Db);
+			if(absolutePathDb.Equals(absolutePathIndex))
+			{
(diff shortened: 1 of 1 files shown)


gregoryyoung:
Heh this one likely been there for many years under the assumption they are not the same location I believe ...

It might be worth revisiting the definition of what to delete overall as opposed to *everything*

src/EventStore.ClusterNode/ClusterVNodeHostedService.cs
```diff
@@ -85,6 +85,13 @@ internal class ClusterVNodeHostedService : EventStoreHostedService<ClusterNodeOp
 					"The given database path starts with a '~'. Event Store does not expand '~'.");
 			}
 
+			string absolutePathIndex = Path.GetFullPath(options.Index);
+			string absolutePathDb = Path.GetFullPath(options.Db);
+			if(absolutePathDb.Equals(absolutePathIndex))
+			{
+				throw new ApplicationInitializationException("The given database and index path points to the same folder");
```
pgermishuys, line 92:
```suggestion
				throw new ApplicationInitializationException($"The given database ({absolutePathDb}) and index ({absolutePathIndex}) paths cannot point to the same directory.");
```

PR #2479, opened 2020-05-12
Remove internal http endpoint

### LINKED ISSUE #2330: [Preview] Gossip and elections over gRPC can be done over both internal/external interfaces
The routes should be segregated based on the origin of the request (internal or external HTTP port).
We also need to take into consideration that `gossip` can be used by external clients to discover a cluster.

hayley-jean:
Removed: Internal http endpoint

Fixes #2330 
This PR requires https://github.com/EventStore/EventStore/pull/2475 to be merged first.
UI update: https://github.com/EventStore/EventStore.UI/pull/249

- Only listen on external http and allow client certificates.
- Remove internal http
- Rename external http to httpEndPoint
- Http port is now set by `--http-port` and `--advertise-http-port-as`
- Remove configuration options to do with manager nodes
- Remove mention of manager nodes in the client and replace them with node gossip port

Rename `OnExt` options:
- AdminOnExt to DisableAdminUi
- StatsOnExt to DisableStatsOnHttp
- GossipOnExt to DisableGossipOnHttp


src/EventStore.Core/Util/Opts.cs
```diff
@@ -259,14 +259,14 @@ public static class Opts {
 		public const string GossipTimeoutMsDescr = "The timeout, in ms, on gossip to another node.";
 		public const int GossipTimeoutMsDefault = 2500;
 
-		public const string AdminOnExtDescr = "Whether or not to run the admin ui on the external HTTP endpoint";
+		public const string AdminOnExtDescr = "Whether or not to run the admin ui on the HTTP endpoint";
```
shaan1337, line 262: @hayley-jean @pgermishuys Do we need to rename these three options: `AdminOnExt`, `GossipOnExt` and `StatsOnExt` ? (since the `OnExt` suffix is no longer applicable)

shaan1337: I think that `GossipOnExt` can be removed altogether since we should not be able to run a cluster if it's set to false.
`AdminOnExt` could be renamed to `DisableAdminUI` or something similar and `StatsOnExt` renamed to `DisableHttpStats`

hayley-jean: `GossipOnExt` could still be used as it could be used to disable other clients from connecting to the gossip (such as the TCP or grpc client) as these still use http gossip, not grpc.

Renaming the options is something we could do, so:
`DisableAdminUI`
`DisableStatsOnHttp`
`DisableGossipOnHttp`

shaan1337: @hayley-jean thanks, yup I missed this point regarding GossipOnExt.

shaan1337:
sorry I mean negate the values in with_default_settings.cs:
```
Assert.AreEqual(Opts.DisableAdminUiDefault, !_settings.AdminOnPublic, "AdminOnPublic");
Assert.AreEqual(Opts.DisableStatsOnHttpDefault, !_settings.StatsOnPublic, "StatsOnPublic");
Assert.AreEqual(Opts.DisableGossipOnHttpDefault, !_settings.GossipOnPublic, "GossipOnPublic");
```

PR #2485, opened 2020-05-13
Replace the usage of DnsEndPoint with an EventStoreEndPoint

pgermishuys:
Changed: Replace the use of `DnsEndPoint` with `EventStoreEndPoint`

We don't really use the `DnsEndPoint` apart from the `host` and `port`. We also see `unspecified` (e.g. `unspecified/localhost:2113`) as the address family in the logs which isn't nice.


shaan1337:
doesn't it defeat the purpose of using the abstract EndPoint class if we add a new EventStoreEndPoint class? IMHO, I find it a bit overkill and it adds more layers of abstractions if the main issue is only around logging.

PR #2487, opened 2020-05-14
More Client API IPEndpoint -> Endpoint changes + add Client cluster connection acceptance tests

shaan1337:
Changed: More changes to support DNS endpoints in the Client API.

Fixes #2478 
This is a follow up PR to #2480:
- Make sure that we resolve the IP address only in the TCPClientConnector
- Additional missing IPEndPoint -> EndPoint changes have been made
- Acceptance tests have been added for Client API cluster connection with IP/DNS gossip seed Endpoints: `EventStore.ClientAPI.Tests.connect_to_cluster`. These tests do not work for v5 yet due to #2497 and #2498 
- Allow setting `CustomHttpMessageHandler=SkipCertificateValidation` in the connection string to skip certificate verification (except for `net452` which does not support `ServerCertificateCustomValidationCallback`)
- Add support for the http scheme before the address in GossipSeeds in the connection string. e.g `GossipSeeds=http://127.0.0.1:2112` or `GossipSeeds=https://127.0.0.1:2112`. Backwards compatibility is maintained: `GossipSeeds=127.0.0.1:2112` also works but uses https by default.
- Removed `HostHeader` from the GossipSeed class
- Resolve the IPv4 address in the ClusterDnsEndPointDiscoverer by adding a `resolveDns` parameter in `endpoint.ToHttpUrl` due to #2501 occurring on Windows.

PR #2505, opened 2020-05-20
Set UseSslConnection=false and ValidateServer=false in connection string where required

shaan1337:
Changed: Set UseSslConnection=false and ValidateServer=false in connection string tests where required

This is a follow up PR to fix failing tests in #2487 which was merged after: #2503 but it was not noticed since CI was run on #2487 before #2503 was merged.


SiJewellPV (commented):
~lgtm~

PR #2513, opened 2020-05-26
Require IP or DNS SAN and CN=eventstoredb-node in client certificate to be assigned system role

shaan1337:
Changed: Require IP or DNS SAN and CN=eventstoredb-node in client certificate to be assigned system role

Fixes https://github.com/EventStore/home/issues/181


pgermishuys:
Consider using `eventstoredb-node` in the CN

shaan1337:
@pgermishuys Ok, I've made the name change from `eventstore-node` to `eventstoredb-node`

PR #2518, opened 2020-05-29
VNodeState in cluster.proto not matching EventStore.Core.Data.VNodeState

shaan1337:
Fixed: VNodeState in cluster.proto not matching EventStore.Core.Data.VNodeState

- Add warning about modifying the VNodeState enum and assign explicit values
- Add DiscoverLeader state to cluster.proto

DiscoverLeader state was added in PR #2386 but we did not update the cluster.proto file.
The reason this was not causing any noticeable issue between servers is that the same server versions were running together in a cluster. During gossip, typecasting EventStore.Core.Data.VNodeState to the protobuf enum would have the wrong value. However, when it would be deserialized on the other server, it would map back again to the correct value.

### DIFF
--- a/src/EventStore.Core/Data/VNodeState.cs
+++ b/src/EventStore.Core/Data/VNodeState.cs
@@ -1,21 +1,22 @@
 namespace EventStore.Core.Data {
+	//WARNING: new states must be added at the bottom of the enum otherwise it may break cluster and client compatibility
 	public enum VNodeState {
-		Initializing,
-		DiscoverLeader,
-		Unknown,
-		PreReplica,
-		CatchingUp,
-		Clone,
-		Follower,
-		PreLeader,
-		Leader,
-		Manager,
-		ShuttingDown,
-		Shutdown,
-		ReadOnlyLeaderless,
-		PreReadOnlyReplica,
-		ReadOnlyReplica,
-		ResigningLeader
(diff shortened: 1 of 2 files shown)


src/EventStore.Core/Data/VNodeState.cs
```diff
@@ -1,21 +1,22 @@
...
-		Leader,
-		Manager,
-		ShuttingDown,
-		Shutdown,
-		ReadOnlyLeaderless,
-		PreReadOnlyReplica,
-		ReadOnlyReplica,
-		ResigningLeader
+		Initializing = 0,
+		DiscoverLeader = 1,
```
pgermishuys, line 5: Is there any particular reason we slotted `DiscoverLeader` as `1` instead of adding it to the end?

shaan1337: @pgermishuys yes, the main reason I've added it at the top is because I've tried to keep it in the logical order of states that a node goes through when starting up

shaan1337: it also doesn't break anything right now, but new states will need to be added to the bottom after the next major release.

pgermishuys: Thanks @shaan1337. The reasoning makes sense.

PR #2519, opened 2020-05-29
Add external gRPC endpoint for gossip

### LINKED ISSUE #2298: Add an external grpc endpoint for gossip
Currently, the cluster connection in the grpc TCP client is done using the existing http gossip endpoint.
It would be better to expose the gossip info over grpc for these clients.

Clients should be able to get the gossip info, but should not be able to send any gossip updates.

It may also be a good idea to allow clients to subscribe to the gossip updates, as suggested here https://github.com/EventStore/EventStore/pull/2306#issuecomment-589712175

Note: We also need to keep external gossip http endpoints as existing TCP clients rely on it to connect to the cluster.

shaan1337:
Added: An external gRPC endpoint for gossip

Fixes #2298
This PR depends on #2518. Thus it has been opened against the `fix-vnodestate` branch so that it's retargeted by github to master once `fix-vnodestate` is merged to master.

The accompanying client-side PR is available here: https://github.com/EventStore/EventStore-Client-Dotnet/pull/27

Note: Concerning authorization, we're using the existing `Node.Gossip.Read` operation here which already allows anonymous access: https://github.com/EventStore/EventStore/blob/e4577510bc5d0d5e9ddaf94c65caeadc969576f5/src/EventStore.Core/Services/Transport/Grpc/Gossip.Read.cs#L16

### DIFF
--- a/src/EventStore.Core/Services/Transport/Grpc/Gossip.Read.cs
+++ b/src/EventStore.Core/Services/Transport/Grpc/Gossip.Read.cs
@@ -13,7 +13,7 @@ using MemberInfo = EventStore.Client.Gossip.MemberInfo;
 
 namespace EventStore.Core.Services.Transport.Grpc {
 	partial class Gossip {
-		private static readonly Operation ReadOperation = new Operation(Plugins.Authorization.Operations.Node.Gossip.Read);
+		private static readonly Operation ReadOperation = new Operation(Plugins.Authorization.Operations.Node.Gossip.ClientRead);
 		public override async Task<ClusterInfo> Read(Empty request, ServerCallContext context) {
 			var user = context.GetHttpContext().User;
 			if (!await _authorizationProvider.CheckAccessAsync(user, ReadOperation, context.CancellationToken).ConfigureAwait(false)) {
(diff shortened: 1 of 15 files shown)


src/EventStore.Core/Services/Transport/Grpc/Gossip.Read.cs
```diff
@@ -0,0 +1,49 @@
...
+using EventStore.Core.Messaging;
+using EventStore.Plugins.Authorization;
+using Grpc.Core;
+using static EventStore.Common.Utils.EndpointExtensions;
+using ClusterInfo = EventStore.Client.Gossip.ClusterInfo;
+using MemberInfo = EventStore.Client.Gossip.MemberInfo;
+
+namespace EventStore.Core.Services.Transport.Grpc {
+	partial class Gossip {
+		private static readonly Operation ReadOperation = new Operation(Plugins.Authorization.Operations.Node.Gossip.Read);
```
pgermishuys, line 16: The `ReadOperation` here is referring to the same resource as the cluster gossip (the resource being `node/gossip`) which means that a client strictly has access to the cluster gossip read.
Would it make more sense to introduce a new resource `client/gossip` or something similar?

cc: @jageall

shaan1337:
@pgermishuys @thefringeninja thanks for your reviews. I've added a new commit here which fixes the above issue: 3cba5ce38cd819f31b91d3754e12c730850974bd. It requires this PR in EventStore.Plugins to be merged and published: https://github.com/EventStore/EventStore.Plugins/pull/6 (i've assumed the next version would be `6.0.0-preview5`)

PR #2553, opened 2020-06-16
Improve support for 20.6 server in v5 client

### LINKED ISSUE #2497: ConnectionSettings.UseCustomHttpClient has no effect in EventStore client 5.0.8
The http client is not being passed as parameter to the ClusterDnsEndpointDiscoverer here:
https://github.com/EventStore/EventStore/blob/v5-master/src/EventStore.ClientAPI/EventStoreConnection.cs#L212

This change was made to be able to connect to a v6 server with gossip seeds.
Related PRs:
https://github.com/EventStore/EventStore/pull/2388
https://github.com/EventStore/EventStore/pull/1959

shaan1337:
Added: Improve support for 20.6 server in v5 client

Fixes #2497  
Currently the 5.0.8 client works with 20.6 server only with valid certificates due to #2497 
The PR fixes this and also adds support for 20.6 using connection strings.

- Pass HttpClient to ClusterDnsEndpointDiscoverer (UseCustomHttpClient previously had no effect)
- Support `CustomHttpClient=SkipCertificateValidation` in connection string (.netstandard2.0 only)
- Support TLS GossipSeeds in connection string (by prefixing gossip seed with `https://` in connection string)

Tests are present in PR #2554

### DIFF
--- a/src/EventStore.ClientAPI/ConnectionSettings.cs
+++ b/src/EventStore.ClientAPI/ConnectionSettings.cs
@@ -37,7 +37,7 @@ namespace EventStore.ClientAPI {
 		/// <summary>
 		/// Allows overriding the HTTPClient <see cref="IHttpClient"/>
 		/// </summary>
-		public IHttpClient CustomHttpClient { get; set; }
+		public IHttpClient CustomHttpClient;
 
 		/// <summary>
 		/// Whether to use excessive logging of <see cref="EventStoreConnection"/> internal logic.
--- a/src/EventStore.ClientAPI/ConnectionString.cs
+++ b/src/EventStore.ClientAPI/ConnectionString.cs
@@ -4,9 +4,12 @@ using System.Data.Common;
 using System.Globalization;
 using System.Linq;
 using System.Net;
+using System.Net.Http;
+using System.Net.Sockets;
 using System.Reflection;
 using System.Text;
 using EventStore.ClientAPI.SystemData;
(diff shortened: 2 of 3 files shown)


src/EventStore.ClientAPI/ConnectionSettings.cs
```diff
@@ -37,7 +37,7 @@ public sealed class ConnectionSettings {
 		/// <summary>
 		/// Allows overriding the HTTPClient <see cref="IHttpClient"/>
 		/// </summary>
-		public IHttpClient CustomHttpClient { get; set; }
+		public IHttpClient CustomHttpClient;
```
jageall, line 40: What reason is there to change this from a property?

shaan1337: `typeof(ConnectionSettings).GetFields(BindingFlags.Instance | BindingFlags.Public)` does not return the `CustomHttpClient` in the fields list here: https://github.com/EventStore/EventStore/pull/2553/files#diff-27fb9d7199f6a8cefc173c874380ed55L99

(It is required for connection strings)

jageall: okay, not ideal, but less trouble than doing anything else i guess

megakid:
Am I going mad? I just downloaded `EventStore.Client 5.0.9` (https://www.nuget.org/packages/EventStore.Client/5.0.9) and this code does not seem to be in there.

PR #2556, opened 2020-06-17
Introduce insecure flag to disable TLS on all TCP & HTTP interfaces to eliminate requirement for certificates

avish0694:
Added: --disable-https flag to disable only HTTPS
Added: --insecure flag to disable TLS on all interfaces (TCP & HTTP) to eliminate requirement for certificates to make it easier to run EventStoreDB

Fixes https://github.com/EventStore/home/issues/197

### DIFF
--- a/src/EventStore.ClusterNode/ClusterNodeOptions.cs
+++ b/src/EventStore.ClusterNode/ClusterNodeOptions.cs
@@ -212,7 +212,10 @@ namespace EventStore.ClusterNode {
 
 		[ArgDescription(Opts.DisableExternalTcpTlsDescr, Opts.InterfacesGroup)]
 		public bool DisableExternalTcpTls { get; set; }
-		
+
+		[ArgDescription(Opts.DisableHttpsDescr, Opts.InterfacesGroup)]
+		public bool DisableHttps { get; set; }
+
 		[ArgDescription(Opts.AuthorizationTypeDescr, Opts.AuthGroup)]
 		public string AuthorizationType { get; set; }
 
@@ -386,6 +389,7 @@ namespace EventStore.ClusterNode {
 
 			DisableInternalTcpTls = Opts.DisableInternalTcpTlsDefault;
 			DisableExternalTcpTls = Opts.DisableExternalTcpTlsDefault;
+			DisableHttps = Opts.DisableHttpsDefault;
 
 			AuthorizationType = Opts.AuthorizationTypeDefault;
 			AuthorizationConfig = Opts.AuthorizationConfigFileDefault;
--- a/src/EventStore.ClusterNode/ClusterNodeOptions.cs
+++ b/src/EventStore.ClusterNode/ClusterNodeOptions.cs
@@ -302,6 +302,9 @@ namespace EventStore.ClusterNode {
 		
 		[ArgDescription(Opts.DevDescr, Opts.AppGroup)]
 		public bool Dev { get; set; }
+
+		[ArgDescription(Opts.InsecureDescr, Opts.AppGroup)]
+		public bool Insecure { get; set; }
 		
 		[ArgDescription(Opts.EnableAtomPubOverHTTPDescr, Opts.InterfacesGroup)]
 		public bool EnableAtomPubOverHTTP { get; set; }
@@ -434,6 +437,7 @@ namespace EventStore.ClusterNode {
 			MaxAppendSize = Opts.MaxAppendSizeDefault;
 
 			Dev = Opts.DevDefault;
+			Insecure = Opts.InsecureDefault;
 		}
 	}
 }
--- a/src/EventStore.ClusterNode/ClusterVNodeHostedService.cs
+++ b/src/EventStore.ClusterNode/ClusterVNodeHostedService.cs
@@ -399,7 +399,7 @@ namespace EventStore.ClusterNode {
 				message += "\nTLS is enabled on at least one TCP/HTTP interface - a certificate is required to run EventStoreDB.";
 			} else {
 				message += "\nTLS is disabled on all TCP/HTTP interfaces - no certificates are required to run EventStoreDB.";
-				message += "\nWe recommended to enable TLS in PRODUCTION to ensure data security.\n";
+				message += "\nIt is recommended to run with TLS enabled in production.\n";
 			}
 
 			Log.Information(message);
--- a/src/EventStore.ClusterNode/ClusterVNodeHostedService.cs
+++ b/src/EventStore.ClusterNode/ClusterVNodeHostedService.cs
@@ -44,49 +44,49 @@ namespace EventStore.ClusterNode {
 				    && x.Source == "<DEFAULT>"
 				    && developmentMode) {
 					x.Value = true;
-					x.Source = "Set by 'Development Mode' mode";
+					x.Source = "Set by 'Development' mode";
 				}
 
 				if (x.Name == nameof(ClusterNodeOptions.CertificateFile)
 				    && x.Source == "<DEFAULT>"
 				    && developmentMode) {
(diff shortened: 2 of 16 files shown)


shaan1337:
@avish0694 I've added a --disable-https option to the PR which is also set to true when --insecure mode is activated.

Out of the box, kestrel does not simultaneously support both Http 1 and Http 2 without TLS. I've worked around it in this commit: a27cd6760eb6f7d47abbd56400fda19b6b695991 by using a modified version of the code written by @hagen93 in the following issue which examines the incoming request to detect the HTTP/2 preface:
https://github.com/KnowitSolutions/Grpc.Helpers/blob/master/Kestrel.ProtocolMultiplexing/ProtocolMultiplexingMiddleware.cs (original issue: https://github.com/dotnet/aspnetcore/issues/13502)

I've made a few small optimizations to `return false` early if there's a mismatch instead of having to wait for the whole connection preface (24 bytes) to be read.

shaan1337:
> What is the value in exposing `DisableHttps`?

@pgermishuys This is more about consistency with other configuration options:
-Insecure is a mode just like -Dev, it does not do anything apart from setting some pre-defined values in the configuration options. -DisableHttps does the actual work to disable HTTPS on the server. So basically -Insecure is a shortcut for disabling TLS on each HTTP and TCP interface.

>  How do we intend this configuration to be used in general as this leaves TLS over TCP enabled for `internal` as well as `external`?
The option can still work properly in the above configuration. It may be possible that a user wants cluster nodes to be mutually authenticated for data security (to prevent a rogue node from replacing all the data) but still wants external clients to be connected on plain HTTP.

src/EventStore.ClusterNode/ClusterVNodeHostedService.cs
```diff
@@ -351,34 +376,39 @@ internal class ClusterVNodeHostedService : EventStoreHostedService<ClusterNodeOp
...
-				builder.WithServerCertificateFromStore(name, options.CertificateSubjectName,
-					options.CertificateThumbprint);
-			} else if (options.CertificateFile.IsNotEmptyString()) {
-				builder.WithServerCertificateFromFile(
-					options.CertificateFile,
-					options.CertificatePrivateKeyFile,
-					options.CertificatePassword);
-			} else if (!options.Dev)
-				throw new Exception("A TLS Certificate is required unless development mode (--dev) is set.");
+			bool insecureMode = options.DisableHttps && options.DisableInternalTcpTls && options.DisableExternalTcpTls;
```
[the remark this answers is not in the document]
shaan1337: Please see my comment here: https://github.com/EventStore/EventStore/pull/2556#issuecomment-651602911, -insecure is a shortcut for setting other options

src/EventStore.ClusterNode/ClusterNodeOptions.cs
```diff
@@ -210,7 +210,10 @@ public class ClusterNodeOptions : IOptions {
 
 		[ArgDescription(Opts.DisableExternalTcpTlsDescr, Opts.InterfacesGroup)]
 		public bool DisableExternalTcpTls { get; set; }
-		
+
+		[ArgDescription(Opts.DisableHttpsDescr, Opts.InterfacesGroup)]
+		public bool DisableHttps { get; set; }
```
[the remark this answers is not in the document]
shaan1337: Noted, we could require certificates only if at least one interface requiring TLS is running instead of just checking the options. I'll address this in a separate commit

jageall:
I am not keen on adding yet another flag. I would rather drop all of the security flags and either run with security on for everything, or security off for everything. Trying to figure out which combinations of flags result in a secure or insecure setup becomes increasingly hard the more flags we add. What happens for example if you set --insecure and --disablehttps false? or --disabletls false? or worse, don't specify --insecure, but do specify the others as true? The issue that should be being addressed here is adding --insecure and disabling all security. If you want to add a flag for disablehttp that can be separate, that needs to be evaluated by the product team because it adds complexity and documentation requirements and can lead to configuration value sets that are hard to reason about.  So adding it should be done by creating an issue for review, not adding it to this PR

src/EventStore.ClusterNode/ClusterNodeOptions.cs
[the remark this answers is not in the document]
shaan1337: Ok, i've addressed these issues in this commit: 7972e059a07b8c275f3b8cf90d7279c0475fdb30

shaan1337:
@jageall @pgermishuys what do you think about the following - i think that the key is transparency to the user to make it more intuitive: 
by default:
```
[23013, 1,12:40:38.243,INF] 
SECURITY
Internal TCP (Replication)
        TLS enabled     : True
HTTP (gRPC / Admin UI)
        TLS enabled     : True
External TCP (Protobuf)
        TLS enabled     : True

TLS is enabled on at least one TCP/HTTP interface - a certificate is required to run EventStoreDB.
```

--insecure:
```
SECURITY
Internal TCP (Replication)
        TLS enabled     : False
HTTP (gRPC / Admin UI)
        TLS enabled     : False

TLS is disabled on all TCP/HTTP interfaces - no certificates are required to run EventStoreDB.
We recommended to enable TLS in PRODUCTION to ensure data security.
```

src/EventStore.ClusterNode/ClusterVNodeHostedService.cs
```diff
@@ -353,34 +378,63 @@ internal class ClusterVNodeHostedService : EventStoreHostedService<ClusterNodeOp
...
 
-			if (!string.IsNullOrEmpty(options.TrustedRootCertificatesPath)) {
-				builder.WithTrustedRootCertificatesPath(options.TrustedRootCertificatesPath);
+			if (requireCertificates) {
+				message += "\nTLS is enabled on at least one TCP/HTTP interface - a certificate is required to run EventStoreDB.";
 			} else {
-				throw new Exception(
-					$"{nameof(options.TrustedRootCertificatesPath)} must be specified unless development mode (--dev) is set.");
+				message += "\nTLS is disabled on all TCP/HTTP interfaces - no certificates are required to run EventStoreDB.";
+				message += "\nWe recommended to enable TLS in PRODUCTION to ensure data security.\n";
```
pgermishuys, line 402:
```suggestion
				message += "\nIt is recommended to run with TLS enabled in production.\n";
```

src/EventStore.ClusterNode/ClusterVNodeHostedService.cs
```diff
@@ -36,41 +36,64 @@ internal class ClusterVNodeHostedService : EventStoreHostedService<ClusterNodeOp
...
+				    && x.Source == "<DEFAULT>"
+				    && insecureMode) {
+					x.Value = true;
+					x.Source =  "Set by 'Insecure' mode";
 				}
 
 				if (x.Name == nameof(ClusterNodeOptions.CertificateFile)
 				    && x.Source == "<DEFAULT>"
-				    && developmentMode) {
+				    && developmentMode && !insecureMode) {
```
hayley-jean, line 73: Could this lead to a situation where a user expects dev certificates to be set up, but they aren't because I'm also running in insecure mode? e.g:
```
EventStore.ClusterNode.exe --dev --insecure --disable-external-tcp-tls=false
```
This would fail further down because no certs were specified but I have a secured interface.

shaan1337: @hayley-jean yup, we can skip these `insecureMode` checks and just set the dev certificates so that the --dev option is more consistent. If --insecure is specified, the certificates would not be used anyway.

shaan1337: This will be addressed when tackling: https://github.com/EventStore/home/issues/213

PR #2557, opened 2020-06-18
Fix Incorrect Reported Docker Container

thefringeninja:


thefringeninja:
@shaan1337 we can address that in a future PR. In fact I am questioning if we need any of those scripts at all. I don't think they are used by CI.

Issue #2559, opened 2020-06-19
Stackoverflow on tcp connections on dotnet core


hayley-jean:
**Describe the bug**
The process (client or server) crashes with a stackoverflow when sending/receiving a large amount of data over a TCP connection.

**To Reproduce**
Steps to reproduce the behavior on the server (20.6.0):
1. Start Event Store with TCP enabled, and TLS over TCP disabled
2. Write a large amount of data with large events
3. Event Store should crash in a few seconds

Steps to reproduce the behavior on the client:
1. Start Event Store with TCP enabled, and TLS over TCP disabled
2. Read a large amount of data with large events
3. The client should crash in a few seconds

**EventStore details**
 - EventStore server version: v20.6.0, 
 - Operating system: Any
 - EventStore client version (if applicable): v5.0.8

**Additional context**
Relevant context from dotnet: https://github.com/dotnet/runtime/issues/24950

PR #2560, opened 2020-06-19
Use a loop instead of recursion for synchronous ReceiveAsync

hayley-jean:
Fixed: Prevent Stackoverflow when accepting too much data over a TCP connection on dotnet core

Fixes: #2559 
Many thanks to @lscpike for the fix!

### DIFF
--- a/src/EventStore.ClientAPI/Transport.Tcp/TcpConnection.cs
+++ b/src/EventStore.ClientAPI/Transport.Tcp/TcpConnection.cs
@@ -200,31 +200,54 @@ namespace EventStore.ClientAPI.Transport.Tcp {
 		}
 
 		private void StartReceive() {
-			var buffer = new ArraySegment<byte>(new byte[TcpConfiguration.SocketBufferSize]);
-			_receiveSocketArgs.SetBuffer(buffer.Array, buffer.Offset, buffer.Count);
+
 			try {
-				NotifyReceiveStarting();
-				bool firedAsync = _receiveSocketArgs.AcceptSocket.ReceiveAsync(_receiveSocketArgs);
-				if (!firedAsync)
-					ProcessReceive(_receiveSocketArgs);
+
+				bool continueReceiveSynchronously = true;
+
+				do {
+
+					var buffer = new ArraySegment<byte>(new byte[TcpConfiguration.SocketBufferSize]);
+					_receiveSocketArgs.SetBuffer(buffer.Array, buffer.Offset, buffer.Count);
+
+					NotifyReceiveStarting();
+					bool firedAsync = _receiveSocketArgs.AcceptSocket.ReceiveAsync(_receiveSocketArgs);
+					if (firedAsync) {
+						continueReceiveSynchronously = false;
+					} else {
+
+						var processReceiveSuccess = ProcessReceive(_receiveSocketArgs);
+
+						if (processReceiveSuccess) {
+							TryDequeueReceivedData();
+						}
+
+						continueReceiveSynchronously = processReceiveSuccess;
+					}
+
+				} while (continueReceiveSynchronously);
+
 			} catch (ObjectDisposedException) {
 				ReturnReceivingSocketArgs();
 			}
 		}
 
 		private void OnReceiveAsyncCompleted(object sender, SocketAsyncEventArgs e) {
-			// No other code should go here.  All handling is the same on async and sync completion.
-			ProcessReceive(e);
+
+			if (ProcessReceive(e)) {
+				StartReceive();
+				TryDequeueReceivedData();
+			}
 		}
 
-		private void ProcessReceive(SocketAsyncEventArgs socketArgs) {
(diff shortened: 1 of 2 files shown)


lscpike:
You're welcome. Worth thinking about if the send side (TrySend) will have the same behaviour/bug. It's the same pattern otherwise.

src/EventStore.ClientAPI/Transport.Tcp/TcpConnection.cs
```diff
@@ -200,31 +200,54 @@ internal class TcpConnection : TcpConnectionBase, ITcpConnection {
...
 				ReturnReceivingSocketArgs();
 			}
 		}
 
 		private void OnReceiveAsyncCompleted(object sender, SocketAsyncEventArgs e) {
-			// No other code should go here.  All handling is the same on async and sync completion.
-			ProcessReceive(e);
+
+			if (ProcessReceive(e)) {
+				StartReceive();
```
shaan1337, line 247: @hayley-jean @lscpike would it make sense to inverse the order here?
i.e `TryDequeueReceiveData()` followed by `StartReceive()`

consider this scenario:
1. StartReceive() will eventually fire asynchronously
2. Let's say that there is new data available immediately and `OnReceiveAsyncCompleted` is called again before the current one has completed
3. This means that there can be multiple concurrent calls to `TryDequeueReceivedData`
4. Only one of these concurrent calls will run due to the lock in `TryDequeueReceivedData` and all other threads will be blocked unnecessarily

However, if we inverse the order it would ensure that we don't run into the above situation

lscpike: I think the idea is there should be an outstanding ReceiveAsync as much of the time as possible to ensure the socket buffer is getting emptied continuously. The TryDequeue is just something that must happen eventually. If anything your question has highlighted the fact that my implementation of the non recursive synchronous receive is not prioritising pulling from the socket above processing the queue. Maybe the StartReceive shouldn't be calling TryDequeue until the firedAsync == true?

shaan1337: @lscpike it makes sense, it appears to be the intention of the original code

it is not too intuitive which method would work best:
- if we dequeue/dispatch immediately, it should reduce the dispatch latency and should prevent "dispatch starvation" if we keep receiving data on the socket.
- if we dispatch after dequeuing a maximum amount of data possible, it gives time for more data to come in, but I think that the data is still being buffered internally at the OS level until we actually hit the OS's TCP receive buffer size at which point data will be discarded. And anyway since we're on a single thread, the total amount of time we spend dequeuing/dispatching the data will be the same so even though we're queuing more data, we will still need to process it later and overall it appears that it would be the same (as long as we're not near the OS's TCP receive buffer size)

i've done some performance tests with `WRFL 1 300000` (run 3 times and averaged) with the test client with only the "receive" commit on the server-side included to avoid being biased by the other changes:
1 - without fix: 34852 writes/sec
2 - with current fix: 36373 writes/sec
3 - with fix + StartReceive/TryDequeueReceivedData inversion: 37115 writes/sec
4 -  with fix + moving TryDequeueReceivedData after while loop : 35439 writes/sec

option 3 seems to be slightly better overall. option 4 also almost immediately prints out lots of messages like this which indicates it's checking out much more memory from the buffer manager to keep the temporary data than options 2 & 3:
```
5666,20,11:07:59.951,DBG] Segments count: 1, buffers count: 512, should be when full: 512
[ 5666,20,11:07:59.958,INF] External TCP connection accepted: [Normal, "127.0.0.1:58770", L"127.0.0.1:1113", {f4e3bf45-af8c-4ff0-ab81-07824a67ac8a}].
[ 5666,21,11:08:00.037,DBG] Segments count: 2, buffers count: 512, should be when full: 1024
[ 5666,21,11:08:00.050,DBG] Segments count: 3, buffers count: 512, should be when full: 1536
[ 5666,21,11:08:00.060,DBG] Segments count: 4, buffers count: 512, should be when full: 2048
...
[ 5666, 6,11:08:00.931,DBG] Segments count: 24, buffers count: 512, should be when full: 12288
```

On another note - it seems that a queue is not even necessary here if we go for option 3 since we always dispatch immediately but we can skip this optimization for now. We're going to deprecate the TCP client soon so it might not be worth optimizing it too much for the client side but the server side code will stay.

lscpike: Nothing quite like empirical data! Top work @shaan1337

hayley-jean: Thanks for the thorough investigation @shaan1337! I'll go ahead and update the PR

shaan1337: thanks for the update @hayley-jean

PR #2572, opened 2020-06-27
Expose option to set the client certificate common name

pgermishuys:
Added: Option to set client certificate common name

Fixes https://github.com/EventStore/home/issues/209

This used to be hardcoded to eventstoredb-node


src/EventStore.Core/Util/Opts.cs
```diff
@@ -218,6 +218,9 @@ public static class Opts {
 
 		public const string CertificateSubjectNameDescr = "The certificate subject name.";
```
shaan1337, line 219: @pgermishuys To avoid any confusion with the certificate loading from store, can we consider renaming:
- --certificate-subject-name to --certificate-store-subject-name? 
- --certificate-thumbprint to --certificate-store-thumbprint?

what do you think?

shaan1337: sorry, forget this, I misread the other option as `CertificateCommonName`

pgermishuys: Renaming the existing options be a breaking change. The best we can do is to describe the new option in a non confusing way.

src/EventStore.Core/Util/Opts.cs
```diff
@@ -218,6 +218,9 @@ public static class Opts {
...
 		public static readonly string CertificateSubjectNameDefault = string.Empty;
+		
+		public const string ClientCertificateCommonNameDescr = "The certificate subject name.";
```
shaan1337, line 222: @pgermishuys I think that the name may be a bit confusing because may support client certificate authentication from external clients later on.

what do you think about `CertificateReservedNodeCommonName` or `CertificateReservedNodeCN` ? with a description of: `The reserved common name to authenticate EventStoreDB nodes/servers from certificates`

pgermishuys: That's fair, i'll rename the option to `CertificateReservedNodeCommonName`

src/EventStore.ClusterNode/ClusterNodeOptions.cs
```diff
@@ -201,6 +201,8 @@ public class ClusterNodeOptions : IOptions {
 
 		[ArgDescription(Opts.CertificateSubjectNameDescr, Opts.CertificatesFromStoreGroup)]
 		public string CertificateSubjectName { get; set; }
+		[ArgDescription(Opts.CertificateReservedNodeCommonNameDescr, Opts.CertificatesFromStoreGroup)]
```
shaan1337, line 204:
```suggestion
		[ArgDescription(Opts.CertificateReservedNodeCommonNameDescr, Opts.CertificateGroup)]
```

src/EventStore.Core/Util/Opts.cs
shaan1337: thanks @pgermishuys

PR #2574, opened 2020-06-29
Improve Docker File

thefringeninja:
Fixed: Log to default directory; args from CLI

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -63,11 +63,16 @@ RUN addgroup --gid ${GID} "eventstore" && \
 
 COPY --from=publish /publish ./
 
+
 RUN mkdir -p /var/lib/eventstore && \
     mkdir -p /var/log/eventstore && \
-    chown -R eventstore:eventstore /opt/eventstore /var/lib/eventstore /var/log/eventstore
+    mkdir -p /etc/eventstore && \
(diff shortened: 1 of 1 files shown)


sigil66:
General best practice for a docker container is to only log to standard out. It's curious we are mounting a volume there, if a customer want's to do this mounting and enabling a log volume should be conditional.

thefringeninja:
@sigil66 I agree in principle, however stats are also logged to that folder (as opposed to being written to a stream in v5).

The other things that are logged to that folder (in json) is just a copy of the console logs (`file`), plus a dedicated error log (`err`). Unfortunately configuring serilog from a file with multiple targets is unnecessarily complicated.

For a future major release, I think a decent compromise is to use configuration to decide which of these logs you want to use, and for the container we turn off `file` and `err`. However that still leaves the `stats` log.

alexeyzimarev:
Are planning to change the user to use the user id (1001) instead of the name?

pgermishuys:
@alexeyzimarev shall we open an issue to deal with that or would you like it addressed in this PR?

Lougarou:
@pgermishuys @alexeyzimarev it looks like we are assigning uid 1000 to eventstore. Does it have to be 1001?
The rest of the changes (log and ip) looks good to me

alexeyzimarev:
I guess 1000 is fine

PR #2580, opened 2020-07-02
Require anonymous access for gossip updates and elections when HTTPS is disabled

shaan1337:
Changed: Instead of always giving system access when running with --insecure (since no client certificate is provided), make gossip/updates policies anonymous when HTTPS is disabled.

Fixes https://github.com/EventStore/home/issues/214


shaan1337:
Closed in favor of https://github.com/EventStore/EventStore/pull/2587

PR #2581, opened 2020-07-02
Review --dev option

shaan1337:
Changed: When --dev is set, disable TLS on all interfaces instead of setting development certificates
Changed: Do not set --mem-db when --dev is set

Fixes https://github.com/EventStore/home/issues/213


alexeyzimarev:
I am confused. Why do we need the `--dev` option if we have `--insecure`?

PR #2590, opened 2020-07-08
Add ability to reload certificates by sending SIGHUP signal or by triggering /admin/reloadconfig endpoint

shaan1337:
Added: Ability to reload certificates by triggering the /admin/reloadconfig endpoint or by sending a SIGHUP signal (linux only)

Fixes https://github.com/EventStore/home/issues/199

This PR needs to wait for https://github.com/EventStore/EventStore.Plugins/pull/8 to be merged and EventStore.Plugins 20.6.1 to be published before being merged.

## Expected work flow:
1 - Update certificates on each node one by one. Update the config file if required (e.g if certificate file names/password/paths have changed)
2. Send  a POST request to the /admin/reloadconfig endpoint (with admin/ops privileges) or a SIGHUP signal to the EventStore process (Linux only)

## All steps below are optional to verify if the new certificates are working fine

3 - Wait for at least 10 minutes to see if there are any gossip errors (due to .NET connection pooling, see this commit: 1e472bee6b477009e45bc605be44863d869194b6)
4 - Decrease node priority of current leader node and trigger a leader resignation on the leader node to ensure TCP connections are dropped and new connections are established
5 - Restart all gRPC clients  to ensure new certificates are being used by the client (Adding `PooledConnectionLifetime` to the gRPC client should not require a restart for unary calls but it must be verified if it works with streaming calls as well)
6 - Restart all TCP clients to ensure new certificates are being used

### DIFF
--- a/src/EventStore.Core/ClusterVNode.cs
+++ b/src/EventStore.Core/ClusterVNode.cs
@@ -208,7 +208,7 @@ namespace EventStore.Core {
 				(endpoint, publisher) =>
 					new EventStoreClusterClient(
 						new UriBuilder(!_vNodeSettings.DisableHttps ? Uri.UriSchemeHttps : Uri.UriSchemeHttp,
-							endpoint.GetHost(), endpoint.GetPort()).Uri, publisher, _internalServerCertificateValidator, _certificate));
+							endpoint.GetHost(), endpoint.GetPort()).Uri, publisher, _internalServerCertificateValidator, _certificateSelector));
 
 			_mainBus.Subscribe<ClusterClientMessage.CleanCache>(_eventStoreClusterClientCache);
 			_mainBus.Subscribe<SystemMessage.SystemInit>(_eventStoreClusterClientCache);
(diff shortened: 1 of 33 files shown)


src/EventStore.Core/ClusterVNode.cs
```diff
@@ -906,6 +932,74 @@ public class ClusterVNode :
...
+			} else {
+				throw new InvalidConfigurationException(
+					$"{nameof(options.TrustedRootCertificatesPath)} was not specified in the configuration.");
+			}
+
+			//no need for a lock here since reference assignment is atomic
+			_certificate = certificate;
+			_trustedRootCerts = trustedRootCerts;
+
+			Log.Information("Certificates have been successfully reloaded");
```
jen20, line 1000: It might be nice to include the thumbprint of both the old and the new certificates here?

src/EventStore.Core/ClusterVNode.cs
```diff
@@ -777,6 +800,8 @@ public class ClusterVNode :
 		}
 
 		public void Handle(SystemMessage.BecomeShuttingDown message) {
+			UnixSignalManager.StopProcessing();
```
jen20, line 803: We should be careful to ensure that this doesn't break SIGINT while shutting down, which should immediately terminate the process.

shaan1337: @jen20 yup, thanks, I've verified that it doesn't break it and does a proper shut down. We're currently handling only signals that we subscribe to in the UnixSignalManager (only SIGHUP currently), any other signal will still use its default handler.

src/EventStore.Core/ClusterVNode.cs
shaan1337: I've added a new commit: f63f989b51c6172af25649bd53d9e3fe6ca291e9 thanks

PR #2599, opened 2020-07-13
Add Transaction Log and Native Projects

condron:
Added: common native file access with positional write and read semantics.
Removed: unbuffered file access
Updated: Unbuffered, Writethrough, and ReduceFileCache command line parameters are now obsolete and will print a warning if set

Added Transaction Log Project and moved related code.


gregoryyoung:
Is a fixed 4096 page size reasonable here on everything?

https://github.com/EventStore/EventStore/pull/2599/commits/842188aa9540f368582db28cf586782a539ddc63#diff-68e31aa1f72dd8ab38c229b0906a1d93R12

condron:
> Is a fixed 4096 page size reasonable here on everything?
> 
> [842188a#diff-68e31aa1f72dd8ab38c229b0906a1d93R12](https://github.com/EventStore/EventStore/commit/842188aa9540f368582db28cf586782a539ddc63#diff-68e31aa1f72dd8ab38c229b0906a1d93R12)

getting the correct disk page size from the various linux variations is not supported in mono.posix and has proven to be non trivial. This page size will work for all currently supported hardware and the intention is to replace it with the actual correct lookup.

gregoryyoung:
@condron we can chat when we are in person but there are some other things which can be done here that could theoretically help significantly (EG 100+% speed up)

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
```diff
@@ -230,13 +208,13 @@ public enum ChunkVersions : byte {
 				_chunkHeader = ReadHeader(reader.Stream);
 				Log.Debug("Opened completed {chunk} as version {version}", _filename, _chunkHeader.Version);
 				if (_chunkHeader.Version != (byte)ChunkVersions.Unaligned &&
-				    _chunkHeader.Version != (byte)ChunkVersions.Aligned)
+					_chunkHeader.Version != (byte)ChunkVersions.Aligned)
 					throw new CorruptDatabaseException(new WrongFileVersionException(_filename, _chunkHeader.Version,
 						CurrentChunkVersion));
 
-				if (_chunkHeader.Version != (byte)ChunkVersions.Aligned && _unbuffered) {
+				if (_chunkHeader.Version != (byte)ChunkVersions.Aligned) {
```
hayley-jean, line 214: This will prevent ES from starting up with chunk files made in ES v3.x which haven't been scavenged on v4 or v5 yet. 

If this is intentional, we should remove the check above that checks for either Unaligned or Aligned chunk versions, and be aware that this introduces an upgrade step

pgermishuys: cc: @condron

condron: Which checks should we we remove above?

hayley-jean:
```csharp
if (_chunkHeader.Version != (byte)ChunkVersions.Unaligned && 
    _chunkHeader.Version != (byte)ChunkVersions.Aligned)	
		throw new CorruptDatabaseException(new WrongFileVersionException(_filename, _chunkHeader.Version, CurrentChunkVersion));
```
At line [211](https://github.com/EventStore/EventStore/pull/2599/files/b3351c7c2bcbcfde11a1ab3d7cb5c559b4eaa901#diff-9906bfb99c1b4f17ed2f491ac6a107caR211) there's a check to ensure that the chunk is either unaligned or aligned

condron: @hayley-jean  Added issue in zenhub for the review on this.

hayley-jean (approved):

pgermishuys (approved):

PR #2600, opened 2020-07-15
Added Index and Storage Projects

condron:
added: Index and Storage Projects


hayley-jean (approved):

src/EventStore.Storage/Storage/EpochManager/EpochManager.cs
```diff
@@ -223,8 +219,8 @@ public class EpochManager : IEpochManager {
 
 			Log.Debug("=== Writing E{epochNumber}@{epochPosition}:{epochId:B} (previous epoch at {lastEpochPosition}). L={leaderId:B}.",
 				epochNumber, epoch.EpochPosition, epochId, lastEpochPosition, epoch.LeaderInstanceId);
-
-			_bus.Publish(new SystemMessage.EpochWritten(epoch));
+			//todo - publish this
```
pgermishuys, line 222: This `todo` can be removed as we are publishing this message from all the `WriteNewEpoch` callers.

src/EventStore.Storage/Storage/EpochManager/EpochManager.cs
```diff
@@ -223,8 +219,8 @@ public class EpochManager : IEpochManager {
...
+			//_bus.Publish(new SystemMessage.EpochWritten(epoch));
```
pgermishuys, line 223: Just a note that this does move the responsibility from the `EpochManager` to publish the `EpochWritten` to the caller of `WriteNewEpoch`.

pgermishuys (approved):
This looks good in general.
Do we want to adjust the namespaces to suite the directories the types have been moved into?

src/EventStore.Storage/Storage/EpochManager/EpochManager.cs
condron: The challenge here is that we are looking to disconnect the system management  functions as expressed by publishing and handling various the `SystemMessages` from the lower level storage implementation. So publishing a system control message directly here directly couples storage into system management. 
Now, currently the system really does need know when an epoch has been written, but once we get back up to reviewing the system management flows I expect this will be better expressed as a slightly different concept than the implied state change from having written the epoch record.

Issue #2598, opened 2020-07-13
Move/replicate logging settings to the ops config


alexeyzimarev:
We need to either move or replicate some of the logging configurations to the server configuration (yaml, env vars and command line). I'd suggest starting with the log level. The setting to disable structured logging is already there, so it would be natural to add the log level there as well.

We also discussed a proposal to create an HTTP endpoint to change the log level on-the-fly, so if there's a need to debug a weird situation, we can change the log level on the fly. What happens now is that we change the logs to be more verbose to debug something and restart the server but after the restart, the problem goes away most of the time.

PR #2602, opened 2020-07-16
Ability to load (or reload) default log level from EventStore config file

shaan1337:
Added: Ability to load (or reload) default log level from EventStore config file

Fixes: #2598 
Note: This PR merges into `certificate-rolling` branch so that it's retarged to master once #2590 is merged.

This PR adds a `LogLevel` configuration option.
Supported values: Default, Verbose, Debug, Information, Warning, Error, Fatal

- Prior to this PR, log levels could already be adjusted dynamically by editing logconfig.json file (located under ` /usr/share/eventstore/` on linux). The change takes effect almost immediately.
- logconfig.json allows editing the log levels for different namespaces: `Grpc`, `System`, `Microsoft` and there's a special one called `Default` for remaining namespaces
- This PR adds a `LogLevel` option to the EventStore configuration to control the `Default` log level.
- The log level can be dynamically reloaded (only if specified in config file) by triggering the /admin/reloadconfig endpoint or sending a SIGHUP signal.
- The option can be specified on the command line (--log-level), environment variable (EVENTSTORE_LOG_LEVEL) or config file (LogLevel). Only config files support dynamic reloading since if the log level is specified on command line/env var they will take precedence over the config file. However, editing logconfig.json directly will still work dynamically even if log level option has been specified on command line/env var.

### DIFF
--- a/src/EventStore.ClusterNode/ClusterVNodeHostedService.cs
+++ b/src/EventStore.ClusterNode/ClusterVNodeHostedService.cs
@@ -15,6 +15,7 @@ using EventStore.Core.PluginModel;
 using EventStore.Core.Services.Monitoring;
 using EventStore.Core.Services.Transport.Http.Controllers;
 using System.Threading.Tasks;
+using EventStore.Common.Log;
 using EventStore.Core.Authentication.InternalAuthentication;
 using EventStore.Core.Authorization;
 using EventStore.Core.Services.PersistentSubscription.ConsumerStrategy;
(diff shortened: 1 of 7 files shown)


shaan1337:
> We can remove the `Log.Information("Reloading the node's configuration since a request has been received on /admin/reloadconfig.");` in `AdminController` to reduce this noise.

@pgermishuys  Thanks I've fixed it in this commit: 0ee2a2b0c07bf4f5f1688b34805f9c2f6764335a on the certificate-rolling branch.

PR #2605, opened 2020-07-17
Introduce the ability to restart persistent subscriptions

pgermishuys:
Added: Introduced the ability to restart the persistent subscriptions service

Fixes https://github.com/eventstore/home/issues/218

### DIFF
--- a/src/EventStore.Core/Services/Transport/Http/Controllers/PersistentSubscriptionController.cs
+++ b/src/EventStore.Core/Services/Transport/Http/Controllers/PersistentSubscriptionController.cs
@@ -40,6 +40,7 @@ namespace EventStore.Core.Services.Transport.Http.Controllers {
 
 		protected override void SubscribeCore(IHttpService service) {
 			Register(service, "/subscriptions", HttpMethod.Get, GetAllSubscriptionInfo, Codec.NoCodecs, DefaultCodecs, new Operation(Operations.Subscriptions.Statistics));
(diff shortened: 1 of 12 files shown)


gregoryyoung:
This likely wants to be a bit more (sadly) complex in how it works.

Ideally its likely wanted to "flag" whether or not to "drain" first. Doing a drain should not be difficult.

pgermishuys:
@gregoryyoung thanks for the input! Agreed, will make sure to consider that as I flesh this out.

src/EventStore.Core/Services/Transport/Http/Controllers/PersistentSubscriptionController.cs
```diff
@@ -677,6 +678,27 @@ public class PersistentSubscriptionController : CommunicationController {
...
+
+			var envelope = new SendToHttpEnvelope(_networkSendQueue, http,
+				(e, message) => e.ResponseCodec.To("Restarting"),
+				(e, message) => {
+					switch (message) {
+						case SubscriptionMessage.PersistentSubscriptionsRestarting _:
+							return Configure.Ok(e.ResponseCodec.ContentType);
+						case SubscriptionMessage.InvalidPersistentSubscriptionsRestart fail:
+							return Configure.BadRequest
+								($"Persistent Subscriptions cannot be restarted as it is in the wrong state: {fail}");
```
hayley-jean, line 694:
```suggestion
								($"Persistent Subscriptions cannot be restarted as it is in the wrong state.");
```
There isn't any relevant information on the message to show in this error message.

PR #2608, opened 2020-07-22
Fix Incorrect Calculation of Checkpoint Interval on Filtered Subcriptions

thefringeninja:
Fixed: wrong calculation of checkpoint interval for filtered subscriptions

### DIFF
--- a/src/EventStore.Core/Services/Transport/Grpc/Enumerators.AllSubscriptionFiltered.cs
+++ b/src/EventStore.Core/Services/Transport/Grpc/Enumerators.AllSubscriptionFiltered.cs
@@ -142,7 +142,7 @@ namespace EventStore.Core.Services.Transport.Grpc {
 				private Position _nextPosition;
 				private ResolvedEvent _current;
 				private Position _currentPosition;
-				private uint _checkpointIntervalCounter;
+				private long _checkpointIntervalCounter;
 
 				public ResolvedEvent Current => _current;
 				public Position CurrentPosition => _currentPosition;
@@ -191,7 +191,7 @@ namespace EventStore.Core.Services.Transport.Grpc {
 					_requiresLeader = requiresLeader;
 					_checkpointReached = checkpointReached;
 					_maxWindowSize = maxWindowSize ?? ReadBatchSize;
-					_checkpointInterval = checkpointIntervalMultiplier;
+					_checkpointInterval = checkpointIntervalMultiplier * _maxWindowSize;
 					_disposedTokenSource = new CancellationTokenSource();
 					_buffer = new ConcurrentQueueWrapper<(ResolvedEvent, Position?)>();
 
@@ -282,8 +282,10 @@ namespace EventStore.Core.Services.Transport.Grpc {
 
 						switch (completed.Result) {
 							case FilteredReadAllResult.Success:
+								 var position = Position.FromInt64(completed.CurrentPos.CommitPosition,
+									completed.CurrentPos.PreparePosition);
 								foreach (var @event in completed.Events) {
-									var position = Position.FromInt64(
+									position = Position.FromInt64(
 										@event.OriginalPosition.Value.CommitPosition,
 										@event.OriginalPosition.Value.PreparePosition);
 									if (position <= _startPosition) {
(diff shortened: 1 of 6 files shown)


mat-mcloughlin:
This works now, but I still have concerns about the fact the interval ends up being `interval x maxWindowSize` that is hard coded to 32 for Grpc client. That means

1 = every 32
2 = every 64

etc

@jageall

src/EventStore.Core/Services/Transport/Grpc/Enumerators.AllSubscriptionFiltered.cs
```diff
@@ -303,10 +304,18 @@ private class CatchupAllSubscription : IStreamSubscription {
 									completed.NextPos.CommitPosition,
 									completed.NextPos.PreparePosition);
 
-								unchecked {
-									if (_checkpointInterval++ % _checkpointIntervalMultiplier == 0) {
-										_buffer.Enqueue((default, _nextPosition));
+								if (completed.Events.Length == 0) {
+									unchecked {
```
pgermishuys, line 308: We can safely remove this as @shaan1337 suggested in his comment as well.

shaan1337:
@thefringeninja thanks for making the changes.
I'm seeing an anomaly, but I'm still investigating why it's happening:

1. Off by one after filtered match
- Filtered reads resulting in no matches look OK - the checkpoints are being triggered at the correct intervals.
- However, filtered reads resulting in matches seem to be triggering the checkpoints off by one:
  - Close your client application
  - Write an event to stream `test`
  - Write 99 events to stream `abcd`
  - Run your client application with a stream filter: `test` and max search window of 1 and checkpoint interval of 100.
  - A checkpoint will be triggered after `test` is read but it shouldn't have been triggered. (if you try with 98 events, it will not be triggered which suggests that an off-by-one error is happening)
  - For some reason, resetting the counter to -1 instead of 0 when `completed.Events.Length != 0` seems to fix the issue but i think that there's another underlying problem.

2. A search window of 1 and checkpoint interval of 1 seems to result in an infinite loop which continuously triggers checkpoints
This seems unrelated to this PR and may have been present before but I'm mentioning it here since it might be related to the issue 1 above.

3.  Also, this is not an issue but an observation:
- Live subscription checkpoints are quite different from paged reads checkpoints since with live subscriptions we are processing event by event while with paged reads we are processing page by page. Would it make sense if we had a `liveCheckpointInterval` and a `pageCheckpointInterval` which the user would need to pass in?

I will post some updates if I figure out what's happening for issue 1 and 2.

shaan1337:
1. and 2. can be fixed with:
```
if (!completed.IsEndOfStream && ++_checkpointIntervalCounter >= _checkpointInterval) {
```

The last empty page sent to indicate the end of stream was not being processed properly.

jageall:
If the number of events in the $all stream is less than 100, I would only expect one checkpoint call when it switches to live. otherwise i would expect 1 checkpoint call every 100 events, and 1 when it switches to live. I am guessing that is not what you are expecting. Also it might be  there is an off by one error with next position that is coming back from the AllReader if you are getting the same event multiple times.

pgermishuys (changes requested):
I am currently seeing
```
Error while processing message "EventStore.Core.Messages.ClientMessage+FilteredReadAllEventsForward" in queued handler '"StorageReaderQueue #3"'.
System.ArgumentOutOfRangeException: The commit position cannot be less than the prepare position (Parameter 'commitPosition')
   at EventStore.Core.Services.Transport.Grpc.Position..ctor(UInt64 commitPosition, UInt64 preparePosition) in C:\source\EventStore\src\EventStore.Core\Services\Transport\Grpc\Position.cs:line 49
   at EventStore.Core.Services.Transport.Grpc.Position.FromInt64(Int64 commitPosition, Int64 preparePosition) in C:\source\EventStore\src\EventStore.Core\Services\Transport\Grpc\Position.cs:line 20
   at EventStore.Core.Services.Transport.Grpc.Enumerators.AllSubscriptionFiltered.CatchupAllSubscription.<>c__DisplayClass23_0.<MoveNextAsync>g__OnMessage|0(Message message) in C:\source\EventStore\src\EventStore.Core\Services\Transport\Grpc\Enumerators.AllSubscriptionFiltered.cs:line 287
   at EventStore.Core.Messaging.CallbackEnvelope.ReplyWith[T](T message) in C:\source\EventStore\src\EventStore.Core\Messaging\CallbackEnvelope.cs:line 15
   at EventStore.Core.Services.Storage.StorageReaderWorker.EventStore.Core.Bus.IHandle<EventStore.Core.Messages.ClientMessage.FilteredReadAllEventsForward>.Handle(FilteredReadAllEventsForward msg) in C:\source\EventStore\src\EventStore.Core\Services\Storage\StorageReaderWorker.cs:line 184
   at EventStore.Core.Bus.MessageHandler`1.TryHandle(Message message) in C:\source\EventStore\src\EventStore.Core\Bus\MessageHandler.cs:line 30
   at EventStore.Core.Bus.InMemoryBus.Publish(Message message) in C:\source\EventStore\src\EventStore.Core\Bus\InMemoryBus.cs:line 285
   at EventStore.Core.Bus.InMemoryBus.Handle(Message message) in C:\source\EventStore\src\EventStore.Core\Bus\InMemoryBus.cs:line 279
   at EventStore.Core.Bus.QueuedHandlerThreadPool.ReadFromQueue(Object o) in C:\source\EventStore\src\EventStore.Core\Bus\QueuedHandlerThreadPool.cs:line 119
```

This looks to be related to the recent addition of converting the `PrevPosition` using `Position.FromInt64`

PR #2635, opened 2020-08-14
Add Dockerfile for Event Store v5

### LINKED ISSUE #2591: Move docker building for pre-v20.6.0 into this repository
As discussed, we want to keep the management of the docker builds contained in this repository.
We will need to ensure that the docker building and publishing have been moved over here before deprecating the [docker repository](https://github.com/EventStore/eventstore-docker).

arwinneil:
- Ubuntu base image updated from 16.04 to 18.04

---
Fixes #2591

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,25 @@
+FROM ubuntu:18.04
+MAINTAINER Event Store Ltd <ops@eventstore.com>
+
+ENV ES_VERSION=5.0.8-1 \
+    DEBIAN_FRONTEND=noninteractive \
+    EVENTSTORE_CLUSTER_GOSSIP_PORT=2112
+
+RUN apt-get update \
+    && apt-get install tzdata curl iproute2 -y \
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM ubuntu:18.04
+MAINTAINER Event Store Ltd <ops@eventstore.com>
+
+ENV ES_VERSION=5.0.8-1 \
```
thefringeninja, line 4: shouldn't this be an ARG instead?

arwinneil: Adapted it from the current v5 Dockerfile from https://github.com/EventStore/eventstore-docker.

 Makes sense to be using ARG here! Updating it

pgermishuys (approved):
This looks good.
There is still some additional admin tasks associated with this to complete this.
- [ ] How and when will we archive the old docker repository?
- [ ] How do we want to handle the open issues on the docker repository? Move them over here?
- [ ] Do we need to do some additional work with regards to documenting or introducing a process to publish this image?

PR #2637, opened 2020-08-14
Allow external clients (e.g UI) to discover supported authentication methods

shaan1337:
Added: Allow external clients to discover supported authentication methods

Related to: https://github.com/EventStore/EventStore.UI/issues/265
This PR adds an `authentication` property to the JSON response of `/info` to allow external clients (e.g UI) to discover supported authentication methods (insecure, internal, oauth, ldaps)

The PR also adds the `WWW-Authenticate` header to 401 responses but with an `X-` in front of the authentication scheme to prevent default browser behaviour (like basic auth popups)

Note: It requires EventStore.Plugins 20.6.1-preview3 to be published (https://github.com/EventStore/EventStore.Plugins/pull/10)

### DIFF
--- a/src/EventStore.Core/ClusterVNode.cs
+++ b/src/EventStore.Core/ClusterVNode.cs
@@ -474,20 +474,20 @@ namespace EventStore.Core {
 
 			var httpAuthenticationProviders = new List<IHttpAuthenticationProvider>();
 
-			foreach (var authenticationScheme in _authenticationProvider.GetSupportedAuthenticationSchemes() ?? Enumerable.Empty<AuthenticationSchemes>()) {
+			foreach (var authenticationScheme in _authenticationProvider.GetSupportedAuthenticationSchemes() ?? Enumerable.Empty<string>()) {
 				switch (authenticationScheme)
 				{
-					case AuthenticationSchemes.Basic:
+					case "Basic":
 						httpAuthenticationProviders.Add(new BasicHttpAuthenticationProvider(_authenticationProvider));
 						break;
-					case AuthenticationSchemes.Bearer:
+					case "Bearer":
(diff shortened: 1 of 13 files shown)


src/EventStore.Core/ClusterVNode.cs
```diff
@@ -472,26 +472,38 @@ public class ClusterVNode :
...
+						httpAuthenticationProviders.Add(new BasicHttpAuthenticationProvider(_authenticationProvider));
+						break;
+					case "Bearer":
+						httpAuthenticationProviders.Add(new BearerHttpAuthenticationProvider(_authenticationProvider));
+						break;
+					case "Insecure":
+						httpAuthenticationProviders.Add(new PassthroughHttpAuthenticationProvider(_authenticationProvider));
+						break;
+					default:
+						Log.Warning($"Unsupported Authentication Scheme: {authenticationScheme}");
```
hayley-jean, line 490: How would Event Store react if none of the authentication schemes were supported? Is that something we want to check after this loop?

shaan1337: @hayley-jean we could log an error in the case where none of the auth schemes are supported, what do you think?

cc @jageall

hayley-jean: Would this be a case where we would want to log an error and shut the node down?

jageall: No supported auth scheme should result in the server shutting down

shaan1337: @hayley-jean @jageall Fixed it in this commit: 0ebef5cf9c1515e918ac8151db8bb931b30ce777, thanks !

hayley-jean (approved):

hayley-jean (approved):

jageall (approved):

PR #2638, opened 2020-08-14
Correctly handle cases when DefaultCredentials are NotAuthenticated in TCP

hayley-jean:
Fixed: Error on TCP operations after default user fails authentication

Fixes https://github.com/EventStore/home/issues/250

- Respond with `NotAuthenticated` when the default user has failed authentication and no credentials have been passed with the TCP package.
- Respond with `NotHandled.NotReady` when the `NotReady` method is called on the `TcpAuthRequest`.

This fixes the following issue:

The default user has the `User` ClaimsPrincipal set when it passes authentication. However, if it fails this remains null.
This change fixes an error in the following situation:
- The connection is using a default user
- The default user failed authentication, either because the node isn't ready or because the credentials are wrong
- The operation does not provide credentials itself

Originally, the above scenario would cause a null argument error as the package doesn't include any authentication tokens:

```Error while processing package. Error: System.ArgumentNullException: Value cannot be null```


Lougarou (approved):
No more errors on server side and on client side receiving exception EventStore.ClientAPI.Exceptions.NotAuthenticatedException

jageall (approved):

PR #2641, opened 2020-08-18
Add options to set the host and ports advertised by the gossip to clients

hayley-jean:
Added: AdvertiseHostToClientAs, AdvertiseHttpPortToClientAs and AdvertiseTcpPortToClientAs to allow setting the gossip and TCP endpoints advertised to clients.

Fixes: https://github.com/EventStore/home/issues/251

Add the options `AdvertiseHostToClientAs`, `AdvertiseHttpPortToClientAs` and `AdvertiseTcpPortToClientAs` options to set the endpoints advertised to clients over gossip.
Expose the above over the gossip to clients connecting to the gRPC GossipClient and to the Http GossipController.


jageall (approved):

alexeyzimarev (approved):

pgermishuys (approved):
This looks good to me @hayley-jean.
An observation which is not a functional issue is that you can't tell from the client gossip which address the nodes are using the gossip with each other.

PR #2642, opened 2020-08-19
Remove Dotnet-ism when Applying Persistent Subscription Settings

thefringeninja:
Changed: MessageTimeout and CheckpointAfter in persistent subscription settings are now expressed in milliseconds.


hayley-jean (commented):
This breaks compatibility with the existing gRPC client, previous versions create the subscriptions successfully but with an incorrect timeout

Would it be possible to add a new field for times in milliseconds, but allow specifying them in ticks for old versions?

thefringeninja:
> Would it be possible to add a new field for times in milliseconds, but allow specifying them in ticks for old versions?

I believe it is possible, we can turn this into a `oneof` without breaking anything.

jageall (approved):

hayley-jean (approved):

PR #2648, opened 2020-08-25
Remove development mode

hayley-jean:
Removed: Development mode

Fixes: https://github.com/EventStore/home/issues/228

With the introduction of `--insecure` and instructions for setting up a cluster in docker, `--dev` doesn't serve much purpose.


pgermishuys (approved):

thefringeninja (approved):

bweston92:
`--dev` actually seemed to do a lot more then `--insecure` what flags can we replicate `--dev` with?

alexeyzimarev:
Yes, like the local cluster set up in Docker Compose as described in the docs https://developers.eventstore.com/server/20.6/server/installation/docker.html#use-docker-compose

bweston92:
I assumed we had to generate certs now, thanks for taking the time to reply.

Issue #2646, opened 2020-08-23
CreatePersistentSubscriptionAsync fails against a 20.6.0 follower when using the 5.0.9 dotnet client


01100010011001010110010101110000:
**Describe the bug**
When executing `CreatePersistentSubscriptionAsync` from the `5.0.9` dotnet client against a Follower 20.6.0 node, the following exception is raised

```
System.ArgumentNullException: Value cannot be null.
Parameter name: ipString
   at System.Net.IPAddress.Parse(String ipString)
   at EventStore.ClientAPI.Messages.ClientMessage.NotHandled.MasterInfo.get_ExternalTcpEndPoint()
   at EventStore.ClientAPI.ClientOperations.OperationBase`2.InspectNotHandled(TcpPackage package)
   at EventStore.ClientAPI.ClientOperations.OperationBase`2.InspectPackage(TcpPackage package)
--- End of stack trace from previous location where exception was thrown ---
   at EventStore.ClientAPI.Internal.EventStoreNodeConnection.CreatePersistentSubscriptionAsync(String stream, String groupName, PersistentSubscriptionSettings settings, UserCredentials credentials)
```

**To Reproduce**
1. Execute `CreatePersistentSubscriptionAsync` against a Follower node in a 20.6.0 cluster using the `5.0.9` client

**Expected behavior**
The client should successfully terminate the connection to the Follower node, connect to master, and create the persistent subscription. This does work from the `5.0.9` client against a slave node in a v5 cluster, or from the `20.6.0` client against a `20.6.0` cluster. 

**Actual behavior**
```
System.ArgumentNullException: Value cannot be null.
Parameter name: ipString
   at System.Net.IPAddress.Parse(String ipString)
   at EventStore.ClientAPI.Messages.ClientMessage.NotHandled.MasterInfo.get_ExternalTcpEndPoint()
   at EventStore.ClientAPI.ClientOperations.OperationBase`2.InspectNotHandled(TcpPackage package)
   at EventStore.ClientAPI.ClientOperations.OperationBase`2.InspectPackage(TcpPackage package)
--- End of stack trace from previous location where exception was thrown ---
   at EventStore.ClientAPI.Internal.EventStoreNodeConnection.CreatePersistentSubscriptionAsync(String stream, String groupName, PersistentSubscriptionSettings settings, UserCredentials credentials)
```

**Config/Logs/Screenshots**
Three node cluster, I can provide logs or a trivial project to repro the problem if that would help. 

[...]

PR #2652, opened 2020-08-25
Handle clusters with only secure tcp endpoints during reconnect

01100010011001010110010101110000:
When sent a reconnect message from a 20.6 cluster (and presumably also from a v5 cluster with `DisableInsecureTcp` enabled, but I've not tested that), `ExternalTcpAddress` is `null` in the message's master info. This allows either `tcpEndPoint` or `secureTcpEndPoint` to be null in the reconnect message, but not both. 

Related to #2646


pgermishuys (approved):
This looks good to me @01100010011001010110010101110000

hayley-jean (approved):

PR #2672, opened 2020-09-08
Add Insecure to builder and hide disable https from embedded client

hayley-jean:
Added: `RunInsecure` option to Embedded Client

Additionally move pass through auth into the builder so that it doesn't need to be set when running insecure in the Embedded Client.

### DIFF
--- a/src/EventStore.ClusterNode/ClusterVNodeHostedService.cs
+++ b/src/EventStore.ClusterNode/ClusterVNodeHostedService.cs
@@ -304,9 +304,7 @@ namespace EventStore.ClusterNode {
 				if(options.DisableInternalTcpTls || options.DisableExternalTcpTls)
 					throw new InvalidConfigurationException($"The '{nameof(options.Insecure)}' option cannot be combined with the '{nameof(options.DisableInternalTcpTls)}' or the '{nameof(options.DisableExternalTcpTls)}' options.");
 
-				builder.DisableInternalTcpTls();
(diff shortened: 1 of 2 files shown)


thefringeninja (approved):

shaan1337 (dismissed):
The changes look good to me - tested with a secure/insecure cluster (HTTP and TCP) and an embedded single node (TCP only)

HTTP does not appear to work on embedded clients right now since the HTTP server is initialized externally in EventStore.ClusterNode/Program.cs

shaan1337 (changes requested):
The changes look good overall but I've noticed the following:
- WithHttpOn() does not appear to work since it seems that the http server is now launched externally. Is it expected?

- There are some discrepancies with the TCP APIs which make it unintuitive. I would suggest the following API changes:
  - Remove WithExternalSecureTcpOn() which is redundant and does not take DisableExternalTcpTls() into consideration
  - Remove WithInternalSecureTcpOn() which is redundant and does not take DisableInternalTcpTls() into consideration
  - In Build() set either _externalSecureTcp or _externalTcp based on whether _disableExternalTcpTls is set (this would also automatically take RunInsecure() into consideration)
  - In Build() set either _internalSecureTcp or _internalTcp based on whether _disableInternalTcpTls is set (this would also automatically take RunInsecure() into consideration)

cc @hayley-jean @thefringeninja  @jageall

PR #2674, opened 2020-09-09
Do not connect to PreReadOnlyReplica nodes

01100010011001010110010101110000:
I encountered an issue where clients would connect to read only replicas which had not yet caught up. This PR will prevent the endpoint discovered from prioritizing such nodes, and [match behavior](https://github.com/EventStore/EventStore-Client-Dotnet/blob/dc6d7dc3941a3c614a5b33aa619eb9306698dc48/src/EventStore.Client/GossipBasedEndpointDiscoverer.cs#L197) with the gRPC client


hayley-jean (approved):
Thanks for the PR, this looks good to me 👍

arwinneil (approved):

avish0694 (approved):
Code review done. Looks good to me.

PR #2677, opened 2020-09-15
Make Follower and Replica Nodes Report Healthy

### LINKED ISSUE #2676: Health Checks Return 503 Service Unavailable on Follower / Replica Nodes

thefringeninja:
Fixes #2676

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,9 +38,10 @@ jobs:
         password: ${{ secrets.GITHUB_TOKEN }}
     - name: Docker Compose Smoke Test
       run: |
+        docker-compose build
         docker-compose up --detach
         printf 'Waiting for cluster'
-        timeout 60 bash -c -- 'until $(curl --output /dev/null --silent --insecure --fail https://localhost:2113/); do printf '.'; sleep 2; done'
+        timeout 60 bash -c -- 'until $(curl --output /dev/null --silent --insecure --fail https://localhost:2113/health/live); do printf '.'; sleep 2; done'
         docker-compose down
   build-container:
     strategy:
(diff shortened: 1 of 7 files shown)


hayley-jean (dismissed):

shaan1337 (commented):
After going through the code, it seems that `SystemReady` was waiting for:
- authentication provider to be initialized (after which `SystemCoreReady` is published)
- all subsystems to be initialized.

In the UserManagementService as well we listen for `BecomeFollower` before publishing `SystemCoreReady`:
https://github.com/EventStore/EventStore/blob/master/src/EventStore.Core/Authentication/InternalAuthentication/UserManagementService.cs#L212

So everything should be working fine at least for leaders and followers (but not replica nodes) but there is a bug that prevents it from working on followers:
- The projection subsystems is initialized only when a node becomes a leader and thus for a follower subsystems are never initialized until it becomes a leader.

With the current fix, we could encounter race conditions in our tests running against containers (e.g attempting to do an operation with admin when auth service is not yet initialized and default users not yet created)

Some fixes I can think of:
- Not ideal: We don't wait for subsystems - only auth provider (by setting _ready when SystemCoreReady is true)
- Ideally: We keep track of an `_initialized` boolean in `ProjectionsSubsystem` and set it to true when the node becomes a leader, follower or read-only-replica. We publish `SystemMessage.SubSystemInitialized` only when first setting `_initialized` to true.

Note: we also need to update UserManagementService to handle BecomeReadOnlyReplica.

shaan1337 (approved):
Looks good :+1: tested with a leader, follower and read-only replica

hayley-jean (approved):

PR #2679, opened 2020-09-16
Echo Back Expected Revision / State to Client

thefringeninja:
Changed: responses to append will include expected revision / state sent from client

### DIFF
--- a/src/EventStore.ClusterNode/EventStore.ClusterNode.csproj
+++ b/src/EventStore.ClusterNode/EventStore.ClusterNode.csproj
@@ -9,15 +9,15 @@
 				<Platforms>x64</Platforms>
 		</PropertyGroup>
 		<ItemGroup>
-				<PackageReference Include="System.ComponentModel.Composition" Version="4.7.0"/>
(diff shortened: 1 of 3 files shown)


George-Payne:
Tested against  node client:

image

shaan1337 (commented):
When testing `StreamState.NoStream` with 20.6 client for backwards compatibility:

```
await client.AppendToStreamAsync("stream", StreamState.NoStream, ...);
```

where `stream` is an existing stream, I'm seeing the following anomaly:

```
EventStore.Client.WrongExpectedVersionException: Append failed due to WrongExpectedVersion. Stream: stream, Expected version: 0, Actual version: 5
```

On the `master` branch, the `Expected version` in the exception is -1 (which seems correct).

After tracing through the code with this PR / 20.6 client, it seems that we're entering  `case AppendReq.Types.Options.ExpectedStreamRevisionOneofCase.NoStream:` but since the old client does not handle `NoStream`, the `Expected version` stays 0.

hayley-jean (approved):

shaan1337 (approved):
Looks good - tested with 20.6 client and this PR: https://github.com/EventStore/EventStore-Client-Dotnet/pull/73

PR #2700, opened 2020-09-28
Fix memory/disk space issues during large cascading index merges / index rebuilds (especially when index cache depth is high)

shaan1337:
Fixed: Memory/disk space issues during large cascading index merges (especially when index cache depth is high)

Thanks to @slav for his assistance with this issue!

This PR fixes multiple memory/disk issues with the indexes:
- During cascading index merges, immediately dispose merged PTables instead of waiting for all the cascading merges to complete. This helps save a lot of memory taken by midpoints (especially when the index cache depth is high) and disk space since the merged PTables can be deleted immediately. Before the PR, around (2 x size of all current PTables on disk) free disk space was required for a cascading merge but now only (1 x size of largest PTable in the cascading merge) free disk space is required.
- Keep midpoints in unmanaged memory and immediately release the memory used by the midpoints when the PTable is disposed.
- Keep midpoints in unmanaged memory instead of List<Midpoint> (which takes excessive memory, particularly when internally doubling the array size) when merging PTables and immediately release the memory when the merge has completed.
- Prevent MemTables from piling up when large index merges are taking place during full index rebuilds by pausing the index rebuild and waiting until the merge has completed.

The following is the rationale behind using unmanaged memory for the midpoints:
- With a high index cache depth, the midpoints would often go to the Large Object Heap and thus they would rarely be garbage collected. From the user-perspective, it would be unclear whether a memory leak is present or whether the midpoints are just waiting to be garbage collected. Thus unmanaged memory gives us much more predictable behaviour.

### DIFF
--- a/src/EventStore.Core/DataStructures/UnmanagedMemoryAppendOnlyList.cs
+++ b/src/EventStore.Core/DataStructures/UnmanagedMemoryAppendOnlyList.cs
@@ -1,0 +1,89 @@
+using System;
+using System.Runtime.InteropServices;
+
+namespace EventStore.Core.DataStructures {
+
+	public interface IAppendOnlyList<in T> {
+		void Add(T item);
+		void Clear();
+		int Count { get; }
+	}
+
+	public sealed class UnmanagedMemoryAppendOnlyList<T> : IAppendOnlyList<T>, IDisposable {
+		private readonly int _maxCapacity;
+		private readonly IntPtr _dataPtr;
+
+		private int _count;
+		private bool _disposed;
+
+		public UnmanagedMemoryAppendOnlyList(int maxCapacity) {
+			if (maxCapacity <= 0) {
+				throw new ArgumentException("maxCapacity must be positive");
+			}
+
+			_maxCapacity = maxCapacity;
+			_count = 0;
+			_dataPtr = IntPtr.Zero;
+			_dataPtr = Marshal.AllocHGlobal(new IntPtr((long)_maxCapacity * Marshal.SizeOf(typeof(T))));
+		}
+
+		~UnmanagedMemoryAppendOnlyList() => Dispose(false);
+
+		public void Add(T item) {
+			if (_count >= _maxCapacity)
+				throw new MaxCapacityReachedException();
+
+			unsafe
+			{
+				new Span<T>(_dataPtr.ToPointer(), _maxCapacity) {
+					[_count] = item
+				};
+			}
+			_count++;
+		}
+
+		public void Clear() {
+			_count = 0;
+		}
+
+		public int Count => _count;
+
+		public T this[int index] {
+			get {
+				if (index < 0 || index >= _count) {
+					throw new IndexOutOfRangeException();
+				}
+
+				unsafe
+				{
+					return new ReadOnlySpan<T>(_dataPtr.ToPointer(), _maxCapacity)[index];
+				}
+			}
+		}
+
+		public void Dispose() {
+			Dispose(true);
+		}
+
+		private void Dispose(bool disposing) {
+			if (_disposed) {
+				return;
+			}
+
+			if (disposing) {
+				//dispose any managed objects here
+			}
+
+			if (_dataPtr != IntPtr.Zero) {
+				Marshal.FreeHGlobal(_dataPtr);
+			}
+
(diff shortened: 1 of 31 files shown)


slav:
Would it make sense to check if there's enough free memory before you allocate memory for merge, and if not enough, abandon the merge? Or, handle out of memory exception during allocation, and if ran out of memory abandon the merge and free up all the memory? That might be better than crashing with OOM in case server runs out of memory.

shaan1337:
@slav Thanks for the suggestions. We will try to tackle this separately since it does not seem to be a straightforward issue :
- If we abandon the merge due to low memory, then we need to determine how to deal with new PTables:
  - Do we block any new merges from happening? This will cause small PTables to accumulate when new writes come in and each PTable will consume a certain amount of memory due to the midpoints which will continuously grow.
  - Do we simulate behaviour similar to MaxAutoMergeIndexLevel and block only merges at a certain level until it is manually triggered by the user?

- We also thought about temporarily releasing the midpoints or degrading the midpoint depth for the existing PTables before doing the merge if we're low on memory but this would temporarily cause performance degradation. The implementation may be a bit complex for this one too.

shaan1337:
Just a small change that needs to be done to this PR: We need to call GC.AddMemoryPressure / GC.RemoveMemoryPressure when allocating/releasing unmanaged memory

src/EventStore.Core/Services/Storage/ReaderIndex/IndexCommitter.cs
```diff
@@ -119,6 +119,16 @@ public class IndexCommitter : IIndexCommitter {
 							(result.RecordPostPosition - startPosition) * 100.0 / (buildToPosition - startPosition));
 						lastTime = DateTime.UtcNow;
 					}
+
+					if (processed % 1000000 == 0) {
+						if (_tableIndex.IsBackgroundTaskRunning) {
+							Log.Debug("Pausing ReadIndex Rebuild due to ongoing index merges.");
+							while (_tableIndex.IsBackgroundTaskRunning) {
```
hayley-jean, line 126: Would this result in large Event Store instances taking longer to come online because all merges must complete before the ReadIndex is available?

shaan1337: @hayley-jean it could happen if all of the following conditions are true:
i) MaxMemTableSize was changed by the user to a value greater than 1M
ii) There are more than 1M events to rebuild when the node starts up
iii) When the node was shut down, there was a pending merge that was not completed.

We could ensure it happens only when the index is behind with something like:
`if(processed % Math.Max(MaxMemTableSize, 1000000) == 0`

shaan1337: @hayley-jean i've fixed it in this commit: cfaf4cbc0cc344d95553d62bd0d06d0392702efb
The index rebuild will be paused during merges only if it's a full index rebuild.

src/EventStore.Core/DataStructures/UnmanagedMemoryAppendOnlyList.cs
```diff
@@ -0,0 +1,110 @@
...
+		private readonly int _maxCapacity;
+		private readonly IntPtr _dataPtr = IntPtr.Zero;
+		private readonly long _dataBytesAllocated;
+
+		private int _count;
+		private bool _disposed;
+
+		public UnmanagedMemoryAppendOnlyList(int maxCapacity) {
+			if (maxCapacity <= 0) {
+				throw new ArgumentException("maxCapacity must be positive");
```
thefringeninja, line 23: Should add `nameof(maxCapacity)` to the exception's constructor

thefringeninja (changes requested):

src/EventStore.Core/DataStructures/UnmanagedMemoryAppendOnlyList.cs
shaan1337: thanks @thefringeninja! fixed in this commit: 44382425ecf32b71ba9db77eae9903197b1583ca

hayley-jean (approved):
We still need to do performance tests on these changes

thefringeninja (approved):

PR #2708, opened 2020-10-04
Added some docs to Uuid

dharmaturtle:
Source: Slack: 
image


thefringeninja:
@dharmaturtle,

Your PR is actually in the wrong repo. Can you add this information here: https://github.com/EventStore/EventStore-Client-Dotnet/blob/master/src/EventStore.Client/Uuid.cs#L6 and resubmit? Thank you!

PR #2711, opened 2020-10-05
Fix Off By One Error in Subscriptions

thefringeninja:
Fixed: off by one error in subscriptions


ahjohannessen:
Do `AllSubscription` and `AllSubscriptionFiltered` also have this bug?

Issue #2659, opened 2020-08-27
Calling /users/ endpoint randomly results in server-side exception


shaan1337:
Calling the /users/ endpoint multiple times results in the following exception on the server:

```
[13234,21,14:57:01.649,ERR] Error while processing message "EventStore.Core.Messages.ClientMessage+ReadStreamEventsBackwardCompleted" in queued handler '"Worker #1"'.
System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
   at System.Collections.Generic.List`1.Enumerator.MoveNext()
   at System.Linq.Enumerable.Any[TSource](IEnumerable`1 source, Func`2 predicate)
   at EventStore.Core.Authentication.InternalAuthentication.AllUsersReader.ReadUserDetailsBackwardCompleted(String loginName, ReadStreamEventsBackwardCompleted result) in /home/shaan/Documents/Work/EventStore/EventStore/src/EventStore.Core/Authentication/InternalAuthentication/AllUsersReader.cs:line 86
   at EventStore.Core.Authentication.InternalAuthentication.AllUsersReader.<>c__DisplayClass12_0.<BeginReadUserDetails>b__0(ReadStreamEventsBackwardCompleted result) in /home/shaan/Documents/Work/EventStore/EventStore/src/EventStore.Core/Authentication/InternalAuthentication/AllUsersReader.cs:line 77
   at EventStore.Core.Helpers.IODispatcher.<>c__DisplayClass25_0.<ReadBackward>b__0(ReadStreamEventsBackwardCompleted res) in /home/shaan/Documents/Work/EventStore/EventStore/src/EventStore.Core/Helpers/IODispatcher.cs:line 200
   at EventStore.Core.Messaging.RequestResponseDispatcher`2.Handle(TResponse message) in /home/shaan/Documents/Work/EventStore/EventStore/src/EventStore.Core/Messaging/RequestResponseDispatcher.cs:line 66
   at EventStore.Core.Messaging.RequestResponseDispatcher`2.EventStore.Core.Bus.IHandle<TResponse>.Handle(TResponse message) in /home/shaan/Documents/Work/EventStore/EventStore/src/EventStore.Core/Messaging/RequestResponseDispatcher.cs:line 48
[...]

PR #2713, opened 2020-10-07
Fixes exception when requesting "https://127.0.0.1:2113/users" multiple times

dyowee:
This is for issue #2659 "Calling /users/ endpoint randomly results in server-side exception". Previously, the current user details is being read asynchronously while at the same time moving forward to the next event from the stream. This causes concurrent access to the internal users list/collection.

The fix was to complete adding the user details to the collection, before moving to the next event in the stream.


dyowee:
Accidentally closed the PR.

thefringeninja (approved):
This seems to fix an issue we're having with a flaky test in the client repos.

avish0694 (approved):
Looks good after testing. No exception was logged after calling `/users/` endpoint multiple times and reading the current user details.

shaan1337 (approved):
Looks good :+1: thanks for this fix @dyowee!

Stress tested with:
```
for i in `seq 1 1000`; do curl --silent -u "admin:changeit" -k 'https://127.0.0.1:2113/users/' |jq -r '.data[].loginName' > ./login_test/$i & ; done
```

Then verified output with:
```
$ cat ./login_test/*|grep -e 'admin' -e 'ops'|wc -l
2000
```

No exceptions on server-side as well

dyowee:
Thanks! :)

PR #2720, opened 2020-10-07
Don't resolve DNS entries to IPs from GossipSeeds

### LINKED ISSUE #2718: v5.0.9 client cannot use a loadbalanced https gossip endpoint with SNI
**Describe the bug**
Connection string of "GossipSeeds=https://my-loadbalanced-https-endpoint:443" does not work on a typical load balancer SNI setup.  Given we cannot use the discover:// protocol with HTTPS in 5.0.9 (without non-connectionstring mods?), this is the next best option but it doesn't work as the DNS name is resolved to an IP address before being a gossip GET is issued.  This prevents the loadbalancer from choosing a certificate using SNI.

**To Reproduce**
1. Expose a https loadbalanced endpoint to your ES cluster using SNI (e.g. normal NGINX behaviour)
2. Try to connect using the connectiong string "GossipSeeds=https://my-loadbalanced-https-endpoint:443"
3. See that it fails to connect as the DNS is resolved to an IP here https://github.com/EventStore/EventStore/blob/dotnet-client-v5.0.9/src/EventStore.ClientAPI/ConnectionString.cs#L60  Unfortunately the code to add a Host header is not enough to allow SNI to work.

**Expected behavior**
An HTTPS load balanced endpoint should work

**Actual behavior**
It fails to connect to the loadbalancer due to using 'https://{ip address}:443' rather than 'https://{dns}:443'

**Additional context**
I understand that this might be fixed in the latest clients so might not an issue worth fixing but thought I'd flag it up (I am trying to find a decent migration 20.6.1 path for our 5.0.9 clients).

megakid:
Not sure what the thinking is here as it seems *very* deliberate but I'll put this here as an RFC.

There are numerous configurations that mean you might want to use TLS (without IP SANs) on your gossip seed addresses.

This fixes #2718 in the v20.6 client

### DIFF
--- a/src/EventStore.ClientAPI/Internal/ClusterDnsEndPointDiscoverer.cs
+++ b/src/EventStore.ClientAPI/Internal/ClusterDnsEndPointDiscoverer.cs
@@ -172,7 +172,7 @@ namespace EventStore.ClientAPI.Internal {
 
 			var url = endPoint.EndPoint.ToHttpUrl(
 				endPoint.SeedOverTls ? EndpointExtensions.HTTPS_SCHEMA : EndpointExtensions.HTTP_SCHEMA,
-				"/gossip?format=json", true);
+				"/gossip?format=json");
 			_client.Get(
 				url,
 				null,


jageall:
IIRC, this is because to make this work we either need to do a reverse lookup or get you to use a SRV record and is under discussion to resolve internally. @shaan1337 or @sigil66 might have a better answer here. but you are right, as of now we only support ips because of that, but we are looking to have a better story around it

megakid:
Would be interested to know the scenarios you're referring to.

I acknowledge this change is probably breaking to some setups but that aside, unless I've missed something I think this change just adds flexibility? 

If you defined GossipSeeds yourself directly, you could choose a DNS (this is the only scenario where different behaviour would occur) or a static IP entry (same behaviour). 
In the case of a discover://, it would result in IP gossip seeds and therefore result in the same behaviour.

shaan1337:
@megakid the reason the hostname had to be resolved in that case was due to mixed IPv4 / IPv6 environments. The main scenario was that `localhost` resolves to both the IPv4 loopback and IPv6 loopback addresses on Windows by default and the HttpClient has no way to filter out IPv6 addresses when making the http request.

The change was introduced in this commit: 670148cff26870387f2c384097a0981dbe99f1ba

Since this would usually be a rare scenario, the simplest solution would be that we could resolve + filter out IPv6 addresses only if it's trying to connect to `localhost` on Windows.

Also, please check out this comment from @davidsh: https://github.com/dotnet/runtime/issues/20876#issuecomment-379585707

SNI should work when setting the Host header in .Net Core 2.1 or above. Could you please confirm what framework you're targeting? (in principle it should work if you're targeting .NET Core 3.1)

megakid:
Thanks @shaan1337 - that explains it, we don't run IPv6 so wouldn't have encountered that issue.  Would it be better to just make localhost a special case then?

Re. SNI I will wireshark it and see what's going on as we're targetting `netcoreapp3.1` and using NGINX plus to give us the certificate etc

shaan1337:
> Re. SNI I will wireshark it and see what's going on as we're targetting `netcoreapp3.1` and using NGINX plus to give us the certificate etc

Thanks for the confirmation, it's strange that it's not working with `netcoreapp3.1`. Please keep us updated about the wireshark results so that we can decide on the way forward.

> Thanks @shaan1337 - that explains it, we don't run IPv6 so wouldn't have encountered that issue. Would it be better to just make localhost a special case then?

Yup, this sounds like the simplest solution so far.

megakid:
I can't reproduce the SNI failure, and I can see on wireshark that SNI host in the initial Client Hello TLSv1.2 message is correctly populated.

shaan1337:
Thanks for the update @megakid. Will discuss with the team on the way forward

shaan1337:
@megakid hello, after discussing with the team, we've decided to leave the code as is and not treat localhost as a special case since we do not have a clear picture of what impact it could have on mixed IPv4/IPv6 networks.

shaan1337:
@megakid we're closing this PR for now since we don't have a repro for the issue yet. please feel free to reopen it if you've been able to reproduce the issue. thanks!

PR #2736, opened 2020-10-28
Prioritize leader selection in elections by replication checkpoint.

condron:
Fixed: EventStore/home#263

Close edge case for isolated deposed leaders being preferred in elections due to an artificially high epoch count.
The prioritizing by replication checkpoint correctly identifies the instance with the most current client acknowledge data.


condron:
Persisting the replication checkpoint creates a performance hit that is not easily addressable, the same effect is being achieved via PR #2742 without the negative performance impact.

Issue #2547, opened 2020-06-11
Not able to start 20.6 on Windows 10 


jmail:
When trying to start:

`docker run --name eventstore2 -it -p 2113:2113 -p 1113:1113 eventstore/eventstore
[    1, 1,07:45:29.764,INF]
"ES VERSION:"             "0.0.0" ("Unknown"/"Unknown", "Unknown")
[    1, 1,07:45:29.779,INF] "OS:"                     Linux ("Unix 4.19.76.0")
[    1, 1,07:45:29.781,INF] "RUNTIME:"                ".NET 3.1.4" (64-bit)
[    1, 1,07:45:29.781,INF] "GC:"                     "3 GENERATIONS"
[    1, 1,07:45:29.782,INF] "LOGS:"                   "/var/log/eventstore"
[    1, 1,07:45:29.790,INF] MODIFIED OPTIONS:

        EXT IP:                   0.0.0.0 (Command Line)
        INT IP:                   0.0.0.0 (Command Line)

DEFAULT OPTIONS:

        HELP:                     False (<DEFAULT>)
        VERSION:                  False (<DEFAULT>)
        LOG:                      /var/log/eventstore (<DEFAULT>)
        CONFIG:                   <empty> (<DEFAULT>)
        WHAT IF:                  False (<DEFAULT>)
        START STANDARD PROJECTIONS: False (<DEFAULT>)
        DISABLE HTTP CACHING:     False (<DEFAULT>)
        HTTP PORT:                2113 (<DEFAULT>)
        ENABLE EXTERNAL TCP:      False (<DEFAULT>)
        INT TCP PORT:             1112 (<DEFAULT>)
        EXT TCP PORT:             1113 (<DEFAULT>)
        EXT HOST ADVERTISE AS:    <empty> (<DEFAULT>)
        EXT TCP PORT ADVERTISE AS: 0 (<DEFAULT>)
        HTTP PORT ADVERTISE AS:   0 (<DEFAULT>)
        INT HOST ADVERTISE AS:    <empty> (<DEFAULT>)
        INT TCP PORT ADVERTISE AS: 0 (<DEFAULT>)
        INT TCP HEARTBEAT TIMEOUT: 700 (<DEFAULT>)
        EXT TCP HEARTBEAT TIMEOUT: 1000 (<DEFAULT>)
        INT TCP HEARTBEAT INTERVAL: 700 (<DEFAULT>)
[...]

PR #2738, opened 2020-10-30
Add option to disable client certificates

BrunoZell:
This adds an option `EVENTSTORE_DISABLE_CLIENT_CERTIFICATE` (boolean) to EventStore. When this option is set to true, the Kestrel option `ClientCertificateMode` will **not** be set to `ClientCertificateMode.AllowCertificate` but instead will default to `ClientCertificateMode.NoCertificate`.

The current implementation will always set this to `ClientCertificateMode.AllowCertificate`. This will result in Google Chrome and curl to automatically send client certificates (esp. when accessed through `localhost`). The client certificate validation will fail within EventStore with the error `The certificate provided by the client failed validation with the following error(s): RemoteCertificateChainErrors (PartialChain)`.

When I configured EventStore for production I constantly hit this error and believed I made a mistake at configuring the server certificate. Event though the error states the client certificate is invalid, it didn't cross my mind that both the browser and curl sending invalid client certificates could be the reason here. And I don't really know how to disable sending client side certificates. Even the `curl -k` option does not work for me and results in `curl: (56) Failure when receiving data from the peer` with the `PartialChain` validation error describe above.

Reading issue #2547, there seem to be quite a few other users confused about this behavior. ([comment](https://github.com/EventStore/EventStore/issues/2547#issuecomment-671036141), [another comment](https://github.com/EventStore/EventStore/issues/2547#issuecomment-698347885)). Due to this it might even be a good idea to disable client certificate authentication by default. However, I opted not to do this in this PR as I don't know what consequences it will bring with it.

Because I can't proceed without this, I have built my own Docker image based in this PR to use as a workaround until this issue is addressed.

### DIFF
--- a/src/EventStore.ClusterNode/Program.cs
+++ b/src/EventStore.ClusterNode/Program.cs
@@ -63,21 +63,24 @@ namespace EventStore.ClusterNode {
 									if (hostedService.Node.DisableHttps) {
 										listenOptions.Use(next => new ClearTextHttpMultiplexingMiddleware(next).OnConnectAsync);
 									} else {
-										listenOptions.UseHttps(new HttpsConnectionAdapterOptions {
-											ServerCertificateSelector = delegate {
+										listenOptions.UseHttps(options => {
+											options.ServerCertificateSelector = delegate {
 												return hostedService.Node.CertificateSelector();
-											},
-											ClientCertificateMode = ClientCertificateMode.AllowCertificate,
-											ClientCertificateValidation = (certificate, chain, sslPolicyErrors) => {
-												var (isValid, error) =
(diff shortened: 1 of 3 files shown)


hayley-jean:
Thanks for the PR @BrunoZell!
We're still discussing this fix internally, but haven't been able to reproduce the error.
Could you provide us with some more information about your setup and the issue itself?

1. Could you post a screenshot of the pop up in Chrome?
2. What OS are you running on and are you using a VPN?
3. If possible, could you provide us with a print out of the environment variables in your command line when running the `curl` command?

ylorph:
We're working on a new UI that will not have that issue.

PR #2741, opened 2020-11-04
Experiment with raft terms

shaan1337:
Fixes https://github.com/EventStore/home/issues/263

Paxos `View numbers` and `Epoch numbers` used in EventStore are very similar to `Election Terms` in the raft consensus algorithm.

This PR is an experiment to:
- Treat a paxos `view number` like a `term` since, like with raft, at most one leader can be elected for a certain view number (term)
- Treat an `epoch number` like a `term`. Any data that comes after an epoch record having epoch number X has term X (till the next epoch record).
- Note: This breaks the continuity of epoch numbers but monotonicity of epoch numbers should be preserved (similar to raft terms)
- Persisting the term checkpoint whenever we see a new view ensures that any newly elected leader will have a larger term (epoch number) than the previous leader.
-  In raft, the node having the highest (term, index) is considered to have the latest data. Here, the node with the (highest epoch number, writer checkpoint) can be considered to have the latest data.


condron:
Superseded by #2745

PR #2742, opened 2020-11-04
Prioritize leader selection in elections by Commit Checkpoint

condron:
Fixed: EventStore/home#263

Close edge case for isolated deposed leaders being preferred in elections due to an artificially high epoch count.

The prioritizing by commit checkpoint correctly identifies the instance with the most current client acknowledge data.

The tests have been enhanced to add support for `epoch number` and differing `last leader` values. This allows recreating the exact scenarios with differing epochs and the disagreement on the last known quorum leader that occurs when isolated nodes rejoin.

Note  this approach also avoids the performance impact of persisting the replication checkpoint


condron:
superseded by #2745

PR #2744, opened 2020-11-06
Add LegacyGossipDiscovery

megakid:
Ideally the `discover://` protocol would stay non-TLS (HTTP) and we'd see the introduction of a `discovers://` (discover secure) scheme that uses HTTPS but I think that would be a breaking change now.

This is next best thing, non-breaking, `discover://` can work with non-TLS gossip endpoints.

This enables a migration path whereby a v20 client can use v5 clusters via `discover://` with non-TLS.

### DIFF
--- a/src/EventStore.ClientAPI.Tests/endpoint_discoverer.cs
+++ b/src/EventStore.ClientAPI.Tests/endpoint_discoverer.cs
@@ -36,7 +36,7 @@ namespace EventStore.ClientAPI.Tests {
 
 			var sut = new ClusterDnsEndPointDiscoverer(new ConsoleLogger(), "dns", 1, 1111,
 				new[] {new GossipSeed(new IPEndPoint(IPAddress.Any, 2113))}, TimeSpan.FromSeconds(1),
-				nodePreference,
+				nodePreference, false,
 				new TestMessageHandler(response.ToJson()));
 
 			var result = await sut.DiscoverAsync(new IPEndPoint(IPAddress.Any, 1113));
--- a/src/EventStore.ClientAPI.Tests/endpoint_discoverer.cs
+++ b/src/EventStore.ClientAPI.Tests/endpoint_discoverer.cs
@@ -37,6 +37,7 @@ namespace EventStore.ClientAPI.Tests {
 			var sut = new ClusterDnsEndPointDiscoverer(new ConsoleLogger(), "dns", 1, 1111,
 				new[] {new GossipSeed(new IPEndPoint(IPAddress.Any, 2113))}, TimeSpan.FromSeconds(1),
 				nodePreference, false,
+				false,
 				new TestMessageHandler(response.ToJson()));
 
 			var result = await sut.DiscoverAsync(new IPEndPoint(IPAddress.Any, 1113));
--- a/src/EventStore.ClientAPI/ConnectionSettings.cs
+++ b/src/EventStore.ClientAPI/ConnectionSettings.cs
@@ -21,6 +21,7 @@ namespace EventStore.ClientAPI {
 			get { return DefaultSettings.Value; }
 		}
 
+
 		/// <summary>
 		/// Creates a new set of <see cref="ConnectionSettings"/>.
 		/// </summary>
@@ -130,6 +131,12 @@ namespace EventStore.ClientAPI {
 		public readonly int MaxDiscoverAttempts;
 
 		/// <summary>
+		/// If true, use non-TLS for discover:// URIs
+		/// Useful for backwards compatibility with v5 clusters
+		/// </summary>
+		public bool InsecureDiscoverGossip;
+
+		/// <summary>
 		/// The well-known endpoint on which cluster nodes are running.
 		/// </summary>
 		public readonly int GossipPort;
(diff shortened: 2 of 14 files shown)


megakid:
I am happy to do the work on the alternative suggestion - let me know if that would be an acceptable solution (e.g. creating `discovers://`).

shaan1337:
@megakid hey, thanks very much for your PR! the general idea looks good to me, i'll test it and keep you updated.

Just thinking out loud: would `InsecureDiscover` be a good name for the option (which is a bit less verbose than `InsecureDiscoverGossip`) ? what do you think @megakid @YoEight ?

shaan1337:
tested and works with the following client connection settings (server = 5.0.9 (2 out of 3-node cluster), client = this branch):

Connection string:
```
EventStoreConnection.Create("ConnectTo=discover://admin:changeit@<cluster dns>:2113;InsecureDiscoverGossip=True;UseSslConnection=false")
```

Connection settings builder:
```
var settings = ConnectionSettings.Create().DisableTls().SetInsecureDiscoverGossip(true);
var clusterSettings = ClusterSettings.Create().DiscoverClusterViaDns().SetClusterDns("<cluster dns>");
EventStoreConnection.Create(settings,clusterSettings)
```

megakid:
> @megakid hey, thanks very much for your PR! the general idea looks good to me, i'll test it and keep you updated.
> 
> Just thinking out loud: would `InsecureDiscover` be a good name for the option (which is a bit less verbose than `InsecureDiscoverGossip`) ? what do you think @megakid @YoEight ?

`InsecureDiscover` sounds good to me.

hayley-jean:
Hi @megakid, sorry for the delay
We're still discussing the naming of this internally because we want to start aligning the clients with each other

megakid:
Thanks @hayley-jean

hayley-jean:
Hi @megakid, thanks for waiting!

After some discussion, we've decided that we'd prefer for this option to be named `LegacyGossipDiscovery` as this hopefully conveys that this should be used to connect to v5.x nodes from the v20.x client. When we're free to break the client in the next major version we can consider adding the `discover` and `discovers` protocols to replace the option like you suggested.

megakid:
Hi @hayley-jean - Thanks for the response, I've renamed the option as requested.

shaan1337 (commented):
The changes no longer seem to work after #2753.
Inspecting the network traces show that it's failing with `Bad Request (Invalid host)` due to the `Host` header being present and it seems that the Mono HTTP server does not accept incoming requests to unknown hosts.

Example request:


```
curl -v http://clusterdnstest.eventstore.com:2112/gossip\?format\=json
*   Trying 127.0.0.3...
* Connected to clusterdnstest.eventstore.com (127.0.0.3) port 2112 (#0)
> GET /gossip?format=json HTTP/1.1
> Host: clusterdnstest.eventstore.com:2112
> User-Agent: curl/7.47.0
> Accept: */*
> 
< HTTP/1.1 400 Bad Request
< Content-Type: text/html; charset=utf-8
< Server: Mono-HTTPAPI/1.0
< Date: Wed, 06 Jan 2021 06:41:07 GMT
< Content-Length: 35
< Connection: close
< 
* Closing connection 0
<h1>Bad Request (Invalid host)</h1>% 
```

megakid:
@shaan1337 are you saying that it works with `https://` but not `http://` ?  The only thing this PR does is allow http calls to the gossip (now unresolved) endpoint.

Would a simple fix would be to remove the `Host` header here?  It isn't needed any more and should have been removed #2753 right?

YoEight (commented):
Not conflicting with #2759

shaan1337:
@megakid The issue should be present with `https` too. It's not an issue with your PR but instead it looks like a regression (when running against v5 server) due to the way DNS discovery has changed with #2753: a direct HTTP request is now made to http://cluster.dns/gossip?format=json letting the http client select any IP address (but the http client will keep the host header). Previously we were doing the DNS resolution ourselves and choosing an IP address to connect with.

shaan1337 (commented):
@YoEight I've tested your latest changes and it's working :) 

CI is just complaining about the following missing XML doc comments on the `EnableVersion5Compatibility` setting:
```
Error: /home/runner/work/EventStore/EventStore/src/EventStore.ClientAPI/ConnectionSettings.cs(164,24): error CS1591: Missing XML comment for publicly visible type or member 'ConnectionSettings.EnableVersion5Compability' [/home/runner/work/EventStore/EventStore/src/EventStore.ClientAPI/EventStore.ClientAPI.csproj]
```
```
Error: D:\a\EventStore\EventStore\src\EventStore.ClientAPI\Transport.Http\HttpAsyncClient.cs(30,90): error CS1573: Parameter 'enableVersion5Compability' has no matching param tag in the XML comment for 'HttpAsyncClient.HttpAsyncClient(TimeSpan, HttpMessageHandler, bool)' (but other parameters do) [D:\a\EventStore\EventStore\src\EventStore.ClientAPI\EventStore.ClientAPI.csproj]
```

shaan1337 (approved):
Changes look good to me :+1: thanks @megakid and @YoEight !

megakid:
The Host issue, to which you've added `EnableVersion5Compability`, is that only present on Mono 5.x servers? or does it also affect .NET Framework 5.x servers?

hayley-jean (approved):

hayley-jean:
> The Host issue, to which you've added `EnableVersion5Compability`, is that only present on Mono 5.x servers? or does it also affect .NET Framework 5.x servers?

The v5.x Event Store server has different requirements for the host header to the v20.x Event Store server, so `EnableVersion5Compability` just sets this correctly for the v5 server. We've added an issue to clarify this option in [here](https://github.com/EventStore/EventStore/issues/2794)

PR #2745, opened 2020-11-09
Add checkpoint tracking for proposed epochs 

condron:
Fixed: EventStore/home#263

Add checkpoint based tracking of proposed epoch numbers on each node in addition to the epoch records in the replicated Log.

Close edge case for isolated deposed leaders being preferred in elections due to subsequent quorums being able to form with a lower epoch number in the case replication is disrupted during a series of election cycles.

More complete fix than #2742 as theoretically possible further edge cases are also closed.
Aligns with paxos terminology better than #2741 

### DIFF
--- a/src/EventStore.Core.Tests/Services/ElectionsService/ElectionServiceUnit.cs
+++ b/src/EventStore.Core.Tests/Services/ElectionsService/ElectionServiceUnit.cs
@@ -18,6 +18,7 @@ namespace EventStore.Core.Tests.Services.ElectionsService {
 		private const int LastCommitPosition = -1;
 		private const int WriterCheckpoint = 0;
 		private const int ChaserCheckpoint = 0;
+		private const int ProposalCheckpoint = -1;
 		private static readonly DateTime InitialDate = new DateTime(2012, 6, 1);
 
 		public ClusterInfo ClusterInfo { get; private set; }
@@ -54,6 +55,7 @@ namespace EventStore.Core.Tests.Services.ElectionsService {
 				clusterSettings.ClusterNodesCount,
 				new InMemoryCheckpoint(WriterCheckpoint),
 				new InMemoryCheckpoint(ChaserCheckpoint),
+				new InMemoryCheckpoint(ProposalCheckpoint),
 				new FakeEpochManager(),
 				() => -1, 0, new FakeTimeProvider());
 			ElectionsService.SubscribeMessages(_bus);
--- a/src/EventStore.Core.Tests/Services/ElectionsService/FakeEpochManager.cs
+++ b/src/EventStore.Core.Tests/Services/ElectionsService/FakeEpochManager.cs
@@ -35,30 +35,34 @@ namespace EventStore.Core.Tests.Services.ElectionsService {
 			}
 		}
 
-		public EpochRecord GetEpoch(int epochNumber, bool throwIfNotFound) {
+		public EpochRecord GetEpochAfter(int epochNumber, bool throwIfNotFound) {
 			lock (_epochs) {
 				var epoch = _epochs.FirstOrDefault(e => e.EpochNumber == epochNumber);
+				if (epoch != null) {
+					var index = _epochs.IndexOf(epoch);
+					epoch = null;
+					if (index + 1 < _epochs.Count) {
+						epoch = _epochs[index + 1];
+					}
+				}
+
 				if (throwIfNotFound && epoch == null)
 					throw new ArgumentOutOfRangeException(nameof(epochNumber), "Epoch not Found");
 				return epoch;
 			}
 		}
 
-		public EpochRecord GetEpochWithAllEpochs(int epochNumber, bool throwIfNotFound) {
-			lock (_epochs) {
-				return GetEpoch(epochNumber, throwIfNotFound);
-			}
-		}
+		
 
 		public bool IsCorrectEpochAt(long epochPosition, int epochNumber, Guid epochId) {
 			throw new NotImplementedException();
 		}
-
-		public void WriteNewEpoch() {
+		
+		public void WriteNewEpoch(int epochNumber) {
 			throw new NotImplementedException();
 		}
 
-		public void SetLastEpoch(EpochRecord epoch) {
+		public void CacheEpoch(EpochRecord epoch) {
 			lock (_epochs) {
 				_epochs.Add(epoch);
 			}
--- a/src/EventStore.Core.Tests/Services/ElectionsService/LeaderNode/ElectionsServiceUnitTests.cs
+++ b/src/EventStore.Core.Tests/Services/ElectionsService/LeaderNode/ElectionsServiceUnitTests.cs
@@ -44,6 +44,7 @@ namespace EventStore.Core.Tests.Services.ElectionsService {
 
 				var writerCheckpoint = new InMemoryCheckpoint();
 				var readerCheckpoint = new InMemoryCheckpoint();
+				var proposalCheckpoint = new InMemoryCheckpoint(-1);
 				var epochManager = new FakeEpochManager();
 				Func<long> lastCommitPosition = () => -1;
 				var electionsService = new Core.Services.ElectionsService(outputBus,
@@ -51,6 +52,7 @@ namespace EventStore.Core.Tests.Services.ElectionsService {
 					3,
 					writerCheckpoint,
 					readerCheckpoint,
+					proposalCheckpoint,
 					epochManager,
 					() => -1, 0, new FakeTimeProvider());
 				electionsService.SubscribeMessages(inputBus);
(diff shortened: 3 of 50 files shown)


shaan1337:
@condron  i don't think we can preserve continuity of epoch numbers but the PR seems to assume epoch numbers are always incremented by one.

here's a counter-example:

the last epoch number is 4:
1. leader of elections is about to propose value 5, so it writes a checkpoint with value 5
2. leader of elections proposes 5
3. the proposal message is lost and it never gets a quorum number of accepts for 5
4. another round of elections starts
5. the same node becomes leader of elections
6. leader of elections proposes 6 (proposal checkpoint + 1)
7. a quorum number of accepts is received
8.  the elected leader writes epoch 6

so in the transaction log, we will have epoch record 4 followed by 6.

this will cause issues when a follower subscribes to the leader here:
https://github.com/EventStore/EventStore/pull/2741/files#diff-439e649bbdaa13b776087135000b10fe10cad91a2d7b09797a97c06ec0f6eac9L263

and that's the reason I had to add a GetNextEpoch() method here to get the next epoch:
https://github.com/EventStore/EventStore/pull/2741/files#diff-ffcebde92419b1e6850e9cbc27c362d75c6791d7e74e413dd0f1e3c142281763R170

shaan1337:
@condron I think that the `LastCommitPosition` should be removed from GetBestLeaderCandidate() as done in #2741 :

```
var best = received.Values
	.OrderByDescending(x => x.EpochNumber)
	.ThenByDescending(x => x.WriterCheckpoint)
	.ThenByDescending(x => x.ChaserCheckpoint)
	.ThenByDescending(x => x.NodePriority)
	.ThenByDescending(x => x.ServerId)
	.FirstOrDefault();
```

In raft, any node having the highest (term, index) is considered to have the latest data. So in our case, it should be (epoch number, writer checkpoint).

Rationale:
We ensure that in any subset of size Q (where Q is the quorum number),  any node having the highest epoch number has been connected to the latest leader.

We don't actually care if some uncommitted data is considered as committed (a client must assume that it is possible a write has been done if there was no response and it will be treated idempotently if retried) but we don't want any committed data to be considered as uncommitted.

So, when we choose the node having the highest (epoch number, writer checkpoint) we ensure that we're selecting the node that has been connected to the latest leader and that has all data replicated to at least a quorum number of nodes (due to the overlapping majorities principle) + possibly some unreplicated data.

src/EventStore.Core.Tests/TransactionLog/TFChunkHelper.cs
```diff
@@ -9,7 +9,10 @@ namespace EventStore.Core.Tests.TransactionLog {
 	public static class TFChunkHelper {
 		public static TFChunkDbConfig CreateDbConfig(string pathName, long writerCheckpointPosition,
 			long chaserCheckpointPosition = 0,
-			long epochCheckpointPosition = -1, long truncateCheckpoint = -1, int chunkSize = 10000,
+			long epochCheckpointPosition = -1,
+			long epochNumberCheckpointPosition = -1,
```
shaan1337, line 13: Should this be `proposalCheckpointPosition` for consistency?

src/EventStore.Core/Services/ElectionsService.cs
```diff
@@ -104,11 +107,18 @@ public class ElectionsService : IHandle<SystemMessage.BecomeShuttingDown>,
...
 			_writerCheckpoint = writerCheckpoint;
 			_chaserCheckpoint = chaserCheckpoint;
+			_proposalCheckpoint = proposalCheckpoint;
 			_epochManager = epochManager;
 			_getLastCommitPosition = getLastCommitPosition;
 			_nodePriority = nodePriority;
 			_timeProvider = timeProvider;
 
+			var lastEpoch = _epochManager.LastEpochNumber;
+			if (_proposalCheckpoint.Read() < lastEpoch) {
```
shaan1337, line 117: _proposalCheckpoint.Read() can be replaced with _proposalCheckpoint.ReadNonFlushed() everywhere since the value is cached in memory.

src/EventStore.Core/Services/Storage/EpochManager/EpochManager.cs
```diff
@@ -203,7 +203,8 @@ public class EpochManager : IEpochManager {
 			// and update EpochManager's state, by adjusting cache of records, epoch count and un-caching 
 			// excessive record, if present.
 			// If we are writing the very first epoch, last position will be -1.
-			var epoch = WriteEpochRecordWithRetry(_lastEpochNumber + 1, Guid.NewGuid(), _lastEpochPosition, _instanceId);
+			
+			var epoch = WriteEpochRecordWithRetry(epochNumber, Guid.NewGuid(), _lastEpochPosition, _instanceId);
```
shaan1337, line 207: Should we add a pre-condition here that epochNumber > _lastEpochNumber and throw otherwise? (to make sure we catch any bugs)

src/EventStore.Core.Tests/Services/ElectionsService/ElectionServiceUnit.cs
```diff
@@ -54,6 +54,7 @@ public class ElectionsServiceUnit {
 				clusterSettings.ClusterNodesCount,
 				new InMemoryCheckpoint(WriterCheckpoint),
 				new InMemoryCheckpoint(ChaserCheckpoint),
+				new InMemoryCheckpoint(ChaserCheckpoint),
```
shaan1337, line 57: Should this be `ProposalCheckpoint`?

src/EventStore.Core.Tests/Services/ElectionsService/FakeEpochManager.cs
```diff
@@ -35,26 +35,30 @@ internal class FakeEpochManager : IEpochManager {
 			}
 		}
 
-		public EpochRecord GetEpoch(int epochNumber, bool throwIfNotFound) {
+		public EpochRecord GetEpochAfter(int epochNumber, bool throwIfNotFound) {
 			lock (_epochs) {
 				var epoch = _epochs.FirstOrDefault(e => e.EpochNumber == epochNumber);
+				if (epoch != null) {
+					var index = _epochs.IndexOf(epoch);
```
shaan1337, line 42: if the epoch is not found, indexOf will return -1 and we should throw an exception here.

src/EventStore.Core/Services/ElectionsService.cs
```diff
@@ -423,14 +456,16 @@ public class ElectionsService : IHandle<SystemMessage.BecomeShuttingDown>,
...
-				.OrderByDescending(x => x.EpochNumber)
-				.ThenByDescending(x => x.LastCommitPosition)
-				.ThenByDescending(x => x.WriterCheckpoint)
-				.ThenByDescending(x => x.ChaserCheckpoint)
-				.ThenByDescending(x => x.NodePriority)
-				.ThenByDescending(x => x.ServerId)
+				.OrderByDescending(x => x.EpochNumber) // latest election
+				.ThenByDescending(x => x.WriterCheckpoint) // log length (if not re-electing leader pick the most complete replica)
+				.ThenByDescending(x => x.ChaserCheckpoint) // in memory index position
+				.ThenByDescending(x => x.LastCommitPosition) // on disk index position
```
shaan1337, line 464: the `LastCommitPosition` is not stored on disk - it is based on the index checkpoint which is updated only after each PTable dump. on startup, the index checkpoint is rebuilt up to the chaser checkpoint so it may be initialized to a value pointing to uncommitted data.

src/EventStore.Core/Services/Storage/EpochManager/EpochManager.cs
```diff
@@ -122,47 +112,79 @@ public class EpochManager : IEpochManager {
 		public EpochRecord[] GetLastEpochs(int maxCount) {
 			lock (_locker) {
 				var res = new List<EpochRecord>();
-				for (int epochNum = _lastEpochNumber, n = maxCount; epochNum >= 0 && n > 0; --epochNum, --n) {
-					EpochRecord epoch;
-					if (!_epochs.TryGetValue(epochNum, out epoch))
-						break;
-					res.Add(epoch);
+				var node = _epochs.Last;
+				while (node != null & res.Count < maxCount) {
```
shaan1337, line 116:
```suggestion
				while (node != null && res.Count < maxCount) {
```

src/EventStore.Core/Services/Storage/EpochManager/EpochManager.cs
```diff
@@ -177,33 +198,30 @@ public class EpochManager : IEpochManager {
...
-			//_checkpoint.Flush();
-
+		public void WriteNewEpoch(int epochNumber) {
 			// Now we write epoch record (with possible retry, if we are at the end of chunk) 
 			// and update EpochManager's state, by adjusting cache of records, epoch count and un-caching 
 			// excessive record, if present.
 			// If we are writing the very first epoch, last position will be -1.
-			var epoch = WriteEpochRecordWithRetry(_lastEpochNumber + 1, Guid.NewGuid(), _lastEpochPosition, _instanceId);
+			if (epochNumber < 0) {
+				throw new ArgumentException($"Cannot add write Epoch with a negative Epoch Number {epochNumber}.",nameof(epochNumber));
```
shaan1337, line 215:
```suggestion
				throw new ArgumentException($"Cannot write Epoch with a negative Epoch Number {epochNumber}.",nameof(epochNumber));
```

src/EventStore.Core/Services/Storage/EpochManager/EpochManager.cs
```diff
@@ -177,33 +198,30 @@ public class EpochManager : IEpochManager {
...
+			}
+			if (epochNumber < LastEpochNumber) {
+				throw new ArgumentException($"Cannot add write Epoch with a Epoch Number {epochNumber} lower than Last Epoch Number {LastEpochNumber}.",nameof(epochNumber));
```
shaan1337, line 218:
```suggestion
				throw new ArgumentException($"Cannot add Epoch with an Epoch Number {epochNumber} lower than Last Epoch Number {LastEpochNumber}.",nameof(epochNumber));
```

src/EventStore.Core/Services/Storage/EpochManager/EpochManager.cs
```diff
@@ -177,33 +198,30 @@ public class EpochManager : IEpochManager {
...
+			}
+			if (epochNumber <= LastEpochNumber) {
+				throw new Exception($"Concurrency failure, Last Epoch Number is {LastEpochNumber} cannot add write new Epoch with a lower Epoch Number {epochNumber}.");
+			}
```
shaan1337, line 222: @condron  The purpose of this `if` statement is unclear, could you please clarify? If assuming that there could be any concurrency issue it could still happen after this if statement and before we call `WriteEpochRecordWithRetry`.

src/EventStore.Core/Services/Storage/EpochManager/EpochManager.cs
```diff
@@ -230,48 +248,49 @@ public class EpochManager : IEpochManager {
...
-				_checkpoint.Write(epoch.EpochPosition);
-				_checkpoint.Flush();
-
-				Log.Debug(
-					"=== Update Last Epoch E{epochNumber}@{epochPosition}:{epochId:B} (previous epoch at {lastEpochPosition}) L={leaderId:B}.",
-					epoch.EpochNumber, epoch.EpochPosition, epoch.EpochId, epoch.PrevEpochPosition, epoch.LeaderInstanceId);
+				if (!_epochs.Contains(epoch)) {
+					_epochs.AddLast(epoch);
+					_lastCachedEpoch = _epochs.Last;
+					_epochs.RemoveFirst();
```
shaan1337, line 280: @condron We should remove an item only if the linked list size was exactly equal to _cacheSize

src/EventStore.Core/Services/Storage/EpochManager/EpochManager.cs
```diff
@@ -122,47 +112,78 @@ public class EpochManager : IEpochManager {
...
+				while (epochNode != null && epochNode.Value.EpochNumber != epochNumber) {
+					epochNode = epochNode.Previous;
 				}
 
-				EpochRecord epoch;
-				if (!_epochs.TryGetValue(epochNumber, out epoch) && throwIfNotFound)
-					throw new Exception(string.Format("Concurrency failure, epoch #{0} should not be null.",
-						epochNumber));
-				return epoch;
+				epoch = epochNode?.Next?.Value;
```
shaan1337, line 140: GetEpochAfter() currently returns `null` even if the value of `epoch` here is not null.

src/EventStore.Core/Services/ElectionsService.cs
```diff
@@ -661,6 +722,7 @@ public class LeaderCandidate {
 			public readonly Guid InstanceId;
 			public readonly EndPoint HttpEndPoint;
 
+			public int ProposalNumber;
```
shaan1337, line 725: @condron During testing, there are cases where ProposalNumber is having a value of 0 after elections are done. It is probably not being set when `LeaderCandidate` is created (since it's not part of the constructor). for safety we should pass `proposalNumber` in the constructor and make it immutable.

src/EventStore.Core/Services/Storage/EpochManager/EpochManager.cs
condron: These if statements are just to give better exception messages, I've reworded the section.

src/EventStore.Core/Services/ElectionsService.cs
condron: We do not Accept masters without the correct epoch and data, this selection here is just for picking the node with the least work to become and active master.

src/EventStore.Core.Tests/Services/ElectionsService/FakeEpochManager.cs
condron: we're inside a lock on the collection and we just found this epoch by searching the collection, so we can trust that it is still there.

src/EventStore.Core/Services/ElectionsService.cs
condron: Fixed

hayley-jean (approved):

src/EventStore.Core/Services/Storage/EpochManager/EpochManager.cs
```diff
@@ -177,101 +211,139 @@ public class EpochManager : IEpochManager {
...
-					"=== Update Last Epoch E{epochNumber}@{epochPosition}:{epochId:B} (previous epoch at {lastEpochPosition}) L={leaderId:B}.",
+				
+				// if we have a first or last and either are equal it's already cached, just return
+				if ((_epochs.Last?.Value.EpochNumber ?? -1) == epoch.EpochNumber ||
+					(_epochs.First?.Value.EpochNumber ?? -1) == epoch.EpochNumber) { return; } 
+
+				//new last epoch written or received, this is the normal case
+				//if the list is empty Last will be null
+				if (_epochs.Last == null || _epochs.Last.Value.EpochNumber < epoch.EpochNumber) {
+					_epochs.AddLast(epoch);
```
shaan1337, line 286: @condron _cacheSize is not being respected here - items should be removed from the head of the list if size > cacheSize

src/EventStore.Core/Services/Storage/EpochManager/EpochManager.cs
```diff
@@ -177,101 +211,139 @@ public class EpochManager : IEpochManager {
...
+					Log.Debug(
+						"=== Cached Epoch E{epochNumber}@{epochPosition}:{epochId:B} (previous epoch at {lastEpochPosition}) L={leaderId:B}.",
+						epoch.EpochNumber, epoch.EpochPosition, epoch.EpochId, epoch.PrevEpochPosition, epoch.LeaderInstanceId);
+					return;
+				}
+
+				//List insert edge case, this is likely a race condition near the last epoch
+				var e = _epochs.Last;
+				while (e?.Previous != null) {
+					if (e.Previous.Value.EpochNumber < epoch.EpochNumber) {
```
shaan1337, line 313: @condron duplicates are not being verified here -  the same epoch record can be added twice to the list

src/EventStore.Core/Services/Storage/EpochManager/EpochManager.cs
condron: addressed

src/EventStore.Core/Services/Storage/EpochManager/EpochManager.cs
condron: addressed

src/EventStore.Core/Services/Storage/EpochManager/EpochManager.cs
```diff
@@ -177,101 +213,139 @@ public class EpochManager : IEpochManager {
...
+			var added = AddEpochToCache(epoch);
 			// Epoch record must have been already written, so we need to make sure it is where we expect it to be.
 			// If this check fails, then there is something very wrong with epochs, data corruption is possible.
-			if (!IsCorrectEpochAt(epoch.EpochPosition, epoch.EpochNumber, epoch.EpochId)) {
-				throw new Exception(string.Format("Not found epoch at {0} with epoch number: {1} and epoch ID: {2}. "
-				                                  + "SetLastEpoch FAILED! Data corruption risk!",
-					epoch.EpochPosition,
-					epoch.EpochNumber,
-					epoch.EpochId));
+			if (added && !IsCorrectEpochAt(epoch.EpochPosition, epoch.EpochNumber, epoch.EpochId)) {
```
shaan1337, line 270:
```suggestion
			if (!added && !IsCorrectEpochAt(epoch.EpochPosition, epoch.EpochNumber, epoch.EpochId)) {
```

@condron should it be `!added` ? the previous code was doing this check if it was not added to the cache.

condron: no, this should check only if this is a new item we did not write in this epoch manager.

src/EventStore.Core.Tests/Services/ElectionsService/LeaderNode/ElectionsServiceUnitTests.cs
```diff
@@ -235,33 +275,34 @@ public class ChoosingLeaderTests {
...
+					nodePriority, epochNumber, previousLeaderId, clusterInfo);
 				prepareOks.Add(prepareOk.ServerId, prepareOk);
 			}
 
-			var lastElectedLeader = tc.LastElectedLeader.HasValue
-				? (Guid?)IdForNode(tc.LastElectedLeader.Value)
-				: null;
+			//var lastElectedLeader = tc.LastElectedLeader.HasValue
+			//	? (Guid?)IdForNode(tc.LastElectedLeader.Value)
+			//	: null;
```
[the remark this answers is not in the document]
condron: updated

shaan1337 (approved):
@condron great work, the changes look good to me :+1:

hayley-jean (approved):

Issue #2549, opened 2020-06-11
RemoteCertificateNameMismatch log entry is not as helpful as it could be


jen20:
A mismatch between the requested hostname and the SANs of a TLS certificate results in:

```
node2_1  | [    1,18,17:56:08.727,ERR] Server certificate validation error: "The certificate provided by the server failed validation with the following error(s): RemoteCertificateNameMismatch (NoError)"
```

It is not straightforward to diagnose what the problematic name is - we should include it in this error message.

kpfaulkner:
I am currently hitting against this as well (3 node cluster on Azure VMs). Have been hunting about for documentation and have resorted to going through the EventStore source but no luck yet. Is there any guidance on this?

alexeyzimarev:
@kpfaulkner the certificate you use for the nodes should contain the IP address and/or DNS name of the node, so it matches with the gossip response (`https://<node>:<http-port>/gossip?format=json`). Also, we expect the nodes to have the same CN for the certificate (`eventstoredb-node`), otherwise, you need to set the `CertificateReservedNodeCommonName` option to the CN of your certificate.

kpfaulkner:
Sorry if this seems a silly question... but does the single cert have all 3 (in my case 3 nodes) IPs in it, or are there 3 separate certificates each with their own IPs? I'd tried a cert with 1 of the IPs (these are internal IPs), also certs with multiple IPs. Still no luck.

Will keep trying.

kpfaulkner:
Well, I suppose its progress.  I can confirm I can hit the gossip endpoint manually (from one machine to another)...  but when I run ES I get a never ending repeat of:

[ 5188,25,01:05:32.071,WRN] Failed authorization check for "(anonymous)" in 00:08:41.5050551 with "node/gossip : update Deny : Policy : Legacy 1 12/31/9999 11:59:59 PM +00:00 : default:denied by default:Deny, $"
[ 5188,11,01:05:32.620,INF] ELECTIONS: (V=516) TIMED OUT! (S=ElectingLeader, M=null).
[ 5188,11,01:05:32.620,INF] ELECTIONS: (V=517) SHIFT TO LEADER ELECTION.
[ 5188,11,01:05:32.620,INF] ELECTIONS: (V=517) VIEWCHANGE FROM ["Unspecified/10.1.2.101:2112", {29d30865-5643-48b9-b274-ee5c72ff9f11}].
[...]

PR #2746, opened 2020-11-12
Adds X509Certificate.Subject to log entry for debugging RemoteCertificateNameMismatch error

dyowee:
Changed: Add the certificate subject to the log message printed when there is a certificate validation error

Related to issue #2549 (RemoteCertificateNameMismatch log entry is not as helpful as it could be)


YoEight (approved):

shaan1337 (approved):
tested, looks good to me :+1: 

thanks @dyowee !

PR #2755, opened 2020-12-14
Expose option for stream info cache capacity

hayley-jean:
Adds option `--stream-info-cache-capacity` to allow setting the cache capacity of the ReadIndex.


jageall (approved):

condron (approved):
looks good

PR #2757, opened 2020-12-16
Introduce low latency heartbeats by relying on TCP ACKs

shaan1337:
Added: Introduce low latency TCP heartbeats by relying on TCP ACKs instead of waiting for the remote party to process heartbeat requests

Fixes: https://github.com/eventstore/home/issues/312

The way heartbeats currently work between nodes is as follows:
- If we have not *received* any data from the remote party for `heartbeat interval` seconds, we enqueue a `heartbeat request` and schedule a heartbeat timeout message locally (Note that we may have *sent* a lot of data in the mean time)
- This `heartbeat request` is enqueued onto the send queue and when it is dequeued (after possibly a lot of other data), it is sent to the remote party over the TCP connection.
- The remote party receives the `heartbeat request` as raw data, enqueues it on the receive queue, dequeues it (after possibly dequeueing a lot of other data), unframes the raw data into a heartbeat request command, processes it and prepares a `heartbeat response` message which is now enqueued on the remote party's send queue and sent.
- If the local party receives the heartbeat response or any other data sent by the remote party before the heartbeat timeout check is triggered, everything is fine otherwise it results in a `HEARTBEAT TIMEOUT` and the TCP connection is closed by the local party.

The above scenario works fine when a connection is idle (there's not much send and receive traffic) or when we are mostly receiving traffic. However, in the case where we are sending a lot of data but not receiving much (e.g on leader nodes), it poses the following problem:
- The `heartbeat request` may need to wait behind a lot of other data on the send queue on the local side before being sent
- The `heartbeat request` may need to wait behind a lot of other data on the receive queue on the remote side before being processed

This PR changes the way heartbeats work by relying on TCP ACKs to determine if the remote party is alive:
- As soon as we have successfully *sent* any data to the remote party, be it the `heartbeat request` or any other data, it implies that we must have received a TCP ACK message from the remote party and we can consider that the remote party is alive.
- Note: We no longer even need to send a heartbeat response for node to node communication but it's been left there for backwards compatibility + it is needed for TCP clients.
- This has been implemented by adding a `TotalAcknowledgedBytes` property on `ITcpConnection` which is equal to `TotalBytesSent` + `TotalBytesReceived` on the TCP connection. If this value changes between a heartbeat request and the heartbeat timeout check, it implies that the node is alive.

### DIFF
--- a/src/EventStore.Core.Tests/Services/Transport/Tcp/TcpConnectionManagerTests.cs
+++ b/src/EventStore.Core.Tests/Services/Transport/Tcp/TcpConnectionManagerTests.cs
@@ -265,6 +265,8 @@ namespace EventStore.Core.Tests.Services.Transport.Tcp {
 			get { return _clientConnectionName; }
 		}
 
+		public long TotalAcknowledgedBytes { get; }
+
 		public bool IsClosed {
 			get { return false; }
 		}
--- a/src/EventStore.Core/Messages/TcpMessage.cs
+++ b/src/EventStore.Core/Messages/TcpMessage.cs
@@ -32,9 +32,9 @@ namespace EventStore.Core.Messages {
 				get { return TypeId; }
 			}
 
-			public readonly int MessageNumber;
+			public readonly long MessageNumber;
 
-			public Heartbeat(int messageNumber) {
+			public Heartbeat(long messageNumber) {
 				MessageNumber = messageNumber;
 			}
 		}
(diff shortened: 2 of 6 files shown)


shaan1337:
Closing this PR as the solution is flawed. This statement is not true:
```
As soon as we have successfully sent any data to the remote party, be it the heartbeat request or any other data, it implies that we must have received a TCP ACK message from the remote party and we can consider that the remote party is alive.
```

A send() on the socket API immediately returns when the data has been put on the outgoing queue - it does not wait for TCP ACKs. Info about acknowledged bytes is not exposed at the application level.

gregoryyoung:
TCP acks are not available through the C# APIs it is as you say ... when its queued which is often.

This is why heartbeats are implemented on increments at a protocol level.

There are some other really weird pathological edge cases around this ...

PR #2759, opened 2020-12-16
Backport DNS patch in v5 ClientAPI.

YoEight:
Fixed: [V5] Allow DNS seed resolution so it works with EventStoreDB Cloud.

### DIFF
--- a/src/EventStore.ClientAPI/ConnectionSettingsBuilder.cs
+++ b/src/EventStore.ClientAPI/ConnectionSettingsBuilder.cs
@@ -306,7 +306,25 @@ namespace EventStore.ClientAPI {
 		/// <exception cref="ArgumentNullException">If <paramref name="clusterDns" /> is null or empty.</exception>
 		public ConnectionSettingsBuilder SetClusterDns(string clusterDns) {
 			Ensure.NotNullOrEmpty(clusterDns, "clusterDns");
-			_clusterDns = clusterDns;
+
+			var splits = clusterDns.Split(':');
+
+			switch (splits.Length) {
+				case 1:
+					_clusterDns = string.Format("{0}:2113", clusterDns);
+					break;
+				case 2:
+					if (int.TryParse(splits[1], out var _)) {
+						_clusterDns = clusterDns;
+						break;
+					} else
+						throw new ArgumentException(string.Format("Specified DNS port is not a number: {0}, DNS definition: {1}",
+							splits[1], clusterDns));
+
+				default:
+					throw new ArgumentException(string.Format("Invalid cluster definition: [{0}]", clusterDns));
+			}
+			
 			return this;
 		}
 
--- a/src/EventStore.ClientAPI/ConnectionSettingsBuilder.cs
+++ b/src/EventStore.ClientAPI/ConnectionSettingsBuilder.cs
@@ -37,7 +37,7 @@ namespace EventStore.ClientAPI {
 		private TimeSpan _clientConnectionTimeout = TimeSpan.FromMilliseconds(1000);
 		private string _clusterDns;
 		private int _maxDiscoverAttempts = Consts.DefaultMaxClusterDiscoverAttempts;
-		private int _gossipExternalHttpPort = Consts.DefaultGossipPort;
+		private int _gossipExternalHttpPort = Consts.DefaultClusterManagerExternalHttpPort;
 		private TimeSpan _gossipTimeout = TimeSpan.FromSeconds(1);
 		private GossipSeed[] _gossipSeeds;
 		private NodePreference _nodePreference = NodePreference.Master;
--- a/src/EventStore.ClientAPI/ConnectionSettingsBuilder.cs
+++ b/src/EventStore.ClientAPI/ConnectionSettingsBuilder.cs
@@ -37,7 +37,7 @@ namespace EventStore.ClientAPI {
 		private TimeSpan _clientConnectionTimeout = TimeSpan.FromMilliseconds(1000);
 		private string _clusterDns;
 		private int _maxDiscoverAttempts = Consts.DefaultMaxClusterDiscoverAttempts;
-		private int _gossipExternalHttpPort = Consts.DefaultClusterManagerExternalHttpPort;
+		private int _gossipExternalHttpPort = Consts.DefaultGossipPort;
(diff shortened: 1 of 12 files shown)


shaan1337 (commented):
@YoEight nice work.

I had a few queries regarding the behaviour below:

1. I can see the same issue mentioned [here](https://github.com/EventStore/EventStore/pull/2744#pullrequestreview-562358658
) when running this version of the client against a v5 server.

here are my server settings:
```
eventstored --db ../db1 --log ../log1 --cluster-size 3 --discover-via-dns=false --gossip-seed 127.0.0.3:2112 --int-ip 127.0.0.2 --ext-ip 127.0.0.2
eventstored --db ../db2 --log ../log2 --cluster-size 3 --discover-via-dns=false --gossip-seed 127.0.0.2:2112 --int-ip 127.0.0.3 --ext-ip 127.0.0.3
```

client settings:
```
using (IEventStoreConnection _conn = EventStoreConnection.Create("ConnectTo=discover://admin:changeit@clusterdnstest.eventstore.com:2113"))
```

These settings work with the previous client but fail with `400 Bad Request` with this version of the client (can be verified with a network trace)

2. There appears to be a breaking change in the way the cluster gossip port is taken

Previously:
```
            var clusterSettings = ClusterSettings.Create()
            .DiscoverClusterViaDns()
            .SetClusterDns("clusterdnstest.eventstore.com")
            .SetClusterGossipPort(2113);
```

Now (previous code no longer works and connects to port 80):
```
            var clusterSettings = ClusterSettings.Create()
            .DiscoverClusterViaDns()
            .SetClusterDns("clusterdnstest.eventstore.com:2113")
```

YoEight:
Nice catch!

src/EventStore.ClientAPI/ConnectionSettingsBuilder.cs
```diff
@@ -306,7 +306,25 @@ public class ConnectionSettingsBuilder {
 		/// <exception cref="ArgumentNullException">If <paramref name="clusterDns" /> is null or empty.</exception>
 		public ConnectionSettingsBuilder SetClusterDns(string clusterDns) {
 			Ensure.NotNullOrEmpty(clusterDns, "clusterDns");
-			_clusterDns = clusterDns;
+
+			var splits = clusterDns.Split(':');
+
+			switch (splits.Length) {
+				case 1:
+					_clusterDns = string.Format("{0}:2113", clusterDns);
```
megakid, line 314: I don't think this port `2113` can be hard coded here.
It must (eventually) use `_gossipExternalHttpPort` but since that can be set (via `SetClusterGossipPort`) after this method is called, the `Build()` method should figure out the best port.

megakid:
> Nice catch!

The `Host` header issue still needs fixing and also ported to the v20 `master` client so we can connect v20 clients to v5 servers.

src/EventStore.ClientAPI/ConnectionSettingsBuilder.cs
YoEight: It is not hard coded per se but rather defaulted if nothing is provided. You are supposed to declare the gossip port with the domain address with this patch.

megakid: So this is a breaking change then?  Given what @shaan1337 said about about this code working before:

        var clusterSettings = ClusterSettings.Create()
            .DiscoverClusterViaDns()
            .SetClusterDns("clusterdnstest.eventstore.com")
            .SetClusterGossipPort(2113);

src/EventStore.ClientAPI/EventStoreConnection.cs
```diff
@@ -107,9 +107,16 @@ public static IEventStoreConnection
 				}
 
 				if (scheme == "discover") {
-					var clusterSettings = new ClusterSettings(uri.Host, connectionSettings.MaxDiscoverAttempts,
-						uri.Port,
+					var port = uri.Port == -1 ? 2113 : uri.Port;
```
shaan1337, line 110: Should we set the default gossip port to `Consts.DefaultClusterManagerExternalHttpPort;` here too? It seems that it would not be a backward-compatibility issue though since uri.Port=-1 was not handled before. But to maintain consistency, we may want to keep it the same.

shaan1337, line 110: It seems that the default port should be set to `connectionSettings.GossipPort` here to maintain backwards compatibility.

src/EventStore.ClientAPI/Internal/ClusterDnsSeed.cs
```diff
@@ -0,0 +1,21 @@
...
+		}
+
+		public string ToHttpUrl() {
+			var scheme = _secure ? "https" : "http";
+
+			return string.Format("{0}://{1}/gossip?format=json", scheme, _clusterDns);
+		}
+
+		public string GetHostHeader() {
+			return "";
```
shaan1337, line 18: Always returning `""` for the host header could be a problem when connecting to version 20.6 or cloud (for TLS connections). I will test it out to confirm. We may need to introduce an `EnableVersion20Compability` flag here to control whether the host header is sent or not.

src/EventStore.ClientAPI/EventStoreConnection.cs
YoEight: Done, tell me if it's better now.

src/EventStore.ClientAPI/Internal/ClusterDnsSeed.cs
YoEight: I originally tested on Cloud with no problem, which was running 20.6.1 at that time.

shaan1337: I can confirm that it results in a certificate validation error with cloud

shaan1337 (changes requested):
I've pushed a few fixes - it now works with v5 (no TLS) and v20.6 (with & without TLS). However it no longer works with cloud due to: https://github.com/EventStore/EventStore/pull/2759#discussion_r561644978.

We need to introduce the v20 compatibility mode and set the host header to `null` if it is enabled in ClusterDnsSeed.

thefringeninja (approved):

hayley-jean:
Closing this in favour of https://github.com/EventStore/EventStore/pull/2827

Issue #2771, opened 2020-12-31
Heartbeat timeout generated on internal TCP connection when forwarding large event from follower to leader


shaan1337:
**Describe the bug**
A heartbeat timeout is generated on the follower node when forwarding a large event from follower to leader over the internal TCP connection. This disrupts the cluster's stability.

**To Reproduce**
1. Start 2 nodes in a cluster of size 3 with default heartbeat intervals/timeouts
2. Write a 10MB event with the EventStore TCP Client to the external TCP port of the follower node (with PerformOnAnyNode())
3. After a few seconds, the follower node will drop the TCP connection to the leader.

The problem occurs for both plaintext TCP connections (with --insecure) and with TLS TCP connections.

**Expected behavior**
A heartbeat timeout should not be generated since both nodes are alive and communicating with each other.

**Actual behavior**
A heartbeat timeout was triggered.

**EventStore details**
 - EventStore server version:  Tested with `master` (8fd491dc83ba151c495b44e66c124fe65a42fb92)
 - Operating system: Tested on Ubuntu 16.04

**Additional context**
See https://github.com/EventStore/EventStore/pull/2757 for the root cause and fix in progress.

PR #2772, opened 2021-01-05
Proactively send heartbeat requests to the remote party if no data was sent within the last heartbeat interval

shaan1337:
Fixed: Proactively send heartbeat requests to the remote party if no data was sent within the last heartbeat interval

This is a follow-up PR to #2757

Fixes: https://github.com/eventstore/home/issues/312
Fixes: https://github.com/EventStore/EventStore/issues/2771

## Current way heartbeats work
The way heartbeats currently work between nodes is as follows:
- If we have not *received* any data from the remote party for `heartbeat interval` seconds, we enqueue a `heartbeat request` and schedule a heartbeat timeout message locally (Note that we may have *sent* a lot of data in the mean time)
- This `heartbeat request` is enqueued onto the send queue and when it is dequeued (after possibly a lot of other data), it is sent to the remote party over the TCP connection.
- The remote party receives the `heartbeat request` as raw data, enqueues it on the receive queue, dequeues it (after possibly dequeueing a lot of other data), unframes the raw data into a heartbeat request command, processes it and prepares a `heartbeat response` message which is now enqueued on the remote party's send queue and sent.
- If the local party receives the heartbeat response or any other data sent by the remote party before the heartbeat timeout check is triggered, everything is fine otherwise it results in a `HEARTBEAT TIMEOUT` and the TCP connection is closed by the local party.

## Problem description
The above scenario works fine when a connection is idle (there's not much send and receive traffic) or when we are mostly receiving traffic. However, in the case where we are sending a lot of data but not receiving any (as in #2771), it poses the following problem:
- The `heartbeat request` may need to wait behind a lot of other data on the send queue on the local side before being sent
- The `heartbeat request` may need to wait behind a lot of other data on the receive queue on the remote side before being processed

## Fix description
The fix in this PR works as follows:
Just like before:
- If we have not *received* any data from the remote party for `heartbeat interval` seconds, we enqueue a `heartbeat request` and schedule a heartbeat timeout message locally

But now, also:
- If we have not *sent* any data to the remote party during the last `heartbeat interval` seconds, we proactively enqueue a `heartbeat request` to generate some data on the connection. Note: We do not schedule a heartbeat timeout check locally in this case.

Assumptions for this method to work in an optimal way:
- Heartbeat intervals must match on both nodes (it is usually the case)
- `Heartbeat interval` + `Latency between nodes` must be lower than the `Heartbeat Timeout` - otherwise in the worst case, if we schedule the heartbeat request `Heartbeat interval` milliseconds late, the other node will already trigger the timeout check. Usually `Latency << Heartbeat Interval`, so it can be reduced to `Heartbeat interval < Heartbeat Timeout` and it is the case by default.

## Additional improvements
- Take received data info directly from TCP connection to receive fresh info: a0008ee4eb7ab5175b02f5a5223acf461fdacea9
- Submit up to MaxSendPacketSize bytes to the socket API - otherwise we receive feedback about data sent only after the whole buffer has been sent (which can be multiple megabytes large): b1653035c1f14265e0111ab174bfce787dd600ef

### DIFF
--- a/src/EventStore.Core.Tests/Services/Transport/Tcp/TcpConnectionManagerTests.cs
+++ b/src/EventStore.Core.Tests/Services/Transport/Tcp/TcpConnectionManagerTests.cs
@@ -265,6 +265,8 @@ namespace EventStore.Core.Tests.Services.Transport.Tcp {
 			get { return _clientConnectionName; }
 		}
 
+		public long TotalBytesSent { get; }
+
 		public bool IsClosed {
 			get { return false; }
 		}
--- a/src/EventStore.Core.Tests/Services/Transport/Tcp/TcpConnectionManagerTests.cs
+++ b/src/EventStore.Core.Tests/Services/Transport/Tcp/TcpConnectionManagerTests.cs
@@ -266,6 +266,7 @@ namespace EventStore.Core.Tests.Services.Transport.Tcp {
 		}
 
 		public long TotalBytesSent { get; }
+		public long TotalBytesReceived { get; }
 
 		public bool IsClosed {
 			get { return false; }
--- a/src/EventStore.Core/Messages/TcpMessage.cs
+++ b/src/EventStore.Core/Messages/TcpMessage.cs
@@ -32,10 +32,12 @@ namespace EventStore.Core.Messages {
 				get { return TypeId; }
 			}
 
-			public readonly int MessageNumber;
+			public readonly long ReceiveProgressIndicator;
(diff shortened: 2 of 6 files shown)


YoEight (approved):
Solid work!

shaan1337:
Thanks @YoEight, would you mind reviewing it again please? I had to rebase to `master` due to the failing build on Windows - looks like its ok now.

jageall (approved):

YoEight (approved):

PR #2775, opened 2021-01-07
Fix performance counter error message on linux / macOS

shaan1337:
Fixed: Performance counter error message on linux / macOS

Fixes: https://github.com/EventStore/home/issues/332

Create CPU & memory performance counters on Windows only since on Linux & macOS use the [HostStat.NET](https://www.nuget.org/packages/HostStat.NET/) library

### DIFF
--- a/src/EventStore.Core/Services/Monitoring/Utils/PerfCounterHelper.cs
+++ b/src/EventStore.Core/Services/Monitoring/Utils/PerfCounterHelper.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Diagnostics;
+using EventStore.Common.Utils;
 using ILogger = Serilog.ILogger;
 
 namespace EventStore.Core.Services.Monitoring.Utils {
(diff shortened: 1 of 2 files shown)


src/EventStore.Core/Services/Monitoring/Utils/PerfCounterHelper.cs
```diff
@@ -14,8 +15,8 @@ internal class PerfCounterHelper : IDisposable {
 		public PerfCounterHelper(ILogger log) {
 			_log = log;
 
-			_totalCpuCounter = CreatePerfCounter("Processor", "% Processor Time", "_Total");
-			_totalMemCounter = CreatePerfCounter("Memory", "Available Bytes");
+			_totalCpuCounter = Runtime.IsWindows ? CreatePerfCounter("Processor", "% Processor Time", "_Total") : null;
+			_totalMemCounter = Runtime.IsWindows ? CreatePerfCounter("Memory", "Available Bytes") : null;
```
jageall, line 19: If we move this check inside CreatePerfCounter then we only need to add a guard clause at the beginning (outside the try block) and if any subsequent counters ever get added, we won't forget that the perfcounters only exist on windows

jageall (changes requested):

thefringeninja (approved):

arwinneil (approved):

PR #2776, opened 2021-01-07
Fix windows build.ps1 script for .NET 5.0

shaan1337:
Fixed: Windows build.ps1 script for .NET 5.0

Note: Runtime identifier parameter has been removed due to: https://github.com/dotnet/sdk/issues/14281

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -5,8 +5,6 @@ Param(
     [Parameter(HelpMessage="Configuration (Debug, Release)")]
     [ValidateSet("Debug","Release")]
     [string]$Configuration = "Release",
-    [Parameter(HelpMessage="The runtime identifier")]
-    [string]$Runtime = "win10-x64",
     [Parameter(HelpMessage="Build UI (yes,no)")]
     [ValidateSet("yes","no")]
     [string]$BuildUI = "no",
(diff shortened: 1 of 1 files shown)


timothycoleman:
I seem to be running into this https://github.com/dotnet/sdk/issues/9964#issuecomment-454989539 where passing `--framework` to dotnet build when building a solution sometimes (frequently for me!)  causes warnings and errors like


```
C:\Program Files\dotnet\sdk\5.0.101\Microsoft.Common.CurrentVersion.targets(2123,5): warning MSB3088: Could not read state file "obj\x64\Debug\net5.0\EventStore.Native.csprojAssemblyReference.cache". The process cannot access the file 'C:\git\EventStore\EventStore3\src\EventStore.Native\obj\x64\Debug\net5.0\EventStore.Native.csprojAssemblyReference.cache' because it is being used by another process. [C:\git\EventStore\EventStore3\src\EventStore.Native\EventStore.Native.csproj]
```

perahps we should remove the `--framework=$NetFramework` from the script (and the corresponding definition of the variable)

shaan1337:
@timothycoleman good catch, thanks! I've pushed the changes.

timothycoleman (approved):
works for me now :)

arwinneil (approved):

PR #2788, opened 2021-01-15
Fix Plugin Tests

thefringeninja:
Fixed: failing test

Attempting to resolve where FakePlugin is in src didn't work as `dotnet test` and the resharper test runner don't output to same folder.


timothycoleman (approved):
This copies the fake plugin source to the build output so that it can be accessed directly by the tests

timothycoleman:
Any idea why `dotnet nuget locals all --clear` helps?
And might it increase the cost of the builds?

jageall (approved):

timothycoleman (approved):

PR #2791, opened 2021-01-20
Backport cluster stability fixes to version 5

shaan1337:
- Backport #2103 - Use gossip to determine node liveness after tcp connection drops
- Backport #2772 - Proactively send heartbeat requests to the remote party if no data was sent within the last heartbeat interval
- Backport #2730 (required by #2772 to avoid merge conflicts) - Use a loop when send completes synchronously

The following scenarios were tested:
For #2103
- Turn off a slave node in a 3-node cluster - the gossip message is sent by the master node and fails. The slave node is correctly marked as dead.
- Turn off a master node in a 3-node cluster - the gossip message is sent by the two slave nodes and fails. The master node is correctly marked as dead.
- Use `tcpkill` to simulate a connection drop between a slave and master. The gossip message is sent from both sides of the connection and they both succeed. The nodes are correctly marked as alive.

For #2772
- Reproduction steps in https://github.com/EventStore/EventStore/issues/2771 were followed (over plaintext TCP only, TLS was not tested). Before the fix, a heartbeat timeout occurs and after the fix it does not occur.


shaan1337:
I'm seeing some regression on the replication speed with the PR - due to b9e0724c57c103b5099d1f242f8b8ede1bf91e96 (tested with two nodes communicating through the loopback address). We also need to verify again if it's not something present on `master` as well.

shaan1337:
I've pushed a commit to address the regression issue: 8a3ee43cc8ba85bea173a39ec7e8374693430bc5
The issue is not present on `master` with .net core 3.1/.net 5.0

YoEight (commented):
@shaan1337 could you extend a bit on the rationals behind `GossipMessage.GetGossipFailed`, `GossipMessage.GetGossipReceived` and `ElectionMessage.ElectionsDone` new messages?

YoEight:
My bad, got all the information I need there https://github.com/EventStore/EventStore/pull/2103

YoEight (dismissed):

shaan1337:
As noticed by @hayley-jean this change: 8a3ee43cc8ba85bea173a39ec7e8374693430bc5 has not been done on the server's TcpConnectionSsl code. I will test if the regression is present there as well, and make the change if necessary.

src/EventStore.Core/Services/Transport/Tcp/TcpConnectionManager.cs
```diff
@@ -410,30 +410,48 @@ public class TcpConnectionManager : IHandle<TcpMessage.Heartbeat>, IHandle<TcpMe
...
+					new TcpMessage.HeartbeatTimeout(receiveProgressIndicator)));
 			}
+			/*
+			 * As a proactive measure, if we have not sent any data to the remote party within the heartbeat interval,
+			 * we also send a heartbeat request just to generate some data for the remote party so that it can clear its heartbeat timeouts.
+			 * This is particularly useful when the remote party's heartbeat request is stuck behind a large (multi-megabyte) TCP message.
+			 */
+			else if (message.SendProgressIndicator == sendProgressIndicator) {
+				SendPackage(new TcpPackage(TcpCommand.HeartbeatRequestCommand, Guid.NewGuid(), null));
+				Log.Trace(
```
hayley-jean, line 432: Trace logs are enabled by default in v5 which means that the logs for an idle cluster may fill up with this message.
Is this log necessary?

shaan1337: I've commented out the Log.Trace line, thanks @hayley-jean

hayley-jean (approved):

YoEight (approved):

PR #2796, opened 2021-01-21
Introduce compatibility mode.

YoEight:
Added: Introduce compatibility mode.

### DIFF
--- a/src/EventStore.ClientAPI/Internal/ClusterDnsEndPointDiscoverer.cs
+++ b/src/EventStore.ClientAPI/Internal/ClusterDnsEndPointDiscoverer.cs
@@ -23,7 +23,8 @@ namespace EventStore.ClientAPI.Internal {
 		private readonly TimeSpan _gossipTimeout;
 
 		private readonly NodePreference _nodePreference;
-		private readonly bool _legacyGossipDiscovery;
+		private readonly ICompatibilityMode _compatibilityMode;
+		private readonly bool _insecureMode;
 
 		public ClusterDnsEndPointDiscoverer(ILogger log,
 			string clusterDns,
@@ -32,9 +33,9 @@ namespace EventStore.ClientAPI.Internal {
 			GossipSeed[] gossipSeeds,
 			TimeSpan gossipTimeout,
 			NodePreference nodePreference,
-			bool legacyGossipDiscovery,
-			bool enableVersion5Compability,
-			HttpMessageHandler httpMessageHandler = null) {
+			ICompatibilityMode compatibilityMode,
+			HttpMessageHandler httpMessageHandler = null,
+			bool insecureMode = false) {
 			Ensure.NotNull(log, "log");
 
 			_log = log;
@@ -44,9 +45,10 @@ namespace EventStore.ClientAPI.Internal {
 			_maxDiscoverAttempts = maxDiscoverAttempts;
 			_gossipSeeds = gossipSeeds;
 			_gossipTimeout = gossipTimeout;
-			_client = new HttpAsyncClient(_gossipTimeout, httpMessageHandler, enableVersion5Compability);
+			_compatibilityMode = compatibilityMode;
+			_client = new HttpAsyncClient(_gossipTimeout, httpMessageHandler, _compatibilityMode);
 			_nodePreference = nodePreference;
-			_legacyGossipDiscovery = legacyGossipDiscovery;
+			_insecureMode = insecureMode;
 		}
 
 		public Task<NodeEndPoints> DiscoverAsync(EndPoint failedTcpEndPoint) {
@@ -103,7 +105,7 @@ namespace EventStore.ClientAPI.Internal {
 			//_log.Debug("ClusterDnsEndPointDiscoverer: GetGossipCandidatesFromDns");
 			var endpoints = _gossipSeeds != null && _gossipSeeds.Length > 0
 				? _gossipSeeds
-				: new[] {new GossipSeed(_clusterDns, !_legacyGossipDiscovery)};
+				: new[] {new GossipSeed(_clusterDns, !(_insecureMode || _compatibilityMode.IsVersion5CompatibilityModeEnabled()))};
 
 			RandomShuffle(endpoints, 0, endpoints.Length - 1);
 			return endpoints;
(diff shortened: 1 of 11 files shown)


YoEight:
Applied @jageall remarks.

megakid:
This is a nice solution but is there anyway we can handle this automatically?

E.g. call `/info` to pull `esVersion` out to determine the cluster version, which can then be used to dictate how the client connects:

```
GET '/info':
{
  "esVersion": "5.0.8.0",
  "state": "master",
  "projectionsMode": "All"
}
```

The difference in http/https could be automatically tried, e.g. default to `https://$ClusterDNS/info`, add a connection string option (Such as CompatibilityMode=5) to enable an automatic fallback to `http://$ClusterDNS/info` if that fails.

hayley-jean:
Thanks for the suggestion @megakid! Automatic compatibility would definitely be a useful feature

I've created an issue to track it separate to this PR here https://github.com/EventStore/EventStore/issues/2801. At the moment, we want to get the compatibility flags out so that we can support connecting v5 clients to v20 servers and vice versa, and then we'll focus on automatic compatibility

shaan1337 (commented):
@YoEight Tested, looks good :+1: 
Just one small change: can you please remove `SetLegacyGossipDiscovery()` from the connection settings builder?

shaan1337:
@YoEight Also, I did not test with connection strings yet - will try it out asap and keep you updated.

src/EventStore.ClientAPI/Internal/ClusterDnsEndPointDiscoverer.cs
```diff
@@ -44,9 +43,9 @@ internal class ClusterDnsEndPointDiscoverer : IEndPointDiscoverer {
 			_maxDiscoverAttempts = maxDiscoverAttempts;
 			_gossipSeeds = gossipSeeds;
 			_gossipTimeout = gossipTimeout;
-			_client = new HttpAsyncClient(_gossipTimeout, httpMessageHandler, enableVersion5Compability);
+			_compatibilityMode = compatibilityMode;
+			_client = new HttpAsyncClient(_gossipTimeout, httpMessageHandler, _compatibilityMode);
```
megakid, line 47: Why is `_compatibilityMode` passed in here? You can just condition the argument you pass to the usage of it (the hostHeader parameter is already there, defaulted to `""`).

megakid:
Please see #2804 for my proposed (minimal) changes to make Auto compatibility mode a reality.

src/EventStore.ClientAPI/Internal/ClusterDnsEndPointDiscoverer.cs
YoEight: Because host header defaulted to an empty string is not enough. The user can still pass something else, hence why we are always overwriting the host header value if version 5 compatibility mode is enabled: https://github.com/EventStore/EventStore/blob/54726446a4ce5815953be1078318f8e58aff2598/src/EventStore.ClientAPI/Transport.Http/HttpAsyncClient.cs#L91

YoEight:
> @YoEight Tested, looks good 
> Just one small change: can you please remove `SetLegacyGossipDiscovery()` from the connection settings builder?

@shaan1337 it's done!

src/EventStore.ClientAPI/Internal/ClusterDnsEndPointDiscoverer.cs
megakid: It is enough.  There is no-where else that the compatibility mode is passed into `HttpAsyncClient` constructor, only in `ClusterDnsEndPointDiscoverer`.  By passing in empty string to the `Get` method, you achieve the exact same behaviour without needlessly bleeding the compatibility mode concept to `HttpAsyncClient`

megakid: Also I feel like that host header compatibility flag should be better described as "MonoCompatibility" or similar since it doesn't affect non-mono servers?

YoEight: I would agree if `HttpAsyncClient` was a user-facing class. I would also argue passing the compatibility mode in this fashion conveys better why we actually have an empty string. You also keep overlooking the fact that having `hostHeader` defaulted to empty string doesn't prevent the user from overwriting that value from the connection setting https://github.com/EventStore/EventStore/blob/711ff91a6c7c6ef80afd488c5611f49b7d5ab232/src/EventStore.ClientAPI/Internal/ClusterDnsEndPointDiscoverer.cs#L188

megakid: Re. your last point, I didn't see that - What path can the user overwrite the header value?

YoEight: No worries, it happens when the user asks to use the DNS to resolve the gossip seeds: https://github.com/EventStore/EventStore/blob/711ff91a6c7c6ef80afd488c5611f49b7d5ab232/src/EventStore.ClientAPI/Internal/ClusterDnsEndPointDiscoverer.cs#L41

YoEight: About the naming, whether it's `MonoCompatibility` or a version number, I agree for the initial issue, your name is more precise. However, we introduce a more stable naming strategy that gives us room for further expansion by using the number version.

YoEight: Having said that, the auto compatibility mode you introduced in your PR is more future-proof and renders the naming strategy useless. So having `MonoCompatibility` might be a better solution after all :smile: !

hayley-jean:
It appears that there's a regression in this PR. The client can only connect to an insecure v20 cluster using `discover://` if it specifies `CompatibilityMode=5`

YoEight:
@hayley-jean I addressed that regression and rebased the PR.

YoEight:
@megakid I merged your PR, solid job!

shaan1337 (approved):
Tested with the following server configs with both connection builder and connection string:
- v20 secure (default)
- v5 no TCP TLS (default)
- v20 --insecure (had to set CompatibilityMode=Auto for it to work)
- v5 TCP TLS

megakid:
@shaan1337 do you mind posting up your test configuration?  I'm curious how you got your v20 secure and v5 TCP TLS working without changing the connection string (or using CompatibilityMode=Auto)

hayley-jean (approved):

PR #2803, opened 2021-01-27
Authentication on Follower Nodes

thefringeninja:
Fixed: Permission Denied when performing privileged commands on a follower


jageall (approved):

hayley-jean (approved):
Looks good and can confirm that the issue is fixed 👍

PR #2804, opened 2021-01-27
Auto compatibility mode

megakid:
This PR is based on the `compatibility-mode` branch that is currently inflight to solve https://github.com/EventStore/EventStore/issues/2801

I aimed for fairly minimal changes here and have tested:

`ConnectTo=discover://localhost:2111;CompatibilityMode=Auto` 

Successfully against a V5 non-TLS and V20 TLS cluster with *no connection string changes*.  

**How it works**

If auto compatibility mode is enabled, it will default to two gossip seeds (first one for v20 TLS, second one for v5 non-TLS).  It will perform a gossip request for those seeds in that order (we want to prioritise v20 clusters here).  This means no pre-flight check is necessary to discover the version of the cluster.

Note: I also fixed a bug here whereby v5 clusters would never get a successful discover during an initial reconnect (e.g. if a disconnect happened, on the reconnect, it tries to use `_oldGossip` which would have defaulted to `seedOverTls = true` which would never work).  Thankfully this never really showed itself given the next attempt would work as it goes back to the configured discover process (e.g. same logic as initial connect discovery)


hayley-jean:
Thanks for the PR @megakid!

We definitely want to get this in, we're just going to wait until the other compatibility PR is merged and will get on to reviewing this one

PR #2809, opened 2021-02-04
Check for old/replayed events and move checkpoint tag forward only if the event passes the source filter [Projections]

### LINKED ISSUE #2785: Exception: failed to handle event because complete TF position required (Parameter 'committedEvent')
**Describe the bug**

Projection may fault after periodically with exception:

`{"@t":"2020-12-09T10:30:55.4407703Z","@mt":"The '{projection}' projection faulted due to '{e}'","@l":"Error","projection":"myprojection","e":"The heading subscription failed to handle an event-996553a5-fca4-4f53-b9c3-bebe300720d2:sampleevent@163301 because complete TF position required (Parameter 'committedEvent')","SourceContext":"EventStore.Projections.Core.Services.Management.ProjectionManager","ProcessId":5448,"ThreadId":23}`

**To Reproduce**
1. Create a projection with code:

`
fromAll().when({
"sampleevent" : function(state,event) {
     emit("myprojection", "EventLtIdAdded",{
     LtId: event.body.LtdId,
     Id: event.body.AggregateId
     })
     }
     }
);
`

2. Add events that will be handle by this custom projection

**Expected behavior**
Projection should not fault. 

**Actual behavior**
Custom Projection faults unexpectedly and require the projection to be reset to resume operations.

**EventStore details**
20.6.1
 

 - Operating system:
 
 Windows Server 2016

shaan1337:
Fixed: Check for old/replayed events only if the event passes the event filter [Projections]

Fixes #2785
This PR fixes a regression introduced by #2428 as from version v20.6 (confirmed with a git bisect)

After removing the `!_eventFilter.PassesSource()` check in #2428, the `_positionTagger.IsMessageAfterCheckpointTag()` check is now always called and it results in the error seen on #2785.

The PR reintroduces the `eventFilter.PassesSource()` check and if it is true, it checks for replayed/old messages and moves the checkpoint tag forward. However, checkpoints are still generated even if the event does not match the source filter (unlike before #2428 was introduced)


jageall (approved):

hayley-jean (dismissed):
The change looks good, and can confirm that this fixes the original issue without bringing back the issue in https://github.com/EventStore/EventStore/issues/2001.

Would you be able to add a test case for the original issue?

hayley-jean (approved):

jageall (approved):

PR #2824, opened 2021-02-24
Prevent a checkpoint from being emitted at same position twice

shaan1337:
Fixed: Prevent a projection checkpoint from being emitted at same position twice

Fix regression introduced due to https://github.com/EventStore/EventStore/pull/2809 and discovered by @timothycoleman 

Co-Authored-By: @hayley-jean 
Fixes: https://github.com/EventStore/home/issues/386


thefringeninja (approved):

condron (approved):
looks good

PR #2825, opened 2021-02-24
Don't write duplicate checkpoints for $by_event_type

hayley-jean:


hayley-jean:
Closing this in favour of https://github.com/EventStore/EventStore/pull/2824 as I believe it's more correct to prevent any duplicate checkpoints

PR #2832, opened 2021-03-02
Add Skip certificate validation connection string option

hayley-jean:
Changed: ValidateServer also sets whether HTTP certificate validation is enabled.

Certificate validation for both TCP and HTTPS can now be disabled with the following:

```
var connectionString = "GossipSeeds=https://localhost:2113;UseSslConnection=true;ValidateServer=false";
```

or

```
var connectionSettings = ConnectionSettings.Create()
    .DisableServerCertificateValidation()
    .SetGossipSeedEndPoints(new IPEndPoint(IPAddress.Loopback, 2113));
```

Ports the functionality of https://github.com/EventStore/EventStore/pull/2585 to the current TCP client.
Fixes https://github.com/EventStore/EventStore/issues/2588

### DIFF
--- a/src/EventStore.ClientAPI/EventStoreConnection.cs
+++ b/src/EventStore.ClientAPI/EventStoreConnection.cs
@@ -1,9 +1,9 @@
 ﻿using System.Net;
 using System.Linq;
-using System.Net.Sockets;
 using EventStore.ClientAPI.Common.Utils;
 using EventStore.ClientAPI.Internal;
 using System;
+using System.Net.Http;
 using System.Runtime.ExceptionServices;
 using EventStore.ClientAPI.Messages;
 using EventStore.ClientAPI.SystemData;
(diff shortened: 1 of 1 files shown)


YoEight (approved):

shaan1337 (commented):
Tested, looks good.

However, I noticed there's already a `DisableServerCertificateValidation()` option for TCP which may be confusing for consumers of the API:
```
var settings =
ConnectionSettings.Create().SkipCertificateValidation().DisableServerCertificateValidation();
```

The repetition is also present in the connection string:
```
ValidateServer=False;SkipCertificateValidation=True
```

Do we want to consolidate these two settings into one?

hayley-jean:
Because `DisableServerCertificateValidation` exists in the previous client version we can't remove it for this version without it being a potentially breaking change.

But we could make `SkipCertificateValidation` disable both HTTP and TCP validation? Both options would still exist, but you'd only have to specify the one.

YoEight (approved):

src/EventStore.ClientAPI/EventStoreConnection.cs
```diff
@@ -193,9 +193,21 @@ public static IEventStoreConnection
...
 			Ensure.NotNull(clusterSettings, "clusterSettings");
 
-			var discoverClient = new HttpAsyncClient(
-				clusterSettings.GossipTimeout,
-				connectionSettings.CustomHttpMessageHandler);
+			var handler = connectionSettings.CustomHttpMessageHandler;
+			if (handler is null) {
+				if (!connectionSettings.ValidateServer) {
+#if NET452
+					connectionSettings.Log.Info(
```
shaan1337, line 200: I was going to suggest that this log message could be converted to a warning, but it looks like the `ILogger` interface does not have `Warn` method.

shaan1337 (approved):

PR #2833, opened 2021-03-02
Configuration Changes

thefringeninja:
Changed: internal configuration system now based on `Microsoft.Extensions.Configuration`

This PR makes the following changes:
- [x] New `ClusterVNode` configuration options record type
- [x] New `TestClient` configuration options based on Dragonfruit
- [x] `--what-if` now exits after printing effective configuration
- [x] OS specific flags style (GNU on Unix, Powershell on Windows) in help message
- [x] All `VNodeBuilders`  are removed
- [x] `EventStore.Rags` is removed

Additional Tasks:
- [x] Test with auth plugins
- [x] Update docs


src/EventStore.Core/ClusterVNodeOptions.cs
```diff
@@ -0,0 +1,634 @@
...
+			[Description("Enables trusted authentication by an intermediary in the HTTP.")]
+			public bool EnableTrustedAuth { get; init; } = false;
+
+			[Description("Whether to disable secure internal TCP communication.")]
+			public bool DisableInternalTcpTls { get; init; } = false;
+
+			[Description("Whether to disable secure external TCP communication.")]
+			public bool DisableExternalTcpTls { get; init; } = false;
+
+			[Description("Enable AtomPub over HTTP Interface.")]
```
hayley-jean, line 574:
```suggestion
			[Description("Enable AtomPub over HTTP Interface."),
			Deprecated("AtomPub over HTTP Interface has been deprecated as of version 20.6.0. It is recommended to use gRPC instead")]
```

src/EventStore.Core/ClusterVNodeOptions.cs
```diff
@@ -0,0 +1,634 @@
...
+		[Description("Interface Options")]
+		public record InterfaceOptions {
+			[Description("Internal IP Address.")] public IPAddress IntIp { get; init; } = IPAddress.Loopback;
+
+			[Description("External IP Address.")] public IPAddress ExtIp { get; init; } = IPAddress.Loopback;
+
+			[Description("The port to run the HTTP server on.")]
+			public int HttpPort { get; init; } = 2113;
+
+			[Description("Whether to enable external TCP communication")]
```
hayley-jean, line 503:
```suggestion
			[Description("Whether to enable external TCP communication"),
			Deprecated("The Legacy TCP Client Interface has been deprecated as of version 20.6.0. It is recommended to use gRPC instead.")]
```

hayley-jean (changes requested):
The new format looks good!

A few issues that have come up in testing so far:
- Command line arguments don't override options set in the config file
- `RunProjections: All` or `RunProjections: System` doesn't start the projections subsystem
- Environment variables don't seem to affect the options (only tested on Windows though)

I've also suggested a few changes to copy over the deprecated messages we currently have on master

hayley-jean (changes requested):
Thanks! The order of precedence, environment variables and subsystems work for me now.

Two additional issues found in testing:

1. If I try to start the server with an auth plugin using a config section rather than a separate file, I get an error saying the option doesn't exist:

```
Error while parsing options: The option LdapsAuth is not a known option. (Parameter 'LdapsAuth')
```

2. The TestClient doesn't print its effective options on startup. `--what-if` also doesn't work.

src/EventStore.ClusterNode/Program.cs
```diff
@@ -1,109 +1,167 @@
...
 					Application.Exit(0, "Cancelled.");
 				};
-
-				if (hostedService.Options.WhatIf) {
-					Log.Information("Exiting with exit code: 0.\nExit reason: WhatIf option specified");
-					return (int)ExitCode.Success;
+				using (var hostedService = new ClusterVNodeHostedService(options)) {
+					using var signal = new ManualResetEventSlim(false);
+					_ = Run(hostedService, signal);
+					// ReSharper disable MethodSupportszzzzzzzzzCancellation
```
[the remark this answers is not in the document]
thefringeninja: Must have fallen asleep on my keyboard 🙄

hayley-jean (approved):

arwinneil (approved):

PR #2819, opened 2021-02-15
Add Keep Alive Configuration to Server

thefringeninja:
Added: Server Support for gRPC Keep Alive

Closes EventStore/home#356

PR #2842, opened 2021-03-09
Backport changes to 20.10.2

hayley-jean:
Added: backports for v20.10.2

Backports the following changes:

- Authentication on Follower Nodes #2803 
- Add Keep Alive Configuration to Server #2819
- Update internal dependencies #2764
  - Fix linux/macOS build.sh script for .NET 5.0 #2774
  - Fix windows build.ps1 script for .NET 5.0 #2776
  - Make --insecure work with .NET 5.0 #2779
  - install both netcoreapp3.1 and net5.0 #2780
- Check retries on messages when connection is lost #2797
- Add timeout when getting gossip during discovery #2821
- Check for old/replayed events and move checkpoint tag forward only if the event passes the source filter [Projections] #2809
- Filter out invalid json events from projections #2812
- Prevent a checkpoint from being emitted at same position twice #2824


pvanbuijtene:
Think there's a bug in the main code `keepAliveInterval` is checked twice instead of `keepAliveTimeout`
https://github.com/EventStore/EventStore/blob/master/src/EventStore.Core/Cluster/Settings/ClusterVNodeSettings.cs#L200

pvanbuijtene:
Some test fixtures are missing from https://github.com/EventStore/EventStore/pull/2809, is this on purpose?

Here: https://github.com/EventStore/EventStore/pull/2842/files#diff-0d7dbd01ca3f959236bc2f27c167a2265e1dd1a675f4b3ee54fe0237f6bffc84R27
There: https://github.com/EventStore/EventStore/pull/2809/files#diff-0d7dbd01ca3f959236bc2f27c167a2265e1dd1a675f4b3ee54fe0237f6bffc84R29

pvanbuijtene (approved):

pvanbuijtene:
@hayley-jean the new pr looks fine, tests locally passing, ci failed on famous `Cannot access a closed file.`.

src/EventStore.ClientAPI/Internal/ClusterDnsEndPointDiscoverer.cs
```diff
@@ -190,19 +203,20 @@ internal class ClusterDnsEndPointDiscoverer : IEndPointDiscoverer {
...
 					}
 
 					completed.Set();
-				},
-				e => {
-					if (e is AggregateException ae)
+				}, e => {
+					if (e is AggregateException ae) 
 						e = ae.Flatten();
 					_log.Error("Failed to get cluster info from [{0}]: request failed, error: {1}.", endPoint, e);
```
shaan1337, line 216: Based on previous changes, this should be `endPoint.EndPoint` in the log error message (it has not been changed on master too)

shaan1337:
#2822 has also been backported to release 20.10.2 branch in PR https://github.com/EventStore/EventStore/pull/2830. Just adding this comment so that we don't forget to include it in the release notes.

src/EventStore.ClientAPI/Internal/ClusterDnsEndPointDiscoverer.cs
```diff
@@ -177,11 +177,24 @@ internal class ClusterDnsEndPointDiscoverer : IEndPointDiscoverer {
...
+								e = ae.Flatten();
+							_log.Error("Failed to get cluster info from [{0}]: deserialization error: {1}.",
+								endPoint.EndPoint, e);
+						}
+					} finally {
 						completed.Set();
-						return;
 					}
 
 					try {
```
shaan1337, line 200: This try-catch-finally block seems redundant (same as above). It is not present on master.

shaan1337:
--keep-alive-interval and --keep-alive-timeout seem to accept a value of -1 (but not -2, etc.). It must be the same on `master` but what's strange is that the code doesn't seem to handle the -1 case as a special case. I'm not sure what the behaviour is when -1 is specified.

shaan1337 (commented):
Changes look good overall :+1:  I've just left a few comments regarding some things i've noticed.

shaan1337 (approved):
- Reviewed code
- Tested: Authentication on Follower Nodes #2803
- Tested: Add Keep Alive Configuration to Server #2819
- Tested: Make --insecure work with .NET 5.0 #2779
- Tested: Check for old/replayed events and move checkpoint tag forward only if the event passes the source filter [Projections] #2809
- Tested: Prevent a checkpoint from being emitted at same position twice #2824

PR #2859, opened 2021-03-16
Make Microsoft.NETFramework.ReferenceAssemblies reference private

megakid:
Currently on nuget, `Microsoft.NETFramework.ReferenceAssemblies` dependency is listed explicitly, this is redundant and should be tidied up.

Using `PrivateAssets="All"` will achieve this.  The only question is I am 99% sure that .NET 5.0 SDK automatically includes this dependency (privately) so might be able to just remove this line altogether.


thefringeninja (approved):

hayley-jean (approved):

hayley-jean:
Thanks @megakid!

PR #2860, opened 2021-03-17
[V5] Extend auto-compatibility mode to gossip seeds configuration.

YoEight:


YoEight:
Tested on v4 cluster with compatibility mode set to `auto` and `5`

YoEight:
Tested on v20.* cluster on EventStoreDB Cloud with mode set to `auto`. Only when the compatibility mode is set to `auto` that we support auto target host detection when SSL is enabled.

YoEight:
Tested auto SSL detection on a v20.* cluster in EventStoreDB Cloud.

src/EventStore.ClientAPI/Internal/ClusterDnsEndPointDiscoverer.cs
```diff
@@ -114,11 +114,38 @@ internal class ClusterDnsEndPointDiscoverer : IEndPointDiscoverer {
...
 			//_log.Debug("ClusterDnsEndPointDiscoverer: GetGossipCandidatesFromDns");
+			
+			IGossipSeed[] endpoints;
 
-			// Safe in this case.
-			IGossipSeed[] endpoints = _gossipSeeds;
+			if ((_gossipSeeds?.Length ?? 0) != 0) {
+				if (_compatibilityMode.IsAutoCompatibilityModeEnabled()) {
+					var tmp = new List<IGossipSeed>();
+					for (var i = 0; i != _gossipSeeds.Length; i++) {
```
shaan1337, line 123:
```suggestion
					for (var i = 0; i < _gossipSeeds.Length; i++) {
```
Just for safety reasons in case the code is modified later by someone else

src/EventStore.ClientAPI/Internal/ClusterDnsEndPointDiscoverer.cs
```diff
@@ -114,11 +114,38 @@ internal class ClusterDnsEndPointDiscoverer : IEndPointDiscoverer {
...
-			if ((endpoints?.Length ?? 0) == 0) {
+					endpoints = tmp.ToArray();
+				} else if (_compatibilityMode.IsVersion5CompatibilityModeEnabled()) {
+					// Safe in this case.
+					endpoints = _gossipSeeds;
+					RandomShuffle(endpoints, 0, endpoints.Length - 1);
+				} else {
+					// Convert to v20.
+					endpoints = new IGossipSeed[_gossipSeeds.Length];
+					for (var i = 0; i != _gossipSeeds.Length; i++) {
```
shaan1337, line 140:
```suggestion
					for (var i = 0; i < _gossipSeeds.Length; i++) {
```

src/EventStore.ClientAPI/GossipSeed.cs
```diff
@@ -24,26 +25,36 @@ public class GossipSeed: IGossipSeed {
...
+				return string.Format("{0}://{1}:{2}/gossip?format=json", scheme, HostHeader, EndPoint.Port);
+			}
+			
 			return EndPoint.ToHttpUrl(SeedOverTls ? EndpointExtensions.HTTPS_SCHEMA : EndpointExtensions.HTTP_SCHEMA,
 				"/gossip?format=json");
 		}
 		
 		public string GetHostHeader() {
-			return HostHeader;
+			return _v20Compatibility ? HostHeader : "";
```
shaan1337, line 57: @YoEight Isn't this a breaking change when _v20Compatibility=false? (for version 5 clients connecting against a version 5 server with a hostname configured)

shaan1337: I believe that if we just return `HostHeader` here it should be ok.

The reason we had to clear the host header out for v20 client to connect against v5 server was due to the change in the DNS discovery mechanism in the v20 client - but it does not apply to IP gossip seeds as far as I can tell.

In this case, it seems that we don't even need the `_v20Compatibility` flag to be passed in the `GossipSeed` and no changes would need to be done to `GossipSeed.cs` at all.

Does this make sense or am I missing a case?

src/EventStore.ClientAPI/GossipSeed.cs
```diff
@@ -24,26 +25,36 @@ public class GossipSeed: IGossipSeed {
...
+		/// <param name="v20Compatibility">If v20 compatibility mode is enabled.</param>
+		public GossipSeed(IPEndPoint endPoint, string hostHeader = "", bool seedOverTls = false, bool v20Compatibility = false) {
 			EndPoint = endPoint;
 			HostHeader = hostHeader == string.Empty ? null : hostHeader;
 			SeedOverTls = seedOverTls;
+			_v20Compatibility = v20Compatibility;
 		}
 		
 		public string ToHttpUrl() {
+			if (_v20Compatibility && !string.IsNullOrEmpty(HostHeader)) {
```
shaan1337, line 46: This change also appears to be unnecessary based on: https://github.com/EventStore/EventStore/pull/2860/files?diff=split&w=1#r600182774

src/EventStore.ClientAPI/Internal/ClusterDnsEndPointDiscoverer.cs
YoEight: Done

src/EventStore.ClientAPI/Internal/ClusterDnsEndPointDiscoverer.cs
YoEight: Done

src/EventStore.ClientAPI/GossipSeed.cs
YoEight: While I agree now that we don't need to change `GetHostHeader` implementation for `GossipSeed`, We do need to send  that host even if we don't perform a DNS discovery in this case. The use-case is when we try to connect to a EventStoreDB Cloud cluster while using gossip seeds. You won't be able to send your gossip request without using the hostname (remember `SkipCertificateServer` is set to `false` by default)

src/EventStore.ClientAPI/GossipSeed.cs
YoEight: I disgree: https://github.com/EventStore/EventStore/pull/2860#discussion_r600227184

src/EventStore.ClientAPI/GossipSeed.cs
YoEight: This is where the header is needed regardless if we do a DNS discovery or not: https://github.com/EventStore/EventStore/blob/6cf09ae8b144ad624f226c6fb784dc1592bacdae/src/EventStore.ClientAPI/Internal/ClusterDnsEndPointDiscoverer.cs#L230-L257

shaan1337 (approved):
great work @YoEight !

I've carried out a few tests and it looks good :+1: 

### Versions tested:
v4 = 4.1.3
v5 = 5.0.10
v20 = 20.10.0

### No Compatibility mode (GossipSeeds=127.0.0.1:2113,127.0.0.2:2113;)
v4 defaults - OK
v5 defaults - OK
v20 defaults - OK (does not connect as expected)

### Compatibility mode = Auto (GossipSeeds=127.0.0.1:2113,127.0.0.2:2113;CompatibilityMode=auto)
v4 defaults - OK
v5 defaults - OK
v20 defaults - OK
v20 defaults + disable external TCP TLS - OK

### Live Upgrades/Downgrades (GossipSeeds=127.0.0.1:2113,127.0.0.2:2113;CompatibilityMode=auto;MaxRetries=1000;MaxReconnections=1000;MaxDiscoverAttempts=1000)
Live cluster upgrade from v4 (defaults) to v20 (defaults) - OK
Live cluster upgrade from v4 (defaults) to v20 (disable TCP TLS) - OK
Live cluster downgrade from v20 (defaults) to v4 (defaults) - OK
Live cluster downgrade from v20 (disable TCP TLS) to v4 (defaults) - OK

### Hostname gossip seed (client):
```
eventstored --db ../db1 --int-ip 127.0.0.1 --ext-ip 127.0.0.1 --discover-via-dns false --gossip-seed 127.0.0.2:2112 --cluster-size 3 --ext-http-prefixes "http://localhost1:2113/"
eventstored --db ../db2 --int-ip 127.0.0.2 --ext-ip 127.0.0.2 --discover-via-dns false --gossip-seed 127.0.0.1:2112 --cluster-size 3 --ext-http-prefixes "http://localhost2:2113/"

var clusterSettings = ClusterSettings.Create()
    .DiscoverClusterViaGossipSeeds()
    .SetGossipSeedEndPoints(
        new GossipSeed[]{
            new GossipSeed(new IPEndPoint(IPAddress.Parse("127.0.0.1"), 2113), "localhost1"),
            new GossipSeed(new IPEndPoint(IPAddress.Parse("127.0.0.2"), 2113), "localhost2")
        }
    );
```

Tested with clients:
 - 5.0.9: OK
 - current PR: OK

hayley-jean (approved):

PR #2869, opened 2021-03-25
Persistent subscriptions to $all

shaan1337:
Added: Persistent subscriptions to $all for gRPC clients

Fixes: https://github.com/EventStore/home/issues/432

This PR introduces persistent subscriptions to $all:
- Subscriptions to $all can be created/updated/deleted via the gRPC API only
- They can be viewed in the UI but not managed there
- Internal stream read timeouts are now handled gracefully by retrying infinitely with exponential backoff and jitter (for both $all and normal persistent subscriptions). Read retries will be logged on the server.
- Read errors will fault the persistent subscription and drop all subscribers (this applies mainly to $all subscriptions - for e.g starting a subscription from an invalid commit/prepare position). Any new subscriber that connects to the subscription will be dropped too.
- Prevents checkpoints and parked messages for persistent subscription to $all from being visible in the $all stream (format: `$persistentsubscription-$all::*-checkpoint`, `$persistentsubscription-$all::*-parked`). Reading or subscribing to the checkpoint/parked message stream should be possible but reading or subscribing $all should not deliver these events. This has been done to prevent these messages from being fed back into the subscription which may cause infinite loops (e.g with a max checkpoint of 1).
- Backwards compatibility is preserved for existing projections allowing direct upgrades.
- Downgrades are possible only if a subscription to $all has not yet written a checkpoint.

### Breaking changes
-  Handle missing case for UpdatePersistentSubscriptionToStreamResult.DoesNotExist: a27b4817c65ed55449386dd64d7f4db70a6412d0. This will result in a different exception on the client side.

### Client changes
.NET gRPC client PR: https://github.com/EventStore/EventStore-Client-Dotnet/pull/108

The tests from the above PR has been run against this branch.

### UI
Following changes need to be made to the web UI:
- Hide edit/delete buttons for $all subscriptions
JSON changes:
- Use `startPosition` instead of `startFrom`
- Use `lastProcessedEventPosition` instead of `lastProcessedEventNumber`
- Use `lastKnownEventPosition` instead of `lastKnownEventNumber`


src/Protos/Grpc/persistent.proto
```diff
@@ -19,7 +19,12 @@ message ReadReq {
 	}
 
 	message Options {
-		event_store.client.shared.StreamIdentifier stream_identifier = 1;
+		event_store.client.shared.StreamIdentifier stream_identifier = 1 [deprecated = true];
+		oneof stream_option {
```
thefringeninja, line 23: Do we expect to add more options to StreamOptions / AllOptions? If not, then according to https://developers.google.com/protocol-buffers/docs/proto#updating, we could do

```
oneof stream_option {
  event_store.client.shared.StreamIdentifier stream_identifier = 1;
  Empty all = 5;
}
```

> Changing a single optional value into a member of a new oneof is safe and binary compatible. Moving multiple optional fields into a new oneof may be safe if you are sure that no code sets more than one at a time. Moving any fields into an existing oneof is not safe.

shaan1337: @thefringeninja thanks, i'll apply this change to `ReadReq` and `DeleteReq` as well :+1:

hayley-jean: @shaan1337 the changes suggested by @thefringeninja look good to me for `ReadReq` and `DeleteReq`, I don't think we're going to need any more options on StreamOptions/AllOptions

shaan1337: thanks @hayley-jean . I've applied the changes.

hayley-jean (approved):
Looks good, my only comment is about the error shown when a read times out

thefringeninja (dismissed):

src/EventStore.Core/Services/Transport/Grpc/PersistentSubscriptions.Read.cs
```diff
@@ -37,8 +37,20 @@ public partial class PersistentSubscriptions {
 			var options = requestStream.Current.Options;
 			var user = context.GetHttpContext().User;
 
+			string streamId = null;
```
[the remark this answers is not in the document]
shaan1337: @thefringeninja i've attached a debugger to the server and can confirm it's going into the `case ReadReq.Types.Options.StreamOptionOneofCase.StreamIdentifier` when using the old client. Just ran the regression tests from the client branch to be on the safe side and they're all passing except: `EventStore.Client.update_non_existent.the_completion_fails_with_not_found` due to https://github.com/EventStore/EventStore/commit/a27b4817c65ed55449386dd64d7f4db70a6412d0

YoEight (dismissed):

hayley-jean (approved):

YoEight (approved):

src/EventStore.Core/Services/Transport/Grpc/PersistentSubscriptions.Read.cs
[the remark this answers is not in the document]
hayley-jean: I can also confirm that the old client can still subscribe to the new server version, the streamId is being populated correctly

PR #2873, opened 2021-03-26
Add updated timestamp to stream metadata result

alexeyzimarev:
Not sure if it needs to be accepted, or I shall consider using direct reads from the meta stream instead (which is possible).

### DIFF
--- a/src/EventStore.ClientAPI.Embedded/EventStoreEmbeddedNodeConnection.cs
+++ b/src/EventStore.ClientAPI.Embedded/EventStoreEmbeddedNodeConnection.cs
@@ -812,27 +812,25 @@ namespace EventStore.ClientAPI.Embedded {
 			return await source.Task.ConfigureAwait(false);
 		}
 
-		public Task<StreamMetadataResult>
-			GetStreamMetadataAsync(string stream, UserCredentials userCredentials = null) {
+		public Task<StreamMetadataResult> GetStreamMetadataAsync(string stream, UserCredentials userCredentials = null) {
 			return GetStreamMetadataAsRawBytesAsync(stream, GetUserCredentials(_settings, userCredentials))
 				.ContinueWith(t => {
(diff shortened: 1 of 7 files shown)


hayley-jean:
@thefringeninja How badly will merging this conflict with https://github.com/EventStore/EventStore/pull/2863?

Should we wait for the TCP client to be moved to [EventStore/EventStoreDB-Client-Dotnet-Legacy](https://github.com/EventStore/EventStoreDB-Client-Dotnet-Legacy) and then open this PR against that repo?

alexeyzimarev:
If we agree that this is acceptable, it might be possible to merge before completing the split? Some of the changes aren't necessary (like string interpolation and nullable things).

hayley-jean:
@alexeyzimarev Yeah, I think it should be fine to merge before the split

However, the CI appears to be failing for legitimate reasons. Would you be able to look into that?

alexeyzimarev:
@hayley-jean the tests fail because `metadata.AsJsonBytes()` returns an empty array for a properly populated object. I was able to follow the writer and it writes all the values, but the writer doesn't flush to the memory stream. I am not sure how it is related to my PR. I've seen a similar issue myself when I was adding new meta using the JSON writer in the replicator.

alexeyzimarev:
I found it. The using was previously around the write and converting it to a larger scope caused late disposal.

hayley-jean (approved):

thefringeninja:
@alexeyzimarev it's moved, can you resubmit your pull request here please? https://github.com/EventStore/EventStoreDB-Client-Dotnet-Legacy/pull/2#pullrequestreview-633733968

hayley-jean:
This PR has been moved to the Legacy Client repo: https://github.com/EventStore/EventStoreDB-Client-Dotnet-Legacy/pull/3

PR #2883, opened 2021-03-30
Extend auto-compatibility mode to gossip seeds configuration

hayley-jean:
Added: Auto-compatibility for gossip seeds to TCP client

Port PR https://github.com/EventStore/EventStore/pull/2860 to v21.

Allow gossip seeds to benefit from auto-compatibility.
This means that the TCP client will try and discover whether a cluster it's gossiping with is secure or insecure.

For example, the following connection string can be used to connect to:
- an insecure v5 cluster
- an insecure v20/21 cluster
- a secure v20/21 cluster:

```
var connectionString = "GossipSeeds=localhost:1113,localhost:2113,localhost:3113;CompatibilityMode=auto;ValidateServer=true;";
```

### DIFF
--- a/src/EventStore.ClientAPI/Internal/EventStoreConnectionLogicHandler.cs
+++ b/src/EventStore.ClientAPI/Internal/EventStoreConnectionLogicHandler.cs
@@ -34,6 +34,7 @@ namespace EventStore.ClientAPI.Internal {
 		private TimeSpan _lastTimeoutsTimeStamp;
 		private readonly OperationsManager _operations;
 		private readonly SubscriptionsManager _subscriptions;
+		private readonly ICompatibilityMode _compatibilityMode;
 
 		private ConnectionState _state = ConnectionState.Init;
 		private ConnectingPhase _connectingPhase = ConnectingPhase.Invalid;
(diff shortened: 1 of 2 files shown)


src/EventStore.ClientAPI/Internal/EventStoreConnectionLogicHandler.cs
```diff
@@ -163,18 +166,28 @@ internal class EventStoreConnectionLogicHandler : IEventStoreConnectionLogicHand
...
-				_settings.UseSslConnection,
+				useSsl,
 				_settings.ValidateServer,
 				_settings.ClientConnectionTimeout,
 				(connection, package) => EnqueueMessage(new HandleTcpPackageMessage(connection, package)),
 				(connection, exc) => EnqueueMessage(new TcpConnectionErrorMessage(connection, exc)),
 				connection => EnqueueMessage(new TcpConnectionEstablishedMessage(connection)),
 				(connection, error) => EnqueueMessage(new TcpConnectionClosedMessage(connection, error)));
+
+			if (_connection.ConnectionId == Guid.Empty) {
```
YoEight, line 186: Won't do anything as `TcpPackageConnection` was patched in `v5-master` because the client used to silently crashed if SSL was enable but `HostHeader` was null or empty. Returning an empty `Guid` was used to communicate that the connection was misconfigured in order to have a proper error. I don't see `Ensure.NotNullOrEmpty` used on `TargetHost` on `master`.

YoEight (changes requested):

src/EventStore.ClientAPI/Internal/EventStoreConnectionLogicHandler.cs
YoEight: `v5-master` context: https://github.com/EventStore/EventStore/blob/v5-master/src/EventStore.ClientAPI/Transport.Tcp/TcpPackageConnection.cs#L53

YoEight (approved):

hayley-jean:
This PR has been moved to the Legacy Client repo: https://github.com/EventStore/EventStoreDB-Client-Dotnet-Legacy/pull/4

PR #2888, opened 2021-03-31
Add gRPC client to EventStore.TestClient

hayley-jean:
Added: gRPC client to EventStore.TestClient

- Split out `TcpTestClient` and `GrpcTestClient`
- Add `WRFLGRPC` command to perform a write flood against gRPC
- Print out usage on startup
- Add `--connection-string` option to specify gRPC connection string
- Make it possible to set stream prefix when running write flood for gRPC

Example command to run a gRPC write flood:

```
wrflgrpc 10 100000 1000 256 1 foo
```

### DIFF
--- a/src/EventStore.TestClient/GrpcCommands/WriteFloodProcessor.cs
+++ b/src/EventStore.TestClient/GrpcCommands/WriteFloodProcessor.cs
@@ -13,7 +13,7 @@ namespace EventStore.TestClient.GrpcCommands {
 		private static readonly UTF8Encoding UTF8NoBom = new UTF8Encoding(false);
 
 		public string Usage {
-			get { return "WRFLGRPC [<clients> <requests> [<streams-cnt> [<size>] [<batchsize>]]]"; }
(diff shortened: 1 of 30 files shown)


condron (dismissed):
looks good

src/EventStore.TestClient/GrpcCommands/WriteFloodProcessor.cs
```diff
@@ -0,0 +1,194 @@
...
+using System.Threading.Tasks;
+using EventStore.Client;
+using EventStore.TestClient.Commands;
+
+namespace EventStore.TestClient.GrpcCommands {
+	internal class WriteFloodProcessor : ICmdProcessor {
+		private static readonly UTF8Encoding UTF8NoBom = new UTF8Encoding(false);
+
+		public string Usage {
+			get { return "WRFLGRPC [<clients> <requests> [<streams-cnt> [<size>] [<batchsize>] <streamNamePrefix>]]"; }
```
timothycoleman, line 16:
```suggestion
			get { return "WRFLGRPC [<clients> <requests> [<streams-cnt> [<size> [<batchsize> [<stream-prefix>]]]]]"; }
<streamNamePrefix>]]"; }
```

PR #2889, opened 2021-04-01
Generalize TF and Index in preparation for LogV3

timothycoleman:
Changed: Generalized TF and Index in preparation for LogV3

In LogV2 the event records (prepare records) contain the stream name as a string. In LogV3 the stream will be identified by a long.
LogV3 also has a different record header format, which does not include the logical position of each record.

We'd prefer to keep using the existing Index and TF machinery for now since it is mature. This PR generalises that machinery to support
both log formats and their different ways of identifying streams.

I've tried to make the changes as obvious as possible, avoiding refactoring or whitespace adjustments in order to make the changes
as easy to read as possible, since it does affect quite a few lines.

### DIFF
--- a/src/EventStore.Core/LogV2/LogV2Sizer.cs
+++ b/src/EventStore.Core/LogV2/LogV2Sizer.cs
@@ -1,0 +1,7 @@
+﻿using EventStore.Core.LogAbstraction;
+
+namespace EventStore.Core.LogV2 {
+	public class LogV2Sizer : ISizer<string> {
+		public int GetSizeInBytes(string t) => 2 * t.Length;
+	}
+}
--- a/src/EventStore.Core/TransactionLog/LogRecords/IPrepareLogRecord.cs
+++ b/src/EventStore.Core/TransactionLog/LogRecords/IPrepareLogRecord.cs
@@ -1,0 +1,23 @@
+﻿using System;
+
+namespace EventStore.Core.TransactionLog.LogRecords {
+	// This interface specifies what the Storage, TF and Index machinery requires
+	// in order to handle a prepare (i.e. data) record.
+	// The V2 prepare implements it trivially
+	public interface IPrepareLogRecord : ILogRecord {
+		PrepareFlags Flags { get; }
+		long TransactionPosition { get; }
+		int TransactionOffset { get; }
+		long ExpectedVersion { get; }
+		Guid EventId { get; }
+		Guid CorrelationId { get; }
+		DateTime TimeStamp { get; }
+		string EventType { get; }
+		ReadOnlyMemory<byte> Data { get; }
+		ReadOnlyMemory<byte> Metadata { get; }
+	}
+
+	public interface IPrepareLogRecord<TStreamId> : IPrepareLogRecord {
+		TStreamId EventStreamId { get; }
+	}
+}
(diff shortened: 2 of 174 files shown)


timothycoleman:
Remind me of the status of the embedded client these days, is it still supported, do we need to worry about breaking changes to the public classes (including for example the vnodebuilder)

gregoryyoung:
"In LogV2 the event records (prepare records) contain the stream name as a string. In LogV3 the stream will be identified by a long.
LogV3 also has a different record header format, which does not include the logical position of each record."

How does this work when there is a hash collision?

timothycoleman:
> How does this work when there is a hash collision?
> Are we getting into ... the probabilities of there being 3 hash collisions is very low? :)

Greetings! The `long` in LogV3 won't be the hash of the stream name, but a unique number that'll be assigned to the stream (a counter, basically).

gregoryyoung:
There are benefits to it being a hash. Drop me a message privately if you want to discuss.

hayley-jean (dismissed):
Looks good 👍

src/EventStore.Core/TransactionLog/LogRecords/IPrepareLogRecord.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+
+namespace EventStore.Core.TransactionLog.LogRecords {
+	// This interface specifies what the Storage, TF and Index machinary requires
```
shaan1337, line 4:
```suggestion
	// This interface specifies what the Storage, TF and Index machinery requires
```

src/EventStore.Core/LogV2/LogV2Sizer.cs
```diff
@@ -0,0 +1,7 @@
+﻿using EventStore.Core.LogAbstraction;
+
+namespace EventStore.Core.LogV2 {
+	public class LogV2Sizer : ISizer<string> {
+		public int GetSizeInBytes(string t) => 2 * t.Length;
```
shaan1337, line 5:
```suggestion
		public int GetInMemorySizeInBytes(string t) => 2 * t.Length;
```

Would `GetInMemorySizeInBytes` be a better name for this method?
Based on the IndexWriter's `_committedEvents` bounded cache's code, it appears to be an estimate for the amount of memory taken by the stream name.

shaan1337 (approved):
Reviewed & Tested, looks good to me - there's just a small typo and a renaming suggestion.

src/EventStore.Core/TransactionLog/LogRecords/IPrepareLogRecord.cs
timothycoleman: good spot!

src/EventStore.Core/LogV2/LogV2Sizer.cs
timothycoleman: hmm, yes good point. I'll rename the class to reflect that rather than the method, since its a property of the implementation really rather than the interface

hayley-jean (approved):

src/EventStore.Core/LogV2/LogV2Sizer.cs
shaan1337: @timothycoleman thanks!

shaan1337 (approved):

PR #2893, opened 2021-04-06
grpc Performance Tuning

condron:
* Increase thread pool limits
* Set window and framing on kestrel for GRPC support
* Set max connections on kestrel for GRPC support


condron:
merged in to the general performance pr

PR #2908, opened 2021-04-14
Add V3 Header and Epoch raw record & infra

timothycoleman:
Added: V3 Epoch Raw Record

recommend reviewing the two commits individually, the fist one is trivial but affects more files

### DIFF
--- a/src/EventStore.LogV3/Raw.cs
+++ b/src/EventStore.LogV3/Raw.cs
@@ -1,0 +1,69 @@
+﻿using System;
+using System.Runtime.InteropServices;
+using EventStore.LogCommon;
+
+namespace EventStore.LogV3 {
+	// todo: alignment, padding
+	public static class Raw {
+		[StructLayout(LayoutKind.Explicit, Size = Size, Pack = 1)]
+		public struct RecordHeader {
+			[FieldOffset(0)] private LogRecordType _type;
+			[FieldOffset(1)] private byte _version;
+			// todo: too much padding for somethng that will occur every record
+			[FieldOffset(8)] private long _ticks;
+			[FieldOffset(16)] private Guid _recordId;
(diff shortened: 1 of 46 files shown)


src/EventStore.LogV3/Raw.cs
```diff
@@ -0,0 +1,69 @@
...
+		[StructLayout(LayoutKind.Explicit, Size = Size, Pack = 1)]
+		public struct RecordHeader {
+			[FieldOffset(0)] private LogRecordType _type;
+			[FieldOffset(1)] private byte _version;
+			// todo: too much padding for somethng that will occur every record
+			[FieldOffset(8)] private long _ticks;
+			[FieldOffset(16)] private Guid _recordId;
+
+			// todo: remove this if possible
+			[FieldOffset(48)] private long _logPosition;
```
shaan1337, line 17:
```suggestion
			[FieldOffset(32)] private long _logPosition;
```

@timothycoleman Is there a specific reason why this field starts at offset 48 instead of 32? (since Guid is 128-bit/16 bytes)

shaan1337 (commented):
nice work :+1: looks good to me, just noticed a small anomaly in one of the field offsets.

src/EventStore.LogV3/Raw.cs
timothycoleman: good spot, there was probably another field before it that got removed :D

shaan1337: great, thanks @timothycoleman . As discussed, the structs will be finalized at a later stage.

shaan1337 (dismissed):

pvanbuijtene (approved):

shaan1337 (approved):

PR #2923, opened 2021-04-20
Add LogV3 SystemStreams abstraction point

timothycoleman:
Added: ISystemStreamLookup abstraction point for LogV3

- 1024 entries reserved for virtual streams (streams that dont have a stream record)
- odd numbers used for metadata streams, which exist implicitly when the normal stream is created
- decorators handle the conversion for meta streams
- determine if a stream is a system stream by looking up its name and checking for $ prefix


src/EventStore.Core/LogV3/LogV3Metastreams.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using EventStore.Core.LogAbstraction;
+
+namespace EventStore.Core.LogV3 {
+	public class LogV3Metastreams : IMetastreamLookup<long> {
+		public bool IsMetaStream(long streamId) =>
+			streamId % 2 == 1;
```
shaan1337, line 7: @timothycoleman Should the stream ID validator also be updated to check for streamId > 1 instead of > 0?
https://github.com/EventStore/EventStore/blob/698eb69c623ce21770c096a47b1cd1d2b4733906/src/EventStore.Core/LogV3/LogV3StreamIdValidator.cs#L9

src/EventStore.Core.Tests.XUnit/LogAbstraction/LogFormatAbstractorV3Tests.cs
```diff
@@ -0,0 +1,95 @@
+﻿using EventStore.Core.LogAbstraction;
+using Xunit;
+
+namespace EventStore.Core.Tests.XUnit.LogAbstraction {
+	// check that lookups of the various combinartions of virtual/normal/meta
```
shaan1337, line 5:
```suggestion
	// check that lookups of the various combinations of virtual/normal/meta
```

shaan1337 (commented):
nice work, looks good to me :+1: just added a suggestion for the stream id validator and to fix a typo.

src/EventStore.Core/LogV3/LogV3Metastreams.cs
timothycoleman: because 1 would be the metastream of 'no stream', which doesn't make any sense, so yep! thats a great spot

shaan1337 (dismissed):

src/EventStore.Core/LogV3/LogV3Metastreams.cs
timothycoleman: changed this to non-negative in the end, need to accept 0 in the same way that the v3 version accepts strings that dont actually correspond to streams in the database

src/EventStore.Core.Tests.XUnit/LogAbstraction/LogFormatAbstractorV3Tests.cs
```diff
@@ -0,0 +1,145 @@
+﻿using EventStore.Core.LogAbstraction;
+using Xunit;
+
+namespace EventStore.Core.Tests.XUnit.LogAbstraction {
+	// check that lookups of the various combinations of virtual/normal/meta
+	// work in both directions and in the stream index.
+	// todo: ideally check that no unexpected (i.e. meta or virtual) streams have been
+	// added to the underlying index after each test.
```
pvanbuijtene, line 8: @timothycoleman can we start using xml comments to make this more ide friendly ?

timothycoleman: yeah fair enough for comments that describe an api. in this case though it's a test class so im not sure there is anywhere useful the xml comments would appear?

pvanbuijtene: I would normalize to one way of documenting classes and methods, will make things a bit easier.

shaan1337 (approved):

pvanbuijtene (approved):

src/EventStore.Core.Tests.XUnit/LogAbstraction/LogFormatAbstractorV3Tests.cs
timothycoleman: i could be persuaded on this, but perhaps generally the xmldoc is for a developer who's using the class's api (instantiating it, calling methods) and ordinary comments are for a developer who is looking at the code inside the class at implementation details, perhaps considering making a code change or hunting a bug

PR #2930, opened 2021-04-25
Log v3: Simple Stream Writes

timothycoleman:
Added: Simple stream writes for LogV3

This supports one event per streamwrite record. TransactionPosition and TransactionOffset are therefore temporarily required, to continue supporting implicit transactions. They have been added to the SystemMetadata.

Tests have been updated to run against both Log formats where appropriate


src/EventStore.Core/Services/Storage/ReaderIndex/IndexReader.cs
```diff
@@ -437,7 +437,7 @@ public class IndexReader<TStreamId> : IndexReader, IIndexReader<TStreamId> {
 		}
 
 		private bool OriginalStreamExists(TFReaderLease reader, TStreamId metaStreamId) {
-			if (_systemStreams.IsSystemStream(metaStreamId)) {
```
timothycoleman, line 440: i wonder if this was just a bug or if theres some reason it was this way

src/EventStore.Core/TransactionLog/LogRecords/LogRecord.cs
```diff
@@ -23,6 +23,10 @@ public abstract class LogRecord : ILogRecord {
 		}
 
 		public static ILogRecord ReadFrom(BinaryReader reader) {
+			return ReadFrom(reader, 0);
```
shaan1337, line 26: should the length be equal to the binary reader's stream length here?

src/EventStore.Core/Services/Storage/ReaderIndex/IndexReader.cs
shaan1337: looks like a bug to me too.
Based on the PR that introduced it: https://github.com/EventStore/EventStore/pull/709, it looks like the error has been repeated here as well (.IsSystemStream instead of IsMetaStream):
https://github.com/EventStore/EventStore/blob/master/src/EventStore.Core/Services/Storage/StorageReaderWorker.cs#L280

src/EventStore.LogV3/Raw.cs
```diff
@@ -66,6 +69,48 @@ public struct EpochHeader {
...
 
+		[StructLayout(LayoutKind.Explicit, Size = Size, Pack = 1)]
+		public struct EventHeader {
+			[FieldOffset(0)] public Guid _eventId;
+			// todo: not used yet
+			[FieldOffset(16)] public int _eventTypeNumber;
+			[FieldOffset(20)] public PrepareFlags _flags;
+			[FieldOffset(24)] public int _eventSize;
+			[FieldOffset(28)] public int _systemMetadataSize;
+			[FieldOffset(32)] public int _dataSize;
```
shaan1337, line 80: does the metadata size need to be in the event header too?

src/EventStore.LogV3/Raw.cs
```diff
@@ -87,5 +132,40 @@ public struct StreamTypeHeader {
...
 			}
 		}
+
+		[StructLayout(LayoutKind.Explicit, Size = Size, Pack = 1)]
+		public struct StreamWriteHeader {
+			[FieldOffset(0)] public long _streamNumber;
+			[FieldOffset(8)] public long _startingEventNumber;
+			[FieldOffset(16)] public short _flags;
+			[FieldOffset(18)] public short _count;
+			[FieldOffset(20)] public int _metadataSize;
```
shaan1337, line 142: is the naming correct?  (based on the log v3 specs, it's called system metadata size)

src/EventStore.LogV3/Raw.cs
timothycoleman: at the moment its implicit from the others (since we know the overall _eventSize and know the sizes of the other two dynamically sized components (data and system meta)

but its possible we'll need it later depending on how we go about doing alignment/padding

shaan1337: @timothycoleman thanks, sounds good :+1:

src/EventStore.Core/TransactionLog/LogRecords/LogRecord.cs
timothycoleman: removed this method

pvanbuijtene (approved):

src/EventStore.Core/Services/Storage/ReaderIndex/IndexReader.cs
shaan1337: i've fixed it in this commit: 5eadcf62f3d061da9716782cda2879b0820d5691

timothycoleman:
I've reviewed and approve @shaan1337 s changes

shaan1337 (approved):

condron (approved):
Looks good

PR #2934, opened 2021-05-03
Add auto-configuration for caching 20.10

hayley-jean:
Added: auto configuration for stream cache, reader threads and worker threads.
Changed: gRPC settings

Merge in performance changes for 20.10.3
This PR contains most of the changes in https://github.com/EventStore/EventStore/pull/2902:

Server improvements:
* Auto-size LRU Cache based on the amount of free memory at startup.
* Auto-configure the number of reader threads based on free memory at startup.
* Auto-configure the worker threads based on the reader thread count

Test Client improvements:
* Add gRPC write flood command
* Add TCP ClientAPI write flood command (as opposed to raw TCP)
* Document Test Client Public methods

**Note**: The commits in this PR need to be preserved as commit b4d0bc8a125d16a44f23c392fea20f2aba73ba23 has been tagged for 20.10.3.


src/EventStore.Core/ClusterVNodeStartup.cs
```diff
@@ -154,7 +155,12 @@ IServiceProvider IStartup.ConfigureServices(IServiceCollection services) => Conf
 						.AddGrpc()
 						.AddServiceOptions<Streams>(options =>
 							options.MaxReceiveMessageSize = TFConsts.EffectiveMaxLogRecordSize)
-						.Services,
+						.Services.Configure<KestrelServerOptions>(options => {
+							options.Limits.MaxConcurrentConnections = 5000;
+							options.Limits.MaxConcurrentUpgradedConnections = 5000;
+							options.Limits.Http2.InitialConnectionWindowSize = 131072 * 1024;
+							options.Limits.Http2.InitialStreamWindowSize = 98304 * 1024;
+						}),
```
jageall, line 163: These need to be moved to the hostbuilder to have an effect

jageall (dismissed):

src/EventStore.Core/ClusterVNodeStartup.cs
jageall: They should really have configuration options too

condron: not sure they should have cong. options

hayley-jean: I've moved the settings and added configuration options in the same place as the keepAlive settings

hayley-jean:
@ahjohannessen We may decide to go with the approach in PR https://github.com/EventStore/EventStore/pull/2949 for configuring kestrel instead of providing options within Event Store itself. If you have any suggestions for a good default for `MaxConcurrentStreams`, could you comment on that PR please?

timothycoleman:
Is it right that the kestrel config changes aren't in the back port? If so I think this is all set

thefringeninja (approved):

hayley-jean:
> Is it right that the kestrel config changes aren't in the back port? If so I think this is all set

I'll open a follow-up PR once this is merged with the kestrel settings changes

timothycoleman (approved):
looks good 👍

Issue #2946, opened 2021-05-09
System.AccessViolation on projection


StevenBlair123:
**Describe the bug**
When a projection hits a specific event, a System.AccessViolationException is thrown, causing the Eventstore to restart.
This process continues until you can manage to disable the projection (which is easier said than done).
I think the problem is related to subscribing to streams which have been truncated.

**To Reproduce**
1. Create a projection subscribing to a stream which has been impacted by a scavenge (for example an $et)
2. Run the projection
3. Once the projection starts processing from this, Exception will be thrown.

**Expected behavior**
I think these "ghost events" (can't think of a better way of describing them)

**Actual behavior**
Projection causes the AccessViolation exception as soon as the ghost event is processed (I have verified this by debugging)

**Config/Logs/Screenshots**
If applicable, please attach your node configuration, logs or any screenshots.

![image (1)](https://user-images.githubusercontent.com/13871941/117563870-a423d600-b0a0-11eb-90ab-572b51c3e295.png)

![image](https://user-images.githubusercontent.com/13871941/117563986-6e332180-b0a1-11eb-8e6f-2b546e99af70.png)

**EventStore details**
 - EventStore server version: 21.2.1.0

 - Operating system: Windows 10

 - EventStore client version (if applicable):

**Additional context**

StevenBlair123:
```
public string GetStatePartition(
	CheckpointTag eventPosition, string category, ResolvedEvent @event) {
	CheckDisposed();


	if (@event == null) throw new ArgumentNullException("event");

	var eventData = GetEventData(@event); //Will return null rather than String.Empty

	if (eventData == null) return null;

	var partition = _query.GetPartition(
		GetEventData(@event),
		new string[] {
			@event.EventStreamId, @event.IsJson ? "1" : "", @event.EventType, category ?? "",
			@event.EventSequenceNumber.ToString(CultureInfo.InvariantCulture), @event.Metadata ?? "",
			@event.PositionMetadata ?? ""
		});
	if (partition == "")
		return null;
	else
		return partition;
}
```

StevenBlair123:
#2947

alexeyzimarev:
Thanks! I will follow up on this one

hayley-jean:
Thank you for reporting this, I've managed to reproduce the issue on my side with the following projection:

```
fromStreams("$et-foo", "$et-test")
.partitionBy(function(e) {
    e.eventType
})
.when({
    $init: function(s,e) {
        return { count: 0 }
[...]

PR #2947, opened 2021-05-09
Prevent scavenged events from being passed to ExecuteHandler

StevenBlair123:
#2946 

Scavenged events (at least from $et) were being passed onto ExecuteHandler causing System.AccessViolation to be thrown (and restarting the Eventstore).
Check is done from GetEventData before deciding to call  _query.GetPartition.


src/EventStore.Projections.Core/Services/v8/V8ProjectionStateHandler.cs
```diff
@@ -126,17 +129,24 @@ public class EmittedEventJsonContract {
...
-				GetEventData(@event),
-				new string[] {
-					@event.EventStreamId, @event.IsJson ? "1" : "", @event.EventType, category ?? "",
-					@event.EventSequenceNumber.ToString(CultureInfo.InvariantCulture), @event.Metadata ?? "",
-					@event.PositionMetadata ?? ""
-				});
+
+			string eventData = GetEventData(@event);
+
+			if (string.IsNullOrEmpty(eventData)) {
```
hayley-jean, line 135: Thanks for the PR @StevenBlair123!

I've tested it locally and it does fix the AccessViolation issue, however there are cases where events with empty data should still be handled.
For example, handling the metadata of an empty non-json event.

It would be better to check the EventType of the ResolvedEvent, which is an empty string for unresolved LinkTos

For example:

```suggestion
			if (string.IsNullOrEmpty(@event.EventType)) {
```

hayley-jean (changes requested):

src/EventStore.Projections.Core/Services/v8/V8ProjectionStateHandler.cs
StevenBlair123: Made the changed and push.

StevenBlair123:
@hayley-jean - we just saw the same problem in the ProcessEvent.

Should we change the code to check the EventType in ProcessPartitionCreated and ProcessEvent

So for ProcessPartitionCreated return true if EventType is null / empty
And for ProcessEvent fall into the if statement if EventType is null / empty?

hayley-jean:
@StevenBlair123 Yeah, we should check the event type in ProcessEvent and ProcessPartitionCreated as well

StevenBlair123:
@hayley-jean - I updated ProcessPartitionCreated and ProcessEvent with the change, but ProcessEvent is causing tests to fail.
Unfortunately, the units are not running locally for me (can't quite work out why) so it's pretty difficult to work out the issue with ProcessEvent.

For now, would the two changes be acceptable (we are really keen to get this in to the next release).
Unless you have any ideas why the ProcessEvent is failing?

hayley-jean (changes requested):
Thanks, we're also keen on getting these changes in the next release!

I've added a couple of suggestions which should get the ProcessEvent and ProcessPartitionCreated methods working properly. 
The tests should pass with those changes

src/EventStore.Projections.Core/Services/v8/V8ProjectionStateHandler.cs
```diff
@@ -10,6 +10,9 @@
 using Serilog;
 
 namespace EventStore.Projections.Core.Services.v8 {
+	using System.Threading;
```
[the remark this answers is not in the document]
StevenBlair123: Fixed

src/EventStore.Projections.Core/Services/v8/V8ProjectionStateHandler.cs
```diff
@@ -126,29 +129,36 @@ public class EmittedEventJsonContract {
...
+		public bool ProcessEvent(string partition, CheckpointTag eventPosition, string category, ResolvedEvent @event, out string newState, out string newSharedState, out EmittedEventEnvelope[] emittedEvents) {
 			CheckDisposed();
 			_eventPosition = eventPosition;
 			_emittedEvents = null;
 			Tuple<string, string> newStates = null;
 
 			var data = GetEventData(@event);
-			if (@event == null || data == null) {
+
+			if (@event == null || data == null ) {
```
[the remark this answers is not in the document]
StevenBlair123: I have a funny feeling I was using data is not empty here rather than EventType by mistake :|

src/EventStore.Projections.Core/Services/v8/V8ProjectionStateHandler.cs
```diff
@@ -170,34 +180,35 @@ public class EmittedEventJsonContract {
...
-			out EmittedEventEnvelope[] emittedEvents) {
+		                                    out EmittedEventEnvelope[] emittedEvents) {
 			CheckDisposed();
 			_eventPosition = createPosition;
 			_emittedEvents = null;
 
 			var data = GetEventData(@event);
-			if (@event == null || data == null) {
+
+			if (@event == null || data == null || string.IsNullOrEmpty(data)) {
```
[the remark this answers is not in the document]
StevenBlair123: This is what happens when I code before my second coffee.

StevenBlair123:
@hayley-jean - seems like the same issue has happened.
I am 99% sure its inside ProcessEvent:

hayley-jean (approved):
Thanks! Looks good to me now 👍

hayley-jean:
> @hayley-jean - seems like the same issue has happened.
> I am 99% sure its inside ProcessEvent:
> 
> if (@event == null || data == null || string.IsNullOrEmpty(@event.EventType)) {
> 
> When I first tried this change, actually changed GetEventData itself to return null, rather than an empty string, and the unit tests were failing.
> 
> Any ideas?

The failing tests don't seem related to this PR, so won't block us merging it once it's approved.

The unit test that originally failed was likely [this one](https://github.com/EventStore/EventStore/blob/master/src/EventStore.Projections.Core.Tests/Services/v8/when_running_with_content_type_validation.cs#L46) which verifies that non-json events with null data can still be processed. This test is passing now

StevenBlair123:
I just made a small tweak to the PR. I had introduced a new property on options inside Projections.js (#2960) and forgot to take it out.

hayley-jean (approved):

src/EventStore.Projections.Core/Services/v8/V8ProjectionStateHandler.cs
```diff
@@ -170,34 +178,35 @@ public class EmittedEventJsonContract {
...
+			
+			            }
+			            catch (InvalidCastException)
+			            {
+			                Console.Error.WriteLine(newState);
+			            }
+			            catch (JsonException)
+			            {
+			                Console.Error.WriteLine(newState);
+			            }*/
```
YoEight, line 196: Safe to say we can delete those lines.

YoEight (changes requested):
Looks good to me. Some minor remarks. Could you format the code properly, please? `dotnet format` should do if you don't use Rider.

src/EventStore.Projections.Core/Services/v8/V8ProjectionStateHandler.cs
StevenBlair123: Fixed. And I have ran "dotnet format"

StevenBlair123:
Can this be closed off now?

YoEight:
Merged: https://github.com/EventStore/EventStore/commit/244170238f56897811f3828a414a67a01e21c375

PR #2951, opened 2021-05-12
Use an interpreter for user projections

jageall:
Added: Interpreter runtime for user projections
Added: Options to switch user runtime back to legacy v8
Changed: Change the user projection runtime to use an interpreter rather than v8

Fixes: #2946 , home#500
Notes:
The option to switch runtimes is marked as deprecated so that if legacy is selected it will print the deprecation warning


hayley-jean (changes requested):
I've added a spec that fails due to the body of the event being different to the legacy runtime

We should also look at adding checks similar to the ones in https://github.com/EventStore/EventStore/pull/2947 to filter out scavenged events

jageall:
> I've added a spec that fails due to the body of the event being different to the legacy runtime

Fixed

YoEight (approved):

src/EventStore.Projections.Core.Tests/Services/Jint/when_running_emitting_js_projection.cs
```diff
@@ -0,0 +1,75 @@
+﻿using System;
+using System.Diagnostics;
+using EventStore.Projections.Core.Services;
+using EventStore.Projections.Core.Services.Processing;
+using NUnit.Framework;
+
+namespace EventStore.Projections.Core.Tests.Services.Jint
+{
```
thefringeninja, line 8: I think the formatting here is wrong

thefringeninja (changes requested):

hayley-jean (approved):

YoEight (approved):

PR #2955, opened 2021-05-12
Log v3/stream records

timothycoleman:
Main work:
- Add Stream records
- Index them
- Use the index for the id -> name lookup

Misc work:
- Move `CopyForRetry` to IPrepareLogRecord<> so that each kind of prepare can create a copy of the same type

Further work:
- resolve //qq's
- unit test adjustments
     - stuff in TransactionLog/LogRecords might need some testing attention (running existing tests against v3)
     - same for Core/Services/Storage
     - the rest should be covered in the tests already


timothycoleman:
accidentally the branch

timothycoleman:
combining this pr with https://github.com/EventStore/EventStore/pull/2959

Issue #2954, opened 2021-05-12
Projection stream position does not move from -1


StevenBlair123:
**Describe the bug**
When creating a projection which reads from a truncated stream ($et), the position on any of streams remains at -1

**To Reproduce**

1. Truncate a stream (tried using a Delete from UI and maxAge) - note the stream tends to have millions of truncated events.
2. Create a projection which uses this truncated stream
3. Projection will remain at Position -1

You can reset the system projections to get the same effect.

**Expected behavior**
Should be able to read from the streams.

**Actual behavior**
Projection does not move position.

**Config/Logs/Screenshots**
If applicable, please attach your node configuration, logs or any screenshots.

**Additional context**
https://discuss.eventstore.com/t/projection-stream-position-does-not-move-from-1/3232
https://github.com/EventStore/EventStore/issues/1872

StevenBlair123:
It's been a while since this was originally spotted, but we think we have found the source of the problem.

Inside `StorageReaderWorker.ReadStreamEventsForward`, the code reads from the stream using `ReadStreamEventsForward`
The problem is when the stream has been truncated, the ReadStreamEventsForward fromEventNumber starts at 0.
When result is returned, the metadata indicates that the stream has indeed been truncated (TruncateBefore) and that seems to be the end of us trying to read events from here.

As a proof of concept, we have have added the following code in after the ReadStreamEventsForward:

```
if (result.Metadata != null && result.Metadata.TruncateBefore != null && result.Records.Length == 0) {
                        result =
                            _readIndex.ReadStreamEventsForward(streamName, streamId, result.NextEventNumber, msg.MaxCount);
                    }
```

So in short, if we have TruncatedBefore and no events, try again from result.NextEventNumber.

After applying this, our projections burst into life!

Any problems with this approach (we are going to clean the code up and create a PR shortly)?

PR #2956, opened 2021-05-13
Handle truncated streams when reading events for projections

StuartFergusonVme:
Fixed: Fix projections getting stuck when reading from truncated streams

Fixes https://github.com/EventStore/EventStore/issues/2954

### DIFF
--- a/src/EventStore.Core.Tests/Http/Streams/feed.cs
+++ b/src/EventStore.Core.Tests/Http/Streams/feed.cs
@@ -415,7 +415,7 @@ namespace EventStore.Core.Tests.Http.Streams {
 			private List<JToken> _entries;
 
 			protected override async Task When() {
-				_feed = await GetJson<JObject>("/streams/" + LinkedStreamName + "/0/forward/10?embed=rich",
+				_feed = await GetJson<JObject>("/streams/" + LinkedStreamName + "/0/forward/10",extra: "embed=rich",
 					accept: ContentType.Json);
 				_entries = _feed != null ? _feed["entries"].ToList() : new List<JToken>();
 			}
--- a/src/EventStore.Core/Services/Storage/StorageReaderWorker.cs
+++ b/src/EventStore.Core/Services/Storage/StorageReaderWorker.cs
@@ -309,6 +309,7 @@ namespace EventStore.Core.Services.Storage {
 
 					var result =
 						_readIndex.ReadStreamEventsForward(streamName, streamId, msg.FromEventNumber, msg.MaxCount);
+					
 					CheckEventsOrder(msg, result);
 					var resolvedPairs = ResolveLinkToEvents(result.Records, msg.ResolveLinkTos, msg.User);
 					if (resolvedPairs == null)
--- a/src/EventStore.Projections.Core/Services/Processing/EventByTypeIndexEventReader.cs
+++ b/src/EventStore.Projections.Core/Services/Processing/EventByTypeIndexEventReader.cs
@@ -251,11 +251,11 @@ namespace EventStore.Projections.Core.Services.Processing {
 						break;
 					case ReadStreamResult.Success:
 						_reader.UpdateNextStreamPosition(message.EventStreamId, message.NextEventNumber);
-						var isEof = message.Events.Length == 0;
-						_eofs[message.EventStreamId] = isEof;
+						_eofs[message.EventStreamId] = (message.Events.Length == 0) && message.IsEndOfStream;
 						EnqueueEvents(message);
 						ProcessBuffersAndContinue(eventStreamId: message.EventStreamId);
 						break;
+				
 					default:
 						throw new NotSupportedException(
 							String.Format("ReadEvents result code was not recognized. Code: {0}", message.Result));
--- a/src/EventStore.Projections.Core/Services/Processing/MultiStreamEventReader.cs
+++ b/src/EventStore.Projections.Core/Services/Processing/MultiStreamEventReader.cs
@@ -22,7 +22,6 @@ namespace EventStore.Projections.Core.Services.Processing {
 		private CheckpointTag _fromPositions;
 		private readonly bool _resolveLinkTos;
 		private readonly ITimeProvider _timeProvider;
-
 		private readonly HashSet<string> _eventsRequested = new HashSet<string>();
 		private readonly Dictionary<string, long?> _preparePositions = new Dictionary<string, long?>();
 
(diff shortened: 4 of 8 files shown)


StuartFergusonVme:
#2954

StevenBlair123:
Any chance we could get this approved to run on the CI process?
Just need to verify tests etc are working

hayley-jean:
Thank you for the PR @StuartFergusonVme!
However, changing the StorageReaderWorker isn't what we'd want to do in this situation.

When a stream has been truncated, the storage reader should return an empty response with NextEventNumber set to the event number of the next available event to read from and IsEndOfStream set to false. It is then up to the client to issue another read from there.

It looks like the projections aren't making use of the IsEndOfStream property of the read, and simply stopping when there are no events.
Would you mind updating the EOF logic in the following readers to use the IsEndOfStream property instead, and see if that fixes the issue?

- [EventByTypeIndexEventReader](https://github.com/EventStore/EventStore/blob/master/src/EventStore.Projections.Core/Services/Processing/EventByTypeIndexEventReader.cs#L254)
- [MultiStreamEventReader](https://github.com/EventStore/EventStore/blob/3b8ca301eb2b2f1a36e7fe2146fe1a03f970cb20/src/EventStore.Projections.Core/Services/Processing/MultiStreamEventReader.cs#L124)
- [StreamEventReader](https://github.com/EventStore/EventStore/blob/master/src/EventStore.Projections.Core/Services/Processing/StreamEventReader.cs#L93)
- [TransactionFileEventReader](https://github.com/EventStore/EventStore/blob/master/src/EventStore.Projections.Core/Services/Processing/TransactionFileEventReader.cs#L66)

StuartFergusonVme:
@hayley-jean updated as requested.

I have added the changes in from @2947 with this as well as without these changes the Access Violation error is hit that has been fixes with that PR.

StevenBlair123:
Could someone approve this so the workflow can kick off?

hayley-jean:
@StuartFergusonVme Thanks for updating! However when I test this locally I see that this change is actually more involved than I initially thought

The projection readers assume that EOF means that there are no events, so don't process events or continue reading once it sets EOF to true - resulting in the projection still getting stuck.

For example, the StreamEventReader only publishes the events to the projections if it's [not EOF](https://github.com/EventStore/EventStore/blob/38ba91a8d7109d39096e7148f5301ce54277deab/src/EventStore.Projections.Core/Services/Processing/StreamEventReader.cs#L110), and the other readers appear to have similar issues.
Would you be willing to fix this in the readers?

StevenBlair123:
@hayley-jean - does that mean we need to check message.events.length

So we would do:

`var eof = (message.events.length == 0) && message.IsEndOfStream`

hayley-jean:
I think that could work @StevenBlair123, it would be the closest to what the readers seem to expect at the moment

StevenBlair123:
OK, we will plug that in just now.

StuartFergusonVme:
@hayley-jean pushed up some more updates as requested, can you approve the CI process please?

StuartFergusonVme:
@hayley-jean just noticed from the last run that the test "EventStore.Core.Tests.ClientAPI.when_connection_drops_messages_that_have_run_out_of_retries_are_not_retried" failed only on ubuntu-18.04 with the error 
Error Message:
2021-05-19T07:14:58.5890544Z      Expected: No Exception to be thrown
2021-05-19T07:14:58.5892880Z   But was:  <System.InvalidOperationException: Subscription group 91101607-5696-46ad-80a2-82449b9fc462 on stream a213f024-620b-47a0-8dfe-3e718915473b already exists

Is this just a timing issue when running the tests as I can't see how the code changes I have made can affect this especially as it is passing on windows?

hayley-jean:
@StuartFergusonVme Thanks! The changes you've made fixes the issue for stream, event type and $all projections 😄

The tests aren't failing because of this PR - they've been rather flaky recently. If this PR breaks anything it would only be in Projections.Core.Tests

The only one left that doesn't work is multi-stream projections, for example:

```fromStreams("foo", "bar")
.when({
    $init: function() {
        return { count: 0 }
    },
    $any: function(s,e) {
        s.count++;
    }
})
```

There's probably something preventing the read from moving on in the case when no events are returned, if you want to have a look at that in the MultiEventReader.
Otherwise I can see if I can push a fix for the multi stream projections later today or tomorrow

StevenBlair123:
@hayley-jean  - we had actually tested this specifically on multi streams (one of the streams was the root of the problem) and it looked like this code was working.
Handle is called for each, and our check should verify if we were at the end of stream.
Clearly, there is something we have missed, but could you give us some guidance on this?

hayley-jean:
@StevenBlair123 Sure, the test I ran was:

1. Write 2000 events to stream `foo`. Leave stream `bar` empty
2. Delete `foo`
3. Write another 20 events to `foo`
4. Create the following projection

```
fromStreams("foo", "bar")
.when({
    $init: function() {
        return { count: 0 }
    },
    $any: function(s,e) {
        s.count++;
    }
})
```

How many truncated events did you test with?
The multi-stream projection reads large batches of events at a time, so it may still have received some events which allowed it to continue running.

As for fixing it, it looks like the positions that are used in the reads are only updated when events are delivered (see [here](https://github.com/EventStore/EventStore/blob/master/src/EventStore.Projections.Core/Services/Processing/MultiStreamEventReader.cs#L329)).
This means that the initial empty read won't update the positions, resulting in the projection reading from position 0 over and over again. So updating the `_fromPositions` on an empty non-eof read should fix this

StevenBlair123:
@hayley-jean  - thanks. we will have a look at this tomorrow.

hayley-jean (approved):
Thanks! This all works on my side now 👍

YoEight (changes requested):
Thanks for your contribution! It looks good to me too. Could you get rid of unnecessary changes and squash your commit too?

StuartFergusonVme:
@YoEight I think I have done what you have asked for but the CI process is blocked now, would you be able to approve the workflow and confirm the change I have done is as you requested.

StevenBlair123:
Could we get someone to accept workflow for CI?

YoEight:
@StuartFergusonVme I expected you squashed your commits too :smile: 

You back merged on your PR branch, let me see what I can do, otherwise, I might have to create another PR to merge this.

YoEight (approved):

YoEight:
@StuartFergusonVme @StevenBlair123 I didn't manage to fix your PR so I opened a new one here: https://github.com/EventStore/EventStore/pull/2979

PR #2958, opened 2021-05-13
Adjust V3 schema

timothycoleman:
Updated: Assorted minor adjustments to V3 schema following discussions

- Use 6 bytes for timestamp at the expense of precision
  (populated by DateTime.UtcNow which is of similar precision)
- PartitionHeader: ushort enum for flags, add reference number
- EventTypeNumber: unsigned
- Add parent topic number
- Add starting event numbers to metadata
  metadata for flexibility, and protobufs has variable length encoding
- Adjust field order so no padding required

### DIFF
--- a/src/EventStore.Core.Tests/TransactionLog/when_writing_prepare_record_to_file.cs
+++ b/src/EventStore.Core.Tests/TransactionLog/when_writing_prepare_record_to_file.cs
@@ -77,7 +77,7 @@ namespace EventStore.Core.Tests.TransactionLog {
 				Assert.AreEqual(p.EventId, _eventId);
 				Assert.AreEqual(p.EventStreamId, streamId);
 				Assert.AreEqual(p.ExpectedVersion, 1234);
-				Assert.AreEqual(p.TimeStamp, new DateTime(2012, 12, 21));
+				Assert.That(p.TimeStamp, Is.EqualTo(new DateTime(2012, 12, 21)).Within(20).Milliseconds);
 				Assert.AreEqual(p.Flags, PrepareFlags.SingleWrite);
 				Assert.AreEqual(p.EventType, "type");
 				Assert.AreEqual(p.Data.Length, 5);
--- a/src/EventStore.Core.Tests/TransactionLog/when_writing_prepare_record_to_file.cs
+++ b/src/EventStore.Core.Tests/TransactionLog/when_writing_prepare_record_to_file.cs
@@ -77,7 +77,7 @@ namespace EventStore.Core.Tests.TransactionLog {
 				Assert.AreEqual(p.EventId, _eventId);
 				Assert.AreEqual(p.EventStreamId, streamId);
 				Assert.AreEqual(p.ExpectedVersion, 1234);
-				Assert.That(p.TimeStamp, Is.EqualTo(new DateTime(2012, 12, 21)).Within(20).Milliseconds);
+				Assert.That(p.TimeStamp, Is.EqualTo(new DateTime(2012, 12, 21)).Within(7).Milliseconds);
 				Assert.AreEqual(p.Flags, PrepareFlags.SingleWrite);
 				Assert.AreEqual(p.EventType, "type");
 				Assert.AreEqual(p.Data.Length, 5);
--- a/src/EventStore.Core/LogV3/LogV3RecordFactory.cs
+++ b/src/EventStore.Core/LogV3/LogV3RecordFactory.cs
@@ -5,6 +5,12 @@ using EventStore.Core.TransactionLog.LogRecords;
 namespace EventStore.Core.LogV3 {
 	public class LogV3RecordFactory : IRecordFactory<long> {
 		public LogV3RecordFactory() {
+			if (!BitConverter.IsLittleEndian) {
+				// to support big endian we would need to adjust some of the bit
+				// operatiosn in the raw v3 structs, and adjust the way that the
+				// v3 records are written/read from disk (currently blitted)
+				throw new NotSupportedException();
+			}
 		}
 
 		public ISystemLogRecord CreateEpoch(EpochRecord epoch) {
--- a/src/EventStore.LogV3.Tests/RecordHeaderTests.cs
+++ b/src/EventStore.LogV3.Tests/RecordHeaderTests.cs
@@ -6,6 +6,7 @@ namespace EventStore.LogV3.Tests {
 		[Fact]
 		public void can_set_time_stamp() {
 			var header = new Raw.RecordHeader {
+				TimeStamp = new DateTime(2000, 03, 21),
 				Type = LogCommon.LogRecordType.EventType,
 				Version = 0x45,
 			};
--- a/src/EventStore.LogV3.Tests/RecordHeaderTests.cs
+++ b/src/EventStore.LogV3.Tests/RecordHeaderTests.cs
@@ -1,0 +1,36 @@
+﻿using System;
+using Xunit;
+
+namespace EventStore.LogV3.Tests {
+	public class RecordHeaderTests {
+		[Fact]
+		public void can_set_time_stamp() {
+			var header = new Raw.RecordHeader {
+				Type = LogCommon.LogRecordType.EventType,
+				Version = 0x45,
+			};
+
+			var now = DateTime.UtcNow;
+			header.TimeStamp = now;
+
+			Assert.Equal(LogCommon.LogRecordType.EventType, header.Type);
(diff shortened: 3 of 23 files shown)


src/EventStore.Core/LogV3/LogV3RecordFactory.cs
```diff
@@ -5,6 +5,12 @@
 namespace EventStore.Core.LogV3 {
 	public class LogV3RecordFactory : IRecordFactory<long> {
 		public LogV3RecordFactory() {
+			if (!BitConverter.IsLittleEndian) {
```
shaan1337, line 8: isn't BinaryWriter (what we use to write/read log records to/from disk) little endian by default? (which should be independent of the memory endianness)

shaan1337: ah nevermind, got it - it's due to the structs

src/EventStore.LogV3/Raw.cs
```diff
@@ -33,8 +31,9 @@ public struct RecordHeader {
 			}
 
 			public DateTime TimeStamp {
-				get => new DateTime(_ticks);
-				set => _ticks = value.Ticks;
+				// little-endian. blank out the 2 LEAST significant bytes
+				get => new(_ticks & SixMostSignificantBytes);
+				set => _ticks |= value.Ticks & SixMostSignificantBytes;
```
shaan1337, line 36: should this be: `_ticks = value.Ticks & SixMostSignificantBytes` ?

timothycoleman: we need the |= so that it doesn't overwrite the two least significant bytes

shaan1337: @timothycoleman  ah ok, got it. the issue I was trying to convey (before I understood the purpose of the OR) is that the 6 most significant bytes from the existing value of ticks are being ORed into the result. so setting _ticks twice could result in an invalid value.

this would prevent it:
`_ticks = (_ticks & ~SixMostSignificantBytes) | (value.Ticks & SixMostSignificantBytes)`

src/EventStore.Core/LogV3/LogV3RecordFactory.cs
timothycoleman: yeah the v3 records are using the binary writer, but not field by field. we're just giving it the whole span of memory for the record. there is similar 'islittleendian' check in the grpc code so it looks like only little endian is supported at the moment anyway, so we decided that if we want to have big-endian later we can address it then. the way we'd support big endian is by adding some translation so that the files are always littleendian, so we aren't producing invalid files in the mean time

src/EventStore.LogV3/Raw.cs
timothycoleman: 🤦🤦🤦 thanks

src/EventStore.LogV3/Raw.cs
```diff
@@ -72,8 +80,21 @@ public struct StreamWriteId {
...
+				}
+				set {
+					if (value == EventNumberDeletedStream) {
+						_startingEventNumber |= SixLeastSignificantBytes;
+						return;
+					}
+
+					if ((value | SixLeastSignificantBytes) != SixLeastSignificantBytes)
+						throw new ArgumentOutOfRangeException(nameof(value), value, null);
+					_startingEventNumber |= value;
```
shaan1337, line 96: Same here:
_startingEventNumber = (_startingEventNumber & ~SixLeastSignificantBytes) | value;

src/EventStore.LogV3/Raw.cs
```diff
@@ -72,8 +80,21 @@ public struct StreamWriteId {
...
-				get => _startingEventNumber;
-				set => _startingEventNumber = value;
+				// little-endian. must not write the 2 MOST significant bytes
+				get {
+					var x = _startingEventNumber & SixLeastSignificantBytes;
+					return x == SixLeastSignificantBytes ? EventNumberDeletedStream : x;
+				}
+				set {
+					if (value == EventNumberDeletedStream) {
+						_startingEventNumber |= SixLeastSignificantBytes;
```
shaan1337, line 90: This one's ok since they are all ones being ORed anyway but for clarity it could be changed to:
_startingEventNumber = (_startingEventNumber & ~SixLeastSignificantBytes) | SixLeastSignificantBytes;

src/EventStore.LogV3/Raw.cs
```diff
@@ -50,11 +49,20 @@ public struct RecordHeader {
 
 		[StructLayout(LayoutKind.Explicit, Size = RecordHeader.RecordIdSize, Pack = 1)]
```
shaan1337, line 50: is there a specific reason for setting `Size = RecordHeader.RecordIdSize` instead of adding a constant `Size = 16` in the `StreamWriteId` struct? (they do not seem related)

src/EventStore.Core.Tests/TransactionLog/when_writing_prepare_record_to_file.cs
```diff
@@ -77,7 +77,7 @@ public class when_writing_prepare_record_to_file<TLogFormat, TStreamId> : Specif
 				Assert.AreEqual(p.EventId, _eventId);
 				Assert.AreEqual(p.EventStreamId, streamId);
 				Assert.AreEqual(p.ExpectedVersion, 1234);
-				Assert.AreEqual(p.TimeStamp, new DateTime(2012, 12, 21));
+				Assert.That(p.TimeStamp, Is.EqualTo(new DateTime(2012, 12, 21)).Within(20).Milliseconds);
```
shaan1337, line 80: Can we use a tighter bound of 7 milliseconds here (or maybe 10)? (16 bits = 65536 ticks ~= 6.5 ms < 7 ms)

shaan1337 (commented):
nice optimizations :+1: i've just added a few suggestions, otherwise the rest looks good to me

src/EventStore.LogV3/Raw.cs
timothycoleman: fixed, nice save that could have been a pain!

src/EventStore.Core.Tests/TransactionLog/when_writing_prepare_record_to_file.cs
timothycoleman: sure, i'll set it to 7

src/EventStore.LogV3/Raw.cs
timothycoleman: it does actually have to be the same size, because it occupies the RecordId section of the RecordHeader instead of the guid

src/EventStore.LogV3/Raw.cs
pvanbuijtene: Shouldn't this be covered by a test?

pvanbuijtene (dismissed):

src/EventStore.Core.Tests/TransactionLog/when_writing_prepare_record_to_file.cs
shaan1337: thanks!

src/EventStore.LogV3/Raw.cs
shaan1337: ok thanks :+1:

src/EventStore.LogV3.Tests/RecordHeaderTests.cs
```diff
@@ -0,0 +1,37 @@
...
+				Type = LogCommon.LogRecordType.EventType,
+				Version = 0x45,
+			};
+
+			var now = DateTime.UtcNow;
+			header.TimeStamp = now;
+
+			Assert.Equal(LogCommon.LogRecordType.EventType, header.Type);
+			Assert.Equal(0x45, header.Version);
+			Assert.Equal(now, header.TimeStamp, TimeSpan.FromMilliseconds(20));
```
shaan1337, line 19:
```suggestion
			Assert.Equal(now, header.TimeStamp, TimeSpan.FromMilliseconds(7));
```

src/EventStore.LogV3/Raw.cs
timothycoleman: @pvanbuijtene it is now, the reason it wasn't before is because is requires two assignments to show the problem and the original test was just setting then checking. in practice the records are immutable after they have been created so calling the setter more than once shouldn't happen, but still its nice to be able to of course

pvanbuijtene (approved):

shaan1337 (approved):

PR #2972, opened 2021-05-26
Add option to switch between v2 and v3 log format

pvanbuijtene:
Added: Option to switch between v2 & v3 log format

_Data folder has to be cleared since formats are not compatible._


timothycoleman (approved):

shaan1337 (approved):
looks good to me :+1:

timothycoleman (approved):

PR #2981, opened 2021-05-31
Improve lookup of first non-expired events in long stream with maxage

### LINKED ISSUE #2973: Reading from a stream with a lot of outdated events throws a DeadlineExceeded exception.
**Describe the bug**
A lot of events older than max age causes reading from start of stream to throw an exception *DeadlineExceeded*

**To Reproduce**
1. Insert at least 750k events with short maxAge
2. Try to read from stream start

**Expected behavior**
Get all the events that are not outdated yet or empty list if there is none

**Actual behavior**
Getting an exception: Grpc.Core.RpcException: Status(StatusCode="DeadlineExceeded", Detail="Deadline Exceeded")

**EventStore details**
 - EventStore server version:
20.6.1; 20.10.2; 21.2.0

**Additional context**
I've tried to mingle with different timeout settings but it didn't help. 
Tried scavenge as well but it also didn't help.
We usually have around 1,5kk events in a stream out of which around 1,1kk are outdated. We try to read 10k at once.
Right now, we have a workaround in place where we look for stream start manually and start reading from there.
Gist with code to reproduce:
https://gist.github.com/jciechowski/a682cec98ffd8023dd6464c087018eeb

jageall:
Added: Faster seek for first non-expired events in long streams with $max-age set 

fixes #2973

### DIFF
--- a/src/EventStore.Core.Tests/Services/Storage/FakeInMemoryTableIndex.cs
+++ b/src/EventStore.Core.Tests/Services/Storage/FakeInMemoryTableIndex.cs
@@ -31,7 +31,7 @@ namespace EventStore.Core.Tests.Services.Storage {
 		{
 		}
 
-		public IEnumerable<IndexEntry> GetRange(TStreamId streamId, long startVersion, long endVersion, int? limit = null)
+		public IReadOnlyList<IndexEntry> GetRange(TStreamId streamId, long startVersion, long endVersion, int? limit = null)
 		{
 			var entries = new List<IndexEntry>();
 			if(_indexEntries.ContainsKey(streamId)){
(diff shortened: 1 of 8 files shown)


thefringeninja (dismissed):

thefringeninja (dismissed):

thefringeninja (approved):

hayley-jean (approved):
Looks good, however if you read a stream with a batch size smaller than the number of expired events then you will receive an empty response and not have a position to continue reading from. This should be addressed in https://github.com/EventStore/EventStore-Client-Dotnet/issues/106.

thefringeninja (approved):

PR #2982, opened 2021-06-02
Create and read root partition

pvanbuijtene:
Added: Initial creation of the LogV3 root partition.

- [x] Initial creation and reading of root partition
- [x] Test coverage
- [x] Test on follower
- [x] Use root partition when creating stream

### DIFF
--- a/src/EventStore.Core/LogV3/LogV3RecordFactory.cs
+++ b/src/EventStore.Core/LogV3/LogV3RecordFactory.cs
@@ -1,14 +1,17 @@
 using System;
 using EventStore.Core.LogAbstraction;
 using EventStore.Core.TransactionLog.LogRecords;
+using EventStore.LogV3;
 using StreamId = System.UInt32;
 
 namespace EventStore.Core.LogV3 {
 	public class LogV3RecordFactory : IRecordFactory<StreamId> {
+		private Guid _rootPartitionId;
+
 		public LogV3RecordFactory() {
 			if (!BitConverter.IsLittleEndian) {
 				// to support big endian we would need to adjust some of the bit
-				// operatiosn in the raw v3 structs, and adjust the way that the
+				// operations in the raw v3 structs, and adjust the way that the
 				// v3 records are written/read from disk (currently blitted)
 				throw new NotSupportedException();
 			}
@@ -28,7 +31,8 @@ namespace EventStore.Core.LogV3 {
 				logPosition: logPosition,
 				timeStamp: timeStamp,
 				streamNumber: streamNumber,
-				streamName: streamName);
+				streamName: streamName,
+				partitionId: _rootPartitionId);
 
 			return result;
 		}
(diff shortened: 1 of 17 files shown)


src/EventStore.Core/LogV3/PartitionManager.cs
```diff
@@ -0,0 +1,124 @@
...
+			_recordFactory.SetRootPartitionId(RootId.Value);
+			
+			_log.Debug("Root partition created, id: " + RootId);
+		}
+
+		private void ReadRootPartition()
+		{
+			SeqReadResult result;
+			_reader.Reposition(0);
+			while ((result = _reader.TryReadNext()).Success)
```
shaan1337, line 91: @pvanbuijtene it looks like we can be more strict here with what types records to expect:
Epoch(0) -> Partition type record -> Partition record

so we can read the records one by one in sequence and throw if any one of them does not match the expected record type

shaan1337: (this can potentially prevent reading through the whole database in case some data is corrupted, although it's very unlikely)

src/EventStore.Core/LogV3/DummyPartitionManager.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace EventStore.Core.LogV3
+{
+	public class DummyPartitionManager : IPartitionManager {
```
timothycoleman, line 5: just for the pattern, this could be called `LogV2PartitionManager` and be moved to the `Core.LogV2` namespace

src/EventStore.Core/LogV3/DummyPartitionManager.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace EventStore.Core.LogV3
+{
```
timothycoleman, line 4: (braces on the end of the previous line, elsewhere also)

src/EventStore.Core/LogV3/LogV3RecordFactory.cs
```diff
@@ -73,5 +77,47 @@ public class LogV3RecordFactory : IRecordFactory<StreamId> {
...
+
+		public ILogRecord CreatePartitionTypeRecord(
+			DateTime timeStamp,
+			long logPosition,
+			Guid partitionTypeId,
+			Guid partitionId,
+			string name) {
+			
+			return new PartitionTypeLogRecord(
+				timeStamp,
```
timothycoleman, line 89: personally i'd add the argument names here, but up to you. just so when reading the code i dont wonder 'is there any chance the guids are back to front'

src/EventStore.Core/LogV3/LogV3RecordFactory.cs
```diff
@@ -73,5 +77,47 @@ public class LogV3RecordFactory : IRecordFactory<StreamId> {
...
+			long logPosition,
+			Guid partitionId,
+			Guid partitionTypeId,
+			Guid parentPartitionId,
+			Raw.PartitionFlags flags,
+			ushort referenceNumber,
+			string name) {
+			
+			return new PartitionLogRecord(
+				timeStamp,
```
timothycoleman, line 108: similarly here

src/EventStore.Core/LogV3/PartitionManager.cs
```diff
@@ -0,0 +1,124 @@
+﻿using System;
+using EventStore.Common.Exceptions;
+using EventStore.Core.TransactionLog;
+using EventStore.Core.TransactionLog.LogRecords;
+using EventStore.LogCommon;
+using ILogger = Serilog.ILogger;
+
+namespace EventStore.Core.LogV3
+{
+	public interface IPartitionManager {
```
timothycoleman, line 10: this could go with the interfaces in `Core.LogAbstraction`

src/EventStore.Core/Services/Storage/StorageChaser.cs
```diff
@@ -5,6 +5,7 @@
 using EventStore.Common.Utils;
 using EventStore.Core.Bus;
 using EventStore.Core.Data;
+using EventStore.Core.LogV3;
```
timothycoleman, line 8:
```suggestion
```

timothycoleman (approved):

shaan1337 (approved):
looks good to me :+1:

shaan1337 (approved):

timothycoleman (approved):

PR #2983, opened 2021-06-02
Add memory-mapped bloom filter to quickly detect streams that do not exist

### LINKED ISSUE #2963: Want to quickly detect if a stream doesn't exist
Currently when writing to a stream we lookup the last event number (and meta). If the stream doesn't exist then it will cause a cache miss followed by a binary search in every index file to establish that the stream doesn't exist.

If we can find a way to (at least _quite often_) determine more cheaply that the stream doesn't exist then that would speed up creation of streams a lot.

Possible implementations

- Existing Bloom filter
    - We already have this for chunks.
    - Populating the bloom filter from the index on startup would likely be too expensive, could persist it somewhere?
    - Written to disk - memory mapped file?
- Bloom filter implemented with Microsoft FASTER
- Use the StreamNameIndex from LogV3

shaan1337:
Added: a bloom filter to quickly check stream existence 

Closes #2963


src/EventStore.Core/DataStructures/ProbabilisticFilter/IProbabilisticFilter.cs
```diff
@@ -4,11 +4,11 @@ namespace EventStore.Core.DataStructures.ProbabilisticFilter {
 	//qq suspicious that we're not accessing the implementations through these interfaces
 	public interface IProbabilisticFilter<in TItem> {
 		void Add(TItem item);
-		bool MayExist(TItem item);
+		bool MightExist(TItem item);
```
bartelink, line 7: I like the name but `CouldContain` might be closer to other names?

timothycoleman: Oh yeah, that’s nice 👍

src/EventStore.Core/DataStructures/ProbabilisticFilter/MemoryMappedFileBloomFilter/MemoryMappedFileBloomFilter.cs
```diff
@@ -0,0 +1,275 @@
...
+			if (create) {
+				FillWithZeros();
+
+				_header = new Header() {
+					Version = Header.CurrentVersion,
+					CorruptionRebuildCount = corruptionRebuildCount,
+					NumBits = _numBits
+				};
+
+				_header.WriteTo(_mmf);
```
jageall, line 101: Do we need to force a flush here to make sure the zeros and the header are definitely on disk before starting to use the file? It would be somewhat unlucky to have a hardware crash in between, but possible I guess

src/EventStore.Core/DataStructures/ProbabilisticFilter/MemoryMappedFileBloomFilter/MemoryMappedFileBloomFilter.cs
```diff
@@ -0,0 +1,275 @@
...
+				uint bytesToClearInBlock = bytesToClear > uint.MaxValue ? uint.MaxValue : (uint)bytesToClear;
+				Unsafe.InitBlock(ref *startAddress, 0, bytesToClearInBlock);
+				startAddress += bytesToClearInBlock;
+				bytesToClear -= bytesToClearInBlock;
+			}
+			Log.Debug("Done zeroing bloom filter");
+		}
+
+		// corruptionThreshold = 0.05 => tolerate up to 5% corruption
+		public void Verify(double corruptionThreshold = 0) {
```
jageall, line 246: Why default to 0 here? Do we need to worry about Epsilon in the comparisons? regardless, it should not accept less than zero. although < 0 would always throw, it wouldn't be clear immediately why

src/EventStore.Core/DataStructures/ProbabilisticFilter/MemoryMappedFileBloomFilter/MemoryMappedFileBloomFilter.cs
```diff
@@ -0,0 +1,275 @@
...
+					corruptedCacheLines++;
+					Log.Verbose("Invalid cacheline starting at {i:N0} / {fileSize:N0}", i, _fileSize);
+				}
+			}
+
+			var totalCacheLines = _fileSize / CacheLineSize;
+
+			var corruptedPercent = (double)corruptedCacheLines / totalCacheLines * 100;
+			if (corruptedPercent > corruptionThreshold * 100)
+				throw new CorruptedHashException(
```
jageall, line 262: Add the threshold to the error message

src/EventStore.Core/DataStructures/ProbabilisticFilter/MemoryMappedFileBloomFilter/MemoryMappedFileBloomFilter.cs
```diff
@@ -0,0 +1,275 @@
...
+					_header.CorruptionRebuildCount,
+					$"{corruptedCacheLines:N0} corruptions detected ({corruptedPercent:N2}%)");
+
+			if (corruptedCacheLines == 0) {
+				Log.Debug("Done verifying bloom filter");
+			} else {
+				Log.Warning("Done verifying bloom filter: {corruptedCacheLines:N0} corruptions detected ({corruptedPercent:N2}%)",
```
jageall, line 269: add the threshold to the log message

jageall (changes requested):

src/EventStore.Core.Tests/DataStructures/bloom_filter_integrity_should.cs
```diff
@@ -0,0 +1,125 @@
...
+			// cacheline should be valid
+			if (expected) {
+				Assert.True(BloomFilterIntegrity.ValidateHash(byteSpan));
+			} else {
+				Assert.False(BloomFilterIntegrity.ValidateHash(byteSpan));
+			}
+		}
+
+		[TestCase]
+		public void rehash_when_one_too_many_bits_are_set() {
```
shaan1337, line 80:
```suggestion
		public void rehash_when_too_many_one_bits_are_set() {
```

src/EventStore.Core/DataStructures/ProbabilisticFilter/MemoryMappedFileBloomFilter/MemoryMappedFileBloomFilter.cs
timothycoleman: the checkpoint is only flushed after a successful flush of the filter, so if we have a checkpoint on start we are guaranteed that the header and zeros wrote correctly. if we don't have a checkpoint then the filter is junked and rebuilt

src/EventStore.Core.XUnit.Tests/LogAbstraction/Common/StreamExistenceFilterTests.cs
```diff
@@ -0,0 +1,182 @@
...
+		public void writes_can_be_read_by_another_thread() {
+			var sut = GenSut();
+			sut.Initialize(new MockExistenceFilterInitializer());
+
+			var theValue = 12345;
+
+			var reader = new Thread(() => {
+				var x = 0;
+				while (!sut.MightContain($"{theValue}")) {
+					x = x * 1;
```
shaan1337, line 172: Can use `SpinWait.SpinUntil(sut.MightContain($"{theValue}"))` just to improve clarity

src/EventStore.Core.Tests/DataStructures/bloom_filter_integrity_should.cs
timothycoleman: it's supposed to describe the case where there is one too many bits set, as in only one extra

src/EventStore.Core/DataStructures/ProbabilisticFilter/MemoryMappedFileBloomFilter/MemoryMappedFileBloomFilter.cs
timothycoleman: ah, yeah i'll remove the optional param

i dont think we need to worry about epsilon, it's a '>' comparison and doesn't nee to be that precise

src/EventStore.Core.XUnit.Tests/LogAbstraction/Common/StreamExistenceFilterTests.cs
timothycoleman: thats tempting, although it's an extra piece of complexity in a fairly subtle condition

src/EventStore.Core.Tests/DataStructures/bloom_filter_integrity_should.cs
jageall: maybe "rehash_when_exactly_one_too_many_bits_are_set" ?

src/EventStore.Core/DataStructures/ProbabilisticFilter/MemoryMappedFileBloomFilter/BloomFilterIntegrity.cs
```diff
@@ -0,0 +1,114 @@
...
+			var cacheToHash = cacheLine[..^HashSize];
+			var hash = _hasher.Hash(cacheToHash);
+			var targetHash = MemoryMarshal.Cast<byte, uint>(cacheLine)[UintHashIndex];
+			if (hash == targetHash)
+				return true;
+
+			// corrupt, try to recover.
+			if (CanRehash(cacheToHash, targetHash)) {
+				WriteHash(cacheLine);
+				Log.Warning("Corrected a hash");
```
jageall, line 58: Should this be a warning? What is anyone going to do about it?

jageall (changes requested):

src/EventStore.Core/DataStructures/ProbabilisticFilter/MemoryMappedFileBloomFilter/BloomFilterIntegrity.cs
jageall: we don't even know what cacheLine it was.

timothycoleman: 👍 i'll remove it

timothycoleman: actually i'll make it verbose

src/EventStore.Core/DataStructures/ProbabilisticFilter/MemoryMappedFileBloomFilter/BloomFilterIntegrity.cs
```diff
@@ -0,0 +1,114 @@
...
+			// corrupt and could not recover, set all bytes to 0xFFs to make safe
+			var ulongs = MemoryMarshal.Cast<byte, ulong>(cacheLine);
+			for (int i = 0; i < ulongs.Length; i++) {
+				ulongs[i] = ulong.MaxValue;
+			}
+
+			return false;
+		}
+
+		// The hash can be wrong simply because the flush happened between when
```
shaan1337, line 71: This also means that a genuine corrupted bit flip from 0->1 in the non-hash part will be accepted. I guess it should be fine since bloom filter will still be safe and it can only cause more false positives.

shaan1337, line 71: This also means that a genuinely corrupted bit (flipped from 0 to 1) will be accepted, but I guess it should be fine since the bloom filter will still be safe and only potentially cause more false positives.

shaan1337: And it's still better than making it all ones

src/EventStore.Core/DataStructures/ProbabilisticFilter/MemoryMappedFileBloomFilter/MemoryMappedFileBloomFilter.cs
```diff
@@ -0,0 +1,279 @@
...
+				// todo: OpenOrCreate can be CreateNew if the mininode closed the file before declaring itself 'stopped'
+				create ? FileMode.OpenOrCreate : FileMode.Open,
+				FileAccess.ReadWrite,
+				FileShare.ReadWrite);
+
+			if (Header.Size > CacheLineSize)
+				throw new Exception("header needs to fit in a cache line");
+
+			// add room for filter bits, rounded up to nearest cacheline
+			_fileSize = GetBytePositionInFile(_numBits) / CacheLineSize * CacheLineSize;
```
shaan1337, line 77: It appears that there are two off-by-ones:
- Should GetBytePositionInFile(_numBits) be GetBytePositionInFile(_numBits - 1) ? (since it's a 0-based index and the largest bit is _numBits - 1)
-  / CacheLineSize * CacheLineSize is taking the floor, so there'll be one cache line missing if (_numBits-1) is not an exact multiple of CacheLineSize (I'm assuming that the line below: `_fileSize += CacheLineSize` is for the header)

src/EventStore.Core/DataStructures/ProbabilisticFilter/MemoryMappedFileBloomFilter/BloomFilterIntegrity.cs
timothycoleman: Yep exactly

src/EventStore.Core/DataStructures/ProbabilisticFilter/MemoryMappedFileBloomFilter/MemoryMappedFileBloomFilter.cs
shaan1337: for the second point, i think that I'm wrong - the header's cache line is being taken into consideration inside GetBytePositionInFile() and the line `_fileSize += CacheLineSize` is actually for the extra cache line if it's not an exact multiple, right?

timothycoleman: Good questions, in reverse order:

- `GetBytePositionInFile` accounts for the header (thats what im trying to indicate with the 'infile' part of the name), so the added cacheline on the next line is inded to account for the fact that we floored it in the calculation, not to make room for the header
- `GetBytePositionInFile(_numBits - 1)` would get the us the position of the last byte in the file that holds filter bits. But to calculate the length needed we want the first byte that doesnt hold filter bits. e.g. if the position of the last byte is 123 then the length is 124 (before rounding up to the nearest cacheline)

src/EventStore.Core/DataStructures/ProbabilisticFilter/MemoryMappedFileBloomFilter/MemoryMappedFileBloomFilter.cs
```diff
@@ -0,0 +1,279 @@
...
+		}
+
+		// corruptionThreshold = 0.05 => tolerate up to 5% corruption
+		public void Verify(double corruptionThreshold) {
+			Ensure.Nonnegative(corruptionThreshold, nameof(corruptionThreshold));
+
+			Log.Debug("Verifying bloom filter...");
+			var corruptedCacheLines = 0;
+
+			for (int i = CacheLineSize; i < _fileSize; i += CacheLineSize) {
```
shaan1337, line 252:
```suggestion
			for (long i = CacheLineSize; i < _fileSize; i += CacheLineSize) {
```

`i` may go out of bounds for large bloom filters

src/EventStore.Core/DataStructures/ProbabilisticFilter/MemoryMappedFileBloomFilter/MemoryMappedFileBloomFilter.cs
```diff
@@ -0,0 +1,279 @@
...
+				var cacheLine = ReadCacheLineFor(i);
+				if (!BloomFilterIntegrity.ValidateHash(cacheLine)) {
+					corruptedCacheLines++;
+					Log.Verbose("Invalid cacheline starting at {i:N0} / {fileSize:N0}", i, _fileSize);
```
shaan1337, line 256:
```suggestion
					Log.Verbose("Invalid cache line starting at {i:N0} / {fileSize:N0}", i, _fileSize);
```

src/EventStore.Core/DataStructures/ProbabilisticFilter/MemoryMappedFileBloomFilter/MemoryMappedFileBloomFilter.cs
```diff
@@ -0,0 +1,279 @@
...
+			}
+
+			var totalCacheLines = _fileSize / CacheLineSize;
+
+			var corruptionThresholdPercent = corruptionThreshold * 100;
+			var corruptedPercent = (double)corruptedCacheLines / totalCacheLines * 100;
+			if (corruptedPercent > corruptionThreshold * 100)
+				throw new CorruptedHashException(
+					_header.CorruptionRebuildCount,
+					$"{corruptedCacheLines:N0} corruptions detected ({corruptedPercent:N2}%). Threshold {corruptionThresholdPercent:N2}%");
```
shaan1337, line 267:
```suggestion
					$"{corruptedCacheLines:N0} corruption(s) detected ({corruptedPercent:N2}%). Threshold {corruptionThresholdPercent:N2}%");
```

src/EventStore.Core/DataStructures/ProbabilisticFilter/MemoryMappedFileBloomFilter/MemoryMappedFileBloomFilter.cs
```diff
@@ -0,0 +1,279 @@
...
+
+			if (corruptedCacheLines == 0) {
+				Log.Debug("Done verifying bloom filter");
+			} else {
+				Log.Warning("Done verifying bloom filter: {corruptedCacheLines:N0} corruptions detected ({corruptedPercent:N2}%). Threshold {corruptionThresholdPercent:N2}%",
```
shaan1337, line 272:
```suggestion
				Log.Warning("Done verifying bloom filter: {corruptedCacheLines:N0} corruption(s) detected ({corruptedPercent:N2}%). Threshold {corruptionThresholdPercent:N2}%",
```

src/EventStore.Core/DataStructures/ProbabilisticFilter/MemoryMappedFileBloomFilter/MemoryMappedFileBloomFilter.cs
timothycoleman: 👍 good catch in case we ever allow a bloom filter that large

shaan1337: @timothycoleman the max size is currently set to 4GB which will overflow an int

src/EventStore.Core/DataStructures/ProbabilisticFilter/MemoryMappedFileBloomFilter/MemoryMappedFileBloomFilter.cs
shaan1337: @timothycoleman thanks for all the clarifications!

regarding `_numBits` vs `_numBits - 1`, i still have some doubts:
take the case where the last filter bit (`_numBit - 1`) lies at a 60-byte boundary just before the 4-byte hash, then the next bit (`_numBit`) will be in the next cache line. With the current code after rounding up, there'll be an extra 64-byte cache line in the file (it should not cause any harm though)

src/EventStore.Core/LogV3/NameIndex.cs
```diff
@@ -86,27 +101,83 @@ public class NameIndex :
...
+
+					// update the streams stream
+					// initialisation of the stream name index caused an entry to be populated in
+					// the last event number cache, now we need to keep it up to date even on initialisation
+					backend.SetStreamLastEventNumber(prepare.EventStreamId, prepare.ExpectedVersion + 1);
+
+					if (catchingUp) {
+						// we are catching up, do not set the last event numbers because they will not be
+						// updated during the catchup if we do write some events to those streams.
+						// thesefore leave the entry blank so it will be (cheaply because in mem) be
```
shaan1337, line 125:
```suggestion
						// therefore leave the entry blank so it will be (cheaply because in mem) be
```

src/EventStore.Core/DataStructures/ProbabilisticFilter/MemoryMappedFileBloomFilter/MemoryMappedFileBloomFilter.cs
timothycoleman: thats true, say _numbits is 60*8=480 then `GetBytePositionInFile(480)` will return 128. which is correct

however since its already exactly a cache line we neededn't add on an extra one (but as you say it doesnt harm if we do so)

shaan1337: @timothycoleman ok, this extra cache line doesn't happen if we use _numBits - 1 since it can be read as:
- Find the cache line to which the last filter bit belongs too
- Round down to the nearest cache line: we're now at the second-to-last cache line
- Add an extra cache line

But since it doesn't cause any harm, i'm marking it as resolved :-)

timothycoleman (approved):

shaan1337:
Reviewed - looks good to me

jageall (approved):

pvanbuijtene (approved):

PR #2990, opened 2021-06-17
Allow loading of untrusted certificates from the windows certificate store

shaan1337:
Fixed: Allow loading of untrusted certificates from the windows certificate store

Currently, certificates can be loaded from the windows certificate store only if it is a trusted certificate due to the `validOnly` parameter being passed as `true`. This is inconvenient since it forces users to add the private CA certificate to the list of global trusted root certificates.


ylorph:
@shaan1337 can we close this one since we can now load the full chain from the windows store ?

hayley-jean:
@shaan1337 I'm going to close this given @ylorph's comment above, but let me know if it's still something that needs to be added

PR #2992, opened 2021-06-17
Ignore operation/task cancelled exceptions

jageall:
Fixed: Aborted http requests are no longer logged in the authentication middleware


thefringeninja (approved):
Looks good, however should we target the release branch with this?

PR #3002, opened 2021-06-21
Make Log Configuration Path Configurable

thefringeninja:
Added: A configuration option to set the path of the log configuration file


hayley-jean (approved):

Lougarou:
Redundant default Log Config option value shown at startup after Log Config has been modified.
logconfigreview
ing parameter.

Lougarou (approved):
looks good and option name issue seems to be a separate issue

Issue #3005, opened 2021-06-23
WrongExpectedVersion when deleting a projection with all delete options


hayley-jean:
Deleting a projection with all delete options checked results in a `WrongExpectedVersion` error and other errors printed in the server logs. The delete request itself reports that it was successful.

## Reproduction Steps:

1. Start an empty Event Store with projections enabled.
2. Create a simple projection. Leave all options default (this fails for both Continuous and OneTime projections):

```
fromStream("test-stream")
.when({
    $init: function() {
        return { count: 0 };
    },
    $any: function(s,e) {
        s.count++;
    }
})
```

3. Delete the projection through the UI, checking all of the delete options: 

![image](https://user-images.githubusercontent.com/5140165/123058111-c3848180-d408-11eb-9cbc-f95b82fb4abf.png)

4. Check the server logs. The following errors should be logged:

```
[15940,16,09:47:20.984,INF] PROJECTIONS: Projection Stream '"$projections-demo-emittedstreams-checkpoint"' deleted
[15940,16,09:47:20.986,INF] PROJECTIONS: Projection '"demo"' deletion has not been written to "$projections-$all". Error: "WrongExpectedVersion"
[15940,16,09:47:20.986,ERR] PROJECTIONS: Got wrong expected version writing projection deletion for "demo".
[15940,16,09:47:20.986,INF] PROJECTIONS: Projection Stream '"$projections-demo-emittedstreams"' deleted
[15940,16,09:47:20.986,ERR] Error while processing message "EventStore.Projections.Core.Messages.ProjectionManagementMessage+Internal+Deleted" in queued handler '"Projections Leader"'.
System.NullReferenceException: Object reference not set to an instance of an object.
   at EventStore.Projections.Core.Services.Management.ProjectionManager.Handle(Deleted message) in C:\Users\hayle\code\EventStore\src\EventStore.Projections.Core\Services\Management\ProjectionManager.cs:line 616
   at EventStore.Core.Bus.MessageHandler`1.TryHandle(Message message) in C:\Users\hayle\code\EventStore\src\EventStore.Core\Bus\MessageHandler.cs:line 30
   at EventStore.Core.Bus.InMemoryBus.Publish(Message message) in C:\Users\hayle\code\EventStore\src\EventStore.Core\Bus\InMemoryBus.cs:line 290
[...]

PR #3014, opened 2021-06-28
Bug/wrong expected when deleting projection

StevenBlair123:
Fixed: WrongExpectedVersion when deleting a projection with all delete options 

#3005 

@hayley-jean - fix applied

I was thinking the method `ManagedProjections.DeleteProjectionStreams` could get changed to check whether checkpoints / emitted streams were originally enabled, irrespective of the deleting options posted up?

So something like:

```
if (PersistedProjectionState.DeleteCheckpointStream && PersistedProjectionState.CheckpointsDisabled) {
	DeleteStream(projectionNamesBuilder.MakeCheckpointStreamName(), DeleteIfConditionsAreMet);
}

if (PersistedProjectionState.DeleteEmittedStreams && PersistedProjectionState.EmitEnabled) {
	if (_emittedStreamsDeleter == null) {
		_emittedStreamsDeleter = new EmittedStreamsDeleter(
			_ioDispatcher,
			projectionNamesBuilder.GetEmittedStreamsName(),
			projectionNamesBuilder.GetEmittedStreamsCheckpointName());
	}

	_emittedStreamsDeleter.DeleteEmittedStreams(DeleteIfConditionsAreMet);
}
```
I am still getting the WrongExpected message (after the deleting the checkpoint stream) but doesn't seem to causing any problems


thefringeninja (approved):

hayley-jean (approved):

hayley-jean:
Thanks for the PR @StevenBlair123!

I agree with you that checking whether checkpoints or emits are enabled before deleting those streams would be a good change as well. Do you want to do that as part of this PR? Otherwise I'm happy to merge this

StevenBlair123:
I can raise another issue for this just so you can get this closed off.

PR #3017, opened 2021-06-29
Fix flakey tests

pvanbuijtene:
Added: `--blame-hang-timeout=5min` in case we have hanging builds.
Added: Improved error messages in case of task timeouts and test setup failures.
Added: More task timeouts to help prevent hanging build runs.
Added: Logging for help investigate flaky tests.
Added: Extra assertions to prevent unclear test failures.
Added: Retry of tests incase of failure triggered by flakiness.
Added: Checks to improve cluster stability before running tests.

"Fixed"... should be reduced :)

### DIFF
--- a/src/EventStore.Core.Tests/ClientAPI/catchup_filtered_subscription.cs
+++ b/src/EventStore.Core.Tests/ClientAPI/catchup_filtered_subscription.cs
@@ -66,12 +66,13 @@ namespace EventStore.Core.Tests.ClientAPI {
 		[Test]
 		public void calls_checkpoint_delegate_during_catchup() {
 			var filter = Filter.StreamId.Prefix("stream-a");
-			var checkpointReached = new CountdownEvent(10);
+			var isV2 = LogFormatHelper<TLogFormat, TStreamId>.IsV2;
+			var checkpointReached = new CountdownEvent(isV2 ? 10 : 12 /* accounting for stream records */);
 			var eventsSeen = 0;
 
 			var settings = new CatchUpSubscriptionFilteredSettings(
-				10000,
-				2,
+				maxLiveQueueSize: 10000,
+				readBatchSize: 2,
 				verboseLogging: false,
 				resolveLinkTos: true,
 				maxSearchWindow: 2,
@@ -95,12 +96,7 @@ namespace EventStore.Core.Tests.ClientAPI {
 				Assert.Fail("Checkpoint reached not called enough times within time limit.");
 			}
 
-			if (LogFormatHelper<TLogFormat, TStreamId>.IsV2) {
-				Assert.AreEqual(10, eventsSeen);
-			} else {
-				// accounting for stream records
-				Assert.AreEqual(7, eventsSeen);
-			}
+			Assert.AreEqual(10, eventsSeen);
 		}
 
 		[Test]
--- a/src/EventStore.Core.Tests/ClientAPI/subscribe_to_all_filtered_should.cs
+++ b/src/EventStore.Core.Tests/ClientAPI/subscribe_to_all_filtered_should.cs
@@ -201,7 +201,6 @@ namespace EventStore.Core.Tests.ClientAPI {
 
 		[Test, Category("LongRunning")]
 		public async Task calls_checkpoint_reached_according_to_checkpoint_message_count() {
-			var isV3 = LogFormatHelper<TLogFormat, TStreamId>.IsV3;
 			var filter = Filter.ExcludeSystemEvents;
 
 			using (var store = BuildConnection(_node)) {
@@ -210,6 +209,11 @@ namespace EventStore.Core.Tests.ClientAPI {
 				var eventsSeen = new List<ResolvedEvent>();
 				var checkpointsSeen = 0;
 
+				if (LogFormatHelper<TLogFormat, TStreamId>.IsV3) {
+					// init stream record so it won't affect the test and prevent flakiness
+					await _conn.AppendToStreamAsync("stream-a", ExpectedVersion.NoStream);
+				}
+				
 				using (await store.FilteredSubscribeToAllAsync(false,
 					filter,
 					(s, e) => {
@@ -222,12 +226,12 @@ namespace EventStore.Core.Tests.ClientAPI {
 						}
 						return Task.CompletedTask;
 					},
-					2)) {
+					checkpointInterval: 2)) {
 					await _conn.AppendToStreamAsync("stream-a", ExpectedVersion.NoStream, _testEvents);
 
 					await appeared.Task.WithTimeout(Timeout);
 
-					Assert.AreEqual(5 /*checkpoints*/ * 2 /*considered events*/ - (isV3 ? 1 : 0) /*filtered out "stream-a" stream record event*/, eventsSeen.Count);
+					Assert.AreEqual(5 /*checkpoints*/ * 2 /*considered events*/, eventsSeen.Count);
 					Assert.True(eventsSeen.All(x => x.Event.EventStreamId == "stream-a"));
 				}
 			}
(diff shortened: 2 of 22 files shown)


hayley-jean (dismissed):

thefringeninja (dismissed):

thefringeninja (dismissed):

src/EventStore.Core.Tests/Http/Cluster/when_requesting_from_follower.cs
```diff
@@ -14,32 +14,43 @@ namespace EventStore.Core.Tests.Http.Cluster {
...
+			var response = await PostEvent(_followerEndPoint, $"streams/{TestDeleteStream}", requireLeader: false);
+			Assert.AreEqual(HttpStatusCode.Created, response.StatusCode);
+			
+			response = await PostEvent(_followerEndPoint, $"streams/{TestStream}", requireLeader: false);
 			Assert.AreEqual(HttpStatusCode.Created, response.StatusCode);
-			leaderIndex = GetLeader().Db.Config.IndexCheckpoint.Read();
-			AssertEx.IsOrBecomesTrue(()=> replica.Db.Config.IndexCheckpoint.Read() >= leaderIndex,
-				msg: $"Waiting for test stream to be synced with follower timed out! ({leaderIndex})");
+
+			AssertEx.IsOrBecomesTrue(() => leader.Db.Config.IndexCheckpoint.Read() > leaderIndex,
```
timothycoleman, line 46: i think we're getting the leader index before appending events to two different streams and then waiting for the index checkpoint to move - but perhaps it could move for the first event and still the second hasn't been processed yet?

if thats right it might be useful to have a helper method that add and waits, and just call that method twice

src/EventStore.Core.Tests/ClientAPI/catchup_filtered_subscription.cs
```diff
@@ -95,12 +96,7 @@ public class catchup_filtered_subscription<TLogFormat, TStreamId> : Specificatio
 				Assert.Fail("Checkpoint reached not called enough times within time limit.");
 			}
 
-			if (LogFormatHelper<TLogFormat, TStreamId>.IsV2) {
-				Assert.AreEqual(10, eventsSeen);
-			} else {
-				// accounting for stream records
-				Assert.AreEqual(7, eventsSeen);
-			}
+			Assert.AreEqual(10, eventsSeen);
```
timothycoleman, line 99: is it known why this was flaky the was it was before?

src/EventStore.Core.Tests/ClientAPI/subscribe_to_all_filtered_should.cs
```diff
@@ -222,12 +226,12 @@ public class subscribe_to_all_filtered_should<TLogFormat, TStreamId> : Specifica
...
 						return Task.CompletedTask;
 					},
-					2)) {
+					checkpointInterval: 2)) {
 					await _conn.AppendToStreamAsync("stream-a", ExpectedVersion.NoStream, _testEvents);
 
 					await appeared.Task.WithTimeout(Timeout);
 
-					Assert.AreEqual(5 /*checkpoints*/ * 2 /*considered events*/ - (isV3 ? 1 : 0) /*filtered out "stream-a" stream record event*/, eventsSeen.Count);
+					Assert.AreEqual(5 /*checkpoints*/ * 2 /*considered events*/, eventsSeen.Count);
```
timothycoleman, line 234: is it known why this was flaky the was it was before?

src/EventStore.Core.Tests/Http/PersistentSubscription/nack.cs
```diff
@@ -29,8 +22,10 @@ class when_nacking_a_message<TLogFormat, TStreamId> : with_subscription_having_e
 				ContentType.CompetingJson,
 				_admin);
 			Assert.AreEqual(HttpStatusCode.OK, _lastResponse.StatusCode);
-			_nackLink = json["entries"].Children().First()["links"].Children()
-				.First(x => x.Value<string>("relation") == "nack").Value<string>("uri");
+			Assert.DoesNotThrow(() => {
```
timothycoleman, line 25: Do the DoesNotThrow calls change the behaviour here?

src/EventStore.Core.Tests/Services/Transport/Tcp/TcpConnectionTests.cs
```diff
@@ -37,16 +38,15 @@ public class TcpConnectionTests {
...
-					},
+					(conn, error) => failedError = error,
 					false);
 
 				var serverSocket = listeningSocket.Accept();
 				var serverTcpConnection = TcpConnection.CreateAcceptedTcpConnection(Guid.NewGuid(),
 					(IPEndPoint)serverSocket.RemoteEndPoint, serverSocket, false);
 
 				mre.Wait(TimeSpan.FromSeconds(3));
+				Assert.Null(failedError);
```
timothycoleman, line 49: since this is set on another thread it'll need some synchronisation otherwise the asserting thread might still have the null from above. a taskcompletion source with RunContinuationsAsynchronously  would do the trick i think

src/EventStore.Core.Tests/Services/Transport/Tcp/when_invalid_data_is_sent_over_tcp.cs
```diff
@@ -43,23 +42,20 @@ public class when_invalid_data_is_sent_over_tcp<TLogFormat, TStreamId> : specifi
...
-			connectedEvent.WaitOne();
+			Assert.True(connectedEvent.Wait(10000));
 			var data = new List<ArraySegment<byte>> {
 				new ArraySegment<byte>(new byte[] {1, 2, 3, 4, 5, 6, 7, 8, 9, 10})
 			};
 			connection.EnqueueSend(data);
-			closedEvent.WaitOne();
+			Assert.True(closedEvent.Wait(10000));
 			connection.Close("intentional close");
+			Assert.Null(failedError);
```
timothycoleman, line 58: same here with respect to threads and taskcompletionsource

src/EventStore.Core.Tests/Http/PersistentSubscription/nack.cs
pvanbuijtene: It prevents the the test host from being aborted caused by an empty collection.

src/EventStore.Core.Tests/Http/Cluster/when_requesting_from_follower.cs
pvanbuijtene: Good point, added it.

src/EventStore.Core.Tests/ClientAPI/catchup_filtered_subscription.cs
pvanbuijtene: The subscription might still continue after reaching the count down resulting in errors like:
```
  Error Message:
     Expected: 7
  But was:  9
```

src/EventStore.Core.Tests/ClientAPI/subscribe_to_all_filtered_should.cs
pvanbuijtene: Same as other test, events might still be counted after reaching the checkpoint count as there are more events present.

src/EventStore.Core.Tests/ClientAPI/catchup_filtered_subscription.cs
timothycoleman: ooh it had processed enough events to trigger the checkpoint but not all of them. got it :+1:

src/EventStore.Core.Tests/Services/Transport/Tcp/TcpConnectionTests.cs
pvanbuijtene: Changed it.

src/EventStore.Core.Tests/Services/Transport/Tcp/when_invalid_data_is_sent_over_tcp.cs
pvanbuijtene: Changed it.

timothycoleman (approved):
👍

thefringeninja (approved):

PR #3021, opened 2021-07-01
fix build instructions

ylorph:
Fixed: Build instructions

* Passing  `--runtime` when building a solution is no longer supported in net 5.0 ,
* Passing `--framework` causes the build to fail due to a race condition inside msbuild , removing since all projects define net5.0 as their target
* start a node in insecure mode, projections on  and atom feed enabled so that the stream browse on the ui works.
* remove instructions for macOS builds (we don't support running on macOS , docker needs to be used ) 
* client libraries builds : links to respective repositories

Todo 
- [x] check & fix `build.cmd` & `build.ps1` scripts on windows
- [x] check & fix `build.sh` on linux

resolves EventStore/documentation#390
resolves EventStore/home#310

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,25 +26,22 @@ We also host native packages for Linux on [Package Cloud](https://packagecloud.i
 
 ## Building EventStoreDB
 
-EventStoreDB is written in a mixture of C#, C++ and JavaScript. It can run on Windows, Linux and macOS using the .NET Core runtime. However, the projections library (which uses the V8 javascript engine) contains platform specific code and it must be built for the platform on which you intend to run it.
+EventStoreDB is written in a mixture of C#, C++ and JavaScript. It can run on Windows, Linux and macOS (using Docker) using the .NET Core runtime. However, the projections library (which uses the V8 javascript engine) contains platform specific code and it must be built for the platform on which you intend to run it.
 
-### Windows / Linux / macOS
+### Windows / Linux
 **Prerequisites**
 - [.NET Core SDK 5.0] (https://dotnet.microsoft.com/download/dotnet/5.0)
 
 ### Build EventStoreDB
 Once you've installed the prerequisites for your system, you can launch a `Release` build of EventStore as follows:
 ```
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -35,11 +35,8 @@ EventStoreDB is written in a mixture of C#, C++ and JavaScript. It can run on Wi
 ### Build EventStoreDB
 Once you've installed the prerequisites for your system, you can launch a `Release` build of EventStore as follows:
 ```
-dotnet build -c Release src/EventStore.sln -f netcoreapp5.0 -r <runtime identifier>
+dotnet build -c Release src/EventStore.sln
```
bartelink, line 38: I think `dotnet build -c Release src` will do the trick as long as there is exactly one `.sln` or `*proj` sufficed file in the folder?

ylorph: I'll try like now :-)

arwinneil (dismissed):

README.md
```diff
@@ -26,25 +26,22 @@ We also host native packages for Linux on [Package Cloud](https://packagecloud.i
 
 ## Building EventStoreDB
 
-EventStoreDB is written in a mixture of C#, C++ and JavaScript. It can run on Windows, Linux and macOS using the .NET Core runtime. However, the projections library (which uses the V8 javascript engine) contains platform specific code and it must be built for the platform on which you intend to run it.
+EventStoreDB is written in a mixture of C#, C++ and JavaScript. It can run on Windows, Linux and macOS (using Docker) using the .NET Core runtime. However, the projections library (which uses the V8 javascript engine) contains platform specific code and it must be built for the platform on which you intend to run it.
```
mat-mcloughlin, line 29: > However, the projections library (which uses the V8 javascript engine) contains platform specific code and it must be built for the platform on which you intend to run it.

Is this still true with the latest changes that James made?

mat-mcloughlin (dismissed):

README.md
ylorph: actually it's pre-built

oskardudycz (approved):

mat-mcloughlin (approved):

PR #3047, opened 2021-07-16
HTTP and AtomPub deprecation

alexeyzimarev:


thefringeninja (approved):

oskardudycz (approved):

src/EventStore.Common/Log/logconfig.json
```diff
@@ -1,10 +1,22 @@
...
-	  "Grpc": "Fatal"
+	"Logging": {
+		"LogLevel": {
+			"Default": "Debug",
+			"System": "Warning",
+			"Microsoft": "Warning",
+			"Grpc": "Fatal"
+		}
+	},
+	"Serilog": {
```
thefringeninja, line 10: If we include a `Serilog` section this means that the default in code logging configuration will get overridden. This is something we definitely do not want.

thefringeninja (changes requested):

src/EventStore.Common/Log/logconfig.json
alexeyzimarev: I am not sure why I changed this file. I didn't have any change planned for it, good catch, I just need to revert it.

thefringeninja (approved):

oskardudycz (approved):

PR #3069, opened 2021-07-22
Dockerfile Improvements

thefringeninja:
Added: Alpine Docker Image

- reduce number of layers
- fail test run when tests fail
- add support for alpine linux
- reduce docker image size by half

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,8 +62,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        container-runtime: [focal, buster-slim]
-        runtime: [linux-x64]
+        container-runtime: [focal, buster-slim, alpine]
     runs-on: ubuntu-latest
     name: ci/github/build-${{ matrix.container-runtime }}
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -62,43 +62,69 @@ jobs:
...
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
     - name: Calculate Version
       run: |
         dotnet tool restore
         version=$(dotnet tool run minver -- --tag-prefix=oss-v)-${{ matrix.container-runtime }}
         echo "VERSION=${version}" >> $GITHUB_ENV
+    - name: Setup Variables
+      id: variables
+      uses: kanga333/variable-mapper@master
```
shaan1337, line 94: @thefringeninja do we want to pin a specific version here so that our CI doesn't break if there are any breaking changes?

.github/workflows/ci.yml
```diff
@@ -62,43 +62,69 @@ jobs:
...
-        docker build \
-          --tag eventstore \
-          --build-arg RUNTIME=${{ matrix.runtime }} \
-          --build-arg CONTAINER_RUNTIME=${{ matrix.container-runtime }} \
-          .
+      uses: docker/build-push-action@v2
+      with:
+        context: .
+        load: true
+        tags: eventstore
```
shaan1337, line 115: @thefringeninja I may be missing something - does `build-args` need to be specified here (and below in `Build Test Container` action) to pass in `matrix.runtime` and `matrix.container-runtime` or is it passed automatically?

shaan1337 (commented):
nice improvements :+1: i've just added two comments for parts which are not clear to me.

oskardudycz (dismissed):

oskardudycz:
Approving in advance, but also second @shaan1337 comments 👍

.github/workflows/ci.yml
thefringeninja: it does, I completely missed that.

shaan1337:
@thefringeninja thanks for making the changes, it appears that we also need to specify CONTAINER RUNTIME here:
https://github.com/EventStore/EventStore/pull/3069/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557L2

I checked here: https://hub.docker.com/_/microsoft-dotnet-sdk?tab=description and 5.0-alpine is available

.github/workflows/ci.yml
shaan1337: as discussed offline with @thefringeninja, variable-mapper doesn't seem to support the standard github action convention (v1,v2, etc.) and any breaking changes would be noticed quickly within a few hours so it should be fine.

shaan1337 (commented):
Just noticed another small change that may need to be done (https://github.com/EventStore/EventStore/pull/3069#issuecomment-888913670), the rest looks good to me.

shaan1337 (approved):

oskardudycz (approved):
Seems legit 👍

PR #3078, opened 2021-07-30
Add a bloom filter to quickly check stream existence

timothycoleman:
Added: a bloom filter to quickly check stream existence

Allows for quickly creating new streams because we avoid searching each index files for the last event of a non-existent stream.

- The bloom filter is stored in a memory mapped file with hashes at the end of each cache line for integrity
- The bloom filter is on by default, set to 256mb which is good for about 256 million streams
- The filter is not compatible with V1 index files. An exception will be thrown on startup.
- In LogV2
    - the stream hashes are what is added to the filter. this allows it to be rebuilt from scratch using the index only
    - the checkpoint is a log position
    - the filter is checked on LRU cache miss
- In LogV3
    - the stream names are what is added to the filter
    - the checkpoint is the event number of the stream record in the streams stream
    - the filter is checked as part of the stream name index

Co-authored-by: Shaan Nobee <sniper111@gmail.com>


timothycoleman:
(this is only draft because i haven't actually run it)

hayley-jean (approved):

thefringeninja (approved):

timothycoleman:
Force pushed, only change was to increase the timeout on those tests

hayley-jean (approved):

PR #3081, opened 2021-08-04
document unknown options error 

ylorph:


docs/configuration/README.md
```diff
@@ -190,3 +190,8 @@ DEFAULT OPTIONS:
      FAULT OUT OF ORDER PROJECTIONS:          False (<DEFAULT>)
 ```
 :::
+
+::: warning
+As of version 21.6 the server will _not start_ when an unknown configuration options is passed in either the configuration file, environment variable or command line.
```
oskardudycz, line 195: I've used a bit different wording in the release notes (https://www.eventstore.com/blog/21.6.0-release-notes):

> 21.6.0 introduced a strict configuration check for settings set through environment variables. It won't allow passing unknown settings. If you were using the obsolete settings from the older versions, it's needed to update or remove them. 
>
> Database logs will provide the failure details, e.g. for `EVENTSTORE_EXT_HTTP_PORT`, you will see the following line `Error while parsing options: The option ExtHttpPort is not a known option. (Parameter 'ExtHttpPort') `. After replacing with the valid param EVENTSTORE_HTTP_PORT database should start properly.

@ylorph thoughts? Might be worth also adding information about the `EVENTSTORE` prefix and issue that @alexeyzimarev had.

ylorph: The log does not provide any error , only the console output

ylorph: rephrased @oskardudycz

oskardudycz (approved):

hayley-jean (approved):

PR #3083, opened 2021-08-04
Always use index tables awaiting merge in the correct order

jageall:
Fixed: Stop using older ptables that are awaiting merge when doing index lookups

When ptables are in the _awaitingMemTables queue, they shouldn't be considered before their position in the index map if they were put in the queue for merging from the index map

### DIFF
--- a/src/EventStore.Core/Index/TableIndex.cs
+++ b/src/EventStore.Core/Index/TableIndex.cs
@@ -100,7 +100,7 @@ namespace EventStore.Core.Index {
 			_indexCacheDepth = indexCacheDepth;
 			_initializationThreads = initializationThreads;
 			_ptableVersion = ptableVersion;
-			_awaitingMemTables = new List<TableItem> {new TableItem(_memTableFactory(), -1, -1, 0)};
+			_awaitingMemTables = new List<TableItem> {new TableItem(_memTableFactory(), -1, -1, 0, false)};
 
 			_lowHasher = lowHasher;
 			_highHasher = highHasher;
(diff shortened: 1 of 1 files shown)


jageall:
This will need to be cherry picked onto 20.10 and master

src/EventStore.Core/Index/TableIndex.cs
```diff
@@ -466,7 +466,7 @@ public class TableIndex : ITableIndex {
 						_awaitingMemTables[j] = new TableItem(ptable,
 							tableItem.PrepareCheckpoint,
 							tableItem.CommitCheckpoint,
-							tableItem.Level);
+							tableItem.Level, false);
```
shaan1337, line 469: @jageall Instead of passing `false` here, would it be better to pass in `tableItem.IsFromIndexMap`, then in the first for loop, skip those awaiting tables where `IsFromIndexMap` is `true`?

shaan1337 (commented):
Just a question regarding the change in `ReclaimMemoryIfNeeded`, the rest looks good to me :+1: 

Also, it appears that we don't need to make changes on `master`/`20.10` since manual index merges no longer modifies the awaiting tables

src/EventStore.Core/Index/TableIndex.cs
jageall: I thought about that, but as the code cannot currently be used, I went with minimal changes. I'd suggest we follow up on this by either removing the method that cannot be called, or making this and some other changes when we test and make this option available to consumers.

shaan1337:
Fully automated reproduction for the issue (tested with 5.0.5): [max_auto_merge_bug_repro_fully_automated.zip](https://github.com/EventStore/EventStore/files/6973695/max_auto_merge_bug_repro_fully_automated.zip)

shaan1337:
Tested repro with:
20.10.2 - bug is not present
v5-master with this fix - bug is no longer present

hayley-jean (approved):

shaan1337 (approved):

PR #3104, opened 2021-08-10
Update standard backup procedure

timothycoleman:
- Added support for including the bloom filter in the backup
- Fixed bug where checkpoints wouldn't be updated on windows (-I flag: see below)
- Fixed bug where index directory would be copied into index directory resulting in backup/index/index/

The -I flag is required on windows for memory mapped files, because the modified date is not reliably updated. We could add a call to set the modified date excplicitly when flushing and remove this flag.
Note that rsync will still be fairly efficient because of its delta transfer algorithm.

### DIFF
--- a/docs/operations/database-backup.md
+++ b/docs/operations/database-backup.md
@@ -63,16 +63,18 @@ of the volume containing the `index\ ` directory and then a snapshot of the volu
 
 ### Backup 
 
-1. Copy the indexes to your backup location (the content of the `index\` directory).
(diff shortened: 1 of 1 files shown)


oskardudycz:
@timothycoleman, do you know if those changes are compatible with previous DB versions (e.g. 5.x, 20.x)?

hayley-jean (dismissed):

shaan1337 (approved):

hayley-jean (approved):

shaan1337 (approved):

PR #3121, opened 2021-08-25
Make Leader Election timeout configurable

hayley-jean:
Added: LeaderElectionTimeoutMs option to allow configuring the timeout for election messages.

Adds the LeaderElectionTimeoutMs option which configures the LiveUntil and Deadlines of messages in ElectionsService

- Deadlines when sending messages over gRPC will be based on the LiveUntil of the messages, rather than setting a deadline separately.
- Update GossipService to use the GossipTimeout as the LiveUntil of gossip messages.

### DIFF
--- a/docs/clustering/gossip.md
+++ b/docs/clustering/gossip.md
@@ -79,3 +79,17 @@ Please note that the `GossipOnSingleNode` option has been deprecated in this ver
 | Environment variable | `EVENTSTORE_GOSSIP_ON_SINGLE_NODE` |
 
 **Default**: `true`
+
+## Leader Election Timeout
+
+The Leader Elections are separate to the node gossip, and are used to elect a node as Leader and assign roles to the other nodes.
+
+In some cases the leader election messages may be delayed, which can result in elections taking longer than they should. If you start seeing election timeouts in the logs or if you've needed to increase the gossip timeout due to a congested network, then you should consider increasing the leader election timeout as well.
+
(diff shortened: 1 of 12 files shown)


thefringeninja (dismissed):

jageall (dismissed):

Lougarou:
The parameter looks to be working as intended. 
Only issue I see is that it is not displayed in Modified Options at startup (the default value is shown). 
Also should it be in second or millisecond given that Gossip messages are in Milliseconds?

Lougarou (approved):

thefringeninja (approved):

jageall (dismissed):

Lougarou (dismissed):

Lougarou:
looks good although all values less than 1 second (for example 999) gives the following error: System.ArgumentOutOfRangeException: leaderElectionTimeout should be positive. (Parameter 'leaderElectionTimeout')

hayley-jean:
> looks good although all values less than 1 second (for example 999) gives the following error: System.ArgumentOutOfRangeException: leaderElectionTimeout should be positive. (Parameter 'leaderElectionTimeout')

Thanks! I've changed it to:

```
System.ArgumentOutOfRangeException: leaderElectionTimeout should be greater than 1 second. (Parameter 'leaderElectionTimeout')
```

thefringeninja (approved):

Lougarou (approved):

PR #3133, opened 2021-08-26
implement non-subscription gRPC filtered reads of $all

### LINKED ISSUE #2980: Add a filter option when reading from a stream

When I want to read the last event in the event store (to measure the "gap/lag" from my read side), I use something like this:

`var events = client.ReadAllAsync(Direction.Backwards, Position.End, 1);`

Great! But, that will also include internal $EventStore events.

I want to be able to include a filter, to avoid getting the internal $xxx events to be included when I query for the last event. So that I in a more correct way can estimate the true "gap" between my read-side projection and the head of the eventstore.

Right now just reading the last event will "corrupt" my estimations.

Adding a filter option like we have for subscriptions would be useful here (SubscriptionFilterOptions).

### LINKED ISSUE #3128: non-subscription ReadReq to `$all` with a filter
:wave: hello!

What do y'all think about being able to use a filter on non-subscription read requests to the `$all` stream?

With the gRPC API we can filter subscription read requests, and with v21.6.0 we can do the same for persistent subscriptions, but for non-subscription read requests we can only read from `$all` with no filter. It looks like the protobufs technically support sending a filter (because of subscriptions) in a non-subscription ReadReq, but the RPC ~returns a gRPC status-code 2 if you do send one~ (on master it returns a status-code 3 with the helpful message: "The combination of (All, Count, Forwards, Filter) is invalid." since [this commit](https://github.com/EventStore/EventStore/commit/89f2000a944ab694fd4dcf2b5057b43b258cbbde#diff-b4a144b42979d2bcf4d8d82ddb732874a39d13b3966f28fe363540f0eadf9893R143-R144)).

Being able to filter with a non-subscription read request would

1. allow you to easily synchronously query the `$all` stream
    - (although I think a client library user could probably build that same functionality on top of (persistent) subscription read requests with a little elbow grease)
2. make the reading RPCs more consistent
3. allow consumers to implement a non-persistent subscription with proper demand based back-pressure to a filtered `$all` stream

<details><summary>With respect to no. 3</summary>

For any other stream, you can write a consumer which reads through a stream in-order with a low memory footprint by having the consumer read chunks of the stream (with non-subscription `ReadReq`s) on-demand until it reaches (near) the end of the stream, where it subscribes.

</details>

I'd like the Streams Read RPC handler to allow a filter to be passed, and return events in the same fashion as a subscription to `$all`+filter emits. I don't have strong opinions about this being implemented for the backwards direction, just forwards is fine.

**Describe alternatives you've considered**

W.r.t. point no. 3 above, without this feature, a consumer is forced to either use a persistent subscription (which gives back-pressure but may not give events in-order, which could be important to the consumer) or to just plain subscribe and allow itself to be flooded by the subscription.

**Additional context**

I (probably naively) think this feature might not be a whole lot of work to implement (in EventStoreDB or in the client libraries; the [change in NFIBrokerage/spear](https://github.com/NFIBrokerage/spear/compare/read-all-stream) would only be 10 lines!) because the protobufs already support passing the filter. That being said I think we'd need a branch in this switch [here](https://github.com/EventStore/EventStore/blob/bbed04100b91b23b307920f115ebf903eb353388/src/EventStore.Core/Services/Transport/Grpc/Streams.Read.cs#L42-L145) to add something like:

```cs
(StreamOptionOneofCase.All, CountOptionOneofCase.Count, ReadDirection.Forwards, FilterOptionOneofCase.Filter) => new Enumerators.ReadAllForwardsFiltered(..)
```

And then it'd be a whole new implementation for Enumerators.ReadAllForwardsFiltered probably based closely on Enumerators.AllSubscriptionFiltered.

There's also a question to be asked about what to do with checkpoints. Since ReadReq for non-subscription takes a count of events to return, I think it makes sense to omit checkpoints entirely.

I'd be happy to attempt a PR for it, but please know my C# skills are very lacking :sweat_smile:

the-mikedavis:
closes #3128

I think this also connects/closes #2980 (didn't see that before I made my issue, sorry for the noise!)

This PR adds enumerators for reading the $all stream forwards and backwards with a filter outside of a subscription in the gRPC API.

The enumerator is a combination of the SubscribeAllFiltered and ReadAllForwards, switching out ReadAllForward's `ClientMessage.ReadAllEventsForward` to the `ClientMessage.FilteredReadAllEventsForward` versions (and same for the Backwards direction).

I've tried out the changes manually against the Elixir community gRPC client here: https://github.com/NFIBrokerage/spear/pull/56#issue-980658981 and it seems to work pretty well!

I'm a total C# noob so if anything here is out of the ordinary, don't be shy to push some changes or let me know :slightly_smiling_face:

### DIFF
--- a/src/EventStore.Core.Tests/Services/Transport/Grpc/StreamsTests/ReadAllBackwardsFilteredTests.cs
+++ b/src/EventStore.Core.Tests/Services/Transport/Grpc/StreamsTests/ReadAllBackwardsFilteredTests.cs
@@ -9,14 +9,15 @@ using NUnit.Framework;
 
 namespace EventStore.Core.Tests.Services.Transport.Grpc.StreamsTests {
 	[TestFixture]
-	public class ReadAllBackwardsTests {
+	public class ReadAllBackwardsFilteredTests {
 		[TestFixture(typeof(LogFormat.V2), typeof(string))]
 		[TestFixture(typeof(LogFormat.V3), typeof(uint))]
-		public class when_reading_all_backwards<TLogFormat, TStreamId> : GrpcSpecification<TLogFormat, TStreamId> {
-			private const string StreamId = nameof(when_reading_all_backwards<TLogFormat, TStreamId>);
+		public class when_reading_all_backwards_filtered<TLogFormat, TStreamId>
+		  : GrpcSpecification<TLogFormat, TStreamId> {
+			private const string StreamId = nameof(when_reading_all_backwards_filtered<TLogFormat, TStreamId>);
 
-			private readonly List<ReadResp> _responses = new();
 			private Position _positionOfLastWrite;
+			private readonly List<ReadResp> _responses = new();
 
 			protected override async Task Given() {
 				var response = await AppendToStreamBatch(new BatchAppendReq {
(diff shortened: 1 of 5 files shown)


hayley-jean:
Thanks for the PR @the-mikedavis!

We'll give this a review, but may have to wait until this issue https://github.com/eventstore/eventstore-client-dotnet/issues/106 is complete before we merge, as there could be more changes needed on the server proto for reads

thefringeninja (approved):

hayley-jean (approved):

hayley-jean:
Looks good to me, the tests are failing for an unrelated reason so I'm happy to merge

JimGoods:
@hayley-jean @thefringeninja Has support for this made it to the .NET client? I've had a good look but can't find it.

PR #3138, opened 2021-08-28
fix BatchAppend deadline race condition

### LINKED ISSUE #3137: race condition in gRPC BatchAppend deadline
**Describe the bug**

It looks like it's possible to accidentally set an infinite deadline when the passed deadline represents a `TimeSpan` less than or equal to `TimeSpan.Zero` because of a race condition in how deadlines are handled in the Streams.BatchAppend receiver.

**To Reproduce**

Write a gRPC AppendBatchReq to the BatchAppend RPC with the [`deadline` timestamp](https://github.com/EventStore/EventStore/blob/a788719bf50b4eb3daa7413407f9b1e7539d513b/src/Protos/Grpc/streams.proto#L221) set to anything in the past. This is can be sent by any gRPC client.

(See example test cases in the next section.)

**Expected behavior**

I would expect deadline timestamps in the past (according to the current server timestamp) to immediately send a BatchAppendResp with the deadline exceeded status. Specifically I want this adaptation of [this test case](https://github.com/EventStore/EventStore/blob/08633f0b1a42083dd7a5d7df711e1ed66597e7c4/src/EventStore.Core.Tests/Services/Transport/Grpc/StreamsTests/AppendBatchToStreamTests.cs#L12-L32) to **fail**:

```cs
public class single_batch<TLogFormat, TStreamId> : GrpcSpecification<TLogFormat, TStreamId> {
	private BatchAppendResp _response;
	protected override Task Given() => Task.CompletedTask;

	protected override async Task When() {
		_response = await AppendToStreamBatch(new BatchAppendReq {
			Options = new() {
				Any = new(),
				StreamIdentifier = new() { StreamName = ByteString.CopyFromUtf8("stream") },
				Deadline = Timestamp.FromDateTime(DateTime.UtcNow.AddSeconds(-1))
			},
			IsFinal = true,
			ProposedMessages = { CreateEvents(1) },
			CorrelationId = Uuid.NewUuid().ToDto()
		});
	}

	[Test]
	public void is_success() {
		Assert.AreEqual(_response.ResultCase, BatchAppendResp.ResultOneofCase.Success);
	}
}
```

(The amount of time behind the current timestamp is not significant, just that the timestamp compares out to `<= TimeSpan.Zero` in the server.)

See also [this test case in NFIBrokerage/spear](https://github.com/NFIBrokerage/spear/pull/57/files#diff-9158c14b2b8bbb4d7deb17a135eada71482fa8170c78d9ca56f5356b3e6be01dR922-R956) which I would expect to pass.

**Actual behavior**

The server attempts to write the given batch as if a deadline had not been declared.

**Config/Logs/Screenshots**

N/A

**EventStore details**

This behavior can be provoked in the test suite. I have also tried it against the stock `eventstore/eventstore:21.6.0-buster-slim` docker image, as well as on master by building and running the usual 

```
dotnet ./src/EventStore.ClusterNode/bin/x64/Release/net5.0/EventStore.ClusterNode.dll --insecure --db ./tmp/data --index ./tmp/index --log ./tmp/log -runprojections all --startstandardprojections --EnableAtomPubOverHttp
```

 - EventStore server version: v21.6.0+

 - Operating system: N/A

 - EventStore client version (if applicable): I'm working off [this branch](https://github.com/NFIBrokerage/spear/pull/57) in NFIBrokerage/spear but it's possible to reproduce this behavior with any client

**Additional context**

BatchAppend correctly computes the deadline `TimeSpan` given the current time and sets up a callback to send a proper timeout BatchAppendResp [here](https://github.com/EventStore/EventStore/blob/08633f0b1a42083dd7a5d7df711e1ed66597e7c4/src/EventStore.Core/Services/Transport/Grpc/Streams.BatchAppend.cs#L261-L270).

Then [in the ClientWriteRequest constructor](https://github.com/EventStore/EventStore/blob/08633f0b1a42083dd7a5d7df711e1ed66597e7c4/src/EventStore.Core/Services/Transport/Grpc/Streams.BatchAppend.cs#L308-L314), the deadline gets checked against `TimeSpan.Zero`. If the deadline has passed according to this check, the callback is invoked, else it's set up to be invoked when the deadline has passed with `Task.Delay`.

If the deadline comes before the current timestamp, then the callback is invoked immediately. This is where the race condition comes in:

```cs
pendingWrites.TryRemove(correlationId, out var pendingWrite)
	? writer.WriteAsync(new BatchAppendResp {
		CorrelationId = Uuid.FromGuid(correlationId).ToDto(),
		StreamIdentifier = new StreamIdentifier {
			StreamName = ByteString.CopyFromUtf8(pendingWrite.StreamId)
		},
		Error = Status.Timeout
	}, cancellationToken)
	: new ValueTask(Task.CompletedTask)
```

The timeout callback attempts to remove the `pendingWrite` from the `ConcurrentDictionary pendingWrites` . This works fine when the deadline is in the future and the task is delayed with `Task.Delay`, but it's a problem when the deadline has already passed because the `ClientWriteRequest pendingWrite` in question is currently being inserted into the `ConcurrentDictionary pendingWrites`:

- [`pendingWrites.AddOrUpdate`](https://github.com/EventStore/EventStore/blob/08633f0b1a42083dd7a5d7df711e1ed66597e7c4/src/EventStore.Core/Services/Transport/Grpc/Streams.BatchAppend.cs#L144-L145) calls the `FromOptions` method to create a new `ClientWriteRequest` to insert into the dictionary
- [`FromOptions`](https://github.com/EventStore/EventStore/blob/08633f0b1a42083dd7a5d7df711e1ed66597e7c4/src/EventStore.Core/Services/Transport/Grpc/Streams.BatchAppend.cs#L261-L270) computes the deadline and calls the `ClientWriteRequest` constructor
- [the `ClientWriteRequest` constructor](https://github.com/EventStore/EventStore/blob/08633f0b1a42083dd7a5d7df711e1ed66597e7c4/src/EventStore.Core/Services/Transport/Grpc/Streams.BatchAppend.cs#L309) immediately invokes the timeout callback when the deadline is in the past
- finally the timeout callback invokes [`pendingWrites.TryRemove`](https://github.com/EventStore/EventStore/blob/08633f0b1a42083dd7a5d7df711e1ed66597e7c4/src/EventStore.Core/Services/Transport/Grpc/Streams.BatchAppend.cs#L262)

`TryRemove` will always fail to remove the key because it is being invoked immediately in the constructor of the object which is about to be inserted, so it always returns `false`, and so the `BatchAppendResp` with `Error = Status.Timeout` will not be emitted in this case; instead the timeout callback will no-op by returning `new ValueTask(Task.CompletedTask)`.

the-mikedavis:
Fixed: potential race condition in BatchAppend

closes #3137 

Adds a test case and fixes a race condition in BatchAppend where deadlines in the past were not being respected as exceeded.

It seems a bit unweildly to pass around the extra `timeout` argument to the `FromOptions` method instead of just recomputing it, but I'm doing it specifically to avoid another potential race condition:  recomputing the `timeout` in `FromOptions` could produce a different TimeSpan which could be exceeded (because `GetRequestTimeout` calls `DateTime.UtcNow`), leading to Task.Delay getting passed a negative TimeSpan. I'd definitely open to suggestions if there's a more idiomatic way to do this :slightly_smiling_face:

### DIFF
--- a/src/EventStore.Core.Tests/Services/Transport/Grpc/StreamsTests/AppendBatchToStreamTests.cs
+++ b/src/EventStore.Core.Tests/Services/Transport/Grpc/StreamsTests/AppendBatchToStreamTests.cs
@@ -1,3 +1,5 @@
+using System;
+using Google.Protobuf.WellKnownTypes;
 using System.Threading.Tasks;
 using EventStore.Client.Streams;
 using EventStore.Core.Services.Transport.Grpc;
@@ -57,6 +59,31 @@ namespace EventStore.Core.Tests.Services.Transport.Grpc.StreamsTests {
 			[Test]
 			public void is_success() {
 				Assert.AreEqual(_response.ResultCase, BatchAppendResp.ResultOneofCase.Success);
+			}
+		}
+
+		[TestFixture(typeof(LogFormat.V2), typeof(string))]
+		[TestFixture(typeof(LogFormat.V3), typeof(uint))]
+		public class exceeded_deadline<TLogFormat, TStreamId> : GrpcSpecification<TLogFormat, TStreamId> {
+			private BatchAppendResp _response;
+			protected override Task Given() => Task.CompletedTask;
+
+			protected override async Task When() {
+				_response = await AppendToStreamBatch(new BatchAppendReq {
+					Options = new() {
+						Any = new(),
+						StreamIdentifier = new() { StreamName = ByteString.CopyFromUtf8("stream") },
+						Deadline = Timestamp.FromDateTime(DateTime.UtcNow.AddSeconds(-1))
+					},
+					IsFinal = true,
+					ProposedMessages = { CreateEvents(1) },
+					CorrelationId = Uuid.NewUuid().ToDto()
+				});
+			}
+
+			[Test]
+			public void is_error() {
+				Assert.AreEqual(_response.ResultCase, BatchAppendResp.ResultOneofCase.Error);
 			}
 		}
 	}
--- a/src/EventStore.Core/Services/Transport/Grpc/Streams.BatchAppend.cs
+++ b/src/EventStore.Core/Services/Transport/Grpc/Streams.BatchAppend.cs
@@ -120,6 +120,8 @@ namespace EventStore.Core.Services.Transport.Grpc {
 							var correlationId = Uuid.FromDto(request.CorrelationId).ToGuid();
 
 							if (request.Options != null) {
+								TimeSpan timeout = Min(GetRequestedTimeout(request.Options), _writeTimeout);
+
 								if (!await _authorizationProvider.CheckAccessAsync(user, WriteOperation.WithParameter(
 									Plugins.Authorization.Operations.Streams.Parameters.StreamId(
 										request.Options.StreamIdentifier)), cancellationToken).ConfigureAwait(false)) {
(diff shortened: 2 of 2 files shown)


thefringeninja (dismissed):
The failing tests have nothing to do with this PR. Looks good!

jageall (dismissed):

the-mikedavis:
Ah whoops I didn't realize the rebase would dismiss the reviews :sweat_smile:, I haven't snuck any extra changes in I promise.

Should I add a line to the changelog as well? And is master the correct base for this or should I be targeting a particular release branch?

thefringeninja (approved):

jageall (approved):

PR #3139, opened 2021-08-28
Make VS nunit adapter happy

timothycoleman:
Fixed: test adapter error when running all tests in visual studio

- give each instance of the parameterized test a unique name (otherwise they aren't all shown/counted(/run?) in VS)
- use chevrons instead of brackets.. using brackets causes an extra test to be shown in VS (with a similar name only without the brackets present) that cannot be run. this does not itself cause an error, BUT causes the following error when run in conjunction with other tests that also have brackets in their names such as in LegacyPolicyVerification

`An exception occurred while invoking executor 'executor://nunit3testexecutor/': Incorrect format for TestCaseFilter Error: Missing '('. Specify the correct format and try again. Note that the incorrect format can lead to no test getting executed.`


thefringeninja (approved):
Looks good on linux!

hayley-jean (approved):

Issue #1744, opened 2018-10-26
Stream soft "undeletion" is not done in a single transaction


shaan1337:
### Scenario:
Consider the following scenario:
1. Soft delete a stream `s`
2. Write an event to stream `s` and wait for the write acknowledgement
3. Immediately read from stream `s` or stream `$$s` (metadata stream) (of the order of milliseconds, e.g. through the Client API)

When reading metadata events from stream `$$s`, truncate before may still be set to long.MaxValue.
Thus when reading from `s`, the read may result in a `StreamNotFound`.

This is because soft "undeletion" is not done in the [same transaction](https://github.com/EventStore/EventStore/blob/master/src/EventStore.Core/Services/Storage/StorageWriterService.cs#L298) as the event write.

The event write may be acknowledged without the metadata being updated on the stream yet.

### Potential issues:
- A node/network failure occurs after the event write but before the metadata update event (to update `truncateBefore`), resulting in the metadata event not being written to disk. The client will receive the write acknowledgement for the event and assume that the write was successful. When the node is back online, it will appear that the stream is still soft deleted (contains no events)

(Note: This failure scenario has not been tested but `EventStore.Core.Tests.ClientAPI.soft_delete` tests confirm the above behaviour)

oskardudycz:
Based on discussion with @hayley-jean, that won't be probably fixed for the current log, it may be included in the Log V3. We also haven't seen that as an issue from customers (so far).

condron:
This will be addressed with the new log features.

PR #3154, opened 2021-09-07
Allow ExpectedVersion.StreamExists for Delete and Tombstone

thefringeninja:
Added: support for ExpectedVersion.StreamExists in Soft Delete and Tombstone stream operations
Changed: Soft Deleting a stream that was never created now results in an error

### DIFF
--- a/src/EventStore.Core.Tests/Services/Transport/Grpc/StreamsTests/DeleteTests.cs
+++ b/src/EventStore.Core.Tests/Services/Transport/Grpc/StreamsTests/DeleteTests.cs
@@ -1,0 +1,116 @@
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using EventStore.Client.Streams;
+using EventStore.Core.Data;
+using EventStore.Core.Services.Transport.Grpc;
+using Google.Protobuf;
+using Grpc.Core;
+using NUnit.Framework;
+
+namespace EventStore.Core.Tests.Services.Transport.Grpc.StreamsTests {
+	[TestFixture]
+	public class DeleteTests {
+		private const string StreamName = nameof(DeleteTests);
+
+		public abstract class DeleteExistingStreamSpecification<TLogFormat, TStreamId>
+			: GrpcSpecification<TLogFormat, TStreamId> {
+			private readonly DeleteReq.Types.Options _options;
+			private Exception _caughtException;
+
+			private DeleteExistingStreamSpecification(DeleteReq.Types.Options options) {
+				_options = new(options) {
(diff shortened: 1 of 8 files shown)


thefringeninja:
Has implications for #1744

pvanbuijtene (dismissed):

src/EventStore.Core.Tests/Services/Transport/Grpc/StreamsTests/DeleteTests.cs
```diff
@@ -0,0 +1,151 @@
...
+				try {
+					await call.ResponseAsync;
+				} catch (Exception ex) {
+					_caughtException = ex;
+				}
+			}
+
+			[Test]
+			public void an_exception_is_thrown() {
+				if (_caughtException is not RpcException ex) {
```
timothycoleman, line 143: I wonder why IsInstanceOf doesn't return the object cast to the specified type

anyway, style thing, but perhaps this would be easier to read:

```
			[Test]
			public void an_exception_is_thrown() {
				Assert.IsInstanceOf<RpcException>(_caughtException);
				var ex = _caughtException as RpcException;
				Assert.AreEqual(StatusCode.FailedPrecondition, ex.StatusCode);
			}
```

src/EventStore.Core.Tests/Services/Transport/Grpc/StreamsTests/DeleteTests.cs
```diff
@@ -0,0 +1,151 @@
...
+
+namespace EventStore.Core.Tests.Services.Transport.Grpc.StreamsTests {
+	[TestFixture]
+	public class DeleteTests {
+		private const string StreamName = nameof(DeleteTests);
+
+		public abstract class DeleteExistingStreamSpecification<TLogFormat, TStreamId>
+			: GrpcSpecification<TLogFormat, TStreamId> {
+			private readonly DeleteReq.Types.Options _options;
+			private DeleteResp _deleteResponse;
```
timothycoleman, line 19: is is assigned to but not checked, is it supposed to be? same for _appendResponse

timothycoleman: (also in tombstone tests)

timothycoleman (approved):
👍

pvanbuijtene (approved):

timothycoleman:
during the review i asked for a change to the storagewriterservice to make sure that the soft delete is accepted if there is metadata for the stream, even if there are no events in the stream.

this was implemented, i just forgot to note it down here so noting it now

PR #3174, opened 2021-09-20
Log V3 : Stream exists if stream name exists in stream name index

shaan1337:


timothycoleman:
Closing for now by association with https://github.com/EventStore/EventStore/pull/3100

PR #3186, opened 2021-09-22
Fix off-by-one error in Index committer service

shaan1337:
Fixed: Off-by-one error in Index committer service

Thanks to @timothycoleman for finding the root cause and @hayley-jean for her assistance.


hayley-jean (approved):

timothycoleman (approved):
The replication checkpoint used to be the preposition of the most recently replicated checkpoint, but a while ago got changed to be the post position (in line with the writer and chaser checkpoints).

PR #3187, opened 2021-09-22
Wait for Replication and Indexing to complete before becoming leader

timothycoleman:
Fixed: race condition where a node becomes leader and immediately writes to streams that have recently been written to but not indexed, resulting in events with incorrect numbers in their streams.

- This change ensures that before becoming leader we successfully write an epoch and get it replicated. Thus demonstrating that we are in a fit state to become leader.
- Therefore the log is successfully replicated before we transition to leader.
- We also wait for the records to be indexed before we transition to leader, otherwise incorrect event numbers can be written because the latest events are not known to the index or IndexWriter

- The InaugurationManager is reponsible for making sure the conditions are met for transitioning to leader.
- In the PreLeader state we now accept replication subscriptions and monitor for quorum, transitioning to Unknown if a quorum is not maintained (exactly as we did in the leader state before)

- Works as a rolling upgrade from 20.10.0

### DIFF
--- a/src/EventStore.Core/Services/Storage/EpochManager/EpochManager.cs
+++ b/src/EventStore.Core/Services/Storage/EpochManager/EpochManager.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using EventStore.Common.Utils;
 using EventStore.Core.Bus;
 using EventStore.Core.Messages;
+using EventStore.Core.Data;
 using EventStore.Core.DataStructures;
 using EventStore.Core.Exceptions;
 using EventStore.Core.TransactionLog;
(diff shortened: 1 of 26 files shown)


timothycoleman:
`LeaderReplicationSerice::ManageRoleAssignments`
will probably start handing out roles while we are in pre-leader state now which is, ok?

src/EventStore.Core/Services/VNode/LastIndexableRecordFinder.cs
```diff
@@ -0,0 +1,43 @@
...
+				}
+
+				// coupled to the types of log record that the index indexes
+				// make this more explicit in future versions
+				if (result.LogRecord.RecordType == LogRecordType.Commit) {
+					_log.Information("Last indexable record is a commit at {position}", result.LogRecord.LogPosition);
+					return result.LogRecord.LogPosition;
+				}
+
+				if (result.LogRecord.RecordType == LogRecordType.Prepare) {
```
shaan1337, line 36: @timothycoleman should the case where a prepare is part of a transaction that has not yet been committed be taken into consideration here? (since it should not be indexable as long as the transaction has not been committed)

timothycoleman: 🤦‍♂️ yes definitely! good catch

timothycoleman: taken a different approach: whenever we write an epoch we immediately write an $epoch-information event to the $epoch-information stream. This allows the InaugurationManager to know where the
index checkpoint is supposed to reach. As a bonus this also gives the users a way to get details about the epochs. These events will not be present in an unfiltered $all read, for backwards
compatibilty, but can be read directly from their stream.

timothycoleman:
force pushed to single commit, full history is on the `preleader-replication-details` branch

src/EventStore.Core/Services/VNode/LastIndexableRecordFinder.cs
shaan1337: nice idea, thanks!

src/EventStore.Core/Services/Storage/EpochManager/EpochManager.cs
```diff
@@ -267,11 +268,90 @@ public class EpochManager : IEpochManager {
...
+					originalLogPosition,
+					retryLogPosition));
+		}
+
+		// we have just written epoch. about to write the $epoch-information for it.
+		// this decides what expected version the $epoch-information should have.
+		// it looks up the previous epoch, finds that epoch's previous information
+		// (which immediately follows it) and gets its event number.
+		bool TryGetExpectedVersionForEpochInformation(EpochRecord epoch, out long expectedVersion) {
+			expectedVersion = default;
```
shaan1337, line 328: @timothycoleman if epoch.PrevEpochPosition is -1, we can directly return false here otherwise it will reposition the reader to -1 (i'm not sure if the reader's behaviour is well defined in this case)

src/EventStore.Core/Services/Storage/ReaderIndex/EventFilter.cs
```diff
@@ -7,6 +7,8 @@
 namespace EventStore.Core.Services.Storage.ReaderIndex {
 	public static class EventFilter {
 		public static IEventFilter None => new AlwaysAllowStrategy();
+		public static IEventFilter DefaultAllFilter { get; } =
+			new NotStrategy(StreamName.Prefixes(SystemStreams.EpochInformationStream));
```
shaan1337, line 11: @timothycoleman we can consider always allowing if the first character is '$' then fall back to the full prefix check. This is much faster for user streams which make up the majority of the transaction log. There's a similar implementation on `master`: https://github.com/EventStore/EventStore/blob/d5fe9fa1d9a72f0f9dcca4edb9efc90397c073bf/src/EventStore.Core/Services/Storage/ReaderIndex/EventFilter.cs#L66

src/EventStore.Core/Services/Storage/ReaderIndex/AllReader.cs
```diff
@@ -48,7 +48,7 @@ public class AllReader : IAllReader {
 		}
 
 		public IndexReadAllResult ReadAllEventsForward(TFPos pos, int maxCount) {
-			return ReadAllEventsForwardInternal(pos, maxCount, maxCount, EventFilter.None);
+			return ReadAllEventsForwardInternal(pos, maxCount, int.MaxValue, EventFilter.DefaultAllFilter);
```
shaan1337, line 51: hmmm, i think that the reason for setting the `maxSearchWindow` to `maxCount` is to add an upper bound to the read, although a multiple of maxCount (e.g 3 x maxCount) sounds more reasonable.

shaan1337, line 51: hmmm, i think that we need to make the change to int.MaxValue on `master` too, otherwise the read may return less than maxCount events

shaan1337: and given that there are only very specific events filtered out, the read should not take that long with int.MaxValue, so it looks good :thinking:

src/EventStore.Core/Services/Storage/ReaderIndex/EventFilter.cs
timothycoleman: made a bespoke filter for the all default that just does the inequality check on the stream name, it does perform a lot better. there doesn't seem to be an additional boost from checking the $ prefix first, probably because here (unlike in master) we are just doing equality and not a prefix check

src/EventStore.Core/Services/Storage/ReaderIndex/AllReader.cs
timothycoleman: yep thats it exactly 👍 . indeed the forward version of this test `be_able_to_read_all_one_by_one_until_end_of_stream` fails without it - it gets immediate endofstream

shaan1337 (approved):
great work :+1:
i've reviewed and tested some scenarios, just a small suggestion:
```
[13173,14,11:30:56.838,INF] "InaugurationManager" in state (PreLeader, WaitingForConditions): Replication "DONE": 681,693,696/681,693,033. Remaining: -663.
[13173,14,11:30:56.838,INF] "InaugurationManager" in state (PreLeader, WaitingForConditions): Index "DONE": 681,693,321/681,693,033. Remaining: -288.
```

for the case when it's done can we hide the progress? (i.e hide : `681,693,321/681,693,033. Remaining: -288`) otherwise it seems a bit strange to see a negative value for `Remaining`

shaan1337 (approved):

timothycoleman:
> great work 👍 i've reviewed and tested some scenarios, just a small suggestion:
> 
> ```
> [13173,14,11:30:56.838,INF] "InaugurationManager" in state (PreLeader, WaitingForConditions): Replication "DONE": 681,693,696/681,693,033. Remaining: -663.
> [13173,14,11:30:56.838,INF] "InaugurationManager" in state (PreLeader, WaitingForConditions): Index "DONE": 681,693,321/681,693,033. Remaining: -288.
> ```
> 
> for the case when it's done can we hide the progress? (i.e hide : `681,693,321/681,693,033. Remaining: -288`) otherwise it seems a bit strange to see a negative value for `Remaining`

yeah you're right it is a bit surprising. ive made it show 0 when there is nothing remaining, instead of negative

hayley-jean (approved):
Looks good 👍

jageall (approved):

PR #3199, opened 2021-09-30
Bringing HTTP into a single doc

mat-mcloughlin:


mat-mcloughlin:
not sure I follow. This is just the HTTP docs?

mat-mcloughlin:
ah, yeah, no text changes, just merging

PR #3254, opened 2021-10-29
Intermediate certificate documentation

pvanbuijtene:
Added documentation for support of intermediate certificate authorities.

### DIFF
--- a/docs/security/configuration.md
+++ b/docs/security/configuration.md
@@ -126,6 +126,34 @@ You need to add the certificate thumbprint setting on Windows so the server can 
 | YAML                 | `CertificateThumbprint` |
 | Environment variable | `EVENTSTORE_CERTIFICATE_THUMBPRINT` |
 
+
+
+### Intermediate CA certificates
+
+Intermediate certificates are supported by loading them from a [PEM](https://www.rfc-editor.org/rfc/rfc1422) or [PKCS #12](https://datatracker.ietf.org/doc/html/rfc7292) bundle. The node's certificate should be first in the bundle, followed by the intermediates (intermediates can be in any order but it would be good to keep it from leaf to root as per the usual convention). The certificate chain is validated on startup with the node's own certificate.
+
+Additionally, in order to improve performance, the server will also try to bypass intermediate certificate downloads, when they are available on the system in the appropriate locations:
+
+#### Steps on Linux:
+The following script assumes EventStoreDB is running under the `eventstore` account.
+```
+sudo su eventstore --shell /bin/bash
+dotnet tool install --global dotnet-certificate-tool
+ ~/.dotnet/tools/certificate-tool add --file /path/to/intermediate.crt
+```
+
+#### Steps on Windows:
+To import the intermediate certificate in the Certificate store, run the following PowerShell under the same account as EventStoreDB is running:
+```PowerShell
+Import-Certificate -FilePath .\path\to\intermediate.crt -CertStoreLocation Cert:\CurrentUser\CA
+```
+
+To import the intermediate certificate in the `Local Computer` store, run the following as `Administrator`:
+```PowerShell
+Import-Certificate -FilePath .\ca.crt -CertStoreLocation Cert:\LocalMachine\CA
+```
+
+
 ## Certificate Generation CLI
 
 Event Store provides the interactive Certificate Generation CLI, which creates self-signed certificates for EventStoreDB. You can use the [configuration wizard](../installation/README.md), that will provide you exact CLI commands that you need to run to generates certificates matching your configuration. 


docs/security/configuration.md
```diff
@@ -126,6 +126,26 @@ You need to add the certificate thumbprint setting on Windows so the server can
...
 
+
+
+### Intermediate CA certificates
+
+Intermediate certificates are supported by loading them from a PEM or PKCS #12 bundle. The node's certificate should be first in the bundle, followed by the intermediates (intermediates can be in any order but it would be good to keep it from leaf to root as per the usual convention). The certificate chain is validated on startup with the node's own certificate.
+
+Additionally, in order to improve performance, the server will also try to bypass intermediate certificate downloads, when they are available on the system in the appropriate locations:
+
+#### Steps on Linux:
```
ylorph, line 137:
```suggestion
#### Steps on Linux:

The following script assumes EventStoreDB is running under the `eventstore` account

```

docs/security/configuration.md
```diff
@@ -126,6 +126,26 @@ You need to add the certificate thumbprint setting on Windows so the server can
...
+```
+sudo su eventstore --shell /bin/bash
+dotnet tool install --global dotnet-certificate-tool
+ ~/.dotnet/tools/certificate-tool add --file /path/to/intermediate.crt
+```
+
+#### Steps on Windows: 
+Right click on the `intermediate certificate` -> `Install` (let windows choose the appropriate location - it should go under `Intermediate Certification Authorities` -> `Certificates` in the current user's certificate store.
```
ylorph, line 145:
```suggestion
In the Certificate store under which EVentStoreDB is running or the `Local Computer` store, right click on the `intermediate certificate` -> `Install` (let windows choose the appropriate location - it should go under `Intermediate Certification Authorities` -> `Certificates` 
```

ylorph (changes requested):

docs/security/configuration.md
```diff
@@ -126,6 +126,26 @@ You need to add the certificate thumbprint setting on Windows so the server can
 | YAML                 | `CertificateThumbprint` |
 | Environment variable | `EVENTSTORE_CERTIFICATE_THUMBPRINT` |
 
+
+
+### Intermediate CA certificates
+
+Intermediate certificates are supported by loading them from a PEM or PKCS #12 bundle. The node's certificate should be first in the bundle, followed by the intermediates (intermediates can be in any order but it would be good to keep it from leaf to root as per the usual convention). The certificate chain is validated on startup with the node's own certificate.
```
oskardudycz, line 133:
```suggestion
Intermediate certificates are supported by loading them from a [PEM](https://www.rfc-editor.org/rfc/rfc1422) or [PKCS #12](https://datatracker.ietf.org/doc/html/rfc7292) bundle. The node's certificate should be first in the bundle, followed by the intermediates (intermediates can be in any order but it would be good to keep it from leaf to root as per the usual convention). The certificate chain is validated on startup with the node's own certificate.
```

docs/security/configuration.md
oskardudycz: @pvanbuijtene question. Cannot we use the `Import-Certificate` PowerShell command for that? See more: https://github.com/EventStore/EventStore-Client-Dotnet/blob/master/samples/secure-with-tls/create-certs.ps1#L5

docs/security/configuration.md
oskardudycz: **QUESTION:** Do we have some recommendations on when to use and when not to use intermediate CA certificates?

Could we also show the example bundle structure or add a link to how to configure such a bundle?

docs/security/configuration.md
pvanbuijtene: Good point, will test it

docs/security/configuration.md
pvanbuijtene: It's always better to use one or more intermediate certificates as this is more secure, since the root CA is not directly used. People who deal with PKI should know this :)

We could show an example once the `es-gencert-cli` is updated, but not sure if those certs should be used in production and if it's wise to show it to people.

oskardudycz: I think that the audience trying to configure those certificates will not always be people dealing with PKI daily, so IMHO, it's always worth adding more context.

Regarding samples, I was more referring to somehow visualising or explaining the structure explained in the description.

docs/security/configuration.md
pvanbuijtene: Successfully tested the PowerShell commands and updated the docs accordingly.

docs/security/configuration.md
pvanbuijtene: Think this info is more for people who are already using intermediate certificates and want to know if it's supported and how. For others, things regarding certificates will just stay as-is.  

Don't think we are in the business of telling how to do PKI, in the mongodb docs they are referring to [prior knowledge](https://docs.mongodb.com/manual/tutorial/configure-ssl/#prerequisites)

Not sure if I'm the right person to make something visual for the website/docs 😬

ylorph (approved):

oskardudycz (approved):

PR #3266, opened 2021-11-05
Add modified differential backup procedure

timothycoleman:
Originally removed in https://github.com/EventStore/EventStore/pull/3126/files
The new procedure
- accounts for the new index bloom filters and stream existence filter
- ensures the last chunk in the backup (which was incomplete at the time of the backup) is re-backed up
- is a bit clearer/simpler about ensuring that the index was backed up successfully during index merges

### DIFF
--- a/docs/operations/database-backup.md
+++ b/docs/operations/database-backup.md
@@ -83,6 +83,40 @@ rsync -a data/*.0* backup
 2. Copy all files to the desired location.
 3. Create a copy of `chaser.chk` and call it `truncate.chk`. This effectively overwrites the restored `truncate.chk`.
 
+## Differential backup & restore
+
+The following procedure is designed to minimize the backup storage space, and can be used to do a full and differential backup.
+
+### Backing up
+
+Backup the index
+
+1. If there are no files in the index directory (apart from directories), go to step 7.
+2. Copy the `index/indexmap` file to the backup. If the source file does not exist, repeat until it does.
+3. Make a list `indexFiles` of all the `index/<GUID>` and `index/<GUID>.bloomfilter` files in the source.
+4. Copy the files listed in `indexFiles` to the backup, skipping file names already in the backup.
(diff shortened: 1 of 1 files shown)


gregoryyoung (commented):
"4. Compare the contents of the `indexmap` file in the source and the backup. If they are different (i.e. the indexmap file has changed since step 2), go back to step 1."

If there is something not in the indexmap but in the backup (EG: there was a merge etc during the backup) this step should be not needed. The file should just be ignored.

timothycoleman:
hey @gregoryyoung, if the indexmap file has changed in step 4 then it could be that it references ptables that we didn't back up cause they might not have existed when we started, so we have to go back. if that makes it any clearer?

docs/operations/database-backup.md
```diff
@@ -83,6 +83,39 @@ rsync -a data/*.0* backup
...
 
+## Differential backup & restore
+
+The following procedure is designed to minimize the backup storage space, and can be used to do a full and differential backup.
+
+### Backing up
+
+Backup the index
+
+1. Make a list `indexFiles` of all the `index/<GUID>` and `index/<GUID>.bloomfilter` files in the source.
```
shaan1337, line 94: @timothycoleman it seems that step 1 and 2 need to be swapped - otherwise if a merge occurs between step 1 and 2, the indexmap file can change before step 2 is executed and the original `indexFiles` list will not be accurate

shaan1337 (commented):
proposed one change but other than that the rest looks good to me :+1:

docs/operations/database-backup.md
timothycoleman: ah yes you're right. i think this change is necessary to be made since we are now comparing the contents of the index map file (instead of the content of the ptable listing like we were before)

timothycoleman:
I swapped the first two steps as suggested, I also adjusted steps 1 and 4 slightly to account for the fact that the indexmap file might not actually exist briefly

shaan1337 (approved):
looks good to me, nice updates :+1:

jageall (approved):

PR #3270, opened 2021-11-08
Fix server crash when multiple threads call IODispatcher

jageall:
Fixed: Exception in scheduled message callback crashes server
Fixed: IODispatcher is now threadsafe for request tracking

### DIFF
--- a/src/EventStore.Core/Helpers/IODispatcher.cs
+++ b/src/EventStore.Core/Helpers/IODispatcher.cs
@@ -15,18 +15,124 @@ using ReadStreamResult = EventStore.Core.Data.ReadStreamResult;
 namespace EventStore.Core.Helpers {
 	
 	public sealed class IODispatcher : IHandle<IODispatcherDelayedMessage>, IHandle<ClientMessage.NotHandled> {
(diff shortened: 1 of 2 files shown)


src/EventStore.Core/Helpers/IODispatcher.cs
```diff
@@ -404,15 +469,12 @@ public readonly
...
-					action(res);
+					if (_requestTracker.RemovePendingRead(res.CorrelationId)) {
+						action(res);
+					}
 				});
 			Delay(TimeSpan.FromMilliseconds(ReadTimeoutMs), () => {
-				RemovePendingRequest(corrId);
-				if (!_pendingReads.IsRegistered(corrId)) return;
-				_pendingReads.Remove(corrId);
+				_requestTracker.RemovePendingRead(corrId);
```
timothycoleman, line 477: need `if` here too

timothycoleman:
i see you added the `if's` for the happy paths :) 
just one more left which i commented above

thefringeninja (approved):

timothycoleman (approved):

PR #3306, opened 2021-11-17
Adding the Vector page

alexeyzimarev:


ylorph (approved):

alexeyzimarev:
Closes https://github.com/EventStore/advocacy/issues/173

oskardudycz (approved):

docs/diagnostics/vector.md
```diff
@@ -0,0 +1,79 @@
+# Vector
+
+> Vector is a lightweight and ultra-fast tool for building observability pipelines.
+> (from Vector website)
+
+You can use [Vector] for extracting metrics or logs from your self-managed EventStore server.
+
+It's also possible to collect metrics from the Event Store Cloud managed cluster or instance, as long as the Vector agent is running on a machine that has a direct connection to the EventStoreDB server. You cannot, however, fetch logs from Event Store Cloud using your own Vector agent.
```
oskardudycz, line 8: **QUESTION:** Do we know why it's impossible to fetch logs from ESC? Is it a temporary limitation?

alexeyzimarev: Because Vector needs access to the file system, it's mentioned on the page. We run Vector in ESC and ship logs to Google Cloud, but it's not managed by the customer.

oskardudycz: Indeed it's below, but it might be worth mentioning also that here to get the full message.

PR #3316, opened 2021-11-25
Fix Bug in Checkpoints on Live Filtered Subscriptions

thefringeninja:
Fixed: use last indexed position of all stream when consumer subscribes to all filtered live

### DIFF
--- a/src/EventStore.Core/Services/Transport/Grpc/Enumerators.AllSubscriptionFiltered.cs
+++ b/src/EventStore.Core/Services/Transport/Grpc/Enumerators.AllSubscriptionFiltered.cs
@@ -289,7 +289,7 @@ namespace EventStore.Core.Services.Transport.Grpc {
 							if (startPosition != Position.End) {
 								ReadHistoricalEvents(startPosition);
 							} else {
-								NotifyCaughtUp(startPosition);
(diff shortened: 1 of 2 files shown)


alexeyzimarev:
Would it fix https://github.com/EventStore/EventStore/issues/3320?

timothycoleman:
some whitespace has changed in the tests that (i think?) was more in keeping with the general style before

src/EventStore.Core/Services/Transport/Grpc/Enumerators.AllSubscriptionFiltered.cs
```diff
@@ -289,7 +289,9 @@ await foreach (var @event in liveEvents.Reader.ReadAllAsync(_cancellationToken)
 							if (startPosition != Position.End) {
 								ReadHistoricalEvents(startPosition);
 							} else {
-								NotifyCaughtUp(startPosition);
+								NotifyCaughtUp(startPosition == Position.End
```
timothycoleman, line 292: `startPosition == Position.End` is necessarily true here because we are in the else?

thefringeninja:
@alexeyzimarev only if you are doing a comparison on the checkpoint position and the last one you saw, in your code

hayley-jean (approved):

jageall (approved):

PR #3352, opened 2021-12-15
Update server protos for persistent subscriptions

hayley-jean:
Added: GetInfo, ReplayParked, List, RestartSubsystem operations to persistent subscription gRPC proto
Added: String ConsumerStrategy property when creating persistent subscriptions over gRPC
Changed: Deprecate NamedConsumerStrategy when creating persistent subscriptions over gRPC

Fixes https://github.com/eventstore/eventstore/issues/3108

### DIFF
--- a/src/EventStore.Core/Services/Transport/Grpc/PersistentSubscriptions.Info.cs
+++ b/src/EventStore.Core/Services/Transport/Grpc/PersistentSubscriptions.Info.cs
@@ -1,0 +1,202 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using EventStore.Client.PersistentSubscriptions;
+using EventStore.Core.Messages;
+using EventStore.Core.Messaging;
+using EventStore.Plugins.Authorization;
+using Grpc.Core;
+
+namespace EventStore.Core.Services.Transport.Grpc {
+	internal partial class PersistentSubscriptions {
+		private static readonly Operation GetInfoOperation = new Operation(Plugins.Authorization.Operations.Subscriptions.Statistics);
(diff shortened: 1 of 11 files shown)


jageall (dismissed):

jageall (dismissed):

timothycoleman (commented):
Looks good, just one question. The XXXCompleted callbacks don't complete the tcs in the case of unexpected messages, is that possible/likely to come up in pratice? If it does I wonder whether the thing awaiting the tcs will wait forever (or maybe something higher up will kill it anyway)

hayley-jean:
> Looks good, just one question. The XXXCompleted callbacks don't complete the tcs in the case of unexpected messages, is that possible/likely to come up in pratice? If it does I wonder whether the thing awaiting the tcs will wait forever (or maybe something higher up will kill it anyway)

Good point. I don't know how likely that is to come up but it's easy enough to complete with an error in those cases and is probably safer that way

src/EventStore.Core/Services/Transport/Grpc/PersistentSubscriptions.Info.cs
```diff
@@ -0,0 +1,193 @@
...
+								RpcExceptions.PersistentSubscriptionFailed(streamId, request.Options.GroupName,
+									completed.ErrorString));
+							return;
+						default:
+							getPersistentSubscriptionInfoSource.TrySetException(
+								RpcExceptions.PersistentSubscriptionFailed(streamId, request.Options.GroupName,
+									$"Unknown result for getting persistent subscription stats: '{completed.Result}'"));
+							return;
+					}
+				}
```
timothycoleman, line 70: optional, but same here?

timothycoleman (approved):
looks good :+1:

jageall (approved):

PR #3363, opened 2021-12-31
Fix Directory.EnumerateFiles regression

shaan1337:
Fixed: Directory.EnumerateFiles regression causing slower startup/truncation times on large databases

Multiple functions namely `GetAllLatestChunkVersions()`, `EnsureNoExcessiveChunks()`, `RemoveOldChunksVersions()` and `TruncateDb()` have a computational complexity of O(n^2) where n = number of chunks.

This was not noticeable earlier since Directory.EnumerateFiles was quite fast but with this ~50x regression in Directory.EnumerateFiles: https://github.com/dotnet/runtime/issues/31214, large databases can take several minutes to load compared to a few seconds in the past.

This PR fixes it by reducing the computational complexity of these methods to O(n lg n).
Existing tests already cover most of these changes - some tests have been strengthened to cover more cases.

## Benchmarks
- Tested with `--skip-index-verify --skip-db-verify` on Ubuntu 20.04/Windows Server 2019 with an AWS EBS gp2 volume on a t2.medium VM. v5-master was tested on Ubuntu 18.04 since I later realized mono runtime we need wasn't available for 20.04.
- Chunk size was adjusted to 256 KB instead of 256 MB
- On linux page cache was cleared between tests with `echo 3 > /proc/sys/vm/drop_caches`. Page cache wasn't cleared on windows.
- Truncation tests were done by simply copying `writer.chk` to `truncate.chk` and starting the database
- the min/max in the current benchmark refers to the time taken without and with the commit: a0940fd7e6c4c7cc50d7b0831bbdc106597761b3. It's basically the difference between loading a fully non-scavenged vs fully scavenged database.

| Test                                              | v5-master (linux)     | v5-master (win) | master (linux) | master (win) | this branch (linux)             | this branch (win)               |
|---------------------------------------------------|-----------------------|-----------------|----------------|--------------|---------------------------------|---------------------------------|
| Opening of chunks (10k 256kb chunks)              | 180 secs              | 7 secs          | 116 secs       | 85 secs      | min: 16 secs / max: 16 secs     | min: 19 secs / max: 25 secs     |
| Opening of chunks (80k 256kb chunks)              | OOM (4GB) & very slow | 4 mins          | 114 mins       | 52 mins      | min: 2.3 mins / max: 3.5 mins   | min: 2 mins / max: 4.6 mins     |
| Check excessive chunks (10k 256kb chunks)         | 101 secs              | 800 ms          | 67 secs        | 53 secs      | 300 ms                          | 700 ms                          |
| Check excessive chunks (80k 256kb chunks)         | Very slow             | 7 secs          | 113 mins       | 74 mins      | 1.5 secs                        | 5 secs                          |
| Remove old chunks (10k 256kb chunks)              | 101 secs              | 600 ms          | 121 secs       | 56 secs      | 250 ms                          | 700 ms                          |
| Remove old chunks (80k 256kb chunks)              | Very slow             | 4.7 secs        | 112 mins       | 104 mins     | 1.7 secs                        | 5 secs                          |
| Total opening time, excl index (10k 256kb chunks) | 6.4 mins              | 8.5 secs        | 5 mins         | 3 mins       | min: 16.5 secs / max: 16.5 secs | min: 20.5 secs / max: 26.5 secs |
| Total opening time, excl index (80k 256kb chunks) | Very slow             | 4.2 mins        | 5.5 hours      | 3.8 hours    | min: 2.3 mins / max: 3.5 mins   | min: 2.2 mins / max: 4.8 mins   |
| Truncate DB (10k 256kb chunks)                    | 115 secs              | 1.6 secs        | 80 secs        | 54 secs      | min: 130 ms / max: 9 secs       | min: 150 ms / max: 22 secs      |
| Truncate DB (80k 256kb chunks)                    | Very slow             | 2.5 mins        | 61 mins        | 52 mins      | min: 1.1s / max: 69 secs        | min: 1s / max: 4.3 mins         |

## Summary
The massive performance boost in loading times is applicable to:
- Users running on linux with any version (assuming that < v5-linux is as slow as v5-linux or slower)
- Users running on windows with v20 or later (v5-win users will not see a significant performance difference)

### DIFF
--- a/src/EventStore.Core/TransactionLog/FileNamingStrategy/VersionedPatternFileNamingStrategy.cs
+++ b/src/EventStore.Core/TransactionLog/FileNamingStrategy/VersionedPatternFileNamingStrategy.cs
@@ -1,14 +1,15 @@
 using System;
+using System.Diagnostics;
 using System.IO;
 using System.Text.RegularExpressions;
 using EventStore.Common.Utils;
 using System.Linq;
 
 namespace EventStore.Core.TransactionLog.FileNamingStrategy {
-	public class VersionedPatternFileNamingStrategy : IFileNamingStrategy {
+	public class VersionedPatternFileNamingStrategy : IVersionedFileNamingStrategy {
 		private readonly string _path;
 		private readonly string _prefix;
-		private readonly Regex _chunkNamePattern;
+		private readonly Regex _pattern;
 
 		public VersionedPatternFileNamingStrategy(string path, string prefix) {
 			Ensure.NotNull(path, "path");
@@ -16,7 +17,7 @@ namespace EventStore.Core.TransactionLog.FileNamingStrategy {
 			_path = path;
 			_prefix = prefix;
 
-			_chunkNamePattern = new Regex("^" + _prefix + @"\d{6}\.\w{6}$");
+			_pattern = new Regex("^" + _prefix + @"\d{6}\.\d{6}$");
 		}
 
 		public string GetFilenameFor(int index, int version) {
@@ -38,15 +39,42 @@ namespace EventStore.Core.TransactionLog.FileNamingStrategy {
 
 		public string[] GetAllVersionsFor(int index) {
 			var versions = Directory.EnumerateFiles(_path, string.Format("{0}{1:000000}.*", _prefix, index))
-				.Where(x => _chunkNamePattern.IsMatch(Path.GetFileName(x)))
+				.Where(x => _pattern.IsMatch(Path.GetFileName(x)))
 				.OrderByDescending(x => x, StringComparer.CurrentCultureIgnoreCase)
 				.ToArray();
 			return versions;
 		}
 
+		public int GetIndexFor(string fileName) {
+			if (!_pattern.IsMatch(fileName))
+				throw new ArgumentException($"Invalid file name: {fileName}");
+
+			var start = _prefix.Length;
+			var end = fileName.IndexOf('.', _prefix.Length);
+			Debug.Assert(end != -1);
+
+			if (!int.TryParse(fileName[start..end], out var fileIndex))
+				throw new ArgumentException($"Invalid file name: {fileName}");
+
+			return fileIndex;
+		}
+
+		public int GetVersionFor(string fileName) {
+			if (!_pattern.IsMatch(fileName))
+				throw new ArgumentException($"Invalid file name: {fileName}");
+
+			var dot = fileName.IndexOf('.', _prefix.Length);
+			Debug.Assert(dot != -1);
+
+			if (!int.TryParse(fileName[(dot+1)..], out var version))
+				throw new ArgumentException($"Invalid file name: {fileName}");
+
+			return version;
+		}
+
 		public string[] GetAllPresentFiles() {
 			var versions = Directory.EnumerateFiles(_path, string.Format("{0}*.*", _prefix))
-				.Where(x => _chunkNamePattern.IsMatch(Path.GetFileName(x)))
+				.Where(x => _pattern.IsMatch(Path.GetFileName(x)))
 				.ToArray();
 			return versions;
 		}
--- a/src/EventStore.Core.Tests/TransactionLog/Truncation/when_truncating_into_the_middle_of_completed_chunk.cs
+++ b/src/EventStore.Core.Tests/TransactionLog/Truncation/when_truncating_into_the_middle_of_completed_chunk.cs
@@ -20,7 +20,7 @@ namespace EventStore.Core.Tests.TransactionLog.Truncation {
 		public override async Task TestFixtureSetUp() {
 			await base.TestFixtureSetUp();
 
-			_config = TFChunkHelper.CreateDbConfigEx(PathName, 1711, 5500, 5500, -1, 1111, 1000, -1);
+			_config = TFChunkHelper.CreateDbConfigEx(PathName, 3711, 5500, 5500, -1, 1111, 1000, -1);
 
 			var rnd = new Random();
 			_file1Contents = new byte[_config.ChunkSize];
@@ -30,6 +30,8 @@ namespace EventStore.Core.Tests.TransactionLog.Truncation {
 
 			DbUtil.CreateSingleChunk(_config, 0, GetFilePathFor("chunk-000000.000001"), contents: _file1Contents);
 			DbUtil.CreateSingleChunk(_config, 1, GetFilePathFor("chunk-000001.000002"), contents: _file2Contents);
+			DbUtil.CreateSingleChunk(_config, 2, GetFilePathFor("chunk-000002.000000"));
+			DbUtil.CreateSingleChunk(_config, 3, GetFilePathFor("chunk-000003.000000"));
 
 			var truncator = new TFChunkDbTruncator(_config);
 			truncator.TruncateDb(_config.TruncateCheckpoint.ReadNonFlushed());
@@ -73,7 +75,7 @@ namespace EventStore.Core.Tests.TransactionLog.Truncation {
 		}
 
 		[Test]
-		public void all_chunks_are_preserved() {
+		public void chunks_after_truncation_point_are_deleted() {
 			Assert.IsTrue(File.Exists(GetFilePathFor("chunk-000000.000001")));
 			Assert.IsTrue(File.Exists(GetFilePathFor("chunk-000001.000002")));
 			Assert.AreEqual(2, Directory.GetFiles(PathName, "*").Length);
--- a/src/EventStore.Core.Tests/TransactionLog/Truncation/when_truncating_into_the_middle_of_multichunk.cs
+++ b/src/EventStore.Core.Tests/TransactionLog/Truncation/when_truncating_into_the_middle_of_multichunk.cs
@@ -20,9 +20,13 @@ namespace EventStore.Core.Tests.TransactionLog.Truncation {
 
 			DbUtil.CreateMultiChunk(_config, 0, 2, GetFilePathFor("chunk-000000.000001"));
 			DbUtil.CreateMultiChunk(_config, 0, 2, GetFilePathFor("chunk-000000.000002"));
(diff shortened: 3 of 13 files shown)


src/EventStore.Core/TransactionLog/FileNamingStrategy/VersionedPatternFileNamingStrategy.cs
```diff
@@ -44,6 +45,59 @@ public class VersionedPatternFileNamingStrategy : IFileNamingStrategy {
...
+				throw new ArgumentException($"Invalid file name: {fileName}");
+
+			return chunkNumber;
+		}
+
+		public void EnumerateAllFiles(
+			Func<string, int> getNextFileNumber,
+			Action<string, int, int> onLatestVersionFound = null,
+			Action<string, int> onOldVersionFound = null,
+			Action<string, int> onFileMissing = null) {
```
jageall, line 67: Prefer not to use optional parameters

src/EventStore.Core/TransactionLog/FileNamingStrategy/IFileNamingStrategy.cs
```diff
@@ -1,9 +1,16 @@
...
 	public interface IFileNamingStrategy {
 		string GetFilenameFor(int index, int version);
 		string DetermineBestVersionFilenameFor(int index);
 		string[] GetAllVersionsFor(int index);
 		string[] GetAllPresentFiles();
+		void EnumerateAllFiles(
+			Func<string, int> getNextFileNumber,
+			Action<string, int, int> onLatestVersionFound = null,
+			Action<string, int> onOldVersionFound = null,
+			Action<string, int> onFileMissing = null);
```
jageall, line 13: Prefer not to use optional parameters

jageall (changes requested):
Initial feedback

src/EventStore.Core/TransactionLog/FileNamingStrategy/VersionedPatternFileNamingStrategy.cs
shaan1337: Fixed in 3c1b84b61a4c4db2d03f367d0f74a1fa12106881

src/EventStore.Core/TransactionLog/FileNamingStrategy/IFileNamingStrategy.cs
shaan1337: Fixed in 3c1b84b61a4c4db2d03f367d0f74a1fa12106881

src/EventStore.Core/TransactionLog/FileNamingStrategy/VersionedPatternFileNamingStrategy.cs
```diff
@@ -44,6 +45,73 @@ public class VersionedPatternFileNamingStrategy : IFileNamingStrategy {
...
+			if (!_chunkNamePattern.IsMatch(fileName))
+				throw new ArgumentException($"Invalid file name: {fileName}");
+
+			var dot = fileName.IndexOf('.', _prefix.Length);
+			Debug.Assert(dot != -1);
+
+			if (!int.TryParse(fileName[(dot+1)..], out var chunkVersion))
+				throw new ArgumentException($"Invalid file name: {fileName}");
+
+			return chunkVersion;
```
jageall, line 74: This is subtly different from before, which is probably ok. The regex allows any character in the version part of the file name and I am not sure I can think of a good reason for it.

shaan1337: Would it be a good time to update the regex to accept only digits? Since this part of the code makes the assumption that it should be a number.

src/EventStore.Core/TransactionLog/FileNamingStrategy/VersionedPatternFileNamingStrategy.cs
```diff
@@ -44,6 +45,73 @@ public class VersionedPatternFileNamingStrategy : IFileNamingStrategy {
...
+		public void EnumerateAllFiles(
+			Func<string, int, int, int> getNextFileNumber,
+			Action<string, int, int> onLatestVersionFound,
+			Action<string, int> onOldVersionFound,
+			Action<string, int> onFileMissing) {
+			var allFiles = GetAllPresentFiles();
+			Array.Sort(allFiles, StringComparer.CurrentCultureIgnoreCase);
+
+			int expectedChunkNumber = 0;
+			for (int i = 0; i < allFiles.Length; i++) {
```
timothycoleman, line 86: since we're iterating through the files that exist rather than up to an expected chunk number, theres a problem if fewer files exist than we are expecting, consider:

```
		[Test]
		public void detect_no_database() {
			var config = TFChunkHelper.CreateSizedDbConfig(PathName, 4000, chunkSize: 1000);
			using (var db = new TFChunkDb(config)) {
				Assert.That(() => db.Open(verifyHash: false),
					Throws.Exception.InstanceOf<CorruptDatabaseException>()
						.With.InnerException.InstanceOf<ChunkNotFoundException>());
			}
		}
```

```
  Message: 
  Expected: instance of <EventStore.Core.Exceptions.CorruptDatabaseException> and property InnerException instance of <EventStore.Core.Exceptions.ChunkNotFoundException>
  But was:  no exception thrown
```

tambeau:
Related Support Ticket https://eventstore.freshdesk.com/a/tickets/110065

src/EventStore.Core/TransactionLog/FileNamingStrategy/VersionedPatternFileNamingStrategy.cs
shaan1337: @timothycoleman great catch! It sounds like passing an expected last file number (derived from the writer checkpoint) should allow us to detect this case.

src/EventStore.Core/TransactionLog/FileNamingStrategy/VersionedPatternFileNamingStrategy.cs
```diff
@@ -44,6 +45,73 @@ public class VersionedPatternFileNamingStrategy : IFileNamingStrategy {
...
+			Debug.Assert(dot != -1);
+
+			if (!int.TryParse(fileName[(dot+1)..], out var chunkVersion))
+				throw new ArgumentException($"Invalid file name: {fileName}");
+
+			return chunkVersion;
+		}
+
+		public void EnumerateAllFiles(
+			Func<string, int, int, int> getNextFileNumber,
```
timothycoleman, line 78: Would recommend using value tuples to name the params here (handy trick i picked up from geall) like

`Func<(string fileName, int sensibleName, int another), int>` for each of the actions so the meaning is clear

timothycoleman, line 78: The logic in here makes this class chunk specific, but it's named generically still.

Recommend taking this method (and the couple of private methods it uses) out to another, chunk specific class that takes a IFileNamingStrategy as a parameter to the constructor. Then it can also enumerate the files into an array during construction and resuse the array for each call to EnumerateAllFiles. On a typical db opening that would reduce the number of EnumerateFiles calls from 3 to 1. (and 2 to 1 in the truncator)

timothycoleman, line 78: Consider either
1. Ensuring that these arguments are not null and passing empty actions, or
2. Assigning empty actions to them if they are null

So that they can be called more naturally `onFileMissing(..)` vs `onFileMissing?.Invoke(..)`

src/EventStore.Core/TransactionLog/FileNamingStrategy/VersionedPatternFileNamingStrategy.cs
shaan1337: Resolved in eb837db3782c45df817204ecbd8b17142448ff1e

src/EventStore.Core/TransactionLog/FileNamingStrategy/VersionedPatternFileNamingStrategy.cs
shaan1337: Resolved in 23a4e1be4d917c382555d07394fa83aa423b4dfa

shaan1337: Resolved in 858121198a22f1e879dfc95d9b7d628e0c1d288f

src/EventStore.Core/TransactionLog/FileNamingStrategy/VersionedPatternFileNamingStrategy.cs
shaan1337: Resolved in 2784d510d333a70dc0996184701cc08fa238bca9

src/EventStore.Core/TransactionLog/FileNamingStrategy/VersionedPatternFileNamingStrategy.cs
shaan1337: Resolved in fb6fc57ee056288bd54f167649fc637a01b876db

shaan1337:
Latest benchmarks on 80,000 256kb chunks database

## Linux / non-scavenged chunks:

Opening: 2 mins 2 secs
Ensure no excessive chunks: 150 ms
Remove old chunks: 150 ms

## Linux / scavenged chunks (simulated by removing version = 0 optimization):

Opening: 2 mins 13 secs
Ensure no excessive chunks: 180 ms
Remove old chunks: 150 ms

## Windows / non-scavenged chunks:

Opening: 2 mins 10 secs
Ensure no excessive chunks: 160 ms
Remove old chunks: 120 ms

## Windows / scavenged chunks (simulated by removing version = 0 optimization):

Opening: 3 mins 58 secs
Ensure no excessive chunks: 185 ms
Remove old chunks: 200 ms

timothycoleman (approved):

PR #3365, opened 2022-01-04
Fix chunk end number not being set correctly

hayley-jean:
Fixed: Set chunk end number to the max between current end number and the added chunk number

This can cause an issue when multiple initialization threads are being used.


timothycoleman (approved):
:(

PR #3398, opened 2022-01-25
Send full certificate chain (excluding root) from client and server side

shaan1337:
Fixed: Send full certificate chain from both server and client side during TLS handshake (requires manually adding intermediate certificates to the store)

Fixes: https://github.com/eventstore/architecture-and-planning/issues/31

According to RFC 5246, https://datatracker.ietf.org/doc/html/rfc5246#section-7.4.2, intermediate certificates should be sent from both the server & client sides when present.

- Tests for HTTP have been done manually since it's not easily feasible with the current structure.

### DIFF
--- a/src/EventStore.ClusterNode/Program.cs
+++ b/src/EventStore.ClusterNode/Program.cs
@@ -1,5 +1,9 @@
 using System;
+using System.Collections.Generic;
+using System.Net.Security;
 using System.Runtime.InteropServices;
+using System.Security.Authentication;
+using System.Security.Cryptography.X509Certificates;
 using System.Threading;
 using System.Threading.Tasks;
 using EventStore.Common.Configuration;
@@ -10,7 +14,6 @@ using EventStore.Core.Services.Transport.Http;
 using EventStore.Core;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Server.Kestrel.Core;
-using Microsoft.AspNetCore.Server.Kestrel.Https;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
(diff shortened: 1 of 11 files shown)


src/EventStore.ClusterNode/Program.cs
```diff
@@ -181,5 +167,41 @@ internal static class Program {
...
 		}
+
+		private static ServerOptionsSelectionCallback CreateServerOptionsSelectionCallback(ClusterVNodeHostedService hostedService) {
+			return ((_, _, _, _) => {
+				var serverOptions = new SslServerAuthenticationOptions {
+					ServerCertificateContext = SslStreamCertificateContext.Create(
+						hostedService.Node.CertificateSelector(),
+						hostedService.Node.IntermediateCertificatesSelector(),
+						offline: true),
+					ClientCertificateRequired = true,
```
thefringeninja, line 178: It seems like this statement and the subsequent comment directly contradict each other...

shaan1337: @thefringeninja well spotted, i'll add a comment to clarify. the name `ClientCertificateRequired` is a bit misleading - it only means "Request a client certificate" but the client has the possibility not to supply one.

shaan1337: The RFC says:

```
7.4.6.  Client Certificate

When this message will be sent:

This is the first message the client can send after receiving a ServerHelloDone message.  This message is only sent if the server requests a certificate. If no suitable certificate is available, the client MUST send a certificate message containing no certificates.
```

shaan1337: Fixed in https://github.com/EventStore/EventStore/pull/3398/commits/0bd39c33e98682d8eee4da32efbd29ce821789f0

thefringeninja (approved):

hayley-jean (commented):
Changes look good, could you squash the commits before merging please @shaan1337?

shaan1337:
@hayley-jean thanks, I've squashed the commits

hayley-jean (approved):

shaan1337:
@hayley-jean @thefringeninja the PR is ready for review again with the following changes:
- Intermediate certificates are no longer automatically added to the certificate store on start up.
- Instead, a verification is done on startup and a warning is logged if the intermediate certificates are not present in the store. (in the next major version, this will be changed to an error otherwise it'll be a breaking change for current versions)
- Intermediate certificates docs have been updated

thefringeninja (approved):

PR #3412, opened 2022-02-02
Return Error if TrackEmittedStreams and !EmitEnabled

thefringeninja:
Changed: Reject gRPC call to create a projection if `TrackEmittedStreams=true` and `EmitEnabled=false`
Changed: Proto names of `emit_enabled` and `track_emitted_streams` swapped for backwards compatibility

This PR swaps the names of EmitEnabled and TrackEmittedStreams in the proto because what the server used to refer to as TrackEmittedStreams is, in fact, the correct behavior for EmitEnabled. This makes the recent change to 'Add EmitEnabled' really mean 'Add TrackEmittedStreams.'

Therefore clients that were relying on the old behavior and just sending `TrackEmittedStreams=true` to enable emitting when creating a projection, will continue to work after upgrading the server.

However, they will run into a change of behavior (because of the bug fix) when they upgrade the client library, so we have made the server throw in this case to make it more obvious that a code change is necessary.

### DIFF
--- a/src/EventStore.Projections.Core/Services/Grpc/ProjectionManagement.Create.cs
+++ b/src/EventStore.Projections.Core/Services/Grpc/ProjectionManagement.Create.cs
@@ -38,31 +38,32 @@ namespace EventStore.Projections.Core.Services.Grpc {
 				ModeOneofCase.Continuous => options.Continuous.EmitEnabled,
 				_ => false
 			};
-			var trackEmittedStreams = (options.ModeCase, emitEnabled) switch {
-				(ModeOneofCase.Continuous, true) => options.Continuous.TrackEmittedStreams,
+			var trackEmittedStreams = (options.ModeCase, emitEnabled, options.Continuous.TrackEmittedStreams) switch {
+				(ModeOneofCase.Continuous, true, true) => true,
+				(ModeOneofCase.Continuous, false, true) =>
+					throw new InvalidOperationException("EmitEnabled must be set to true to track emitted streams."),
 				_ => false
 			};
-			var checkpointsEnables = options.ModeCase switch {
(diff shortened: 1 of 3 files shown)


src/EventStore.Projections.Core/Services/Grpc/ProjectionManagement.Create.cs
```diff
@@ -40,7 +40,7 @@ internal partial class ProjectionManagement {
 			};
 			var trackEmittedStreams = (options.ModeCase, emitEnabled) switch {
 				(ModeOneofCase.Continuous, true) => options.Continuous.TrackEmittedStreams,
-				_ => false
+				_ => throw new InvalidOperationException("EmitEnabled must be set to true to track emitted streams.")
```
timothycoleman, line 43: i dont think this is quite the one

timothycoleman:
code looks good, for the changelog entry perhaps something like

```
Changed: Reject gRPC call to create a projection if TrackEmittedStreams=true and EmitEnabled=false
Changed: Proto names of emit_enabled and track_emitted_streams swapped for backwards compatibility

This PR swaps the names of EmitEnabled and TrackEmittedStreams in the proto because what the server used to refer to as TrackEmittedStreams is in fact the correct behaviour for EmitEnabled. This makes the recent change to 'Add EmitEnabled' really mean 'AddTrackEmittedStreams'

Therefore clients that were relying on the old behaviour and just sending TrackEmittedStreams=True to enable emitting when creating a projeciton, will continue to work after upgrading the server.

However, they will run into a change of behaviour (because of the bug fix) when they upgrade the client library, so we have made the server throw in this case to make it more obvious that a code change is necessary
```

timothycoleman (approved):

hayley-jean (approved):

PR #3419, opened 2022-02-08
IODispatcher bug fix

### LINKED ISSUE #3374: Change Password API Returns 401
**Describe the bug**
The `/users/admin/command/change-password` endpoint returns a 401 on 21.10.1.

Works as expected on older versions.

**EventStore details**
 - EventStore server version: 21.10.1

**Additional context**
Support ticket: 110099

shaan1337:
Fixed: IODispatcher bug causing password change notifications to be missed sometimes

Reproduction steps are available in the ticket: https://eventstore.freshdesk.com/a/tickets/110099

Fixes #3374

### DIFF
--- a/src/EventStore.Core/Helpers/IODispatcher.cs
+++ b/src/EventStore.Core/Helpers/IODispatcher.cs
@@ -894,7 +894,9 @@ namespace EventStore.Core.Helpers {
 		}
 
 		public void Handle(ClientMessage.NotHandled message) {
-			_requestTracker.RemovePendingRead(message.CorrelationId);
+			// we do not remove the pending read here but only the pending request.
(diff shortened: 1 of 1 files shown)


timothycoleman:
To summarize, @shaan1337 found that the `PasswordChangeNotificationReader` `ReadNotificationsFrom` pump was stalling. If the `ReadStreamEventsForward` results in a `NotHandled` message (e.g. because the cluster is still starting up), the IODispatcher's handling of that message tidied up the request, preventing the timeout handling from being able to find the request so it assumes the request was handled successfully and does not call the timeout callback. This prevents the pump from knowing to issue the next read.

hayley-jean (approved):

PR #3425, opened 2022-02-14
Add fsyncing bloomfilter

timothycoleman:
todo: testing and more testing


src/EventStore.Core/DataStructures/ProbabilisticFilter/PersistentBloomFilters/BloomFilterAccessor.cs
```diff
@@ -0,0 +1,226 @@
...
+			// first cache line is reserved for the header
+			bytePositionInFile += CacheLineSize;
+
+			return bytePositionInFile;
+		}
+
+		// converts page number into byte position of the page in the file
+		public (long PositionInFile, int PageSize) GetPagePositionInFile(int pageNumber) {
+			// first cache line is reserved for the header
+			var positionInFile = CacheLineSize + (pageNumber * PageSize);
```
shaan1337, line 117: @timothycoleman it looks like the multiplication could overflow an int for a 4GB bloom filter

src/EventStore.Core/DataStructures/ProbabilisticFilter/PersistentBloomFilters/BloomFilterAccessor.cs
```diff
@@ -0,0 +1,226 @@
...
+
+			var pageSize = Math.Min(PageSize, (int)(FileSize - positionInFile));
```
shaan1337, line 119: here too, it looks like the cast to `int` may cause an overflow when positionInFile is 0 and FileSize is large.

src/EventStore.Core/DataStructures/ProbabilisticFilter/PersistentBloomFilters/BloomFilterAccessor.cs
```diff
@@ -0,0 +1,226 @@
...
+			}
+		}
+
+		public ref byte ReadByte(long bytePositionInFile) {
+			EnsureBounds(bytePositionInFile, length: 1);
+			return ref *(Pointer + bytePositionInFile);
+		}
+
+		public Span<byte> ReadCacheLineFor(long bytePositionInFile) {
+			bytePositionInFile = bytePositionInFile / CacheLineSize * CacheLineSize;
```
shaan1337, line 145: suggestion: use .RoundDownToMultipleOf(CacheLineSize); for consistency

src/EventStore.Core/DataStructures/ProbabilisticFilter/PersistentBloomFilters/BloomFilterAccessor.cs
```diff
@@ -0,0 +1,226 @@
...
+			var byteValue = ReadByte(bytePositionInFile);
+			return byteValue.IsBitSet(bitPosition % 8);
+		}
+
+		public void SetBit(long bitPosition) {
+			var bytePositionInFile = GetBytePositionInFile(bitPosition);
+			ref var byteValue = ref ReadByte(bytePositionInFile);
+			byte oldByte = byteValue;
+			byteValue = byteValue.SetBit(bitPosition % 8);
+			var cacheLine = ReadCacheLineFor(bytePositionInFile);
```
shaan1337, line 165: suggestion: also read the cache line and recompute the hash only if the byte's value has changed.

src/EventStore.Core/DataStructures/ProbabilisticFilter/PersistentBloomFilters/BloomFilterAccessor.cs
timothycoleman: 🤦‍♂️ thats a jolly good spot

src/EventStore.Core/DataStructures/ProbabilisticFilter/PersistentBloomFilters/BloomFilterAccessor.cs
timothycoleman: yep this one too 👍

src/EventStore.Core/DataStructures/ProbabilisticFilter/PersistentBloomFilters/BloomFilterAccessor.cs
timothycoleman: 👍

src/EventStore.Core/DataStructures/ProbabilisticFilter/PersistentBloomFilters/BloomFilterAccessor.cs
timothycoleman: ah yeah good point

gregoryyoung:
@timothycoleman are you adding an unbuffered stream version of this as well?

If so is there a significant difference in how the caching needs to work?

timothycoleman:
@gregoryyoung not for the minute, but once we're on net6 the new random file access stuff in there could have a bearing i think

gregoryyoung:
There is the current unbufferedstream in our codebase, do you want me to spend a bit tomorrow or Friday to see if I can just quickly work it in as a spike?

timothycoleman:
This is holding up a patch release at the moment so we'll probably take the shortest path to getting that out (i.e. the FileStream at least for the short term) but that said, the UnbufferedFileStream stuff looks interesting and like it might be a pretty easy fit. Feel free to take a look of course, but no promises on when I might get back to it. I'm back on to scavenging after this

src/EventStore.Core/DataStructures/ProbabilisticFilter/FileStreamPersistence.cs
```diff
@@ -0,0 +1,275 @@
...
+						pageNumber++;
+						if (pageNumber == DataAccessor.NumPages)
+							goto Done;
+					}
+				}
+			}
+
+
+			Done:
+			fileStream.FlushToDisk();
```
shaan1337, line 193: may be worth considering flushing in small batches instead of doing one big flush: see https://github.com/EventStore/EventStore/pull/1918

src/EventStore.Core/DataStructures/ProbabilisticFilter/FileStreamPersistence.cs
```diff
@@ -0,0 +1,275 @@
...
+					var cacheLine = remaining[..BloomFilterIntegrity.CacheLineSize];
+					cacheLine.CopyTo(localCacheLine);
+					cacheLine.Clear();
+				}
+
+				foreach (var @byte in localCacheLine) {
+					for (var bitOffset = 0; bitOffset < 8; bitOffset++) {
+						if (@byte.IsBitSet(bitOffset)) {
+							WritePage(pageNumber, fileStream);
+							flushedPages++;
```
shaan1337, line 174: @timothycoleman i may be missing something but the code doesn't seem to unset the dirty bit after writing the page

src/EventStore.Core/DataStructures/ProbabilisticFilter/BloomFilterIntegrity.cs
```diff
@@ -4,12 +4,15 @@
...
 
-namespace EventStore.Core.DataStructures.ProbabilisticFilter.MemoryMappedFileBloomFilter {
+namespace EventStore.Core.DataStructures.ProbabilisticFilter {
 	// This class is responsible for maintaining and validating a 4 byte hash at the end
 	// of the cache line.
 	public class BloomFilterIntegrity {
 		public const int CacheLineSize = 64;
 		public const int HashSize = 4;
+		//qq this is the flush unit, discuss with james whether it should be configurable and what the default should be.
+		public const int PageSize = 8 * 1024;
```
shaan1337, line 14: it could also be looked up at runtime with `sysconf(_SC_PAGESIZE)` on linux and `GetSystemInfo()` on windows with a fallback of the usual default (4096)

shaan1337: if it is configurable, the value 0 could be used for auto-configuration (similar to how it is for other auto-configurable options)

src/EventStore.Core/DataStructures/ProbabilisticFilter/FileStreamPersistence.cs
shaan1337: ohh nevermind, i've now seen that it's cleared after being copied.

src/EventStore.Core/DataStructures/ProbabilisticFilter/FileStreamPersistence.cs
timothycoleman: woops yes i meant to add a call to flushtodisk in the 'pause'

gregoryyoung: @shaan1337 do you mean to flush(false) multiple times then flush(true) at the end?

src/EventStore.Core/DataStructures/ProbabilisticFilter/FileStreamPersistence.cs
```diff
@@ -0,0 +1,292 @@
...
+
+			activelyFlushing.Stop();
+
+			var flushedBytes = flushedPages * DataAccessor.PageSize;
+			var flushedMegaBytes = (float)flushedBytes / 1000 / 1000;
+			var activeFlushRateMBperS = flushedMegaBytes / activelyFlushing.Elapsed.TotalSeconds;
+
+			Log.Information(
+				"Flushed {pages:N0} pages out of {totalPages:N0}. {bytes:N0} bytes. " +
+				"Delay {delay} per batch. Total delay {totalDelay:N0}ms. " +
```
shaan1337, line 208:
```suggestion
				"Delay {delay} ms per batch. Total delay {totalDelay:N0} ms. " +
```

nit: add unit and space

src/EventStore.Core/DataStructures/ProbabilisticFilter/FileStreamPersistence.cs
```diff
@@ -0,0 +1,292 @@
...
+				"Actively flushing: {activeFlushTime} {activeFlushRate:N2}MB/s. ",
```
shaan1337, line 209:
```suggestion
				"Actively flushing: {activeFlushTime} {activeFlushRate:N2} MB/s. ",
```

src/EventStore.Core/DataStructures/ProbabilisticFilter/BloomFilterAccessor.cs
```diff
@@ -0,0 +1,234 @@
...
+			if (pageNumber >= NumPages)
+				throw new ArgumentOutOfRangeException(nameof(pageNumber), pageNumber, "");
+
+			var positionInFile = pageNumber * PageSize;
+
+			var pageSize = Math.Min(PageSize, FileSize - positionInFile);
+
+			if (pageNumber == 0) {
+				pageSize -= 64;
+				positionInFile += 64;
```
shaan1337, line 124:
```suggestion
				pageSize -= CacheLineSize;
				positionInFile += CacheLineSize;
```

src/EventStore.Core/DataStructures/ProbabilisticFilter/AlignedMemory.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using System.Runtime.InteropServices;
+
+namespace EventStore.Core.DataStructures.ProbabilisticFilter {
+	// net6 can do aligned allocations directly
+	// but we need to overallocate and do our own alignment.
+	public sealed unsafe class AlignedMemory : IDisposable {
```
shaan1337, line 7: @timothycoleman does a finalizer need to be added as a precaution to free up the allocated memory in case someone forgets to call Dispose() ?

src/EventStore.Core/DataStructures/ProbabilisticFilter/BloomFilterIntegrity.cs
shaan1337: note: if it is configurable, it should be a multiple of the cache line size (64) and disk block size (4096)

gregoryyoung (commented):
With the BloomFilterAccessor which is using off heap memory has it been considered to use buffers from BufferManager?

This would put all of that memory usage through the same known mechanism as other things are using.

This would have both some pros and cons and is not by any mean a "this is a great idea" type thing, more I am wondering if it was considered.

src/EventStore.Core/DataStructures/ProbabilisticFilter/AlignedMemory.cs
```diff
@@ -0,0 +1,44 @@
...
+		public AlignedMemory(long size, int alignTo) : this(new IntPtr(size), alignTo) {
+		}
+
+		/// contents of memory are not initialized
+		public AlignedMemory(IntPtr size, int alignTo) {
+			_size = (long)size;
+			var bytesToAllocate = size + alignTo;
+			_intPtr = Marshal.AllocHGlobal(bytesToAllocate);
+			_bytesAllocated = (long)bytesToAllocate;
+			GC.AddMemoryPressure(_bytesAllocated);
```
gregoryyoung, line 23: I dont think this is actually needed.

Also why is size being passed as an IntPtr then converted to a long?! This is just ODD!

jageall (dismissed):

src/EventStore.Core/DataStructures/ProbabilisticFilter/AlignedMemory.cs
timothycoleman: Yeah you’re right, I’ll take a look, probable make use of a safe handle

timothycoleman:
@gregoryyoung I hadn’t considered the buffer manager, I’ll take a look to find out more about it, thanks for the pointer 

the intptr/long thing is because allochglobal takes an intptr and addmemorypressure takes a long. Although, I wonder if we really need both constructors. I’ll check when I’m in the office

src/EventStore.Core/DataStructures/ProbabilisticFilter/AlignedMemory.cs
timothycoleman: Actually I think let’s get this in as is if you’re happy and we can circle back round after the release

jageall (approved):

shaan1337 (approved):

PR #3427, opened 2022-02-15
Add support for DNS discovery with a secure cluster having only DNS SANs (including wildcard SANs)

shaan1337:
Added: Support for DNS discovery with a secure cluster having only DNS SANs (including wildcard SANs)

### Summary:
- The PR adds support for DNS discovery with a secure cluster having only DNS SANs, basically:
```
DiscoverViaDns: True
ClusterDns: test.cluster.dns
Insecure: False
+ Certificate containing DNS SAN (test.cluster.dns) or wildcard DNS SAN (*.cluster.dns) but no IP SAN
```
-  This is useful for situations where certificates are generated by a public certificate authority (which cannot easily contain IP SANs).
- There still needs to be a DNS A record pointing towards the IP addresses of each node. We do not yet support full hostname DNS discovery (e.g via DNS SRV records) which would eliminate the need to configure the IP addresses.

### Fix description:
- The fix works by adding flexibility to support validation of multiple names against a certificate (instead of a single one - usually the target host used in a HTTP URI) when connecting to a TLS-enabled endpoint.
- The cluster DNS name is then supplied as an additional name used for validation.
- By default, hostname validation against a certificate will still be done by the .NET libraries but when it fails, we do some additional validation of "extra" hostnames with our own implementation.
- The PR should not cause any breaking changes since it is only additive.

This required implementation of:
- our own hostname validation logic based on [RFC 6125](https://tools.ietf.org/html/rfc6125) since there is currently no such API in .NET 5 - probably because it depends on the system's libraries (e.g openssl on linux) for name validation.
- our own SAN parsing logic from ASN.1 encoded data based on [RFC 5280](https://datatracker.ietf.org/doc/html/rfc5280). The existing code for SAN parsing depends on the ASNEncodedData.Format() function which is not ideal since this method should in principle be used only for pretty-printing.

### Scenarios tested
The following settings were common across all scenarios:
- DiscoverViaDns: True
- ClusterDns: test.esdb-qa.eventstore.com
- DNS A Record for `test.esdb-qa.eventstore.com` which points to 127.0.0.1, 127.0.0.2, 127.0.0.3

### Certificate with DNS SAN only :heavy_check_mark: 
- CN=eventstoredb-node
- DNS SAN = test.esdb-qa.eventstore.com

### Certificate with wildcard DNS SAN only :heavy_check_mark: 
- CN=eventstoredb-node
- DNS SAN = *.esdb-qa.eventstore.com

### Certificate with IP SAN only :heavy_check_mark: 
- CN=eventstoredb-node
- IP SAN = 127.0.0.1 / 127.0.0.2 / 127.0.0.3 respectively

### Certificate with CN and no DNS SAN :x:
- CN=test.esdb-qa.eventstore.com
- DNS SAN = `none`
- CertificateReservedNodeCommonName: `test.esdb-qa.eventstore.com`
- Although it is a valid certificate, this does not work since we currently [require at least one DNS SAN or IP address in the certificate.](https://github.com/EventStore/EventStore/blob/bd962bcddbf524bccb03166dd431f5b59a84c5a2/src/EventStore.Core/Services/Transport/Http/Authentication/ClientCertificateAuthenticationProvider.cs#L53) However this should not be a problem since the use of CNs has been deprecated since a long time ago in RFC 2818: `Although the use of the Common Name is existing practice, it is deprecated and Certification Authorities are encouraged to use the dNSName instead.`

### Certificate with both CN / SAN  :heavy_check_mark: 
- In situations where there are both a CN and SAN, the SAN takes precedence and the CN should be ignored. ([extract from RFC 6125](https://datatracker.ietf.org/doc/html/rfc6125#section-6.4.4))

### Rolling upgrade - setting DiscoverViaDns to True when it was previously False :heavy_check_mark: 
- Rolling upgrades will work but the node being upgraded will still need to have an IP SAN
- Rolling upgrades to a certificate with no IP SAN will require two steps (suggested by @hayley-jean):
  - First do a rolling upgrade of each node to a certificate with `<IP SAN, Cluster DNS SAN>`
  - Do a second rolling upgrade of each node to a certificate having only the `<Cluster DNS SAN>`

### DIFF
--- a/src/EventStore.Common/Utils/CertificateExtensions.cs
+++ b/src/EventStore.Common/Utils/CertificateExtensions.cs
@@ -1,0 +1,141 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Formats.Asn1;
+using System.Globalization;
+using System.Linq;
+using System.Net;
+using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
+
+namespace EventStore.Common.Utils {
+	public static class CertificateExtensions {
+		public static IEnumerable<(string name, string type)> GetSubjectAlternativeNames(this X509Certificate certificate) {
+			// Implemented based on RFC 5280 (https://datatracker.ietf.org/doc/html/rfc5280)
+			// - Reads IP addresses and DNS names from the Subject Alternative Names extension
+			// - Does not support other name types yet
+
+			if (certificate is not X509Certificate2 certificate2) {
+				using var c2 = new X509Certificate2(certificate);
+				certificate2 = c2;
+			}
+
+			X509ExtensionCollection extensions;
+			try {
+				extensions = certificate2.Extensions;
+			} catch (CryptographicException) {
+				return null;
+			}
+
+			var sans = new List<(string, string)>();
+			foreach (var extension in extensions) {
+				if (extension.Oid?.Value != "2.5.29.17") continue; // Oid for Subject Alternative Names extension
+				var asnReader = new AsnReader(extension.RawData, AsnEncodingRules.DER).ReadSequence();
+				while (asnReader.HasData) {
+					Asn1Tag tag;
+					try {
+						tag = asnReader.PeekTag();
+					} catch (AsnContentException) {
+						break;
+					}
+
+					switch (tag.TagValue) {
+						case 2: // dNSName [2] IA5String
+							sans.Add((asnReader.ReadCharacterString(UniversalTagNumber.IA5String, tag), CertificateNameType.DnsName));
+							break;
+						case 7: // iPAddress [7] OCTET STRING
+							sans.Add((new IPAddress(asnReader.ReadOctetString(tag)).ToString(), CertificateNameType.IpAddress));
+							break;
+						default:
+							asnReader.ReadEncodedValue();
+							break;
+					}
+				}
+			}
+
+			return sans;
+		}
+
+		public static bool MatchesName(this X509Certificate certificate, string name) {
+			// Implemented based on RFC 6125 (https://datatracker.ietf.org/doc/html/rfc6125) with the following changes:
+			// - Does not support SRV-ID and URI-ID identifier types yet
+			// - Partial wildcard support is not implemented since it has been deprecated in most major browsers
+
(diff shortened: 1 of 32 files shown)


thefringeninja (approved):

thefringeninja (approved):

src/EventStore.Common/Utils/CertificateExtensions.cs
```diff
@@ -0,0 +1,141 @@
...
+
+			if (string.IsNullOrEmpty(certName) ||
+			    string.IsNullOrEmpty(name))
+				return false;
+
+			// if at least one of the names is an IP address, do an exact match
+			if (certNameType == CertificateNameType.IpAddress ||
+			    IPAddress.TryParse(certName, out _) ||
+			    IPAddress.TryParse(name, out _))
+				return name.EqualsOrdinalIgnoreCase(certName);
```
hayley-jean, line 107: We document that it is a requirement that if an IP address is supplied in the certificate then it has to match the machine's IP address regardless of the DNS name provided in the SAN.

hayley-jean (approved):
Tested this with an upgrade between old and new versions. As long as the ip address is still provided in the certificate, the nodes will be able to communicate with nodes running an older version.

src/EventStore.Common/Utils/CertificateExtensions.cs
shaan1337: @hayley-jean i didn't follow - actually, we do check for all SANs here:

https://github.com/EventStore/EventStore/blob/803c2b55ce8834918aaaafff090099233f75555a/src/EventStore.Common/Utils/CertificateExtensions.cs#L71

shaan1337: but if either the san or the target host is an ip address then we do an exact check (since a hostname cannot contain only digits according to RFCs 952 & 1123)

hayley-jean: Thanks, I didn't notice that. Then yeah, all we would need to do is add the fact that you can have wildcard certs with dns discovery to the documentation and leave it at that.

PR #3435, opened 2022-02-18
Fix projections subsystem getting stuck in a stopping state when reads timeout

hayley-jean:
Fixed: Prevent projections subsystem from getting stuck in a stopping state due to read timeouts
Changed: Reads operations in IODispatcher are no longer tracked if they don't have a timeoutAction
Added: Logging around read timeouts in projection emitted streams and emitted stream trackers

Fixes https://github.com/EventStore/EventStore/issues/3431

- Use the timeout functionality from the `IODispatcher` when recovering a projection from a checkpoint so that the reads can be properly tracked.
- Add `timeoutAction` to `ReadEvent` in the `IODispatcher`.
- Log an error when reads time out while reading linked events in emitted streams.
- Log an error when reads time out while emitted streams into the cache of the emitted stream tracker on timeout.

### DIFF
--- a/src/EventStore.Core/Helpers/IODispatcher.cs
+++ b/src/EventStore.Core/Helpers/IODispatcher.cs
@@ -390,20 +390,31 @@ namespace EventStore.Core.Helpers {
 			long fromEventNumber,
 			ClaimsPrincipal principal,
 			Action<ClientMessage.ReadEventCompleted> action,
-			Guid? corrId = null) {
-			corrId ??= Guid.NewGuid();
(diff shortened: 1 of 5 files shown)


timothycoleman:
I think the only potential risk here is that the subsystem will now shutdown while requests without timeouts are in flight, so as long as that isn't a problem (?) then this is a nice way to solve the getting stuck

thefringeninja (approved):

timothycoleman (approved):

PR #3446, opened 2022-03-04
[Cherry-pick] Send full certificate chain (excluding root) from client and server side

hayley-jean:
Fixed: Send full certificate chain from both server and client side during TLS handshake (requires manually adding intermediate certificates to the store)

Cherry-picks https://github.com/EventStore/EventStore/pull/3398 to release 21.10


thefringeninja (approved):

shaan1337 (approved):
Basic test done with 3 node cluster

PR #3453, opened 2022-03-13
Request manager optimizations -wip

condron:
pr to run the tests

### DIFF
--- a/src/EventStore.Core/Services/RequestManager/CommitSource.cs
+++ b/src/EventStore.Core/Services/RequestManager/CommitSource.cs
@@ -1,8 +1,12 @@
 ﻿using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading;
+using System.Threading.Channels;
+using System.Threading.Tasks;
+using EventStore.Common.Utils;
 using EventStore.Core.Bus;
 using EventStore.Core.Messages;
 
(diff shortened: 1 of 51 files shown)


src/EventStore.Core/Services/RequestManager/CommitSource.cs
```diff
@@ -1,72 +1,34 @@
...
-			if (!notificationDictionary.TryGetValue(position, out var actionList)) {
-				actionList = new List<Action> { target };
-				notificationDictionary.TryAdd(position, actionList);
-			} else {
-				lock (actionList) {
-					actionList.Add(target);
-				}
-			}
+		public Task WaitForReplication(long position, CancellationToken token) {
+			return new Task(async () => {
```
gfoidl, line 23: The returned Task is in the "created" state and won't never complete, i.e. `await` that task and you have a deadlock.

Why not
```c#
return _replicatedTracker.Waitfor(position);
```
?
The CancellationToken isn't passed in, but the same problem -- amongst the deadlock -- has the current state of PR too.
Ideally the CancellationToken should be passed down to `_replicatedTracker.Waitfor`.

src/EventStore.Core/Services/RequestManager/LogNotificationService.cs
```diff
@@ -0,0 +1,110 @@
...
+			_name = name;
+		}
+		public void UpdateLogPosition(long position) {
+			if (Interlocked.Read(ref _nextLogPosition) <= position) {
+				Interlocked.Exchange(ref _nextLogPosition, position);
+				_positionUpdated.Set();
+			}
+		}
+		private void Notify() {
+			while (!_canceled.IsCancellationRequested) {
```
gfoidl, line 46: I'm not familiar with what this code should do*, but it seems this is quite heavy on sync-primitives.
Can something like [Channels](https://devblogs.microsoft.com/dotnet/an-introduction-to-system-threading-channels/) be used to do the same thing but leass heavy?

<sub>\* just started to dig into the code base a little bit</sub>

timothycoleman:
spring cleaning: closing this for now

PR #3454, opened 2022-03-15
Add Duration Option to Batch Append Deadline

thefringeninja:
21.10 introduced a new feature: Batch Appends, which allows the gRPC client to open a single bi-directional stream to the server and multiplex append operations through it. As the call must remain open indefinitely, leveraging gRPC deadlines will not work. Instead we added a `Timestamp` option to `BatchAppendReq`.

This type is the equivalent of `DateTime` in dotnet, and operations using it will fail if the clock skew between client and server is greater than the deadline. Instead, we add a new `Duration` option (equivalent to `TimeSpan`) to set the deadline.

Client SDKs should only send `Duration` if Server Capabilities reports that the `batchappend` method feature contains `deadlineduration`.


src/EventStore.Core/Services/Transport/Grpc/ServerFeatures.cs
```diff
@@ -26,19 +29,22 @@ internal partial class ServerFeatures
 						MethodName = x.Method.Name.ToLower(),
 						ServiceName = x.Method.ServiceName.ToLower(),
 					};
+
 					if (x.Method.ServiceName.Contains("PersistentSubscriptions")) {
 						method.Features.AddRange(new[] {"stream", "all"});
 					} else if (x.Method.ServiceName.Contains("Streams") && x.Method.Name.Contains("Read")) {
 						method.Features.AddRange(new[] {"position", "events"});
+					} else if (x.Method.ServiceName.Contains("Streams") && x.Method.Name.Contains("BatchAppend")) {
+						method.Features.Add("deadlineduration");
```
timothycoleman, line 38: should this have some kind of separation between the words (casing, or hyphen), or have we already got other ones where the words are just concatenated

src/EventStore.Core/Services/Transport/Grpc/ServerFeatures.cs
```diff
@@ -5,19 +5,22 @@
...
 using Grpc.Core;
 using Microsoft.AspNetCore.Routing;
-
+using VersionInfo = EventStore.Common.Utils.VersionInfo;
 namespace EventStore.Core.Services.Transport.Grpc {
 	internal partial class ServerFeatures
 		: EventStore.Client.ServerFeatures.ServerFeatures.ServerFeaturesBase {
 		public const int ApiVersion = 1;
-		private readonly EndpointDataSource _endpointDataSource;
+		private readonly Task<SupportedMethods> _supportedMethods;
```
timothycoleman, line 13: don't think there is a need to cache it as a Task, directly as `SupportedMethods` will do? then the `GetSupportedMethods` that the constructor calls can just return `SupportedMethods`

proto.lock
```diff
@@ -5259,8 +5259,13 @@
                   },
                   {
                     "id": 6,
-                    "name": "deadline",
+                    "name": "deadline_21_10_0",
```
timothycoleman, line 5262: not that important, but consider `deadline_as_date_time` since the generated name `Deadline21100` is a bit weird

src/EventStore.Core/Services/Transport/Grpc/ServerFeatures.cs
thefringeninja: I think underscore is the way to go here

src/EventStore.Core/Services/Transport/Grpc/ServerFeatures.cs
timothycoleman: retracted. joao pointed out this would result in allocating a task (from result) when the grpc method is called, so there is an advantage to caching the task

timothycoleman (approved):

timothycoleman:
looks good :+1:

YoEight (approved):

PR #3458, opened 2022-03-25
Retry establishing a TCP connection to the leader

hayley-jean:
Fixed: Attempt to reconnect to the leader every second if the node fails to establish a connection (for example due to DNS lookup timeout).

Fixes https://github.com/EventStore/home/issues/723

To reproduce the original issue, edit the code to throw an exception in [ResolveDnsToIpAddress](https://github.com/EventStore/EventStore/blob/master/src/EventStore.Common/Utils/EndpointExtensions.cs#L37) and then try and have the node subscribe to the leader. The follower will fail to subscribe and will get stuck in a state where it does not replicate from the leader.

- Catch exceptions when calling `ConnectToLeader` in `ReplicaService` and publish a `ReplicaMessage.LeaderConnectionFailed` message.
- Handle `LeaderConnectionFailed` in the `VNodeController` by attempting to `ReconnectToLeader`.
- Create a separate `ConnectionCorrelationId`, which covers establishing the connection to the leader. This is separate to the `StateCorrelationId` because subscriptions need to be retried independently of connections.
- Only attempt to create a subscription if the connection is not null.

### DIFF
--- a/src/EventStore.Core/Messages/ReplicationMessage.cs
+++ b/src/EventStore.Core/Messages/ReplicationMessage.cs
@@ -139,12 +139,30 @@ namespace EventStore.Core.Messages {
 			}
 
 			public readonly MemberInfo Leader;
-			public readonly Guid StateCorrelationId;
+			public readonly Guid ConnectionCorrelationId;
 
-			public ReconnectToLeader(Guid stateCorrelationId, MemberInfo leader) {
-				Ensure.NotEmptyGuid(stateCorrelationId, "stateCorrelationId");
-				Ensure.NotNull(leader, "leader");
-				StateCorrelationId = stateCorrelationId;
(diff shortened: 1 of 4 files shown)


hayley-jean:
>I wonder if there is a better name than _connectionCorrelationId in the controller. maybe _leaderConnectionCorrelationId, at risk of being pretty long?

I prefer the longer name actually. There's not a whole lot of references to this so it shouldn't be too cumbersome

> Not sure if there might be a good place to hook in a test or not

Sure, I'll see if I can find a place to do a test for this

> Which versions should we backport to

I think it could be backported to both 20.10 and 21.10. I can't think of anything off the top of my head that'll prevent the backport to 20.10

YoEight (approved):

timothycoleman (approved):
looks good :+1:

PR #3474, opened 2022-04-21
Want to make reads that are after the LastEventNumber efficient

timothycoleman:
Added: Fast path for forward reads that start after the LastEventNumber

Use case: Some clients want to be able to read a stream up to latest from a follower (say, receive events 0-100) but then do an additional read on the leader starting with event 101 in case the follower is stale.

This PR introduces a fast path for when a read starts after the LastEventNumber. We get the LastEventNumber (which may be cached anyway) and then there is no further need to query the index.

On my machine with (a fairly contrived) 100 ptables, this PR makes the ReadIndex access in this scenario 6000x faster

Note this only applies to reading streams forward. If reading backwards it is not clear that such an optimisation could apply since the maxcount would apply from the last actual event downwards, meaning we'd be returning events anyway.


hayley-jean (approved):

pvanbuijtene (approved):

pvanbuijtene:
Merge conflict during cherry pick resolved in https://github.com/EventStore/EventStore/pull/3480

PR #3476, opened 2022-04-21
Fix overflow when scavenge has saved a lot of space

hayley-jean:
Fixed: overflow when trying to complete a scavenge after the scavenge has saved a lot of space.

Fixes https://github.com/EventStore/home/issues/727

### DIFF
--- a/src/EventStore.Core.Tests/Services/Storage/Scavenge/ScavengeLogManager/ScavengeLogHelper.cs
+++ b/src/EventStore.Core.Tests/Services/Storage/Scavenge/ScavengeLogManager/ScavengeLogHelper.cs
@@ -1,0 +1,75 @@
+using System;
+using System.Collections.Generic;
+using EventStore.Core.Data;
+using EventStore.Core.Services;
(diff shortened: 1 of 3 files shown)


StevenBlair123:
Beat me to it.
I have a branch here with almost the same change (I am just testing it on a real scavenge on production just now)

One little though:

You have this:

`incompleteScavengeStats.MaxChunkScavenged = Math.Max(incompleteScavengeStats.MaxChunkScavenged, (int)(long)chunkEndNumberEntry);`

And I did it this way (MaxChunkScavenged  is now a long):

`incompleteScavengeStats.MaxChunkScavenged = Math.Max(incompleteScavengeStats.MaxChunkScavenged, (long)chunkEndNumberEntry);`

timothycoleman (approved):
Looks good. Seems the default behaviour of newtonsoft when deserialising any number to an object is to use int64 (or BigInteger if necessary). I guess that means it didn't even have to overflow the int32 to fail to cast

avish0694 (approved):
The changes look good to me.

PR #3500, opened 2022-06-08
Ensure no pending writes can be incorrectly acked or published when going offline for truncation. 

### LINKED ISSUE #3472: Writes incorrectly acked during leader election
If a leader has pending writes that have not yet been replicated, loses quorum, re-connects to the cluster, finds that a new leader has been elected, and finds that the replicated log position has been advanced by the new leader to beyond the pending writes, and those pending writes have not expired - then the old leader will (incorrectly) ack those writes back to the client before (correctly) going offline for truncation of those same writes.

Possibly these writes could appear in a subscription just before the node goes offline, too

jageall:
Fixed: Ensure no pending writes can be incorrectly acked or published when going offline for truncation. 

This is sent back to the client as a WriteTimeout or CommitTimeout which indicates that the client should retry the request with the exact same values.

Fixes #3472

### DIFF
--- a/src/EventStore.Core/Services/RequestManager/RequestManagementService.cs
+++ b/src/EventStore.Core/Services/RequestManager/RequestManagementService.cs
@@ -167,7 +167,15 @@ namespace EventStore.Core.Services.RequestManager {
 		
 		
 		public void Handle(SystemMessage.StateChangeMessage message) {
-			//TODO(clc): if we have become resigning leader should all requests be actively disposed?
+			
+			if (_nodeState == VNodeState.Leader && message.State is not VNodeState.Leader or VNodeState.ResigningLeader) {
+				var keys = _currentRequests.Keys;
+				foreach (var key in keys) {
+					if (_currentRequests.Remove(key, out var manager)) {
+						manager.Dispose();
+					}
+				}
+			}
 			_nodeState = message.State;
 		}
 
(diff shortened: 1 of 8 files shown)


hayley-jean (dismissed):

hayley-jean (approved):
Code looks good, and while I've not reproduced the original issue, I can confirm that the `ReplicationTrackingService` now shuts down when an offline truncation is about to take place

src/EventStore.Core/Services/RequestManager/RequestManagementService.cs
```diff
@@ -167,7 +167,15 @@ public class RequestManagementService :
 		
 		
 		public void Handle(SystemMessage.StateChangeMessage message) {
-			//TODO(clc): if we have become resigning leader should all requests be actively disposed?
+			
+			if (_nodeState == VNodeState.Leader && message.State is not VNodeState.Leader or VNodeState.ResigningLeader) {
```
timothycoleman, line 171: do we still want the change in this file? the node is just about to go offline for truncation so this might just be extra noise now

timothycoleman (changes requested):
looks good, just question

timothycoleman (approved):
:+1:

timothycoleman:
Able to reproduce the bug and confirmed that this fixes it.

I set up a three node docker cluster, with three writers, each talking to one node. Each writer writes an event and attempts to read it back in a loop. Used https://github.com/worstcase/blockade to partition the network such that the leader and its writer are separated in their own partition (irl this could correspond to an availability zone), and upon repairing the partition the bug was observed.

Had to artificially widen the window of failure by setting long timeouts (30s) and having the cluster storage writer service sleep for 5s after detecting trunction is required, to give the replication path time to reliably execute. in both cases smaller values would probably have sufficed

PR #3532, opened 2022-07-22
Prevent gRPC subscriptions from hanging while catching up

timothycoleman:
Fixed: Prevent gRPC subscriptions from hanging while catching up

Stream subscriptions, $all subscriptions, and filtered $all subscriptions issue reads while they are catching up. Before this fix, if any of those reads were discarded because they had expired (i.e. were already 10s old by the time they reached the front of the queue), no further read would be attempted and the subscription would stall.

This change causes a reply to be sent back with status Expired if the read is being discarded due to expiry, and the subscription will retry the read.

The gRPC subscription explicitly opts in to this reply behaviour. For everything else no reply is sent back to make load shedding as quick as possible. Other callers manage their own retries.


jageall (approved):

pvanbuijtene (changes requested):
Finalizing 21.10.6, done.

thefringeninja (approved):

Issue #3512, opened 2022-07-04
[Persistent Subscription] Deletion PS did not handled properly


dhingrak:
**Describe the bug**

The issue happened if the Persistent subscription configuration fails to update during the deletion or update process. In this case, the config for deleted PS would still be there and the user will be able to create duplicate persistent subscriptions. 

For more details on this issue, refer to ticket https://eventstore.freshdesk.com/a/tickets/110355

This issue happened at the following line of code for the above ticket: 
https://github.com/EventStore/EventStore/blob/5837eee3cd2199c891c93349163456ce2d567b7d/src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs#L699

https://github.com/EventStore/EventStore/blob/feaaae634ad661f8f19221c569e669d73ea6e831/src/EventStore.Core/Services/PersistentSubscription/ConsumerStrategy/PinnedState/PinnedConsumerState.cs#L48

We are not handling exceptions thrown by the subscriber.shudown process

@mt":"Error while processing message {message} in queued handler '{queue}'.","@l":"Error","@i":1948184534,"@x":"System.InvalidOperationException: Operation is not valid due to the current state of the object.\n   at EventStore.Core.Services.PersistentSubscription.ConsumerStrategy.PinnedState.PinnedConsumerState.DisconnectNode(Guid nodeId) in /home/runner/work/TrainStation/TrainStation/build/oss-eventstore/src/EventStore.Core/Services/PersistentSubscription/ConsumerStrategy/PinnedState/PinnedConsumerState.cs:line 48\n   at EventStore.Core.Services.PersistentSubscription.PersistentSubscriptionClientCollection.ShutdownAll() in /home/runner/work/TrainStation/TrainStation/build/oss-eventstore/src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionClientCollection.cs:line 37\n   at EventStore.Core.Services.PersistentSubscription.PersistentSubscription.Shutdown() in /home/runner/work/TrainStation/TrainStation/build/oss-eventstore/src/EventStore.Core/Services/PersistentSubscription/PersistentSubscription.cs:line 344\n

**Expected behavior**
User not able create duplicate Persistent Subscription

**Actual behavior**
User able to create duplicate Persistent Subscription

PR #3533, opened 2022-07-22
InvalidOperationException thrown while removing the subscribers in delete persistent subscription process. Modify the update persistent subscription strategy

dhingrak:
Fixed:  #3512  duplicate PS in the EventStoreDB UI

### DIFF
--- a/src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
+++ b/src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
@@ -218,13 +218,9 @@ namespace EventStore.Core.Services.PersistentSubscription {
 				string user
 		) {
 			if (!_started) return;
-			var key = BuildSubscriptionGroupKey(eventSource.ToString(), groupName);
+			var stream = eventSource.ToString();
+			var key = BuildSubscriptionGroupKey(stream, groupName);
 			Log.Debug("Creating persistent subscription {subscriptionKey}", key);
-
-			if (_subscriptionsById.ContainsKey(key)) {
-				onExists($"Group '{groupName}' already exists.");
-				return;
-			}
 
 			if (!_consumerStrategyRegistry.ValidateStrategy(namedConsumerStrategy)) {
 				onFail($"Consumer strategy {namedConsumerStrategy} does not exist.");
@@ -235,8 +231,8 @@ namespace EventStore.Core.Services.PersistentSubscription {
 				onFail(startFromValidationError);
 				return;
 			}
-
-			CreateSubscriptionGroup(eventSource,
+			
+			var result = TryCreateSubscriptionGroup(eventSource,
 				groupName,
 				resolveLinkTos,
 				startFrom,
@@ -253,11 +249,14 @@ namespace EventStore.Core.Services.PersistentSubscription {
 				ToMessageTimeout(messageTimeoutMilliseconds)
 			);
 
+			if (!result) {
+				onExists($"Group '{groupName}' already exists.");
+				return;
+			}
+
 			Log.Debug("New persistent subscription {subscriptionKey}", key);
-			_config.Updated = DateTime.Now;
-			_config.UpdatedBy = user;
-			_config.Entries.Add(new PersistentSubscriptionEntry {
-				Stream = eventSource.ToString(), //'Stream' name kept for backward compatibility
+			var createEntry = new PersistentSubscriptionEntry {
+				Stream = stream, //'Stream' name kept for backward compatibility
 				Filter = EventFilter.ParseToDto(eventSource.EventFilter),
 				Group = groupName,
 				ResolveLinkTos = resolveLinkTos,
@@ -276,7 +275,8 @@ namespace EventStore.Core.Services.PersistentSubscription {
 				StartFrom = startFrom is PersistentSubscriptionSingleStreamPosition x ? x.StreamEventNumber : long.MinValue,
 				#pragma warning restore 612
 				StartPosition = startFrom.ToString()
-			});
+			};
+			UpdateSubscriptionConfig(user, stream, groupName, createEntry);
 			SaveConfiguration(() => onSuccess(""));
 		}
 
@@ -403,7 +403,8 @@ namespace EventStore.Core.Services.PersistentSubscription {
 				string user
 		) {
 			if (!_started) return;
-			var key = BuildSubscriptionGroupKey(eventSource.ToString(), groupName);
+			var stream = eventSource.ToString();
+			var key = BuildSubscriptionGroupKey(stream, groupName);
 			Log.Debug("Updating persistent subscription {subscriptionKey}", key);
 
 			if (!_subscriptionsById.ContainsKey(key)) {
@@ -420,31 +421,32 @@ namespace EventStore.Core.Services.PersistentSubscription {
 				onFail(startFromValidationError);
 				return;
 			}
+			
+			var subscription = new PersistentSubscription(
+				new PersistentSubscriptionParams(
+					resolveLinkTos,
+					key,
+					eventSource,
+					groupName,
+					startFrom,
+					recordStatistics,
+					ToMessageTimeout(messageTimeoutMilliseconds),
+					maxRetryCount,
+					liveBufferSize,
+					bufferSize,
+					readBatchSize,
+					ToCheckPointAfterTimeout(checkPointAfterMilliseconds),
+					minCheckPointCount,
+					maxCheckPointCount,
+					maxSubscriberCount,
+					_consumerStrategyRegistry.GetInstance(namedConsumerStrategy, key),
+					_streamReader,
+					_checkpointReader,
+					new PersistentSubscriptionCheckpointWriter(key, _ioDispatcher),
+					new PersistentSubscriptionMessageParker(key, _ioDispatcher)));
 
-			RemoveSubscription(eventSource.ToString(), groupName);
-			RemoveSubscriptionConfig(user, eventSource.ToString(), groupName);
-
-			CreateSubscriptionGroup(eventSource,
-				groupName,
-				resolveLinkTos,
-				startFrom,
-				recordStatistics,
-				maxRetryCount,
-				liveBufferSize,
-				bufferSize,
-				readBatchSize,
-				ToCheckPointAfterTimeout(checkPointAfterMilliseconds),
-				minCheckPointCount,
-				maxCheckPointCount,
-				maxSubscriberCount,
-				namedConsumerStrategy,
-				ToMessageTimeout(messageTimeoutMilliseconds)
-			);
-
-			_config.Updated = DateTime.Now;
-			_config.UpdatedBy = user;
-			_config.Entries.Add(new PersistentSubscriptionEntry {
-				Stream = eventSource.ToString(), //'Stream' name kept for backward compatibility
+			var updateEntry = new PersistentSubscriptionEntry {
+				Stream = stream, //'Stream' name kept for backward compatibility
 				Group = groupName,
 				Filter = EventFilter.ParseToDto(eventSource.EventFilter),
 				ResolveLinkTos = resolveLinkTos,
@@ -460,10 +462,15 @@ namespace EventStore.Core.Services.PersistentSubscription {
 				MessageTimeout = messageTimeoutMilliseconds,
 				NamedConsumerStrategy = namedConsumerStrategy,
 				#pragma warning disable 612
-				StartFrom = startFrom is PersistentSubscriptionSingleStreamPosition x ? x.StreamEventNumber : long.MinValue,
+				StartFrom = startFrom is PersistentSubscriptionSingleStreamPosition x
+					? x.StreamEventNumber
+					: long.MinValue,
 				#pragma warning restore 612
 				StartPosition = startFrom.ToString()
-			});
+			};
+			
+			UpdateSubscription(stream, groupName, subscription);
+			UpdateSubscriptionConfig(user, stream, groupName, updateEntry);
 			SaveConfiguration(() => onSuccess(""));
 		}
 
@@ -566,9 +573,8 @@ namespace EventStore.Core.Services.PersistentSubscription {
 				message.User?.Identity?.Name
 			);
 		}
-
-
-		private void CreateSubscriptionGroup(IPersistentSubscriptionEventSource eventSource,
+		
+		private bool TryCreateSubscriptionGroup(IPersistentSubscriptionEventSource eventSource,
 			string groupName,
 			bool resolveLinkTos,
 			IPersistentSubscriptionStreamPosition startFrom,
@@ -583,13 +589,12 @@ namespace EventStore.Core.Services.PersistentSubscription {
 			int maxSubscriberCount,
 			string namedConsumerStrategy,
 			TimeSpan messageTimeout) {
-			var key = BuildSubscriptionGroupKey(eventSource.ToString(), groupName);
-			List<PersistentSubscription> subscribers;
-			if (!_subscriptionTopics.TryGetValue(eventSource.ToString(), out subscribers)) {
-				subscribers = new List<PersistentSubscription>();
-				_subscriptionTopics.Add(eventSource.ToString(), subscribers);
-			}
+			var stream = eventSource.ToString();
+			var key = BuildSubscriptionGroupKey(stream, groupName);
 
+			if (_subscriptionsById.ContainsKey(key))
+				return false;
+			
 			var subscription = new PersistentSubscription(
 				new PersistentSubscriptionParams(
 					resolveLinkTos,
@@ -612,8 +617,10 @@ namespace EventStore.Core.Services.PersistentSubscription {
 					_checkpointReader,
 					new PersistentSubscriptionCheckpointWriter(key, _ioDispatcher),
 					new PersistentSubscriptionMessageParker(key, _ioDispatcher)));
-			_subscriptionsById[key] = subscription;
-			subscribers.Add(subscription);
+			
+			UpdateSubscription(stream, groupName, subscription);
+			return true;
+
 		}
 
 		private void DeletePersistentSubscription(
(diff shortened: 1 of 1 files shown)


shaan1337:
hey @dhingrak,

just wanting to make sure i understand the problem and solution correctly:

- my (rough) understanding after reading #3512 , is that the root cause is that saving the configuration to the config stream is done asynchronously and this opens up the possibility for a persistent subscription to be deleted in the _subscription* dictionaries but not yet in the config stream - so creating a new one could result in a duplicate. however, what's not clear to me, is that even if the above scenario occurred, _config should not be able to contain two subscriptions with the same name because the messages handled in the persistent subscription service are handled one at a time (see comment below about concurrency in solution section).

- to reproduce the issue, how did you simulate step 2 in #3512 ?
 
- we also have the issue where deleting/cleaning up the checkpoint and parked messages streams are done asynchronously and it may never actually happen if there is a timeout (but that's another problem)

solution:
- based on the code changes, it seems that you're relying on a lock to prevent a concurrent update (if i'm understanding it correctly). however a lock shouldn't be required since all the messages handled by the persistent subscriptions service are serialized in a queue and executed one at a time.

dhingrak:
@shaan1337 
Thanks a lot Shaan for your comments. Actually, I forgot to update the problem statement for https://github.com/EventStore/EventStore/issues/3512. 
After more investigation and discussion with Hayley we figured out it could be the create subscription part that are not handling the simultaneously request from clients.

I have tested the below scenario in my local environment. To reproduce this issue I have added a delay of 500 ms in createSubscriptionGroup section code and am able to create duplicate subscriptions. By adding a delay the check for existing persistent subscriptions passes when there are simultaneous requests and we have a duplicate subscription in UI. 

I have added a lock on the _subscriptionsById so the two clients are not able to update the state of _subscriptiionsById simultaneously to avoid the duplicate PS creation. 

**Update persistent subscription changes:** 

I have modify the code as suggested by Hayley so the update PS process completes in a single step rather than deleting the PS and then create a new PS

shaan1337:
@dhingrak thanks for all the clarifications :+1: i'll give it a try

shaan1337:
hey @dhingrak, i've checked the repro code you've posted in https://github.com/EventStore/EventStore/issues/3512

by adding `Task.Delay(...).ContinueWith...`, it is possible that the continuation will be run on another thread, which would cause the PersistentSubscriptionService to longer be thread-safe and could subsequently allow creation of duplicate persistent subscriptions.

would you be able to try with `Thread.Sleep(1000);` instead?

dhingrak:
hey @shaan1337 

I have tested the scenario with Thread.Sleep(1000) and I am not able to create duplicate PS. Indeed, the `Task.Delay(...).ContinueWith..` running on another thread in my local testing and that's why there are duplicate PS

dhingrak:
The above PR fixes the issue for InvalidOperationException thrown during the subscriber.shutdown process. Also, as suggested by Hayley we have modify the logic for updating the persistent subscriptions. Currently, the update process consists of delete and then recreate the persistent subscription. This PR will update the persistent subscription without deleting and recreating it.

src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
```diff
@@ -399,18 +402,19 @@ string user
...
 		) {
 			if (!_started) return;
-			var key = BuildSubscriptionGroupKey(eventSource.ToString(), groupName);
+			var stream = eventSource.ToString();
+			var key = BuildSubscriptionGroupKey(stream, groupName);
 			Log.Debug("Updating persistent subscription {subscriptionKey}", key);
 
 			if (!_subscriptionsById.ContainsKey(key)) {
-				onNotExist($"Group '{groupName}' does not exist.");
+				onNotExist($"Group '{groupName}' does not exists.");
```
shaan1337, line 414:
```suggestion
				onNotExist($"Group '{groupName}' does not exist.");
```

src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
```diff
@@ -626,22 +675,45 @@ string user
...
 			Log.Debug("Deleting persistent subscription {subscriptionKey}", key);
 
 			PersistentSubscription subscription;
+			
+			List<PersistentSubscription> subscribers;
+
+
 			if (!_subscriptionsById.TryGetValue(key, out subscription)) {
-				onNotExist($"Group '{groupName}' does not exist.");
+				onNotExist($"Group '{groupName}' does not exists.");
```
shaan1337, line 688:
```suggestion
				onNotExist($"Group '{groupName}' does not exist.");
```

src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
```diff
@@ -399,18 +402,19 @@ string user
...
+			var key = BuildSubscriptionGroupKey(stream, groupName);
 			Log.Debug("Updating persistent subscription {subscriptionKey}", key);
 
 			if (!_subscriptionsById.ContainsKey(key)) {
-				onNotExist($"Group '{groupName}' does not exist.");
+				onNotExist($"Group '{groupName}' does not exists.");
 				return;
 			}
-
+
```
shaan1337, line 417:
```suggestion
```

src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
```diff
@@ -612,8 +647,22 @@ string user
 					_checkpointReader,
 					new PersistentSubscriptionCheckpointWriter(key, _ioDispatcher),
 					new PersistentSubscriptionMessageParker(key, _ioDispatcher)));
+			
+			if (_subscriptionsById.ContainsKey(key)) {
```
shaan1337, line 651: suggestion: move this check before creating the PersistentSubscription object, otherwise it's not used.
it should also be ok to remove the Log.Information line since we don't currently log these errors:

```
			if (_subscriptionsById.ContainsKey(key))
				return false;
```

src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
```diff
@@ -612,8 +647,22 @@ string user
...
+				Log.Information("GroupName already exists: " + groupName);
+				return false;
+			}
+
 			_subscriptionsById[key] = subscription;
+
```
shaan1337, line 657: suggestion:
move `List<PersistentSubscription> subscribers;` down here

src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
```diff
@@ -420,31 +424,55 @@ string user
...
 
-			_config.Updated = DateTime.Now;
-			_config.UpdatedBy = user;
-			_config.Entries.Add(new PersistentSubscriptionEntry {
-				Stream = eventSource.ToString(), //'Stream' name kept for backward compatibility
+						subscribers[i] = subscription;
+						break;
+					}
+				}
+			} else {
```
shaan1337, line 468: it seems that `subscription` is not being added to the `subscribers` list in the `else` part

actually, should we throw an exception in the `else` part? since the subscription should exist in `_subscriptionTopics` at this point.

src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
dhingrak: Done

src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
dhingrak: Done

src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
dhingrak: Made the changes

src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
dhingrak: Done

src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
```diff
@@ -719,29 +791,6 @@ string user
 			);
 		}
 
-		private void RemoveSubscriptionConfig(string username, string eventSource, string groupName) {
```
shaan1337, line 722: suggestion:
to avoid code repetition in the update & remove methods, could we keep these two methods and add a parameter named: `replaceBy`  ?

```
RemoveSubscriptionConfig(string username, string eventSource, string groupName, PersistentSubscriptionEntry replaceBy);
RemoveSubscription(string eventSource, string groupName, PersistentSubscription replaceBy)
```

when `replaceBy` is null, we call RemoveAt(index) and when it's not null, we can replace the value at the index with that subscription or config.

shaan1337: the replacement null check would also need to be done for `_subscriptionsById.Remove(key);`

src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
shaan1337: edit: this may no longer be relevant if we go for the change suggested here: https://github.com/EventStore/EventStore/pull/3533/files#r987722268

src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
shaan1337: for more clarity, maybe we could then rename `RemoveSubscriptionConfig` to `UpdateSubscriptionConfig` and `RemoveSubscription` to `UpdateSubscription`

dhingrak: Thanks a lot Shaan for the suggestions. Changes updated

src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
```diff
@@ -460,10 +463,15 @@ string user
...
-				StartFrom = startFrom is PersistentSubscriptionSingleStreamPosition x ? x.StreamEventNumber : long.MinValue,
+				StartFrom = startFrom is PersistentSubscriptionSingleStreamPosition x
+					? x.StreamEventNumber
+					: long.MinValue,
 				#pragma warning restore 612
 				StartPosition = startFrom.ToString()
-			});
+			};
+			
+			UpdateSubscription(stream,groupName,subscription);
```
shaan1337, line 473:
```suggestion
			UpdateSubscription(stream, groupName, subscription);
```

src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
```diff
@@ -718,24 +737,47 @@ string user
...
 			_config.UpdatedBy = username;
 			var index = _config.Entries.FindLastIndex(x => x.Stream == eventSource && x.Group == groupName);
-			_config.Entries.RemoveAt(index);
+			
+			// Updating the persistent subscription if replaceBy is not null, otherwise delete the persistent subscription
+			if (replaceBy != null) {
+				_config.Entries[index] = replaceBy;
+			} else {
+				_config.Entries.RemoveAt(index);
+			}
```
shaan1337, line 751:
```suggestion
			// Update the persistent subscription if replaceBy is not null, otherwise delete the persistent subscription
			if (replaceBy != null)
				_config.Entries[index] = replaceBy;
			else
				_config.Entries.RemoveAt(index);
```

src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
```diff
@@ -718,24 +737,47 @@ string user
...
 			var key = BuildSubscriptionGroupKey(eventSource, groupName);
-			_subscriptionsById.Remove(key);
+			
+			// Updating the persistent subscription if replaceBy is not null, otherwise delete the persistent subscription
+			if(replaceBy != null)
+			{
+				_subscriptionsById[key] = replaceBy;
+			} else {
+				_subscriptionsById.Remove(key);
+			}
```
shaan1337, line 764:
```suggestion
			// Update the persistent subscription if replaceBy is not null, otherwise delete the entry
			if (replaceBy != null)
				_subscriptionsById[key] = replaceBy;
			else
				_subscriptionsById.Remove(key);
```

src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
```diff
@@ -718,24 +737,47 @@ string user
...
+							sub.Shutdown();
+						} catch (Exception ex) {
+							Log.Error(ex, "Failed to shut down subscriber with id: {subscriptionId}", sub.SubscriptionId);
+						}
+
+						if (replaceBy != null) {
+							subscribers[i] = replaceBy;
+						} else {
+							subscribers.RemoveAt(i);
+						}
```
shaan1337, line 780:
```suggestion
						if (replaceBy != null)
							subscribers[i] = replaceBy;
						else
							subscribers.RemoveAt(i);
```

src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
```diff
@@ -718,24 +737,47 @@ string user
...
 			if (_subscriptionTopics.TryGetValue(eventSource, out subscribers)) {
 				for (int i = 0; i < subscribers.Count; i++) {
 					var sub = subscribers[i];
 					if (sub.SubscriptionId == key) {
-						sub.Shutdown();
-						subscribers.RemoveAt(i);
+						try {
+							sub.Shutdown();
+						} catch (Exception ex) {
+							Log.Error(ex, "Failed to shut down subscriber with id: {subscriptionId}", sub.SubscriptionId);
```
shaan1337, line 773:
```suggestion
							Log.Error(ex, "Failed to shut down subscription with id: {subscriptionId}", sub.SubscriptionId);
```

src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
```diff
@@ -399,18 +400,19 @@ string user
 				Action<string> onSuccess,
 				Action<string> onFail,
 				Action<string> onNotExist,
-				Action<string> onAccessDenied,
+				Action<string> onAccessDenied,
```
shaan1337, line 403:
```suggestion
				Action<string> onAccessDenied,
```

src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
```diff
@@ -583,13 +590,13 @@ string user
...
-			if (!_subscriptionTopics.TryGetValue(eventSource.ToString(), out subscribers)) {
-				subscribers = new List<PersistentSubscription>();
-				_subscriptionTopics.Add(eventSource.ToString(), subscribers);
-			}
+			var stream = eventSource.ToString();
+			var key = BuildSubscriptionGroupKey(stream, groupName);
 
+			if (_subscriptionsById.ContainsKey(key)) {
+				return false;
+			}
```
shaan1337, line 598:
```suggestion
			if (_subscriptionsById.ContainsKey(key))
				return false;
```

src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
```diff
@@ -253,11 +249,16 @@ string user
 				ToMessageTimeout(messageTimeoutMilliseconds)
 			);
 
+			if (!result) {
+				onExists($"Group '{groupName}' already exists.");
+				return;
+			}
+
 			Log.Debug("New persistent subscription {subscriptionKey}", key);
 			_config.Updated = DateTime.Now;
```
shaan1337, line 258: similar to: https://github.com/EventStore/EventStore/pull/3533/files#r988645440
we can refactor this part and call `UpdateSubscriptionConfig(user, stream, groupName, createEntry);`

src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
shaan1337: thanks very much Kunal. I've made a few formatting suggestions and some additional refactoring suggestions for the creation part.

src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
shaan1337: thanks!

src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
shaan1337: thanks!

dhingrak:
Thanks a lot Shaan for helping me in refactoring the code.

src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs
```diff
@@ -235,8 +231,8 @@ string user
 				onFail(startFromValidationError);
 				return;
 			}
-
-			CreateSubscriptionGroup(eventSource,
+			
+			var result = CreateSubscriptionGroup(eventSource,
```
shaan1337, line 235: suggestion:
1. we can rename `CreateSubscriptionGroup` to `TryCreateSubscriptionGroup` 
2. [here](https://github.com/EventStore/EventStore/blob/0f4a2db987faeec93d642c2dcc6cf7d63b54837f/src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionService.cs#L1028), we can log a warning when TryCreateSubscriptionGroup() returns false:
```
var key = BuildSubscriptionGroupKey(eventSource.ToString(), groupName);
Log.Warning("A duplicate persistent subscription: {subscriptionKey} was found in the configuration. Ignoring it.", key);
```

shaan1337 (approved):

thefringeninja (approved):

PR #3543, opened 2022-07-28
Terminate live gRPC subscription if it falls behind

timothycoleman:
Fixed: Slow EventCommitted handling. Live gRPC subscriptions that fall behind are dropped and invited to resubscribe.

The client can resubscribe from where it received up to.
The slowness was due to lock contention.
Longer term the subscription will be able to correctly switch between catchup and live smoothly but this requires more thought.

### DIFF
--- a/src/EventStore.Core/Services/Transport/Grpc/Enumerators.AllSubscription.cs
+++ b/src/EventStore.Core/Services/Transport/Grpc/Enumerators.AllSubscription.cs
@@ -264,7 +264,9 @@ namespace EventStore.Core.Services.Transport.Grpc {
 								caughtUpSource.TrySetResult(position);
 							}
 
+#pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
 							async Task OnHistoricalEventsMessage(Message message, CancellationToken ct) {
+#pragma warning restore CS1998
 								if (message is ClientMessage.NotHandled notHandled &&
 								    RpcExceptions.TryHandleNotHandled(notHandled, out var ex)) {
 									Fail(ex);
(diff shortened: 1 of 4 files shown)


thefringeninja (approved):

jageall (approved):

hayley-jean (approved):

ahjohannessen:
This change is a breaking semantic change! Updating my cluster from 21.10.5 to 21.10.7 I would not expect this behaviour. You put the burden on clients with this. Instead of this hack, you could have opted for a setting that was more forgiving, e.g. opt-in on server side or opt-in in the subscribe grpc message.

ahjohannessen:
If you are not going to change this to what it was previously with some backoff, I suggest that the exception is something more specific than:

```
io.grpc.StatusRuntimeException: ABORTED: Operation timed out: Consumer too slow to handle event while live. Client resubscription required.
```
Perhaps more like the existing rpc expections `maximum-subscribers-reached` or `persistent-subscription-dropped`. This way, I as a client, can parse headers for, say, `resubscription-required` instead of looking for status code ABORTED and parse the message for `Client resubscription required` in order to get a specific exception.

jageall:
@ahjohannessen Clients have always had to handle errors coming back from the server and re-subscribe for example due to network issues or a node restarting, so although it does put the burden of resubscription on to the clients in the event of the new error case, there are arguments both ways as to whether or not this is a breaking semantic change. 

This was an urgent fix as it was causing server production outages. Adding in another rpc exception specifically for this was going to force a revision of the clients for the affected parties and not an option. 

We are working on a better fix, but it needs more time.  It should be also noted that the server does not use semantic versioning. Server versions are based on support cycles and patches. we will look at updating the release notes to make the effect of this change clearer.

ahjohannessen:
I am well aware of that I need to handle retries when a node is restarting or network issues, handling a new grpc status ABORTED for subscriptions was not something I expected.

I have already updated my client and put it into production in order to handle ABORTED grpc status for subscriptions. If this behaviour is for the long term, I think adding in another rpc exception is the most principled approach.

I suggest that you try solve the problem on server side with catch up / read and backoff until you finally give up after a configurable retries and then return an exception to the client.

PR #3548, opened 2022-08-01
Update changelog for 21.10.7

hayley-jean:


Lougarou (approved):

Lougarou (commented):
is it worth mentioning https://github.com/EventStore/EventStore/pull/3542?

pvanbuijtene (approved):

hayley-jean:
> is it worth mentioning #3542?

@Lougarou I don't think so, since it was added and then reverted so it's not actually in the release

PR #3569, opened 2022-08-18
Dynamically resize stream info caches based on available memory

shaan1337:
Added: Dynamically resize stream info caches based on available memory

This PR adds support for caches that can be automatically resized based on free memory. Right now only one cache is dynamic: the `StreamInfo` cache but others can be added later on. A user can still make the cache static by specifying --stream-info-cache-capacity.

The rationale behind the changes is that a static cache which is scaled to free memory available on startup would not be able to efficiently deal with these scenarios:
- growth of the database with time
- heavy operations like index merges, scavenging etc. (an estimate could be done but it would waste memory during non-peak hours)
- launching multiple db instances on a single machine (not recommended)
- growth in the number of clients connected
- spike in memory usage of other processes on the system

## Memory allocation ratio
The `StreamInfo` cache is a logical cache comprised of the `LastEventNumber` and `StreamMetadata` cache. Currently 60% of the memory allocated to the logical cache is reserved to the `LastEventNumber` cache and 40% to `StreamMetadata` cache (this is based on empirical testing).

## Cache resizer hierarchy
The cache resizers are modeled as a tree structure. When resizing caches, we specify the capacity that we wan't to allocate to the root of the tree. The children (and sub-children) will then derive their capacities based on their weights and the amount allocated to them by their parent.

## Resizing algorithm

The resizing algorithm is as follows:
- Every 15 seconds:
  - first check if the amount of memory available for caching has changed by more than 200MB since the last resize. If this condition is true, continue to the next step, otherwise don't do anything.
  - check if there's less than 25% or 6 GiB free memory. If this condition is true, resize the caches, otherwise go to next step. The total free memory takes these factors into consideration
    - free system memory
    - free memory on the .NET heap
    - memory deleted by the caches but not yet garbage collected. 
  - check if 10 minutes has passed since the last resize. If condition is true, resize caches, otherwise do nothing.
 
The 200 MB change threshold ensures that:
- We don't resize too often if not much has changed.

Resizing the caches every 10 mins ensures that:
- the caches can grow again when more memory is available
- we don't abruptly decrease the cache sizes by a large value (unless there's an abrupt free memory increase within past 10 minutes)

Checking every 15 seconds ensures that:
- we can quickly act if we exceed the 25% / 6 GiB free memory threshold before it's too late and we run out of memory

## New stats
The following stats were added under `.stats.es`:

```
  "cache": {
    "StreamInfo": {
      "LastEventNumber": {
        "name": "LastEventNumber",
        "count": 266339,
        "sizeBytes": 45833588,
        "capacityBytes": 2032332494,
        "utilizationPercent": 2.2552209412245907
      },
      "Metadata": {
        "name": "Metadata",
        "count": 133169,
        "sizeBytes": 34591868,
        "capacityBytes": 1354888329,
        "utilizationPercent": 2.553115799995942
      },
      "name": "StreamInfo",
      "count": 399508,
      "sizeBytes": 80425456,
      "capacityBytes": 3387220824,
      "utilizationPercent": 2.3743788840145603
    },
    "name": "cache",
    "count": 399508,
    "sizeBytes": 80425456,
    "capacityBytes": 3387220824,
    "utilizationPercent": 2.3743788840145603
  }
```

`.stats.sys.totalMem` - total system memory

`.stats.proc.gc.fragmentation` - what percent of the heap is fragmented

## TODOs in later PRs
- Convert the chunks cache to a dynamic cache (e.g if ChunksCacheSize is set to -1)
- Add a `StreamInfoCacheSize` parameter (measured in bytes) which deprecates `StreamInfoCacheCapacity` (breaking change)


shaan1337:
Based on tests with a large instance, it seems that the size calculation is lower by ~25% than the actual size, i'll update the PR to make the calculation more precise (LRU cache overhead is not properly accounted for)

src/EventStore.Core/Caching/MemSizer.cs
```diff
@@ -0,0 +1,31 @@
...
+
+		public static int SizeOf(string[] arr)
+		{
+			if (arr == null)
+				return 0;
+
+			var size = 0;
+			size += ArraySize; // string array
+			size += arr.Length * Unsafe.SizeOf<string>(); // string refs
+			size += arr.Sum(SizeOf); // strings
```
timothycoleman, line 27:
```suggestion
			size += arr.Sum(static x => SizeOf(x)); // strings
```

adding the explicit argument x causes the compiler to cache the func rather than allocating a new one each time. adding the keyword static cause a compiler error if some future code change will prevent it from doing that

timothycoleman:
i pushed a bunch of trivial adjustments to a branch `dynamic-cache-sizeing-tim-suggestions`, consider these suggestions you can merge/squash in whichever ones you like

src/EventStore.Core/Caching/MemSizer.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace EventStore.Core.Caching {
+	public class MemSizer {
+		public const int ObjectHeaderSize = 16;
+		public const int ArraySize = 24;
+		public const int StringSize = 20;
+
+		public static int SizeOf(string s)
+		{
+			if (s.Length == 0)
```
timothycoleman, line 14: consider checking for null string

src/EventStore.Core/Caching/DynamicCacheManager.cs
```diff
@@ -0,0 +1,200 @@
...
+			return availableMem;
+		}
+
+		private long CalcMemAllocation(long availableMem, int cacheWeight, long minMemAllocation) {
+			return Math.Max(availableMem * cacheWeight / _totalWeight, minMemAllocation);
+		}
+
+		private void Tick() {
+			_bus.Publish(
+				TimerMessage.Schedule.Create(
```
timothycoleman, line 194: the schudule message could be a member and reused each time (same as in InaugurationManager)

src/EventStore.Core/Caching/DynamicCacheManager.cs
```diff
@@ -0,0 +1,200 @@
...
+				sw.Stop();
+				Log.Debug(
+					"{name} cache resized to ~{allocatedMem:N0} bytes in {elapsed}.",
+					cacheSettings.Name, allocatedMem, sw.Elapsed);
+				_maxMemAllocation[cacheIndex] = allocatedMem;
+			}
+		}
+
+		private long CalcAvailableMemory(long freeMem, long cachedMem) {
+			var keepFreeMem = Math.Max(_keepFreeMemBytes, _totalMem * _keepFreeMemPercent / 100);
```
timothycoleman, line 172: dunno when machines will get enough memory to make this overflow but :)

src/EventStore.Core/Caching/DynamicCacheManager.cs
```diff
@@ -0,0 +1,200 @@
...
+						"{name} cache size auto-configured to ~{allocatedMem:N0} bytes.",
+						cacheSettings.Name, allocatedMem);
+					_maxMemAllocation[cacheIndex] = allocatedMem;
+				}
+			}
+		}
+
+		private void ResizeCachesIfNeeded() {
+			var freeMem = _getFreeMem();
+			var keepFreeMem = Math.Max(_keepFreeMemBytes, _totalMem * _keepFreeMemPercent / 100);
```
timothycoleman, line 124: consider refactor with the same line in CalcAvailableMemory

src/EventStore.Core/Caching/DynamicCacheManager.cs
```diff
@@ -0,0 +1,200 @@
...
+			          "Keep free bytes: {keepFreeMemBytes:N0} bytes\n\n" +
+			          "Memory available for caching: ~{availableMem:N0} bytes\n",
+				freeMem, _totalMem, cachedMem,
+				_keepFreeMemPercent, _keepFreeMemBytes, availableMem);
+
+			return availableMem;
+		}
+
+		private long CalcMemAllocation(long availableMem, int cacheWeight, long minMemAllocation) {
+			return Math.Max(availableMem * cacheWeight / _totalWeight, minMemAllocation);
```
timothycoleman, line 189: potential overflow

src/EventStore.Core/Caching/DynamicLRUCache.cs
```diff
@@ -0,0 +1,32 @@
...
+using EventStore.Core.DataStructures;
+
+namespace EventStore.Core.Caching {
+	public class DynamicLRUCache<TKey, TValue> : LRUCache<TKey, TValue>, IDynamicLRUCache<TKey, TValue> {
+		const int ResizeBatchSize = 100000;
+
+		public DynamicLRUCache(long capacity, Func<TKey, TValue, int> calculateItemSize = null)
+			: base(capacity, calculateItemSize) { }
+
+		public void Resize(long newCapacity, out int removedCount, out long removedSize) {
```
timothycoleman, line 12: tbh i'd say this operation is invasive/coupled enough to the implementation of the LRUCache that it warrants putting it in the LRUCache itself, and get rid of this derived class (and then probably the dynamic interface too)

src/EventStore.Core/Data/StreamMetadata.cs
```diff
@@ -310,5 +311,17 @@ public class StreamMetadata {
 				jsonWriter.WriteEndArray();
 			}
 		}
+
+		public int Size =>
```
timothycoleman, line 315: consider renaming these `Size` getters to something like `ApproximateSize` or `SizeForCache`, since this isn't for marshalling/size required in a buffer etc

src/EventStore.Core/DataStructures/ILRUCache.cs
```diff
@@ -1,8 +1,13 @@
 using System;
 
 namespace EventStore.Core.DataStructures {
-	public interface ILRUCache<TKey, TValue> {
+	public interface ILRUCache {
+		long Size { get; }
```
timothycoleman, line 5: consider making these explicitly bytes since the same property names in the generic collections are measured in items

timothycoleman:
could update LogV2Sizer to have a more accurate calculation like the memsizer, since we're using the LogV2Sizer to do part of the size calculation

src/EventStore.Core/Services/Storage/ReaderIndex/IndexBackend.cs
```diff
@@ -24,18 +29,69 @@ public interface IIndexBackend<TStreamId> : IIndexBackend {
...
+			_streamIdSizer.GetSizeInBytes(streamId) + metadataCached.Size + MetadataCached.DictionaryEntryOverhead;
+
+		private void UpdateMaxMemoryAllocation(long allocatedMem) {
+			UpdateCapacity(_streamLastEventNumberCache, LastEventNumberCacheName,
+				allocatedMem * LastEventNumberCacheRatio / 100);
+			UpdateCapacity(_streamMetadataCache, StreamMetadataCacheName,
+				allocatedMem * StreamMetadataCacheRatio / 100);
+		}
+
+		private delegate void LogAction(string msg, params object[] args);
```
timothycoleman, line 83: can use `Log.Write` and pass in the level rather than using a delegate to switch between two methods

shaan1337:
hey @timothycoleman thanks for the feedback! will go through it asap.

i've pushed a couple more commits for more precise calculation and also some fixes (where memory wasn't being properly released)

src/EventStore.Core/Caching/MemSizer.cs
shaan1337: fixed in: 57a260ecf1f37450ebc517760d4b8ce7682e764b

src/EventStore.Core/Caching/MemSizer.cs
shaan1337: fixed in: 57a260ecf1f37450ebc517760d4b8ce7682e764b

src/EventStore.Core/Caching/DynamicCacheManager.cs
shaan1337: fixed in: 58281fc4812c1c8c80ada6f0ddc0c3716254f7a5

src/EventStore.Core/Caching/DynamicCacheManager.cs
shaan1337: fixed in: c2d1aac9ea0cc79ff2f0ae338223c45a936a4e82

src/EventStore.Core/Caching/DynamicCacheManager.cs
shaan1337: fixed in: c2d1aac9ea0cc79ff2f0ae338223c45a936a4e82

src/EventStore.Core/Caching/DynamicCacheManager.cs
shaan1337: fixed in: c2d1aac9ea0cc79ff2f0ae338223c45a936a4e82

src/EventStore.Core/Caching/DynamicLRUCache.cs
shaan1337: fixed in: 890b0f3b3c73cc857bcd9868411d82cb01a38727

src/EventStore.Core/Data/StreamMetadata.cs
shaan1337: fixed in: f3d24f346c8b6815142904b38393fb1a9bfb97ba

src/EventStore.Core/Services/Storage/ReaderIndex/IndexBackend.cs
shaan1337: fixed in: e06f87cac61d79fde696581e1a760d0cf14fbb23

src/EventStore.Core/DataStructures/ILRUCache.cs
shaan1337: the `Size` and `Capacity` properties are flexible: they can represent just an item count or also size in bytes (or any other arbitrary value).

For e.g with the default constructor of LRUCache, calculateSize just returns 1 (which represents item count)

timothycoleman: ah i see, yep fair enough :+1:

src/EventStore.Core/Services/Storage/ReaderIndex/IndexBackend.cs
```diff
@@ -24,18 +29,69 @@ public interface IIndexBackend<TStreamId> : IIndexBackend {
...
-		private readonly ILRUCache<TStreamId, MetadataCached> _streamMetadataCache;
+		private readonly IDynamicLRUCache<TStreamId, EventNumberCached> _streamLastEventNumberCache;
+		private readonly IDynamicLRUCache<TStreamId, MetadataCached> _streamMetadataCache;
+		private readonly ISizer<TStreamId> _streamIdSizer;
+
+		// very rough approximation of memory taken by one item in the stream info cache.
+		// used only when cache is configured in terms of "max items" rather than "max bytes".
+		public const long StreamInfoCacheUnitSize = 1000;
+
+		private const int LastEventNumberCacheRatio = 60;
```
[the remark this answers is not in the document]
shaan1337: fixed after refactoring (with composite cache resizer)

hayley-jean (approved):
Will this need to be rebased due to the dotnet6 upgrade?

hayley-jean (approved):

timothycoleman (approved):

PR #3570, opened 2022-08-18
Add better error messages for the gossip seed configuration

jamesmintram:
Displays a human readable error when the port is missing
Displays a human readable error when the port is not a valid int

Also updated the name of ParseEndPoint to ParseGossipEndPoint as
the error messages created are specific to the Gossip seed.

Proposed fix for: https://github.com/EventStore/EventStore/issues/3565


timothycoleman (approved):
thanks!

hayley-jean (approved):

hayley-jean:
Looks good, thanks you!

PR #3571, opened 2022-08-23
Update tracing packages to fix memory leak

PaskeS:
Fixed: Memory leak caused by EventPipeEventSource

When doing some performance analysis I found that memory was growing at a rate of about ~4mb/hr

Example ES left running over a weekend (57 hours)
image

Example ES that I pulled from a UAT box (long uptime)
image

Tested locally after upgrading packages finding after 16 hours of running under load that EventPipeEventSource had used 8mb, this was still climbing at a rate of 0.5mb/hr

It was caused by `AddCallbackOnDotNetRuntimeLoad` that was only used to count gc collections. This in itself was actually incorrect as it would count a generation 2 gc but dismissed the fact that ment there was a generation 0 and 1 gc. I have replaced this in place with `gen-x-gc-count` from dotnet-counters that counts this over a time interval. This means the stat wont be as responsive (as its counted over an interval) but realistically not an issue for this stat. 

If we want the stat to use the "old" values I am happy to change `GetGcStats` so the counter will negate higher levels gcs e.g. gen 0 = gen 0 - gen 1 - gen2


thefringeninja:
@PaskeS great work! Just one thing, we are going to need to keep the original behavior. Please use the values from `dotnet-counters` in newly defined fields. Thank you!

PaskeS:
> @PaskeS great work! Just one thing, we are going to need to keep the original behavior. Please use the values from `dotnet-counters` in newly defined fields. Thank you!

Have updated to keep to the previous behaviour as it was before

timothycoleman (changes requested):
Hi @PaskeS

Great find on the memory leak!

Sorry, there was a misunderstanding about the previous behaviour. It's fine for this PR to correct the undercounting of gen0 and gen1. So your last commit can be removed.

The behaviour that is important to preserve is that the stats show the total number of GCs rather than GCs for the period. It looks like your code does attempt to do this in the block starting line 59, but it flows down into the subsequent code and gets overwritten. Would you mind fixing this?

Then just make sure the commit history is tidy and we should be good to go

PaskeS:
> Hi @PaskeS
> 
> Great find on the memory leak!
> 
> Sorry, there was a misunderstanding about the previous behaviour. It's fine for this PR to correct the undercounting of gen0 and gen1. So your last commit can be removed.
> 
> The behaviour that is important to preserve is that the stats show the total number of GCs rather than GCs for the period. It looks like your code does attempt to do this in the block starting line 59, but it flows down into the subsequent code and gets overwritten. Would you mind fixing this?
> 
> Then just make sure the commit history is tidy and we should be good to go

Understood, forgot to return. Will sort it out now and rebase to one commit

PaskeS:
@timothycoleman I have fixed up, have rebased and changed history to one commit. However being rejected when force pushing my branch, do you have any branch protections in place that would stop this? I have committed to a second branch (https://github.com/PaskeS/EventStore/tree/tracing-memory-leak-rb) and can PR this if easier

jageall:
@PaskeS  There shouldn't be anything stopping you force pushing to your branch as it is in your fork, so that is a bit weird. I'd like to get this moving so if it is easier to pr your second branch then that is probably best

PaskeS:
@jageall https://github.com/EventStore/EventStore/pull/3578 -> lost my patience with it

PaskeS:
Closing in favour of above PR

Issue #3547, opened 2022-07-30
Deleting a stream using gRPC produces an incorrect error


andriykohut:
**Describe the bug**
Calling `Streams.Delete` for a stream that doesn't exist or using an unknown revision number produces incorrect error message:

```
StatusCode.FAILED_PRECONDITION
details = "Append failed due to WrongExpectedVersion. Stream: ..., Expected version: -1, Actual version: 0"
```

**To Reproduce**
1. Use `Streams.Delete` on an unknown stream with `Stream exists` expected state, or call a delete on an existing stream with revision that doesn't exist

**Expected behavior**
The error should mention Delete and not Append

**Config/Logs/Screenshots**
Log example:
```
es-client-eventstore.db-1         | [    1,14,20:29:33.633,INF] Error status code 'FailedPrecondition' raised.
es-client-eventstore.db-1         | Grpc.Core.RpcException: Status(StatusCode="FailedPrecondition", Detail="Append failed due to WrongExpectedVersion. Stream: 3aadc515-4027-4823-89bf-b7e0f92f6a0e, Expected version: -1, Actual version: 0")
es-client-eventstore.db-1         |    at EventStore.Core.Services.Transport.Grpc.Streams`1.DeleteInternal(String streamName, Int64 expectedVersion, ClaimsPrincipal user, Boolean hardDelete, Boolean requiresLeader, CancellationToken cancellationToken) in /home/runner/work/TrainStation/TrainStation/build/oss-eventstore/src/EventStore.Core/Services/Transport/Grpc/Streams.Delete.cs:line 152
es-client-eventstore.db-1         |    at EventStore.Core.Services.Transport.Grpc.Streams`1.Delete(DeleteReq request, ServerCallContext context) in /home/runner/work/TrainStation/TrainStation/build/oss-eventstore/src/EventStore.Core/Services/Transport/Grpc/Streams.Delete.cs:line 48
es-client-eventstore.db-1         |    at Grpc.Shared.Server.UnaryServerMethodInvoker`3.AwaitInvoker(Task`1 invokerTask, GrpcActivatorHandle`1 serviceHandle)
es-client-eventstore.db-1         |    at Grpc.Shared.Server.UnaryServerMethodInvoker`3.AwaitInvoker(Task`1 invokerTask, GrpcActivatorHandle`1 serviceHandle)
[...]

PR #3583, opened 2022-09-06
Fix incorrect error message when deleting a stream using gRPC

321zer0:
Fixed: Incorrect error message when deleting a stream using gRPC. Fixes issue #3547

### DIFF
--- a/src/EventStore.Core/Services/Transport/Grpc/RpcExceptions.cs
+++ b/src/EventStore.Core/Services/Transport/Grpc/RpcExceptions.cs
@@ -70,13 +70,14 @@ namespace EventStore.Core.Services.Transport.Grpc {
 
 
 		public static RpcException WrongExpectedVersion(
+			string operation,
 			string streamName,
 			long expectedVersion,
 			long? actualVersion = default) =>
 			new(
 				new Status(
 					StatusCode.FailedPrecondition,
-					$"Append failed due to WrongExpectedVersion. Stream: {streamName}, Expected version: {expectedVersion}, Actual version: {actualVersion}"),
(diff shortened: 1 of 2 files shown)


timothycoleman (dismissed):

src/EventStore.Core/Services/Transport/Grpc/RpcExceptions.cs
```diff
@@ -76,7 +76,7 @@ public static class RpcExceptions {
 			new(
 				new Status(
 					StatusCode.FailedPrecondition,
-					$"Append failed due to WrongExpectedVersion. Stream: {streamName}, Expected version: {expectedVersion}, Actual version: {actualVersion}"),
+					$"Delete failed due to WrongExpectedVersion. Stream: {streamName}, Expected version: {expectedVersion}, Actual version: {actualVersion}"),
```
hayley-jean, line 79: This changes the `RpcException.WrongExpectedVersion` method to be specific to the `Delete` operation, but there's nothing in the method signature that makes this obvious. While delete is currently the only operation using this, if someone were to add another operation that returns a `WrongExpectedVersion` exception then they may use this without realising that it'll return the wrong error message.

You could pass in an operation name to use in the error message (my preference), or perhaps change the method name to reflect that it's specific for `Delete`.

@timothycoleman what do you think?

hayley-jean, line 79: Or just change the message to `Operation` so that it works for every case:

```suggestion
					$"Operation failed due to WrongExpectedVersion. Stream: {streamName}, Expected version: {expectedVersion}, Actual version: {actualVersion}"),
```

timothycoleman:
just needs squashing to a single commit

timothycoleman (approved):

hayley-jean (approved):

PR #3584, opened 2022-09-07
Respond to request with NotReady if the node is not yet able to authenticate users, instead of ignoring.

timothycoleman:
Fixed: gRPC calls can no longer stall if received during server initialization.

Previously, authentications (such as for read requests) attempted during the Initializating state would stall.

Also
- Fixes a race condition where the HttpContext could be accessed after it had been disposed.
- Add proper headers/trailers when failing authentication for a gRPC call.
- Log correct duration when logging a failed authorization.

Likely will fix
- https://github.com/EventStore/EventStore/issues/3535
- https://github.com/EventStore/EventStore/issues/3575

### DIFF
--- a/src/EventStore.Core/Authorization/PolicyAuthorizationProvider.cs
+++ b/src/EventStore.Core/Authorization/PolicyAuthorizationProvider.cs
@@ -56,10 +56,11 @@ namespace EventStore.Core.Authorization {
 				if (result.Grant == Grant.Allow && _logSuccesses)
 					_logger.Information(
 						"Successful authorization check for {identity} in {duration} with {evaluationResult}",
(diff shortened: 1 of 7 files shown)


src/EventStore.Core/Authorization/PolicyAuthorizationProvider.cs
```diff
@@ -59,7 +59,7 @@ public class PolicyAuthorizationProvider : IAuthorizationProvider {
 						cp.FindFirst(ClaimTypes.Name).Value, sw.Elapsed.Subtract(startedAt), result);
```
jageall, line 59: this line needs the null check on the name the same as below

jageall (changes requested):

jageall (approved):

PR #3593, opened 2022-09-16
Update docs about persistent subscriptions to $all over HTTP

321zer0:
Changed: Deleting persistent subscriptions to $all over HTTP is not supported

### DIFF
--- a/docs/http-api/persistent.md
+++ b/docs/http-api/persistent.md
@@ -10,6 +10,8 @@ The Administration UI includes a _Competing Consumers_ section where you are abl
 
 Before interacting with a subscription group, you need to create one. You receive an error if you try to create a subscription group more than once. This requires [admin permissions](security.md).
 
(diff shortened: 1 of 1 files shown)


hayley-jean (changes requested):
The `Update` and `Create` operations also don't support persistent subscriptions to $all. So either we need a similar warning on each operation, or one at the top of the document that says that $all persistent subscriptions aren't supported over the HTTP API, and you should use gRPC instead.

I also think these should be in a warning block to make them more visible, like the TCP warning [here](https://developers.eventstore.com/server/v21.10/#tcp-protocol). You can do this by putting the text in a `::: warning` block like [this](https://github.com/EventStore/EventStore/blob/4e56eacb954814d9bb47b0ac9616d084688e6848/docs/README.md#tcp-protocol)

321zer0:
@hayley-jean Thanks for the review. I've added warning blocks and also instructions for the `update` and create `operations` as well

Lougarou (approved):

hayley-jean (approved):

PR #3598, opened 2022-10-03
Improve nunit test logging

timothycoleman:
Changed: Improved unit test logging 

- Configure logger to output Warnings from the production code to the console, and Debug messages from the test code.
- Log a warning if node shutdown times out
- Add overload of IsOrBecomesTrue that can generate the message upon giving up
- MiniNode and MiniClusterNode (debug) log on state change
- SpecificationWithDirectory(PerTestFixture) debug logs on setup/teardown, makes it easier to tell which test is actually running

Significant advantage here is that if the mininode experiences an error we will be able to see it and which test it is associated with in the test output.

### DIFF
--- a/src/EventStore.Core.Tests/Helpers/MiniClusterNode.cs
+++ b/src/EventStore.Core.Tests/Helpers/MiniClusterNode.cs
@@ -55,7 +55,7 @@ namespace EventStore.Core.Tests.Helpers {
 		public Task Started => _started.Task;
 		public Task AdminUserCreated => _adminUserCreated.Task;
 
-		public VNodeState NodeState = VNodeState.Unknown;
+		public VNodeState NodeState { get; private set; }
 		private readonly IWebHost _host;
 
 		private readonly TestServer _kestrelTestServer;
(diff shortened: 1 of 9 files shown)


pvanbuijtene (dismissed):

pvanbuijtene (dismissed):

src/EventStore.Core.Tests/Helpers/MiniClusterNode.cs
```diff
@@ -55,7 +55,7 @@ public class MiniClusterNode<TLogFormat, TStreamId> {
 		public Task Started => _started.Task;
 		public Task AdminUserCreated => _adminUserCreated.Task;
 
-		public VNodeState NodeState = VNodeState.Unknown;
+		public VNodeState NodeState;
```
shaan1337, line 58:
```suggestion
		public VNodeState NodeState { get; private set; }
```

timothycoleman: :+1:

pvanbuijtene (dismissed):

shaan1337 (commented):
Given that most of the time, the tests will pass (except some flaky ones occasionally), do we want to add an environment variable to control whether we want to turn the test output on? By default, we could then set this environment variable on CI.

Someone running tests locally may only want to run a few tests to see if they pass, and he/she would need to scroll through the output to check the results

timothycoleman:
good idea, will come back to this after the server release

PR #3599, opened 2022-10-04
Node not going into ready state due to synchronization bug in UserManagementService

shaan1337:
Fixed: Node not going into ready state due to synchronization bug in UserManagementService

`NotifyInitialized()` can be called from different threads but there was no synchronization around `_numberOfStandardUsersToBeCreated`.

When a node transitions to leader, it creates the admin & ops users if they don't already exist. This is done asynchronously. If both the admin & ops users happen to finish being created at the same time, NotifyInitialized() will be called from two different threads and a dirty read of `_numberOfStandardUsersToBeCreated` can happen - both threads will see the value as 2 and decrement it to 1.

The node thus never transitions to the `Ready` state.

Luckily, the issue does not have an impact on the node's status on every election but only on the node's startup (since once [_ready](https://github.com/EventStore/EventStore/blob/13828a891d1bc4e683bc3cdc6af2556a8e520f12/src/EventStore.Core/ClusterVNodeStartup.cs#L50) is true, it stays true)

## Reproduction steps
1. On master, run the test: `EventStore.Core.Tests.ClientOperations.when_starting_a_transaction_expecting_no_stream` 100 times in a loop, it will hang at one point:
```
$ dotnet build -c Release src/EventStore.Core.Tests/EventStore.Core.Tests.csproj
$ for i in `seq 1 100`; do dotnet test -c Release --no-build src/EventStore.Core.Tests/EventStore.Core.Tests.csproj --filter FullyQualifiedName~EventStore.Core.Tests.ClientOperations.when_starting_a_transaction_expecting_no_stream; done
```

2. Attach a debugger (e.g with Rider), you'll see that the thread is stuck at `_node.StartAsync` in `CreateTestNode()`

Do the above with this branch - none of the 100 test runs will hang.

### DIFF
--- a/src/EventStore.Core/Authentication/InternalAuthentication/UserManagementService.cs
+++ b/src/EventStore.Core/Authentication/InternalAuthentication/UserManagementService.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Linq;
 using System.Security.Claims;
+using System.Threading;
 using System.Threading.Tasks;
 using EventStore.Common.Utils;
 using EventStore.Core.Bus;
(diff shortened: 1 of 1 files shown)


timothycoleman (approved):
Nice spot :+1:

pvanbuijtene (approved):

PR #3609, opened 2022-10-07
Authentication handles NotHandled message rather than waiting for timeout

timothycoleman:
Fixed: Quick responses for authentication requests when the node is not ready

Now the authentication handles the NotHandled message which is generated when the node is not in a state to be able to repond to reads (such as while becoming leader after an election).

Prior to this change the authentication would not fail until it timed out 10s later, causing the client to be waiting an unnecessarily long time.

Also
- Handle timeouts via RequestResponseDispatcher so that it can tidy up the entry in its own map, avoiding leak.
- Pass the whole RequestResponseDispatcher into the schedule messages rather than allocating lots of closures that will end up in gen2

### DIFF
--- a/src/EventStore.Core/Helpers/IODispatcherDelayedMessage.cs
+++ b/src/EventStore.Core/Helpers/IODispatcherDelayedMessage.cs
@@ -10,22 +10,22 @@ namespace EventStore.Core.Helpers {
 		}
 
 		private readonly Guid _correlationId;
-		private readonly Action _action;
+		private readonly ICorrelatedTimeout _timeout;
 		private readonly Guid? _messageCorrelationId;
 
-		public IODispatcherDelayedMessage(Guid correlationId, Action action) {
-			_action = action;
+		public IODispatcherDelayedMessage(Guid correlationId, ICorrelatedTimeout timeout) {
+			_timeout = timeout;
 			_correlationId = correlationId;
 		}
 
-		public IODispatcherDelayedMessage(Guid correlationId, Action action, Guid? messageCorrelationId) {
-			_action = action;
+		public IODispatcherDelayedMessage(Guid correlationId, ICorrelatedTimeout timeout, Guid messageCorrelationId) {
+			_timeout = timeout;
 			_correlationId = correlationId;
 			_messageCorrelationId = messageCorrelationId;
 		}
 
-		public Action Action {
-			get { return _action; }
+		public void Timeout() {
+			_timeout.Timeout(_messageCorrelationId ?? Guid.Empty);
 		}
 
 		public Guid CorrelationId {
--- a/src/EventStore.Core/Messaging/RequestResponseAltDispatcher.cs
+++ b/src/EventStore.Core/Messaging/RequestResponseAltDispatcher.cs
@@ -1,0 +1,65 @@
+﻿using System;
+using EventStore.Core.Bus;
+using EventStore.Core.Helpers;
+using EventStore.Core.Messages;
+
+namespace EventStore.Core.Messaging {
+	// Just a typedef
+	public class ReadDispatcher :
+		RequestResponseAltDispatcher<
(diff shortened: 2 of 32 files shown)


src/EventStore.Core/Messaging/RequestResponseAltDispatcher.cs
```diff
@@ -0,0 +1,65 @@
...
+				getResponseCorrelationId,
+				defaultReplyEnvelope,
+				cancelMessageFactory) {
+			_getResponseAltCorrelationId = getResponseAltCorrelationId;
+		}
+
+		void IHandle<TResponseAlt>.Handle(TResponseAlt message) {
+			var correlationId = _getResponseAltCorrelationId(message);
+
+			// if we dont handle the alternative message, then dont remove the handler
```
shaan1337, line 56:
```suggestion
			// if we don't handle the alternative message, then don't remove the handler
```

src/EventStore.Core/Messaging/RequestResponseAltDispatcher.cs
```diff
@@ -0,0 +1,65 @@
...
+		where TResponseAlt : Message
+		where THandler : IHandle<TResponse>, IHandleAlt<TResponseAlt>, IHandleTimeout {
+
+		private readonly Func<TResponseAlt, Guid> _getResponseAltCorrelationId;
+
+		public RequestResponseAltDispatcher(
+			IPublisher publisher,
+			Func<TRequest, Guid> getRequestCorrelationId,
+			Func<TResponse, Guid> getResponseCorrelationId,
+			Func<TResponseAlt, Guid> getResponseAltCorrelationId,
```
shaan1337, line 42: i'm not sure i understand why there's a need for the response correlation IDs, don't they always have the same value as the request correlation id?

src/EventStore.Core/Messaging/RequestResponseDispatcher.cs
```diff
@@ -49,26 +80,38 @@ public sealed class RequestResponseDispatcher<TRequest, TResponse> : IHandle<TRe
...
-				action(message);
+				handler.Handle(message);
 				return true;
 			}
 
 			return false;
 		}
 
+		public void Timeout(Guid correlationId) {
+			// if we dont handle timeouts then don't remove the handler
```
shaan1337, line 96:
```suggestion
			// if we don't handle timeouts then don't remove the handler
```

src/EventStore.Core/Helpers/IODispatcherDelayedMessage.cs
```diff
@@ -10,22 +10,22 @@ public sealed class IODispatcherDelayedMessage : Message {
 		}
 
 		private readonly Guid _correlationId;
-		private readonly Action _action;
+		private readonly ICorrelatedTimeout _timeout;
 		private readonly Guid? _messageCorrelationId;
 
-		public IODispatcherDelayedMessage(Guid correlationId, Action action) {
-			_action = action;
+		public IODispatcherDelayedMessage(Guid correlationId, ICorrelatedTimeout timeout) {
```
shaan1337, line 16: suggestion: rename `IODispatcherDelayedMessage` to `IODispatcherTimeoutMessage`

previously the code was handling a general, scheduled message, so the name `IODispatcherDelayedMessage` was appropriate but now it handles timeouts only

src/EventStore.Core/Messaging/RequestResponseAltDispatcher.cs
shaan1337: ahh ok got it, i thought we knew the request correlation id upfront, but it's generated only after the read request is sent (otherwise we could just pass in a single id here instead of 3 functions)

shaan1337 (approved):

shaan1337 (approved):

hayley-jean (approved):

PR #3610, opened 2022-10-07
WIP: handle io dispatcher timeouts better

timothycoleman:
- handle timeouts via the RequestResponseDispatchers so they can tidy up the map
- pass the whole RequestResponseDispatcher in the schedule messages rather than allocating closures that will end up in gen2

needs some testing, will need rebasing (it is only the second commit not the first)


timothycoleman:
encorporated into https://github.com/EventStore/EventStore/pull/3609

PR #3643, opened 2022-11-01
Log warnings and errors when close to the max chunk number limit

shaan1337:
Changed: Log warnings and errors when close to the max chunk number limit

### DIFF
--- a/src/EventStore.Core/TransactionLog/Chunks/TFChunkManager.cs
+++ b/src/EventStore.Core/TransactionLog/Chunks/TFChunkManager.cs
@@ -9,6 +9,9 @@ namespace EventStore.Core.TransactionLog.Chunks {
 	public class TFChunkManager : IDisposable {
 		private static readonly ILogger Log = Serilog.Log.ForContext<TFChunkManager>();
 
+		// MaxChunksCount is currently capped at 400,000 since:
+		// - the chunk file naming strategy supports only up to 6 digits for the chunk number.
+		// - this class uses a fixed size array to keep the chunk list
 		public const int MaxChunksCount = 400_000; 
 
 		public int ChunksCount {
--- a/src/EventStore.Core/TransactionLog/Chunks/TFChunkWriter.cs
+++ b/src/EventStore.Core/TransactionLog/Chunks/TFChunkWriter.cs
@@ -2,6 +2,8 @@
 using EventStore.Common.Utils;
 using EventStore.Core.TransactionLog.Checkpoint;
 using EventStore.Core.TransactionLog.LogRecords;
+using Serilog;
+using Serilog.Events;
 
 namespace EventStore.Core.TransactionLog.Chunks {
 	public class TFChunkWriter : ITransactionFileWriter {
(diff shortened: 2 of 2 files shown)


thefringeninja (dismissed):

timothycoleman:
A downside here is that the next limit is the filename pattern and before this pr the chunk limit served as a monitor for how close we were getting to that

We could
- raise the limit on the chunk naming too (but this could be more effort than it is really worth at the moment)
- maybe log errors when getting above a certain limit? and shutdown when above an even higher limit?
- other?

hayley-jean:
We should:
- Log a warning when creating a new chunk over 450,000. (Warn that the server will shut down at 500,000).
- Shut down the server with an error when the chunk count reaches 500,000

src/EventStore.Core/TransactionLog/Chunks/TFChunkWriter.cs
```diff
@@ -18,6 +19,13 @@ public class TFChunkWriter : ITransactionFileWriter {
 
 		private TFChunk.TFChunk _currentChunk;
 
+		// MaxChunkNumber is currently capped at 500,000 since the chunk file naming strategy supports only 6 digits for the chunk number.
+		// It can safely be increased up to 1,000,000 if required.
```
timothycoleman, line 23: 999,999 rather?

timothycoleman: ah nevermind, i see it is exclusive

src/EventStore.Core/TransactionLog/Chunks/TFChunkManager.cs
```diff
@@ -80,7 +81,7 @@ public class TFChunkManager : IDisposable {
 				chunkNum = chunk.ChunkHeader.ChunkStartNumber - 1;
 			}
 
-			for (int chunkNum = lastChunkToCache; chunkNum < _chunksCount;) {
+			for (int chunkNum = lastChunkToCache; chunkNum < _chunks.Count;) {
 				var chunk = _chunks[chunkNum];
```
timothycoleman, line 85: we are outside of the lock here, is it possible that there are threading problems if we happen to write while we read? if not, say, because of the specific way we are using the list, it would be worth a comment where the list is declared explaining

shaan1337:
As discussed, we're taking a more simplistic approach for now: only log warnings and errors when close to the limit.

The PR did not originally cater for the filename pattern limit mentioned by Tim here: https://github.com/EventStore/EventStore/pull/3643#issuecomment-1309984189.

Furthermore using the list is slightly more complex than initially expected due to the synchronization required on the list count (and one of the purposes of the PR was to simplify the code)

src/EventStore.Core/TransactionLog/Chunks/TFChunkManager.cs
shaan1337: well spotted, thanks Tim. As discussed, i've simplified the PR for now: https://github.com/EventStore/EventStore/pull/3643#issuecomment-1340724427

timothycoleman (approved):

thefringeninja (approved):

PR #3645, opened 2022-11-07
Improve new scavenge based on cloud testing

timothycoleman:
Added: Tuning new scavenge based on cloud tests

Performance
- Configurable page size
- Configurable hash users lru cache size
- ensure WAL stays small during Calculation by not having open reads while writing
- Cluster the HashUsers table by ROWID instead of hash to speed up inserts
- Connect to the db off of the main queue

Logging
- extra stats logging
- extra error logging
- log the cache size correctly when it is >2gb

### DIFF
--- a/src/EventStore.Core.XUnit.Tests/Scavenge/Infrastructure/Scenario.cs
+++ b/src/EventStore.Core.XUnit.Tests/Scavenge/Infrastructure/Scenario.cs
@@ -374,7 +374,7 @@ namespace EventStore.Core.XUnit.Tests.Scavenge {
 					index: calculatorIndexReader,
 					chunkSize: dbConfig.ChunkSize,
 					cancellationCheckPeriod: cancellationCheckPeriod,
-					checkpointPeriod: checkpointPeriod,
+					buffer: new(checkpointPeriod),
 					throttle: throttle);
 
 				IChunkExecutor<TStreamId> chunkExecutor = new ChunkExecutor<TStreamId, ILogRecord>(
--- a/src/EventStore.Core/TransactionLog/Scavenging/Sqlite/SqliteScavengeBackend.cs
+++ b/src/EventStore.Core/TransactionLog/Scavenging/Sqlite/SqliteScavengeBackend.cs
@@ -18,8 +18,11 @@ namespace EventStore.Core.TransactionLog.Scavenging.Sqlite {
 		//    persisted checkpoint.
 		private const string SqliteWalJournalMode = "wal";
 		private const int SqliteNormalSynchronousValue = 1;
+		private const int WalAutocheckpointDisabled = 0;
+
 		private const int DefaultSqliteCacheSize = 2 * 1024 * 1024;
-		private const int SqlitePageSize = 16 * 1024;
+		private const int DefaultSqlitePageSize = 16 * 1024;
+		private readonly int _pageSizeInBytes;
 		private readonly long _cacheSizeInBytes;
 		private SqliteBackend _sqliteBackend;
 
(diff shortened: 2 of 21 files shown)


src/EventStore.Core/TransactionLog/Scavenging/Sqlite/SqliteScavengeBackend.cs
```diff
@@ -18,8 +18,10 @@ public class SqliteScavengeBackend<TStreamId> : SqliteScavengeBackend, IScavenge
 		//    persisted checkpoint.
 		private const string SqliteWalJournalMode = "wal";
 		private const int SqliteNormalSynchronousValue = 1;
+		private const int WalAutocheckpointDisabled = 0;
+
 		private const int DefaultSqliteCacheSize = 2 * 1024 * 1024;
-		private const int SqlitePageSize = 16 * 1024;
+		private const int SqlitePageSize = 64 * 1024;
```
shaan1337, line 24: on ZFS, the default logical block size is 128 kb (recordsize parameter)

- could it be the reason why better performance was observed with 64kb on cloud? (and would 128kb work even better?)
- can it result in worse performance on other file systems when doing many updates? (with the standard 4KB logical block sizes)

src/EventStore.Core.XUnit.Tests/Scavenge/Infrastructure/Scenario.cs
```diff
@@ -374,7 +374,7 @@ public class Scenario<TLogFormat, TStreamId> : Scenario {
 					index: calculatorIndexReader,
 					chunkSize: dbConfig.ChunkSize,
 					cancellationCheckPeriod: cancellationCheckPeriod,
-					checkpointPeriod: checkpointPeriod,
+					buffer: new(checkpointPeriod),
```
shaan1337, line 377: is this parameter correct? :thinking:

shaan1337: clarified by Tim - not a typo

src/EventStore.Core/TransactionLog/Scavenging/Sqlite/SqliteScavengeBackend.cs
timothycoleman: yeah fair questions, decided to make this configurable with a default of 16k as it was before

shaan1337: thanks!

src/EventStore.Core/TransactionLog/Scavenging/Sqlite/SqliteScavengeBackend.cs
```diff
@@ -55,16 +62,19 @@ public class SqliteScavengeBackend<TStreamId> : SqliteScavengeBackend, IScavenge
 			var collisionStorage = new SqliteCollisionScavengeMap<TStreamId>();
 			CollisionStorage = collisionStorage;
 
-			var hashes = new SqliteScavengeMap<ulong, TStreamId>("HashUsers");
+			var hashes = new SqliteScavengeMap<ulong, TStreamId>("HashUsers", "INT");
```
shaan1337, line 65: is it because the "HashUsers" map is backed by a cache that we're able to use an autoincrementing rowid key here to speed up accumulation? otherwise, queries done during accumulation would be slower.

shaan1337 (approved):
Looks good :+1: 

Just noticed something (which isn't serious but could help improve usability): if someone's running with an invalid scavenge configuration, they will learn about it only when the next scavenge is started and not on startup. for example, if someone changes the sqlite page size and they start up EventStoreDB (with an existing scavenge.db), no error will be printed until the next scavenge is started. however, they can still delete the scavenging directory without the need to restart the database.

pvanbuijtene (approved):

PR #3646, opened 2022-11-07
Fix edge cases in maxage read fast path

timothycoleman:
Fixed: Edge cases in maxage read fast path.

These could result in the nextEventNumber being set incorrectly, leading to gRPC reads going into an infinite loop reading from e.g. 5 then 7 then 5, looking for events, for as long as the client keept the gRPC call open.

Two places relied on the index entries not having gaps in the version numbers for a particular streamhash, but it is possible for this to happen:

Old scavenge:
- During the course of a regular index scavenge, since it starts with the newest tables and works backwards
- If you scavenge starting from a chunk > 0, removing some records, then scavenge the index, it removes from the index whatever is no longer present in the log
- If there is a hash collision between two streams and the union of version numbers has gaps in

New scavenge:
- If the stream has a maybeDiscardPoint due to maxage, keeping some entries in the index that are not in the log, but some later entries are not kept because the node is restarted while they are still in the memtables

### DIFF
--- a/src/EventStore.Core/Services/Storage/ReaderIndex/IndexReader.cs
+++ b/src/EventStore.Core/Services/Storage/ReaderIndex/IndexReader.cs
@@ -314,10 +314,11 @@ namespace EventStore.Core.Services.Storage.ReaderIndex {
 				//involve an additional set of reads for no good reason
 				while (indexEntries.Count == 0) {
 					// we didn't find any indexEntries, and we already early returned in the case that startEventNumber > lastEventNumber
-					// so we must be reading before the oldest entry for this stream hash.
+					// so we want to find the oldest entry that is after startEventNumber.
+					// startEventNumber may be before the start of the stream, or it may be in a gap caused by scavenging.
 					// this will generally only iterate once, unless a scavenge completes exactly now, in which case it might iterate twice
-					if (tableIndex.TryGetOldestEntry(streamId, out var oldest)) {
-						startEventNumber = oldest.Version;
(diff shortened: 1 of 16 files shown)


thefringeninja (dismissed):

src/EventStore.Core/Services/Storage/ReaderIndex/IndexReader.cs
```diff
@@ -380,6 +381,16 @@ public class IndexReader<TStreamId> : IndexReader, IIndexReader<TStreamId> {
 					indexEntries = tableIndex.GetRange(streamId, mid, mid + maxCount - 1);
 					if (indexEntries.Count > 0) {
 						nextEventNumber = indexEntries[0].Version + 1;
+					} else {
```
shaan1337, line 384: the comment above needs adjustment from `two` to `three`:
... `two` explicit continue cases
... the `two` continue cases always narrow

shaan1337 (approved):
looks good, just found some minor missing adjustments to a comment

src/EventStore.Core/Services/Storage/ReaderIndex/IndexReader.cs
shaan1337: (or just remove `two`)

timothycoleman: thanks, done!

thefringeninja (approved):

PR #3655, opened 2022-11-10
Fix projection progress report.

### LINKED ISSUE #2820: Projections reports their % done incorrectly in the UI
**Describe the bug**
Projections will report its percentage % completion of events processed wrong and if you write an event, then the projections will trigger and go to 100%, but without it, the projections don't do that recalculation.

**To Reproduce**
1. Start a EventStoreDB 20.10.2 node 
2. Enable all projections to see incorrect % done status

**Expected behavior**
 Process events and calculate percentage done correctly.  

**Actual behavior**

 Projection stats UI showing arbitrary numbers % done 

**Config/Logs/Screenshots**

image
**EventStore details**
 - EventStore server version: 
 
 Event Store 20.10.2.0

 - Operating system: Mac/ Linux /windows

 - EventStore client version (if applicable):
 
 N/A

**Additional context**
N/A

YoEight:
Fixed: Fix projection progress report.

The issue was about how `ReaderSubscriptionBase` reported progression. `ReaderSubscriptionBase` uses a time window to decide if we should report progress downstream. If that time windows was missed, progression was not reported.

The new solution reports progression asynchronously, every 500ms, to match previous time window.

Fixes #2820

### DIFF
--- a/src/EventStore.Projections.Core/Messages/ReaderSubscriptionMessage.cs
+++ b/src/EventStore.Projections.Core/Messages/ReaderSubscriptionMessage.cs
@@ -273,7 +273,7 @@ namespace EventStore.Projections.Core.Messages {
 				get { return TypeId; }
 			}
 
-			public ReportProgress(Guid correlationId, object source = null) : base(correlation, null, source) { }
+			public ReportProgress(Guid correlationId, object source = null) : base(correlationId, null, source) { }
 		}
 	}
 }
--- a/src/EventStore.Projections.Core/Messages/ReaderSubscriptionMessage.cs
+++ b/src/EventStore.Projections.Core/Messages/ReaderSubscriptionMessage.cs
@@ -216,7 +216,7 @@ namespace EventStore.Projections.Core.Messages {
 			private readonly long? _safeTransactionFileReaderJoinPosition;
 			private readonly float _progress;
 
-			//NOTE: committed event with null event _data means - end of the source reached.  
+			//NOTE: committed event with null event _data means - end of the source reached.
 			// Current last available TF commit position is in _position.CommitPosition
 			// TODO: separate message?
 
@@ -264,6 +264,16 @@ namespace EventStore.Projections.Core.Messages {
 			public string Reason {
 				get { return _reason; }
 			}
+		}
+
+		public class ReportProgress : SubscriptionMessage {
+			private static readonly int TypeId = System.Threading.Interlocked.Increment(ref NextMsgId);
+
+			public override int MsgTypeId {
+				get { return TypeId; }
+			}
+
+			public ReportProgress(Guid correlation, object source = null) : base(correlation, null, source) { }
 		}
 	}
 }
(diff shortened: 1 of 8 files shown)


src/EventStore.Projections.Core/Messages/ReaderSubscriptionMessage.cs
```diff
@@ -265,5 +265,15 @@ public class Faulted : SubscriptionMessage {
...
 		}
+
+		public class ReportProgress : SubscriptionMessage {
+			private static readonly int TypeId = System.Threading.Interlocked.Increment(ref NextMsgId);
+
+			public override int MsgTypeId {
+				get { return TypeId; }
+			}
+
+			public ReportProgress() : base(Guid.NewGuid(), null, null) { }
```
thefringeninja, line 276: IMO we should follow the convention established by the other implementations of `SubscriptionMessage`. `correlationId` and an optional `source` should come from the constructor.

src/EventStore.Projections.Core/Services/Processing/EventReaderCoreService.cs
```diff
@@ -243,7 +245,17 @@ public class EventReaderCoreService :
...
-		
+		public void Handle(ReaderSubscriptionMessage.ReportProgress message) {
+			if (_stopped)
+				return;
+
+			foreach (var subscription in _subscriptions.Values) {
+				subscription.Handle(message);
+			}
+
+			_publisher.Publish(TimerMessage.Schedule.Create(TimeSpan.FromMilliseconds(500), new SendToThisEnvelope(this), new ReaderSubscriptionMessage.ReportProgress()));
```
thefringeninja, line 256: See above comment. These messages should share the same correlation id

thefringeninja: Also, should we be creating a new envelope every time?

thefringeninja (changes requested):

src/EventStore.Projections.Core/Messages/ReaderSubscriptionMessage.cs
YoEight: Done

src/EventStore.Projections.Core/Services/Processing/EventReaderCoreService.cs
YoEight: Done

hayley-jean (changes requested):
This is reporting the progress better, but it still doesn't dependably reach 100%

The projection only updates its progress when it handles an event that passes its filter. If there are a large number of events that don't pass the projection's filters at the end of the stream/$all then the progress will get stuck again.

Probably the easiest way to test this is to write a bunch of system events at the end of the log. The projections will report the progress correctly while the events are being written, but if you restart the node afterwards, the progress will be incorrect again.

Setting the progress to 100% once the reader receives the `EventReaderIdle` message might be the simplest way to solve this, since the reader should be caught up when it goes idle. However, this needs to be tested in case the readers also enter idle if it's paused to wait for a checkpoint.

hayley-jean (approved):

thefringeninja (approved):

PR #3656, opened 2022-11-11
Scavenging Quality of Life improvements

timothycoleman:
Added: Improvements to Scavenge HTTP API (query if scavenge is running, stop any running scavenge)
Added: ScavengeId to log context

HTTP API:
- `GET` the current scavenge (id and link for deleting) if any, from `/admin/scavenge/current`
- `DELETE` `/admin/scavenge/current` to stop the current scavenge regardless of its id.
- Stopping a scavenge now waits for the scavenge to stop before responding to the request

Logging:
- Add `ScavengeId` to the log context


dhingrak:
Made calls in Postman.
GET  /admin/scavenge/current and DELETE /admin/scavenge/current endpoint messages are working fine. One question, why we are returning the ScavengeId even if the DELETE endpoint returns 404.
For example:- /admin/scavenge/testing returns the current scavengeId with 404.

timothycoleman:
Thanks @dhingrak. I think it was like that from before. To be honest I hadn't noticed - so there was a way to query the current scavenge id after all! but at least after this pr it is more obvious

dhingrak (approved):

hayley-jean (approved):

PR #3668, opened 2022-11-25
Revert sqlite patch

pvanbuijtene:
Changed: Revert the Sqlite patch, will be fixed in build


timothycoleman (approved):

shaan1337 (approved):

timothycoleman:
For posterity, the patch was reverted because it caused dependency problems in the mono packaging process. Instead for V5 we left the nuget packages the same and dropped in the newer sqlite native library during the packaging process

PR #3671, opened 2022-12-05
Replace google.rpc.Status

intale:
# Original problem

When using protos to generate a code using `google-protobuf` - collisions happen when using that code in one project along with other libraries which define `google.rpc.Status` and `google.rpc.Code` objects. Such collisions make it impossible to use ES-generated code and another libraries together. Example of such error:

```
Google::Protobuf::TypeError: Unable to build file to DescriptorPool: duplicate symbol 'google.rpc.Status'
```

Simply saying, if one lib defines `google.rpc.Status`, then, when another lib tries to do the same - error raises. Of course, another libraries should follow this principle as well - by defining those objects in their own namespace. But, at least, we can make sure that EventStore protos are not affected by this issue.

# Solution

Replace `google.rpc.Status` and `google.rpc.Code` objects with `event_store.client.Status` and `event_store.client.Code` objects. This way we are simply keeping all ES-related objects under our own namespace.


intale:
@thefringeninja hey Joao. It seems there is some issue with tests. Unfortunately, my skills in C# end here - I have no idea how to fix that. Could you please take a look at it when you would have free time? Attaching tests results.

[test-results.zip](https://github.com/EventStore/EventStore/files/10210719/test-results.zip)

intale:
Message from João in discord:
```
so i discussed this with the team... the change is too destructive as it would require changing too much
what we recommend is simply copying all the proto files by hand when they change, which isn't that often. we are taking great pains to maintain backwards (and forwards when possible) compatibility
```

Closing this PR as it is not going to be fixed and merged

PR #3672, opened 2022-12-05
Add `AllowUnknownOptions` configuration option

timothycoleman:
Note this works in a slightly different way to the equivalent change in master/21.10/22.10 because of the old Rags processing
Note also that 20.10 is out of support now


timothycoleman:
Fortunately this isn't needed in the end

PR #3674, opened 2022-12-06
TFChunk disposal synchronization improvements

shaan1337:
Fixed: Improved synchronization during TFChunk disposal

These are some small synchronization issues I've found in TFChunk disposal while implementing https://github.com/EventStore/EventStore/pull/3669

The most important one, which might have an impact in practice, is: b91a6c432fb68777603044d013ee169fa6657c13. From my understanding, writing to a `volatile` variable is equivalent to adding a barrier before the write ("release semantics" - https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/classes#1454-volatile-fields), not after it.

Regarding 7f2ecca1048a30cd5df0205bf639217e7c40d70d, cleaning up twice may not actually have an impact in practice but the change makes the code more correct. A simpler alternative could be to make sure that CleanUpFileStreamDestruction() and FreeCachedData() are idempotent but it would require explaining it clearly in the comments.

39a9a9138d6a87356e7701204e7bcdcc19f751e5 does not have a big impact in practice since:
1. it applies only to mem db
2. the call to TFChunk.CompleteScavenge() and TFChunk.Dispose()/TryClose() are probably mutually exclusive in practice

but i added the change mainly for the sake of completeness.

### DIFF
--- a/src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
+++ b/src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
@@ -934,6 +934,9 @@ namespace EventStore.Core.TransactionLog.Chunks.TFChunk {
 
 		public bool TryClose() {
 			_selfdestructin54321 = true;
+
+			Thread.MemoryBarrier();
+
 			bool closed = true;
 			closed &= TryDestructFileStreams();
 			closed &= TryDestructMemStreams();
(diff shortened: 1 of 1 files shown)


pvanbuijtene (dismissed):

timothycoleman:
I've rebased and resolved the (trivial) conflict.

I'm not totally convinced these changes are all necessary, but I don't think they do any harm and I expect this whole area will become quite a lot simpler when we can switch to net6 style random io

pvanbuijtene (approved):

timothycoleman (approved):

PR #3684, opened 2022-12-15
Use a source generator to generate the dynamic ids for message types

timothycoleman:
Added: Source generator for dynamic message type ids

This saves on boilerplate code when adding new messages, but is also anticipated to be of use with the enhanced telemetry work


src/EventStore.SourceGenerators/Messaging/Attributes.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+
+namespace EventStore.SourceGenerators.Messaging {
+
+	public class RootMessageAttribute : Attribute {
```
gfoidl, line 5: From a consumption perspective it's better to have the attributes defined in the "main"-code, not generated.

E.g. when you start a new project, then there's no attribute before the generator creates it. That can cause troubles with VS (even today).
Thus the common pattern (at least in dotnet-org) seems to be to have the needed attributes be parts of the shipped binaries, so the generator just "consumes" them.

src/EventStore.SourceGenerators/Messaging/StatsMessageSourceGenerator.cs
```diff
@@ -0,0 +1,279 @@
...
+using Microsoft.CodeAnalysis.Text;
+
+// for speed and simplicity this does all its generation based on the syntax tree and not on the semantic model
+// there are essentially three phases
+// 1. use a SyntaxReceiver to detect and record uses of the attributes that we will generate code for
+// 2. Register each usage, discovering how they are related to each other in the abstract syntax tree bottom-up
+// 3. Process the resulting tree top-down, adding in the extra generated code that we want.
+namespace EventStore.SourceGenerators.Messaging {
+	[Generator]
+	public class StatsMessageSourceGenerator : ISourceGenerator {
```
gfoidl, line 18: The generator should use [incremental version](https://github.com/dotnet/roslyn/blob/main/docs/features/incremental-generators.md).

The `ISourceGenerator` runs -- put simply -- on every keystroke, thus can slow down dev-experience on medium / large projects significantly. The incremental generator is more advanced, so that only on actual relevant change it's active. IMO also the programming model is more natural and aligned to how compilers work (parse and emit is more separated).

src/EventStore.SourceGenerators/Messaging/Attributes.cs
timothycoleman: thanks! ive adjusted this

src/EventStore.SourceGenerators/Messaging/StatsMessageSourceGenerator.cs
timothycoleman: thanks i'll take a look

thefringeninja (approved):

pvanbuijtene (approved):

PR #3695, opened 2023-01-06
Add duration histograms for gRPC reads and writes

timothycoleman:
Added: Histograms for gRPC reads and appends

The changes in the gRPC services are only wrapping existing code with the timing mechanism. No changes to the logic.


timothycoleman:
image

YoEight (approved):
LGTM

hayley-jean (approved):

PR #3701, opened 2023-01-19
Improve handling of slow persistent subscriptions consumers

hayley-jean:
Fixed: Slow persistent subscription consumer no longer slows down other subscribers

Fixes https://github.com/EventStore/home/issues/885
Fixes https://github.com/EventStore/home/issues/830

Set the gRPC Persistent Subscription channel to `DropWrite` when handling `StreamEventAppeared` messages and the channel is full. The dropped messages will timeout and be retried by the `PersistentSubscriptionsService`.

This prevents the Persistent Subscriptions thread from getting locked waiting for the channel to empty when pushing new messages to gRPC clients.


timothycoleman (approved):
fab :+1:

timothycoleman:
although, we could consider a more user-facing changelog message here, something like

Fixed: slow persistent subscription consumer no longer slows down other subscribers

pvanbuijtene (approved):

PR #3702, opened 2023-01-19
Ensure persistent subscriptions messages don't continue on the main queue

hayley-jean:
Changed: Publish messages from the persistent subscriptions IODispatcher to the Persistent Subscriptions queue rather than the main queue

This prevents continuations which could take a long time to process from happening on the main queue.

Fixes https://github.com/EventStore/home/issues/885


timothycoleman (approved):
great :+1:

There is only one other IODispatcher that runs its continuations on the main queue, but it seems not to actually be used at all. For completeness I have removed it here https://github.com/EventStore/EventStore/pull/3709

pvanbuijtene (approved):

PR #3703, opened 2023-01-19
Use a stopwatch instead of DateTime.UtcNow for durations

timothycoleman:
Added: a process-wide stopwatch for measuring durations

- The stopwatch won't be affected by system clock changes
- The stopwatch is 1/3 faster
- The stopwatch can be used by other parts of the system that need to time things, too

### DIFF
--- a/src/EventStore.Core/Telemetry/Instant.cs
+++ b/src/EventStore.Core/Telemetry/Instant.cs
@@ -1,0 +1,36 @@
+﻿using System;
+using System.Diagnostics;
+
+namespace EventStore.Core.Telemetry;
+
+// this provides stronger typing than just passing a long representing the number of ticks
+// and provides us a place to change the resolution and size if long ticks is overkill.
+public struct Instant {
+	private static readonly double _secondsPerTick = 1 / (double)Stopwatch.Frequency;
+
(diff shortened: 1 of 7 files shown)


src/EventStore.Core/Telemetry/Instant.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Diagnostics;
+
+namespace EventStore.Core.Telemetry;
+
+// this provides stronger typing than just passing a long representing the number of ticks
+// and provides us a place to change the resolution and size if long ticks is overkill.
+public struct Instant {
+	private static readonly double _secondsPerTick = 1 / (double)Stopwatch.Frequency;
+
+	private static readonly Stopwatch _stopwatch = Stopwatch.StartNew();
```
gfoidl, line 11: Instead of the `static readonly`-field you could use `Stopwatch.GetTimestamp()` (the static method on Stopwatch).
* avoids the fields access
* the other methods on SW also call `GetTimestamp` internally -- so should be minimally faster

timothycoleman: oh yeah, thanks!

thefringeninja (approved):

pvanbuijtene (approved):

PR #3713, opened 2023-01-26
Implement external redaction APIs

shaan1337:


src/EventStore.Core/Synchronization/IExclusiveLock.cs
```diff
@@ -0,0 +1,6 @@
+namespace EventStore.Core.Synchronization;
+
+public interface IExclusiveLock {
```
timothycoleman, line 3: I could be persuaded, but is this abstraction helping - I suspect that the different implementations would have have quite different properties and not really be substitutable i.e. the consumer would likely be relying on it being a semaphoreslim in particular

src/EventStore.Core/Services/RedactionService.cs
```diff
@@ -0,0 +1,244 @@
...
+		public void Handle(RedactionMessage.SwitchChunkUnlock message) {
+			if (_switchChunksLock.TryRelease())
+				message.Envelope.ReplyWith(
+					new RedactionMessage.SwitchChunkUnlockCompleted(SwitchChunkUnlockResult.Success));
+			else
+				message.Envelope.ReplyWith(
+					new RedactionMessage.SwitchChunkUnlockCompleted(SwitchChunkUnlockResult.Failed));
+		}
+
+		public void Handle(RedactionMessage.SwitchChunk message) {
```
timothycoleman, line 106: it would probably be good if the service ensured that the lock was held by whomever is sending the request.

perhaps the grpc call should generate an id (guid) and include it in lock/switch/unlock messages. the redaction service could then check that the lock is held and by the right person when a switchchunk message is received, and it might help in the grpccall itself to be able to 'finally unlock' safely even if the lock was not successfully acquired

src/EventStore.Core/Services/RedactionService.cs
```diff
@@ -0,0 +1,244 @@
...
+			ThreadPool.QueueUserWorkItem(_ => {
+				try {
```
timothycoleman, line 108: if we had the lock at the time this was scheduled, is it possible that we released it before this actually runs

im not certain, but it might be sensible to give the redaction service its own queue (channel?) similar to the writerservice so that it can just be conceptually single threaded and process off of the main bus. then the lock/switch/unlock steps would be performed strictly one at a time

src/EventStore.Core/Services/RedactionService.cs
```diff
@@ -0,0 +1,244 @@
...
+				} catch (Exception ex) {
+					Log.Error(ex, "An error has occurred when trying to switch chunk: {targetChunk} with chunk: {newChunk}.",
+						message.TargetChunkFile, message.NewChunkFile);
+					message.Envelope.ReplyWith(
+						new RedactionMessage.SwitchChunkCompleted(SwitchChunkResult.UnexpectedError));
+				}
+			});
+		}
+
+		private void SwitchChunk(string targetChunkFile, string newChunkFile, IEnvelope envelope) {
```
timothycoleman, line 119: it's just slightly tricky here to be sure that we dont reply twice (i wonder if we have (or should have) some coding standard around whether envelopes should deal with multiple replies or whether the person replying should be responsible for not replying more than once (unless they are supposed to). the actual implementation of the envelope calls Set rather than TrySet on the tcs so it will be sad if we do reply more than once currently

src/EventStore.Core/Services/Transport/Grpc/Redaction.SwitchChunk.cs
```diff
@@ -0,0 +1,94 @@
...
+
+		private async Task SwitchChunksLock() {
+			var lockTcs = new TaskCompletionSource<RedactionMessage.SwitchChunkLockCompleted>();
+			_bus.Publish(new RedactionMessage.SwitchChunkLock(
+				new CallbackEnvelope(msg => lockTcs.SetResult(msg as RedactionMessage.SwitchChunkLockCompleted))
+			));
+
+			var completionMsg = await lockTcs.Task.ConfigureAwait(false);
+
+			// it's important not to throw any exceptions beyond this point in this method
```
timothycoleman, line 39: if we make it so that we can always safely send the unlock message even if we dont hold the lock then put the 'SwitchChunksLock' call above inside the try, and then not worry about exceptions being thrown here

src/EventStore.Core/Services/Transport/Grpc/Redaction.SwitchChunk.cs
```diff
@@ -0,0 +1,94 @@
...
+			// (except to indicate that acquiring the lock failed), otherwise we can end up
+			// in a situation where the lock remains permanently acquired.
+
+			if (completionMsg is null) // should never happen
+				return;
```
timothycoleman, line 44: probably throw an exception

src/EventStore.Core/Services/Transport/Grpc/Redaction.SwitchChunk.cs
```diff
@@ -0,0 +1,94 @@
...
+				throw RpcExceptions.RedactionLockFailed();
+		}
+
+		private async Task SwitchChunks(
+			ServerCallContext context,
+			IAsyncStreamReader<SwitchChunkReq> requestStream,
+			IServerStreamWriter<SwitchChunkResp> responseStream) {
+
+			// write empty headers to indicate that acquiring the lock was successful
+			await context.WriteResponseHeadersAsync(new Metadata()).ConfigureAwait(false);
```
timothycoleman, line 57: i dont mind particularly, but our of curiosity is this better as a header than it would have been as a message?

src/EventStore.Core/Services/Transport/Grpc/Redaction.SwitchChunk.cs
```diff
@@ -0,0 +1,94 @@
...
+				await SwitchChunks(context, requestStream, responseStream).ConfigureAwait(false);
+			} finally {
+				await SwitchChunksUnlock().ConfigureAwait(false);
+			}
+		}
+
+		private async Task SwitchChunksLock() {
+			var lockTcs = new TaskCompletionSource<RedactionMessage.SwitchChunkLockCompleted>();
+			_bus.Publish(new RedactionMessage.SwitchChunkLock(
+				new CallbackEnvelope(msg => lockTcs.SetResult(msg as RedactionMessage.SwitchChunkLockCompleted))
```
timothycoleman, line 34: maybe we could do with a TcsEnvelope that creates the tcs itself and knows how to set it on reply

src/EventStore.Core/Services/Transport/Grpc/Redaction.SwitchChunk.cs
```diff
@@ -0,0 +1,94 @@
...
+			await SwitchChunksLock().ConfigureAwait(false);
+			try {
+				await SwitchChunks(context, requestStream, responseStream).ConfigureAwait(false);
+			} finally {
+				await SwitchChunksUnlock().ConfigureAwait(false);
+			}
+		}
+
+		private async Task SwitchChunksLock() {
+			var lockTcs = new TaskCompletionSource<RedactionMessage.SwitchChunkLockCompleted>();
```
timothycoleman, line 32: TaskCreationOptions.RunContinuationsAsynchronously

src/EventStore.Core/Synchronization/IExclusiveLock.cs
shaan1337: yup, good point, i'll remove the interface

src/EventStore.Core/Services/RedactionService.cs
shaan1337: good idea :+1:

src/EventStore.Core/Services/RedactionService.cs
```diff
@@ -0,0 +1,244 @@
...
+			if (!IsValidSwitchChunkRequest(targetChunkFile, newChunkFile, out var newChunk, out var failReason)) {
+				envelope.ReplyWith(new RedactionMessage.SwitchChunkCompleted(failReason));
+				return;
+			}
+
+			_db.Manager.SwitchChunk(
```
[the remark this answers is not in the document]
shaan1337: yes, that's correct.

it'll be caught at a higher level here and the reply UnexpectedError will be returned to the client:
https://github.com/EventStore/EventStore/blob/838e840df29eb78e4bc14fe4165977d131b95096/src/EventStore.Core/Services/RedactionService.cs#L110

src/EventStore.Core/Services/Transport/Grpc/Redaction.SwitchChunk.cs
shaan1337: ok :+1: i was initially worried that it could unlock when not holding the lock but the GUID thing will prevent that from happening.

src/EventStore.Core/Services/Transport/Grpc/Redaction.SwitchChunk.cs
shaan1337: :+1:

src/EventStore.Core/Services/Transport/Grpc/Redaction.SwitchChunk.cs
shaan1337: actually, the comment may be a bit misleading:
to be able to throw an RPC exception before any message has been sent on the channel, we need to send the headers (that will contain the exception details in the metadata)

i'll rephrase it.

i like it mainly for its simplicity:
- the proto file is simpler
- locking is a pre-requisite to be able to switch chunks so, using the headers fits this purpose well

src/EventStore.Core/Services/Transport/Grpc/Redaction.SwitchChunk.cs
shaan1337: :+1:

src/EventStore.Core/Services/Transport/Grpc/Redaction.SwitchChunk.cs
shaan1337: ok, i'll give this idea a try

src/EventStore.Core/Services/RedactionService.cs
shaan1337: hmm yup true, it would help make the code easier to read

for this case, an easy way to reason about it is that the envelope reply is the last line of code that's executed before the function returns from the thread created in `Handle(RedactionMessage.SwitchChunk message)`

src/EventStore.Core/Services/RedactionService.cs
shaan1337: i'm not sure i followed, do you mean:
- lock acquired
- chunk switch has started (but envelope not yet replied)
- lock released due to gRPC heartbeat timeout (e.g if the redactor app exits)
- chunk switch completes, replying on the envelope

in this case, it seems that it wouldn't cause an issue on the gRPC side, since the tcs would be set but nothing would be awaiting it any more

src/EventStore.Core/Synchronization/IExclusiveLock.cs
shaan1337: fixed

shaan1337: fixed in: d39ec1679772c86b11779b9e5c5ef07c4e0ff630

src/EventStore.Core/Services/Transport/Grpc/Redaction.SwitchChunk.cs
shaan1337: fixed in: https://github.com/EventStore/EventStore/pull/3713/commits/1f7be5d0cfc215e2c3c0cb0a511e01f31d5071c8

src/EventStore.Core/Services/Transport/Grpc/Redaction.SwitchChunk.cs
shaan1337: fixed in 1f7be5d0cfc215e2c3c0cb0a511e01f31d5071c8 but eventually overridden by TcsEnvelope changes (that throws if the resulting message type is not correct)

src/EventStore.Core/Services/Transport/Grpc/Redaction.SwitchChunk.cs
shaan1337: fixed in: 1bc7a6c4dfc0d37d6db9f204654b89ccbd5487eb

src/EventStore.Core/Services/Transport/Grpc/Redaction.SwitchChunk.cs
shaan1337: fixed in TcsEnvelope changes: 1bc7a6c4dfc0d37d6db9f204654b89ccbd5487eb

src/EventStore.Core/Services/Transport/Grpc/Redaction.SwitchChunk.cs
shaan1337: the headers are now written in the SwitchChunksLock() method since it's better placed there and the comment has been updated (it wasn't possible to put it there previously since it could thrown an exception): https://github.com/EventStore/EventStore/pull/3713/commits/1f7be5d0cfc215e2c3c0cb0a511e01f31d5071c8

src/EventStore.Core/Services/RedactionService.cs
shaan1337: fixed in: 56fadc4f319ef10d126217f138452cb3e4d809ef. for some reason, the queue stats don't seem to work, i'm looking into it.

src/EventStore.Core/Services/RedactionService.cs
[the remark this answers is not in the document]
timothycoleman: oh right, thank i missed that :+1:

src/EventStore.Core/Services/Transport/Grpc/Redaction.SwitchChunk.cs
timothycoleman: makes sense i think, im just making this unresolved for the minute to remind me to come back to it in a bit

timothycoleman:
I pushed some suggestions to https://github.com/EventStore/EventStore/tree/simple-redaction-suggestions that you can encorporate if you wish

src/EventStore.ClusterNode/Program.cs
```diff
@@ -231,6 +228,44 @@ internal static class Program {
...
 
+		private static void ConfigureHttpOptions(ListenOptions listenOptions, ClusterVNodeHostedService hostedService, bool useHttps) {
+			if (useHttps)
+				listenOptions.UseHttps(CreateServerOptionsSelectionCallback(hostedService), null);
+			else
+				listenOptions.Use(next =>
+					new ClearTextHttpMultiplexingMiddleware(next).OnConnectAsync);
+		}
+
+		private static void TryListenOnUnixSocket(ClusterVNodeHostedService hostedService, KestrelServerOptions server) {
```
timothycoleman, line 239: what security concerns do we potentially have here
connections from other local processes that aren't the eventstore user?
maybe its ok because its in the db directory, i wonder how it compares to the way we set permissions for checkpoints and chunks etc.

src/EventStore.Core/Data/Redaction/GetEventPositionResult.cs
```diff
@@ -0,0 +1,15 @@
+namespace EventStore.Core.Data.Redaction {
+	public enum GetEventPositionResult {
+		Success = 0,
```
timothycoleman, line 3: Probably worth having a 'None' as the zero value to avoid mistaking uninitialized for success at some point. Same for the other enums in the same folder

src/EventStore.Core/Messaging/TcsEnvelope.cs
```diff
@@ -0,0 +1,20 @@
...
+
+		public TcsEnvelope() {
+			_tcs = new TaskCompletionSource<TResult>(TaskCreationOptions.RunContinuationsAsynchronously);
+		}
+
+		public void ReplyWith<T>(T message) where T : Message {
+			if (message is not TResult result)
+				throw new ArgumentException("Unexpected message type.", nameof(message));
+
+			_tcs.SetResult(result);
```
timothycoleman, line 17:
```suggestion
			_tcs.TrySetResult(result);
```

perhaps better to use TrySetResult, after all it isn't the fault of the thing replying if something else has also subscribed to the message and already replied

src/EventStore.Core/Messaging/TcsEnvelope.cs
```diff
@@ -0,0 +1,20 @@
...
+	public class TcsEnvelope<TResult> : IEnvelope where TResult : class {
+		private readonly TaskCompletionSource<TResult> _tcs;
+		public Task<TResult> Task => _tcs.Task;
+
+		public TcsEnvelope() {
+			_tcs = new TaskCompletionSource<TResult>(TaskCreationOptions.RunContinuationsAsynchronously);
+		}
+
+		public void ReplyWith<T>(T message) where T : Message {
+			if (message is not TResult result)
```
timothycoleman, line 14: we it intentional to move the exception to being thrown on the one sending the reply? we could consider TrySetException

src/EventStore.Core/Services/RedactionService.cs
```diff
@@ -0,0 +1,261 @@
...
+			IQueuedHandler queuedHandler,
+			TFChunkDb db,
+			IReadIndex<TStreamId> readIndex,
+			SemaphoreSlimLock switchChunksLock) {
+			Ensure.NotNull(queuedHandler, nameof(queuedHandler));
+			Ensure.NotNull(db, nameof(db));
+			Ensure.NotNull(readIndex, nameof(readIndex));
+			Ensure.NotNull(switchChunksLock, nameof(switchChunksLock));
+
+			_queuedHandler = queuedHandler;
```
timothycoleman, line 41: perhaps don't need to pass this in here solely to stop it on shutdown. since the bootstrapper is doing the subscriptions for us it could probably also take care of shutting down the queue too?

src/EventStore.Core/Services/RedactionService.cs
```diff
@@ -0,0 +1,261 @@
...
+				return false;
+			}
+
+			if (!newChunkFooter.IsCompleted) {
+				failReason = SwitchChunkResult.NewChunkNotCompleted;
+				return false;
+			}
+
+			try {
+				newChunk = TFChunk.FromCompletedFile(
```
timothycoleman, line 234: could add a comment saying that the chunkmanager will re-open the file with its own settings. when i first saw this i was worried that this TFChunk would be added to the manager with a maxreadercount of 1

src/EventStore.Core/Services/RedactionService.cs
```diff
@@ -0,0 +1,261 @@
...
+					optimizeReadSideCache: false,
+					reduceFileCachePressure: true);
+			} catch (HashValidationException) {
+				failReason = SwitchChunkResult.NewChunkHashInvalid;
+				return false;
+			} catch {
+				failReason = SwitchChunkResult.NewChunkOpenFailed;
+				return false;
+			} finally {
+				newChunk?.Dispose();
```
timothycoleman, line 249: is this intentional?

src/Protos/Grpc/redaction.proto
```diff
@@ -0,0 +1,39 @@
...
+message GetEventPositionResp {
+	repeated EventPosition event_positions = 1;
+}
+
+message EventPosition {
+	uint64 log_position = 1;
+	ChunkInfo chunk_info = 2;
+}
+
+message ChunkInfo {
```
timothycoleman, line 26: bumping a little on this being called 'ChunkInfo' but then having an event specific offset in it. would it be senseible to just encororate these fields directly into the eventposition?

src/Protos/Grpc/redaction.proto
```diff
@@ -0,0 +1,39 @@
+syntax = "proto3";
+package event_store.client.redaction;
+option java_package = "com.eventstore.dbclient.proto.redaction";
+
+import "shared.proto";
+
+service Redaction {
```
timothycoleman, line 7: i assume it is intended that the caller will open the switchchunks call to acquire the lock before sending any GetEventPositionReqs, might be worth a quick comment here to explain the intended usage

timothycoleman (changes requested):
pretty straight forward comments in line, but have a look at the suggestions branch before those

src/EventStore.Core/Services/Storage/StorageScavenger.cs
```diff
@@ -120,6 +132,17 @@ public class StorageScavenger :
...
+			lock (_lock) {
+				switchChunksLockId = _switchChunksLockId;
+			}
+
+			try {
+				_switchChunksLock.TryRelease(switchChunksLockId);
+			} catch (Exception ex) {
+				logger.Error(ex, "SCAVENGING: Unexpected error when releasing the chunk switch lock.");
+			}
+
```
timothycoleman, line 145: I haven't thought about this carefully, but is there a reason to acquire and release the _lock and then acquire it again below? possibly we could just hold the _lock

src/EventStore.Core/Services/RedactionService.cs
```diff
@@ -0,0 +1,261 @@
...
+			var streamId = _readIndex.GetStreamId(streamName);
+			var result = _readIndex.ReadEventInfo_KeepDuplicates(streamId, eventNumber);
+
+			var eventPositions = new EventPosition[result.EventInfos.Length];
+
+			for (int i = 0; i < result.EventInfos.Length; i++) {
+				var eventInfo = result.EventInfos[i];
+				var logPos = eventInfo.LogPosition;
+				var chunk = _db.Manager.GetChunkFor(logPos);
+				var localPosition = chunk.ChunkHeader.GetLocalLogPosition(logPos);
```
timothycoleman, line 68: not certain, but do we need to increase the maxreadercount for this

src/EventStore.Core/Data/Redaction/GetEventPositionResult.cs
shaan1337: fixed :+1:

src/EventStore.Core/Messaging/TcsEnvelope.cs
shaan1337: fixed :+1:

src/EventStore.Core/Messaging/TcsEnvelope.cs
shaan1337: fixed :+1:

src/EventStore.Core/Services/RedactionService.cs
shaan1337: fixed :+1:

src/Protos/Grpc/redaction.proto
shaan1337: fixed :+1:

src/EventStore.Core/Services/Storage/StorageScavenger.cs
shaan1337: yes, the main reason is for code readability - i think that it makes the critical section(s) much clearer compared to taking the lock on large blocks of code. also it avoids holding the lock for longer than is actually required (although performance's not really critical here)

src/EventStore.Core/Services/RedactionService.cs
shaan1337: yup, i agree, it's not too neat in this case. i've actually followed the pattern used by other services: e.g SubscriptionsService and PersistentSubscriptionService. it seems that the advantage of doing it this way is that you can stop other internal things before stopping the queue (e.g as done in PersistentSubscriptionService)

src/EventStore.Core/Services/RedactionService.cs
shaan1337: yes, if we move `newChunk.Dispose();` to the try block, and calling Dispose() throws, then the reply reason will be: `SwitchChunkResult.NewChunkOpenFailed`. but right now if Dispose() throws, it'll return `SwitchChunkResult.UnexpectedError` which is more appropriate

shaan1337: hmmm for more clarity, i'm moving it out of the finally block, since it's not necessary (as newChunk will not be assigned anyway when exceptions are thrown)

src/EventStore.Core/Services/RedactionService.cs
shaan1337: well spotted, fixed :+1:

we actually need it for _readIndex.ReadEventInfo_KeepDuplicates() and chunk.GetActualRawPosition(), so i've added 1 to GetPTableMaxReaderCount()

src/EventStore.ClusterNode/Program.cs
shaan1337: > connections from other local processes that aren't the eventstore user?
hmmm good point, i need to verify what permissions are set by kestrel on the unix socket by default. at first glance it seems that the owner and group have read/write permissions and others have only read permissions.

shaan1337: the default permission is to allow the owner and group to connect to the socket file, which seems good (in terms of security). we would thus need to document how to run the redactor app as the eventstore user.

shaan1337: the only downside is that root/sudo access is required to switch to the `eventstore` user since it's not an account that can be logged into

src/EventStore.Core/Services/RedactionService.cs
timothycoleman: aha ok :+1:

src/EventStore.Core/Services/Storage/StorageScavenger.cs
timothycoleman: fair enough :+1:

src/EventStore.ClusterNode/Program.cs
shaan1337: i think it's ok, because root/sudo access is required anyway for management purposes - for example to restart the eventstore service with `systemctl`. and as we're planning to use unix sockets for other management purposes (e.g resetting admin passwords), it seems to be fine.

src/EventStore.Core/Services/RedactionService.cs
shaan1337: agreed, fixed in: https://github.com/EventStore/EventStore/pull/3713/commits/a9533ba7848f78ead50abbf304e3dce504430b53 and added a comment

timothycoleman: still suspicious to pass a disposed chunk in to switch chunk? 😅

src/EventStore.Core/Services/Storage/StorageScavenger.cs
gregoryyoung: > yes, the main reason is for code readability - i think that it makes the critical section(s) much clearer compared to taking the lock on large blocks of code (but there may be a better way). also it avoids holding the lock for longer than is actually required (although performance's not really critical here)

Just as another one. Its not a mine field. What if on a later refactor a blocking operation were to be added to something called inside of the block as I am sure we have all had happen at some point :-D I have seen some really sneaky examples of this ... even things like Console.WriteLine causing a problem ...

src/EventStore.Core/Services/RedactionService.cs
shaan1337: hehe, fair enough, i've removed the Dispose() call :)

src/EventStore.Core/Services/Storage/StorageScavenger.cs
shaan1337: @gregoryyoung agreed, that's a strong point :+1:

timothycoleman: @gregoryyoung yeah good point, agreed

timothycoleman (dismissed):
awesome :+1:

hayley-jean (dismissed):

hayley-jean (approved):

pvanbuijtene (approved):

PR #3718, opened 2023-01-27
Cancel reads already in the reader queues when the gRPC call is cancelled

timothycoleman:
Fixed: Cancel reads already in the reader queues when the gRPC call is cancelled

Before this commit cancelling the gRPC call would prevent it issuing additional reads but would not cancel the read that is already in the queue. If the queue is long this can be an important difference.

### DIFF
--- a/src/EventStore.Core/Messages/ClientMessage.cs
+++ b/src/EventStore.Core/Messages/ClientMessage.cs
@@ -81,9 +81,11 @@ namespace EventStore.Core.Messages {
 			public readonly ClaimsPrincipal User;
 
 			public readonly DateTime Expires;
+			public readonly CancellationToken CancellationToken;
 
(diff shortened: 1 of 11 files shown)


thefringeninja (approved):
Looks good but one question, is this something we care about for TCP clients?

timothycoleman:
cheers. i'd say grpc is the priority, but do you think we could do something straight forwardly for tcp?

thefringeninja:
Probably not :|

alexeyzimarev:
Maybe a stupid thought, but maybe making the queues upper bound and denying requests would prevent overflowing queues and enable back pressure handling?

timothycoleman:
@alexeyzimarev yeah for sure, we do want to add some kind of backpressure/load shedding soon. in some of the recent metrics work we have started to track the length of the queues in seconds, which might help such a mechanism. would be good to discuss

hayley-jean (approved):

PR #3721, opened 2023-01-31
Implement gossip-read processing time tracking.

YoEight:
Added:  Metrics for gossip

- PullFromPeer (off by default) tracks the time it takes the node to get gossip from another. this is uncommon and only used when tcp connections drop.
- PushToPeer (on by default) tracks the time it takes the node to push gossip to another. this is the common gossip-sharing mechanism.
- ProcessingPushFromPeer (on by default) is how long it takes the node to process the gossip pushed to it from another.
- ProcessingRequestFromPeer (off by default) is how long it takes the node to respond to the request for gossip from another.
- ProcessingRequestFromHttpClient (off by default) is how long it takes the node to respond to the request for gossip from the http endpoint.
- ProcessingRequestFromGrpcClient (off by default) is how long it takes the node to respond to the request for gossip from the grpc client endpoint.

All are histograms. The general idea is to be able to tell how effectively the nodes are able to gossip with each other. are they in danger of timing out.

### DIFF
--- a/src/EventStore.Core/Services/Transport/Grpc/Gossip.Read.cs
+++ b/src/EventStore.Core/Services/Transport/Grpc/Gossip.Read.cs
@@ -5,6 +5,7 @@ using EventStore.Client.Gossip;
 using EventStore.Client;
 using EventStore.Core.Messages;
 using EventStore.Core.Messaging;
+using EventStore.Core.Telemetry;
 using EventStore.Plugins.Authorization;
 using Grpc.Core;
 using ClusterInfo = EventStore.Client.Gossip.ClusterInfo;
@@ -19,13 +20,15 @@ namespace EventStore.Core.Services.Transport.Grpc {
 				throw RpcExceptions.AccessDenied();
 			}
 			var tcs = new TaskCompletionSource<ClusterInfo>();
-			_bus.Publish(new GossipMessage.ClientGossip(new CallbackEnvelope(msg => GossipResponse(msg, tcs))));;
+			var duration = _tracker.Start();
+			_bus.Publish(new GossipMessage.ClientGossip(new CallbackEnvelope(msg => GossipResponse(msg, tcs, duration))));;
 			return await tcs.Task.ConfigureAwait(false);
 		}
 
-		private void GossipResponse(Message msg, TaskCompletionSource<ClusterInfo> tcs) {
+		private void GossipResponse(Message msg, TaskCompletionSource<ClusterInfo> tcs, Duration duration) {
 			if (msg is GossipMessage.SendClientGossip received) {
 				tcs.TrySetResult(ToGrpcClusterInfo(received.ClusterInfo));
+				duration.Dispose();
 			}
 		}
 
--- a/src/EventStore.ClusterNode/telemetryconfig.json
+++ b/src/EventStore.ClusterNode/telemetryconfig.json
@@ -28,6 +28,15 @@
 		"StreamTombstone": ""
 	},
 
+	"GossipTrackers": [
+//		"PullFromPeer",
+		"PushToPeer",
+		"ProcessingPushFromPeer"
+//		"ProcessingRequestFromPeer",
+//		"ProcessingRequestFromGrpcClient",
+//		"ProcessingRequestFromHttpClient"
+	],
+
 	// this specifies what name to track each queue under, according to regular expressions to
 	// match the queue names against
 	"Queues": [
--- a/src/EventStore.Common/Configuration/TelemetryConfiguration.cs
+++ b/src/EventStore.Common/Configuration/TelemetryConfiguration.cs
@@ -58,6 +58,15 @@ namespace EventStore.Common.Configuration {
 			StreamTombstone,
 		}
 
+		public enum Gossip {
+			PullFromPeer = 1,
+			PushToPeer,
+			ProcessingPushFromPeer,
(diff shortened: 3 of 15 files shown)


src/EventStore.Core/GossipStatsCollector.cs
```diff
@@ -0,0 +1,110 @@
...
+
+public class GossipStatsCollectors {
+	public static readonly GossipStatsCollectors Default = new GossipStatsCollectors();
+
+	private readonly ConcurrentDictionary<string, GossipStatsCollector> _collectors =
+		new ConcurrentDictionary<string, GossipStatsCollector>();
+
+	private GossipStatsCollectors() {}
+
+	public GossipStatsCollector New(string name) {
```
thefringeninja, line 33: Minor nitpick, this method should be called `Create` since `new` is for constructors

src/EventStore.Core/GossipStatsCollector.cs
```diff
@@ -0,0 +1,110 @@
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+
+namespace EventStore.Core;
+
+public readonly struct GossipStats {
```
thefringeninja: You can simplify this by using a record

```
public record struct GossipStats(string Name, double MovingAverage, double Max, double Min, double Last);
```

thefringeninja (changes requested):

src/EventStore.Core/GossipStatsCollector.cs
YoEight: Thanks. Turned out I no longer need that record anymore

thefringeninja (approved):

timothycoleman (changes requested):
It would be cool to track the duration of how long it takes on the calling node side to retrieve gossip from another node, to see how close to the timeout we are

I think some way to turn it on/off in the telemetryconfig would be handy, I'm not sure at what granularity though

src/EventStore.Core/GossipStatsCollector.cs
```diff
@@ -0,0 +1,36 @@
...
+	internal GossipStatsCollector(DurationMetric metric, string name) {
+		_tracker = new DurationTracker(metric, name);
+	}
+
+	public IDisposable StartRequest() {
+		return _tracker.Start();
+	}
+}
+
+public class GossipStatsCollectors {
```
timothycoleman, line 19: to fit the pattern this class might be better of being called `GossipTrackers` (and having its own file)

src/EventStore.Core/GossipStatsCollector.cs
```diff
@@ -0,0 +1,36 @@
...
+	public static readonly GossipStatsCollectors Default = new GossipStatsCollectors();
+	private Meter _coreMeter;
+	private DurationMetric _metric;
+
+	public Meter CoreMeter {
+		set {
+			_coreMeter = value;
```
timothycoleman, line 26: do we need the meter in here, it may be sufficient to just have hte metric, and to pass it in in the constructor

src/EventStore.Core/Services/Transport/Grpc/Gossip.Read.cs
```diff
@@ -13,19 +13,22 @@
...
 		private static readonly Operation ReadOperation = new Operation(Plugins.Authorization.Operations.Node.Gossip.ClientRead);
+		private static readonly GossipStatsCollector Collector = GossipStatsCollectors.Default.Create("Grpc");
 		public override async Task<ClusterInfo> Read(Empty request, ServerCallContext context) {
 			var user = context.GetHttpContext().User;
 			if (!await _authorizationProvider.CheckAccessAsync(user, ReadOperation, context.CancellationToken).ConfigureAwait(false)) {
 				throw RpcExceptions.AccessDenied();
 			}
 			var tcs = new TaskCompletionSource<ClusterInfo>();
-			_bus.Publish(new GossipMessage.ClientGossip(new CallbackEnvelope(msg => GossipResponse(msg, tcs))));;
+			var tracker = Collector.StartRequest();
```
timothycoleman, line 23: to fit the pattern of the others i think this would be a 'duration'

src/EventStore.Core/GossipStatsCollector.cs
```diff
@@ -0,0 +1,36 @@
...
+namespace EventStore.Core;
+
+public class GossipStatsCollector {
+	private readonly DurationTracker _tracker;
+
+	internal GossipStatsCollector(DurationMetric metric, string name) {
+		_tracker = new DurationTracker(metric, name);
+	}
+
+	public IDisposable StartRequest() {
```
timothycoleman, line 14: if we return a IDisposable here the Duration will be boxed

timothycoleman:
I've rebased and pushed additional suggestions

timothycoleman (approved):
fab :+1:

hayley-jean (changes requested):
This tracks gossip messages from clients (gRPC and HTTP), but isn't tracking the gossip between the nodes. The calls in `SendGossip` also need to be tracked [here](https://github.com/EventStore/EventStore/blob/8513cef3e6835792abadeedaee3f1756104748ec/src/EventStore.Core/Cluster/EventStoreClusterClient.Gossip.cs#L40) for intra-node gossip

hayley-jean (approved):

Issue #3697, opened 2023-01-10
Filtered subscription to all fires CheckpointReached with incorrect positions once "live"


rcknight:
**Describe the bug**

Wasn't sure if this should be reported against the client or the server, so apologies if this is in the wrong place!

When subscribing to all events, with the server side event filter set to `EventTypeFilter.ExcludeSystemEvents()` (other filters not tested), the checkpointReached callback works as expected while the subscription is catching up historic events. As "live" events then arrive, checkpointReached continues to fire, but the position parameter is always the same position. 

We caught this during testing, but we had our checkpointReached handler naively updating our last processed checkpoint, so this could have caused an incorrect checkpoint and events being unexpectedly reprocessed by the system following a restart.

**To Reproduce**
Run the following example code https://gist.github.com/rcknight/6c850894c9c199e2f6e1bc30e079e1cb

**Expected behavior**
CheckpointReached fires with consistently increasing positions.

**Actual behavior**
CheckpointReached fires every 32 events with the same position over and over once the subscription is live.

**EventStore details**

 - EventStore server version:
22.10.0

 - Operating system:
 Docker / MacOS (though other team members have seen this on docker on x64 machines so I don't think this relates to the alpha image)

 - EventStore client version (if applicable):
 22.0.0

alexeyzimarev:
@rcknight It's getting fixed, but I could suggest another workaround as yours seems quite complex. I would just remove the filter and accept additional traffic and compute to filter on the client side, but the code will remain trivial and checkpointing will work reliably in both catch up and live modes.

PR #3726, opened 2023-01-31
fix - filtered subscription checkpoint

dhingrak:
Fixes: #3697 

- This PR will update checkpoint when the subscription becomes live


thefringeninja (approved):

timothycoleman:
Oh I see, woops I think I introduced this in this commit 3e13ade306650fe17945c0de97d3f441505718c0

There is some code on line 445 that should have been moved along with the rest

```
var position = Position.FromInt64(checkpointReached.Position.Value.CommitPosition,
	checkpointReached.Position.Value.PreparePosition);
```

and we can use this to get the position from the checkpoint itself rather than having to read it out of the events. that way it remains the job of the SubscriptionsService to generate the correct checkpoint, and the enumerator just respects it

timothycoleman (approved):

thefringeninja (approved):

PR #3731, opened 2023-02-08
Avoid a gen2 allocation when scheduling a timer message

timothycoleman:
Changed: Adjustments to reduce allocations

### DIFF
--- a/src/EventStore.Core/Services/TimerService/TimerService.cs
+++ b/src/EventStore.Core/Services/TimerService/TimerService.cs
@@ -23,7 +23,10 @@ namespace EventStore.Core.Services.TimerService {
 		}
 
 		public void Handle(TimerMessage.Schedule message) {
-			_scheduler.Schedule(message.TriggerAfter, OnTimerCallback, message);
+			_scheduler.Schedule(
+				message.TriggerAfter,
+				static (scheduler, state) => OnTimerCallback(scheduler, state),
+				message);
 		}
 
 		private static void OnTimerCallback(IScheduler scheduler, object state) {


timothycoleman:
erm, ok second commit doesn't actually work

shaan1337 (commented):
I'm not sure if i'm interpreting this correctly:

I've checked the IL and it seems that an action object is still created: 
```
    IL_0021: newobj       instance void class [System.Runtime]System.Action`2<class EventStore.Core.Services.TimerService.IScheduler, object>::.ctor(object, native int)
```

it appears that it first pushes a pointer to the required function on the stack (with `ldftn`), then always creates the action object.

Given that the `OnTimerCallback` function is not an anonymous function, it will have a predefined location on the heap, so it won't make a new allocation (whether or not it was a static method doesn't matter in this case)

it seems that the optimization would only apply if we were using an anonymous method

also, it seems that it would be a gen0 allocation here (but i think the title says gen2 because of the previous commit)

shaan1337:
> erm, ok second commit doesn't actually work, removed it

just out of curiosity, why doesn't it work? (the code looked good)

shaan1337 (approved):
looks good after discussions with @timothycoleman :+1: 

the `newobj` instruction is actually executed only once after the change

timothycoleman:
ya in a nutshell:

- it's (likely) gen2 because the action was allocated when the message is scheduled and then collected after the callback is called (some-many seconds later)
- there is still an allocation but the compiler caches it in a property so it is only once rather than one per schedule
- the other optimisation didn't work because Reply is actually generic. the original code called Reply<T> and the change actually called Reply<Message> and then the cast in the SendToThisEnvelope didn't work

thefringeninja (approved):

PR #3737, opened 2023-02-14
Implement Event/Byte read tracking.

YoEight:
Added: Metrics for count of events being read/written and bytes being read

- bytes written can be inferred from the movement of the writer checkpoint
- events written works by counting the number of events indexed since (i) from the users perspective that is when the event is written in the sense that that is when it can be read and (ii) the mechanism can work the same on the leader and followers

image
image

### DIFF
--- a/src/EventStore.Core/MetricsBootstrapper.cs
+++ b/src/EventStore.Core/MetricsBootstrapper.cs
@@ -1,12 +1,15 @@
-﻿using System;
+using System;
+using System.Collections.Generic;
 using System.Diagnostics.Metrics;
 using System.Linq;
+using EventStore.Common.Utils;
 using EventStore.Core.TransactionLog.Chunks;
 using EventStore.Core.TransactionLog.Checkpoint;
 using EventStore.Core.Index;
 using EventStore.Core.Messaging;
 using EventStore.Core.Services.VNode;
 using EventStore.Core.Telemetry;
+using EventStore.Core.TransactionLog;
 using EventStore.Core.TransactionLog.Scavenging;
 using Conf = EventStore.Common.Configuration.TelemetryConfiguration;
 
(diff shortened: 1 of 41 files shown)


timothycoleman:
Added an explanation of the pattern here to clarify things and facilitate discussion https://github.com/EventStore/architecture-and-planning/pull/127

src/EventStore.Core/StorageTracker.cs
```diff
@@ -0,0 +1,37 @@
...
+using EventStore.Core.TransactionLog.LogRecords;
+
+namespace EventStore.Core;
+
+public class StorageTracker : IStorageTracker {
+	private readonly Counter<long> _readBytes;
+	private readonly Counter<long> _readEvents;
+	private readonly Counter<long> _indexedEvents;
+
+	public StorageTracker(Meter meter) {
```
timothycoleman, line 12: generally i've passed in a 'Metric' rather than a 'Meter' to the trackers, where a Metric is a wrapper around the instrument... In this case I think the metric would be sufficiently trivial to leave it out, but i'd still prefer to pass the instruments into the tracker rather than having the tracker create them, so that we can see all the instruments that we are creating in the bootstrapper itself

src/EventStore.Core/StorageTracker.cs
```diff
@@ -0,0 +1,37 @@
...
+		_readBytes = meter.CreateCounter<long>("ReadBytes", "bytes");
+		_readEvents = meter.CreateCounter<long>("ReadEvents", "events");
+		_indexedEvents = meter.CreateCounter<long>("IndexedEvents", "events");
```
timothycoleman, line 15: open to opinions, should _readEvents and _indexedEvents be dimensions of one instrument? perhaps we should see how the opentelemetry instrumentation libraries break down their time series into instruments/dimensions

src/EventStore.Core/MetricsBootstrapper.cs
```diff
@@ -54,6 +55,8 @@ public static class MetricsBootstrapper {
 		var queueQueueingDurationMaxMetric = new DurationMaxMetric(coreMeter, "eventstore-queue-queueing-duration-max");
 		var queueProcessingDurationMetric = new DurationMetric(coreMeter, "eventstore-queue-processing-duration");
 
+		trackers.StorageTracker = new StorageTracker(coreMeter);
```
timothycoleman, line 58: need some way to enable/disable in telemetryconfig

timothycoleman (changes requested):
looks good, comments in line. i also merged master in to resolve a conflict, we can squash later

timothycoleman:
if you're happy with the last adjustments ive made we can squash and approve this

hayley-jean (approved):

PR #3738, opened 2023-02-15
Parse custom password for admin and ops user through Environment Variables.

dhingrak:
Added: The ability to set default admin and ops passwords on first run of the database

Fixes #https://github.com/EventStore/home/issues/897

Currently the user does not have any option to set the default password for Admin and Ops user. This PR provides the ability to set default admin and ops passwords on first run of the database. It will not impact the existing password for the admin and ops user. 

For this to work, you can set the `DefaultAdminPassword`  and `DefaultOpsPassword` option using environment variables:

| Format               | Syntax                |
| Environment variable | `EVENTSTORE_DEFAULT_ADMIN_PASSWORD` |
| Environment variable | `EVENTSTORE_DEFAULT_OPS_PASSWORD` |

**Default**: `changeit`

**Warning**
The above two options can only be set through environment variables due to security reasons. The user will receive the error message if they try to pass the options using command line or config file.

### DIFF
--- a/src/EventStore.Core.Tests/Common/ClusterNodeOptionsTests/when_building/with_cluster_node_and_custom_settings.cs
+++ b/src/EventStore.Core.Tests/Common/ClusterNodeOptionsTests/when_building/with_cluster_node_and_custom_settings.cs
@@ -1,8 +1,15 @@
 using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Linq;
 using System.Net;
+using EventStore.Common.Configuration;
 using EventStore.Common.Utils;
 using EventStore.Core.Certificates;
 using EventStore.Core.LogAbstraction;
+using EventStore.Core.Services;
+using Microsoft.Extensions.Configuration;
 using NUnit.Framework;
 
 namespace EventStore.Core.Tests.Common.ClusterNodeOptionsTests.when_building {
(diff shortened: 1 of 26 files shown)


src/EventStore.Core.Tests/Common/ClusterNodeOptionsTests/when_building/with_cluster_node_and_custom_settings.cs
```diff
@@ -261,4 +268,94 @@ public class
...
+		[Test]
+		public void should_set_the_custom_admin_and_ops_user_password() {
+			Assert.AreEqual(_adminPassword,_options.DefaultUser.DefaultAdminPassword);
+			Assert.AreEqual(_opsPassword, _options.DefaultUser.DefaultOpsPassword);
+		}
+	}
+	
+	[TestFixture(typeof(LogFormat.V2), typeof(string))]
+	[TestFixture(typeof(LogFormat.V3), typeof(uint))]
+	public class with__cluster_custom_settings_check_for_environment_only_options<TLogFormat, TStreamId> : SingleNodeScenario<TLogFormat,
```
hayley-jean, line 292:
```suggestion
	public class with_cluster_custom_settings_check_for_environment_only_options<TLogFormat, TStreamId> : SingleNodeScenario<TLogFormat,
```

src/EventStore.Core.Tests/Common/ClusterNodeOptionsTests/when_building/with_cluster_node_and_custom_settings.cs
```diff
@@ -261,4 +268,94 @@ public class
...
+		[Test]
+		public void should_return_error_when_default_password_options_pass_through_command_line() {
+
+			var args = new string[] {
+				"--DefaultAdminPassword=Admin2023#",
+				"--DefaultOpsPassword=Ops2023#"
+			};
+			_configurationRoot = new ConfigurationBuilder()
+				.Add(new DefaultSource(new Dictionary<string, object> {
+					[nameof(Flag)] = false,
```
hayley-jean, line 324: These configuration root options aren't used in these tests, so they could be removed and the default source set to just the options you're interested in, for example :

```
new Dictionary<string, object> {
	[nameof(ClusterVNodeOptions.DefaultUser.DefaultAdminPassword)] = "changeit",
	[nameof(ClusterVNodeOptions.DefaultUser.DefaultOpsPassword)] = "changeit",
}
```

src/EventStore.Core.Tests/Common/ClusterNodeOptionsTests/when_building/with_single_node_and_custom_settings.cs
```diff
@@ -149,4 +156,94 @@ public class with_custom_advertise_as<TLogFormat, TStreamId> : SingleNodeScenari
...
+		[Test]
+		public void should_return_error_when_default_password_options_pass_through_command_line() {
+
+			var args = new string[] {
+				"--DefaultAdminPassword=Admin#",
+				"--DefaultOpsPassword=Ops#"
+			};
+			_configurationRoot = new ConfigurationBuilder()
+				.Add(new DefaultSource(new Dictionary<string, object> {
+					[nameof(Flag)] = false,
```
hayley-jean, line 212: Same comment as above

src/EventStore.Core.Tests/Services/Transport/Tcp/TcpClientDispatcherTests.cs
```diff
@@ -27,17 +27,21 @@ public class TcpClientDispatcherTests {
 
 		private ClientTcpDispatcher _dispatcher;
 		private TcpConnectionManager _connection;
+		private ClusterVNodeOptions.DefaultUserOptions _defaultUserOptions;
 
 		[OneTimeSetUp]
 		public void Setup() {
 			_dispatcher = new ClientTcpDispatcher(2000);
-
+			_defaultUserOptions = new ClusterVNodeOptions.DefaultUserOptions() {
```
hayley-jean, line 35: Perhaps it would be worthwhile adding the default user options to a common location and reusing them from there? For example, you could add them to DefaultData: https://github.com/EventStore/EventStore/blob/master/src/EventStore.Core.Tests/DefaultData.cs

hayley-jean (changes requested):

hayley-jean (approved):
Before merging, could you update your changelog entry in the PR description to something more descriptive, and move the issue fix statement to a new line? For example:

> Added: The ability to set default admin and ops passwords on first run of the database

Also, could you include the new option names in the PR description, as well as add them to the security docs at https://github.com/EventStore/EventStore/blob/master/docs/security.md

src/EventStore.Core/ClusterVNodeOptions.Framework.cs
```diff
@@ -8,6 +8,8 @@
 using EventStore.Common.Configuration;
 using EventStore.Common.Utils;
 using Microsoft.Extensions.Configuration;
+using Enum = System.Enum;
```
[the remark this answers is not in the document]
dhingrak: Actually, there is no need of these using statements. Let me remove that.

hayley-jean (dismissed):

src/EventStore.Core/ClusterVNodeOptions.Framework.cs
```diff
@@ -60,6 +60,34 @@ where defaultValues.TryGetValue(option.Name, out var defaultValue)
...
+					.Where(option => option.GetCustomAttribute<EnvironmentOnlyAttribute>() != null)
+					.Select(option => option)
+					.ToArray();
+
+				var errorBuilder = new StringBuilder();
+
+				foreach (var provider in ConfigurationRoot.Providers) {
+					var source = provider.GetType();
+
+					if (source.Name is "Default" or "EnvironmentVariables") continue;
```
pvanbuijtene, line 76: This will fail if the underlying type gets renamed, maybe better to check the type?

pvanbuijtene (changes requested):
Nice change 👍 

There's one other thing, currently the help output `EventStore.ClusterNode.exe --help` shows the default password.

```
...

Default User Options
-DefaultAdminPassword=changeit                                                                              Admin Default password
-DefaultOpsPassword=changeit                                                                                Ops Default password

Logging Options
...
```

Think it will be better to show it like
```
...

Default User Options
-DefaultAdminPassword=****                                                                              Admin Default password
-DefaultOpsPassword=****                                                                                Ops Default password

Logging Options
...
```

Can you have a look at this?

_Update:_
We had some discussion during the standup and since it will be removed it might not be an issue.

src/EventStore.Core/ClusterVNodeOptions.Framework.cs
ylorph: my 2 cents: 
- Never show a password, by definition 

- that setting shouldn't printed out as the other one since it's _only_ throught environment variable and that might confuse people if  `-DefaultAdminPassword` is printed out as the rest of the settings are 

so --help should really heve another section 
"ENV varaible only" 
 `EVENTSTORE_DEFAULT_ADMIN_PASSWORD` `sets the password of the admin account on first start of the DB`
 
as for the default value: we don't advertise them today in --help, so I would not change anything there 

Question : is the value printed in the dump of the config  at startup in the log ?

ylorph:
my 2 cents:

    Never show a password, by definition

    that setting shouldn't printed out as the other one since it's only throught environment variable and that might confuse people if -DefaultAdminPassword is printed out as the rest of the settings are

so --help should really heve another section
"ENV varaible only"
EVENTSTORE_DEFAULT_ADMIN_PASSWORD sets the password of the admin account on first start of the DB

as for the default value: we don't advertise them today in --help, so I would not change anything there

Question : is the value printed in the dump of the config at startup in the log ?

alexeyzimarev:
If we are printing the values, we can annotate configuration properties with some attribute like `Sensitive` and mask the value when printing. It will keep the code generic, and allow us to mask particular options when being printed.

hayley-jean:
@alexeyzimarev @ylorph the option is already marked as sensitive and will be masked when printed out in the logs. But the `--help` dump doesn't take the `Sensitive` attribute into account because all of those values are the ones already set in the code.

I agree with @ylorph's approach to have a separate section for environment variables only, where the option is shown but the default value is not

alexeyzimarev:
Do you mean `--help` or `--whatif`? I don't think `--whatif` can reliably print anything in this case as it should not touch the database, so it doesn't know if the password has previously been set.

If that is not a concern, I am ok with printing out things on the console when it's limited to having access to the machine console. Rightfully mentioned, the value is already in the config file, so hiding it makes no sense. Or does it?

What if a logged in user has no access to the config file, but can run `eventstored --whatif`?

One thing I know for sure is that no password should be printed to logs, but it seems to be covered already.

Another question is the ops user password. Would it be the same? We can also just set it to a random value and tell the admin to change it.

hayley-jean:
@alexeyzimarev 

> Do you mean --help or --whatif? I don't think --whatif can reliably print anything in this case as it should not touch the database, so it doesn't know if the password has previously been set.

The discussion above was only about what is printed in `--help` when it shows the default values of the new `EVENTSTORE_DEFAULT_ADMIN_PASSWORD` and `EVENTSTORE_DEFAULT_OPS_PASSWORD` options. So `--help` would only ever have printed `changeit`.

All other places that print out configuration (`--what-if`, startup logs, etc) will obscure the values, as the default password options are marked as sensitive.

Also, the options only change the password for the first run of the database. If you have an existing database, then setting these options do nothing.

> If that is not a concern, I am ok with printing out things on the console when it's limited to having access to the machine console. Rightfully mentioned, the value is already in the config file, so hiding it makes no sense. Or does it?
> What if a logged in user has no access to the config file, but can run eventstored --whatif?

This PR disallows setting the default admin and ops passwords in the config file or from the command line, it can only be set by an environment variable so hiding the set value still makes sense.

> Another question is the ops user password. Would it be the same? We can also just set it to a random value and tell the admin to change it.

The ops password is set in the same way as the admin password for now.
But doing it this way may be a good option when we eventually remove `changeit` so that users can _just_ provide the admin option through environment variable if they'd prefer.

pvanbuijtene (approved):
Works great, nice change 👍

hayley-jean (approved):

PR #3741, opened 2023-02-17
Create and use a separate iodispatcher for scavenge log

hayley-jean:
Fixed: Use a separate IODispatcher for scavenge log

This fixes a bug introduced by https://github.com/EventStore/EventStore/pull/3702, which changed the IODispatcher used by persistent subscriptions to subscribe and publish to the psub-specific queues and buses.

However, the scavenge log was using the same IODispatcher, so the messages for scavenging were being sent to the wrong queue.

### DIFF
--- a/src/EventStore.Core/ClusterVNode.cs
+++ b/src/EventStore.Core/ClusterVNode.cs
@@ -1190,16 +1190,16 @@ namespace EventStore.Core {
 			var perSubscrBus = new InMemoryBus("PersistentSubscriptionsBus", true, TimeSpan.FromMilliseconds(50));
 			var perSubscrQueue = new QueuedHandlerThreadPool(perSubscrBus, "PersistentSubscriptions", _queueStatsManager,
 				trackers.QueueTrackers, false);
-			var ioDispatcher = new IODispatcher(_mainQueue, new PublishEnvelope(perSubscrQueue));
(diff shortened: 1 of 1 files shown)


src/EventStore.Core/ClusterVNode.cs
```diff
@@ -1258,6 +1258,16 @@ public class ClusterVNode<TStreamId> :
 
 			// STORAGE SCAVENGER
 			ScavengerFactory scavengerFactory;
+			var scavengerDispatcher = new IODispatcher(_mainQueue, new PublishEnvelope(MainQueue));
```
timothycoleman, line 1261: oh good catch - could just be worth checking if we need all of these subscriptions, although.. if its not obvious maybe err on the side of keeping them

hayley-jean: It looks like the scavenging IODispatcher only ever reads backwards or writes events, so I've removed the forward reads

dhingrak (approved):

timothycoleman (approved):
Looks good :+1:

For posterity, the risk of the IO Dispatcher publishing on to the mainqueue is that the registered callbacks will be called on the main bus which is dangerous if they aren't guaranteed to be quick. I've checked that the callbacks on this iodispatcher are quick and they all look fine except for `TFChunkScavengerLogManager::GatherIncompleteScavenges` which is potentially longer but only happens once on startup so I think this is fine for now

Issue #3486, opened 2022-05-03
Group the dump of configuration settings at startup


hayley-jean:
As an improvement to PR https://github.com/EventStore/EventStore/pull/3475, it would be useful to sort the dump into the configuration groups, and have the settings arranged alphabetically within the groups.

For example, the `--help` option prints out the groups:

```
OPTION                                                                                               DESCRIPTION

Application Options
-Help=False                                                                                          Show help.
-Version=False                                                                                       Show version.
-Config=C:\Users\hayle\code\EventStore\src\EventStore.ClusterNode\bin\x64\Release\net5.0             Configuration files.
-WhatIf=False                                                                                        Print effective configuration to console and then exit.
-DisableHttpCaching=False                                                                            Disable HTTP caching.
-StatsPeriodSec=30                                                                                   The number of seconds between statistics gathers.
-WorkerThreads=0                                                                                     The number of threads to use for pool of worker services. Set to '0' to scale automatically (Default)
-EnableHistograms=False                                                                              Enables the tracking of various histograms in the backend, typically only used for debugging, etc.
-LogHttpRequests=False                                                                               Log Http Requests and Responses before processing them.
-LogFailedAuthenticationAttempts=False                                                               Log the failed authentication attempts.
-SkipIndexScanOnReads=False                                                                          Skip Index Scan on Reads. This skips the index scan which was used to stop reading duplicates.
-MaxAppendSize=1048576                                                                               The maximum size of appends, in bytes. May not exceed 16MB.
-Insecure=False                                                                                      Disable Authentication, Authorization and TLS on all TCP/HTTP interfaces.

Logging Options
-Log=C:\Users\hayle\code\EventStore\src\EventStore.ClusterNode\bin\x64\Release\net5.0\logs           Path where to keep log files.
-LogConfig=logconfig.json                                                                            The name of the log configuration file.
-LogLevel=Default                                                                                    Sets the minimum log level. For more granular settings, please edit logconfig.json. (Default, Verbose, Debug, Information, Warning, Error, Fatal)
-LogConsoleFormat=Plain                                                                              Which format (plain, json) to use when writing to the console. (Plain, Json)
-LogFileSize=1073741824                                                                              Maximum size of each log file.
-LogFileInterval=Day                                                                                 How often to rotate logs. (Infinite, Year, Month, Day, Hour, Minute)
-LogFileRetentionCount=31                                                                            How many log files to hold on to.
-DisableLogFile=False                                                                                Disable log to disk.

[...]

PR #3752, opened 2023-02-27
Dump the configuration settings in a group at EventStore start up

dhingrak:
Fixed: #3486 

This PR dump the configuration settings in their respective groups at start up. Also, the configuration settings arranged alphabetically within the groups.

Before :

sortBy dump options

After:

groupBy config options

### DIFF
--- a/src/EventStore.Common/Configuration/OptionsDumper.cs
+++ b/src/EventStore.Common/Configuration/OptionsDumper.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Linq;
 using System.Reflection;
 using System.Text;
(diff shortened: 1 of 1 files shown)


timothycoleman (changes requested):
From the screenshot, is it intentional that there are two groups for `Certificate Options` - one just plain and one with `(from file)`

Also itd be neat if the groups themselves were sorted alphabetically I think

dhingrak:
@timothycoleman I have grouped the options based on DescriptionAttribute (ClusterVNodeOptions) so that's why we have two groups for certificate options. It is similar to the grouping that we get from --help command

> Also itd be neat if the groups themselves were sorted alphabetically I think
Yeah that's right, groups should also be sorted alphabetically.

timothycoleman:
oh! righty i see what you mean, thats fine

src/EventStore.Common/Configuration/OptionsDumper.cs
```diff
@@ -8,15 +9,14 @@
...
 			_optionSections = optionSections;
-			_sortOrder = _optionSections.SelectMany(optionSection =>
-					optionSection.GetProperties()
-						.Select(p => NameTranslators.CombineByPascalCase(p.Name, " ").ToUpper()))
-				.OrderBy(x => x)
-				.ToList();
+			_sortGroupBy = _optionSections.SelectMany(type => type.GetProperties().OrderBy(x => x.Name).GroupBy(
+					option =>
+						option.DeclaringType?.GetCustomAttribute<DescriptionAttribute>()?.Description ?? string.Empty))
+				.OrderBy(option => option.Key).ToList();
```
timothycoleman, line 19: getting the declaring type for each property is feeling a bit odd because we already have it (it is `type`) because we are, indeed, iterating through its properties

it might be more natural to to map each optionSection to a tuple containing its description and the names, in order, of the properties.

then _sortGroupBy would be of type `List<(string, List<string>)>` (and might have a different name), and you wouldn't need the `SelectMany` or the `GroupBy`

I _think_ you can call `CombineByPascalCase` in here like before, instead of having to call `RemoveWhiteSpacesToCombine` later

src/EventStore.Common/Configuration/OptionsDumper.cs
```diff
@@ -26,19 +26,39 @@ public class OptionsDumper {
...
+				dumpOptions.Append($"{name.ToUpper()} OPTIONS:").AppendLine().AppendLine();
+				foreach (var groupOptions  in _sortGroupBy) {
+					bool oneTimeCheckForGroupOptions = true;
+					
+					foreach (var option in groupOptions) {
+						var result = (from configOption in configOptions
+							where string.Equals(
+								NameTranslators.RemoveWhiteSpacesToCombine(configOption.Key, string.Empty), option.Name,
+								StringComparison.CurrentCultureIgnoreCase)
+							select configOption).FirstOrDefault();
```
timothycoleman, line 44: it feels like what we could do with here is a dictionary lookup rather than searching through the `configOptions` each time. the funny thing is in `Dump`, `info` already _is_ a dictionary. So perhaps this will turn out nicely if we pass `info` in here as `configOptions` and also pass a `Func<Type, bool>` to act as a filter.

Out in `Dump` the two filters can be defined something like like 

```
			bool Default(Type x) => x == typeof(Default);
			bool Modified(Type x) => x != typeof(Default);
```

timothycoleman:
looks good :+1: i've pushed some minor additional adjustments for your consideration

thefringeninja (approved):

PR #3782, opened 2023-03-08
Move EventStoreDB start up validations to separate validation service.

dhingrak:
Added: Create a new class for EventStoreDB start up validations. 

The goal of this PR is to accumulate the EventStoreDB start up validations into a single validation service.


hayley-jean (approved):

src/EventStore.Core/ClusterValidationService.cs
timothycoleman, line 1: theres a mixture of tabs and spaces indenting in here

timothycoleman: it isn't really a service, perhaps we could call it something like ClusterVNodeOptionsValidator

src/EventStore.Core/ClusterValidationService.cs
```diff
@@ -0,0 +1,136 @@
...
+           "The given ClusterSize is set to 1 but GossipSeeds are multiple. We will never be able to sync up with this configuration.");
+     }
+     
+     if (options.Cluster.ReadOnlyReplica && options.Cluster.ClusterSize <= 1) {
+        throw new InvalidConfigurationException(
+           "This node cannot be configured as a Read Only Replica as these node types are only supported in a clustered configuration.");
+     }
+  }
+
+  public static bool StartUpValidations(ClusterVNodeOptions options) {
```
timothycoleman, line 106: better to give methods imperative names, here we could have 'ValidateForStartup'

timothycoleman (approved):
:+1:

PR #3783, opened 2023-03-10
update doc for persistent subscription's ordering guarantee

yehohanan7:


yehohanan7:
> Just as a note: I believe you **CAN** run persistent subscriptions with an assured ordering guarantee.
> 
> It is highly unlikely that you would ever actually **want** to do this but it can be done.
> 
> The way to do this is to set your buffer/in flight/etc sizes to 1 and only use a single subscriber. Unless something has changed this will end up with the ordering guarantee (and horrible throughput! :-D +-10 messages/second)

@gregoryyoung I think i misunderstood the way `Pinned` strategy works, after re-reading the doc we hash the "event id" to determine the bucket, If we hash other attribute like the "aggregate id" then we could achieve ordering guarantee for the events generated for a specific aggregate right?

gregoryyoung:
Pinned strategy does not use eventid it uses stream id ... you want to provide locality of the stream.

You can find the code for this here: https://github.com/EventStore/EventStore/blob/master/src/EventStore.Core/Services/PersistentSubscription/ConsumerStrategy/PinnablePersistentSubscriptionConsumerStrategy.cs#L79

Issue #3784, opened 2023-03-11
Provide suggestions for unknown/invalid configuration parameters


matsientst:
**Is your feature request related to a problem? Please describe.**
When a user makes a minor mistake passing a parameter the system doesn't provide any attempts to guide them to the correct answer. If the user enters a configuration option like `--cluster-sze` the system just ignores it and lists it as an unknown option. 

**Describe the solution you'd like**
Use the Levenshtein distance algorithm to find the closest default configuration name match. It should look to see if the distance is up to 5 characters off and provide the best suggestion. If it's wildly off we need to just let it go rather than suggesting something completely inaccurate

hayley-jean:
Closed by https://github.com/EventStore/EventStore/pull/3785

PR #3785, opened 2023-03-11
Matsientst/invalid config closest match

matsientst:
Added: Suggestions for unknown/invalid configuration parameters (#3784)

This PR adds the feature to suggest a parameter when the user passes an incorrect parameter. For example, if the user passes a parameter like: `--cluster-sze` the system would suggest:
`The option "ClusterSze" is not a known option. Did you mean "ClusterSize"`

### DIFF
--- a/src/EventStore.ClusterNode/Program.cs
+++ b/src/EventStore.ClusterNode/Program.cs
@@ -84,11 +84,15 @@ namespace EventStore.ClusterNode {
 					? LogEventLevel.Information
 					: LogEventLevel.Fatal;
 
-				foreach (var key in options.Unknown.Keys) {
-					Log.Write(level, "The option {key} is not a known option.", key);
+				foreach (var (option, suggestion) in options.Unknown.Options) {
+					if (string.IsNullOrEmpty(suggestion)) {
+						Log.Write(level, "The option {option} is not a known option.", option);	
+					} else {
+						Log.Write(level, "The option {option} is not a known option. Did you mean {suggestion}?", option, suggestion);
+					}
 				}
 
-				if (options.Unknown.Keys.Any() && !options.Application.AllowUnknownOptions) {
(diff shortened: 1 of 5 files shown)


ylorph:
I'd suggest either 
*  removing the dump of --help configuration settings when we detect something wrong
* inverting order : dump --help and the `FTL` message 
Eg: the following is correctly detected, though i have to scroll up quite a bit to be able to see the problem
.\EventStore.ClusterNode.exe --dev  --Advertise-Host-To-Client-s example.com

image

That's using the console ofc.
when using the logs it would help as well , as the default for most people is to look at the tail of the log first

ylorph:
Also when using `allow unknown options` : `--dev  --Advertise-Host-To-Client-s example.com --allow-unknown-options true`
Shouldn't really be the log level WRN instead of INF  ? otherwise it will never be acted upon 

[24716, 1,10:21:19.195,INF] The option "AdvertiseHostToClientS" is not a known option. Did you mean "AdvertiseHostToClientAs"

hayley-jean (approved):
Looks good and works locally

I agree with @ylorph about the log changes, but I'm happy for that to be in a separate PR if you prefer

timothycoleman:
Looks good to me too, and also agreed about yves suggestions for a separate PR (for context though, we originally chose INF rather tha WRN when unknown options are allowed because we were expecting it to be made use of in cloud and didn't want customers to have warnings that weren't their fault. i think this may no longer be applicable though.

I've pushed some suggestions, they're pretty minor just idomatic stuff mostly. if you're happy @matsientst then they can be squashed in

timothycoleman (changes requested):
pushed some suggestions ^

matsientst:
@timothycoleman Looks great! Changes squashed. Am I good to merge?

hayley-jean (approved):

PR #3787, opened 2023-03-15
Do not allow anonymous user to access some endpoints and streams in EventStoreDB

dhingrak:
Added: Options to restrict EventStoreDB access for anonymous users

Related to https://github.com/EventStore/home/issues/899
 
This PR adds a functionality to restrict the anonymous access to some endpoints and streams in EventStoreDB by setting the following options:- 
 

Format               | Syntax                                     |
|:---------------------|:-------------------------------------------|
| Command line         | `--allow-anonymous-stream-access`          |
| YAML                 | `AllowAnonymousStreamAccess`               |
| Environment variable | `EVENTSTORE_ALLOW_ANONYMOUS_STREAM_ACCESS`

Default: true

Note: In the future, the value will likely be disabled by default

Format               | Syntax                                       |
|:---------------------|:---------------------------------------------|
| Command line         | `--allow-anonymous-endpoint-access`          |
| YAML                 | `AllowAnonymousEndpointAccess`               |
| Environment variable | `EVENTSTORE_ALLOW_ANONYMOUS_ENDPOINT_ACCESS` |

Default: true

### DIFF
--- a/docs/security.md
+++ b/docs/security.md
@@ -75,6 +75,34 @@ For this to work, you can use the `DefaultOpsPassword` option:
 Due to security reasons the DefaultAdminPassword and DefaultOpsPassword options can only be set through environment variables. The user will receive the error message if they try to pass the options using command line or config file.
 :::
 
+### Allow anonymous access to some endpoints and streams
+
+Historically, anonymous users with network access have been allowed to read/write streams that do not have access control lists.
+
+This anonymous access to streams can be disabled, and in the future will likely be disabled by default.
+
+| Format               | Syntax                                     |
+|:---------------------|:-------------------------------------------|
+| Command line         | `--allow-anonymous-stream-access`          |
+| YAML                 | `AllowAnonymousStreamAccess`               |
+| Environment variable | `EVENTSTORE_ALLOW_ANONYMOUS_STREAM_ACCESS` |
+
+**Default**: `true`
+
+Similarly to streams above, anonymous access has historically been available to the `/stats` and `/gossip` endpoints, and the `HTTP OPTIONS` method.
+
+This can be disabled with the following option. Anonymous access will still be granted to `/ping`, `/info`, the static content of the UI, and http redirects
+
+For this to work, you can disable the `AllowAnonymousEndpointAccess` option:
+
+| Format               | Syntax                                       |
+|:---------------------|:---------------------------------------------|
+| Command line         | `--allow-anonymous-endpoint-access`          |
+| YAML                 | `AllowAnonymousEndpointAccess`               |
+| Environment variable | `EVENTSTORE_ALLOW_ANONYMOUS_ENDPOINT_ACCESS` |
+
+**Default**: `true`
+
 ### Certificates configuration
 
 In this section, you can find settings related to protocol security (HTTPS and TLS).
--- a/src/EventStore.Core/Authorization/LegacyAuthorizationProviderFactory.cs
+++ b/src/EventStore.Core/Authorization/LegacyAuthorizationProviderFactory.cs
@@ -10,6 +10,8 @@ using Serilog;
 namespace EventStore.Core.Authorization {
 	public class LegacyAuthorizationProviderFactory : IAuthorizationProviderFactory {
 		private readonly IPublisher _mainQueue;
+		private readonly bool _allowAnonymousEndpointAccess;
+		private readonly bool _allowAnonymousStreamAccess;
 
 		private static readonly Claim[] Admins =
 			{new Claim(ClaimTypes.Role, SystemRoles.Admins), new Claim(ClaimTypes.Name, SystemUsers.Admin)};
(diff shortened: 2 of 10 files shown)


alexeyzimarev:
Does it make sense to protect the `ping` endpoint?

dhingrak:
> Does it make sense to protect the `ping` endpoint?

I don't think there is any need to protect `ping` endpoint. Currently, I have locked down most of the endpoints until Hayley has a discussion with Tim and you regarding which endpoints will remain available to anonymous requests.

hayley-jean:
I think we should split this into two options:

| Name | Description | Default |
|--------|--------------|---------|
| AllowAnonymousEndpointAccess | Allows anonymous access to the gossip and stats endpoints | true |
| AllowAnonymousStreamAccess | Allows anonymous reads and writes to streams | ??? |

`Ping` should always allow anonymous access

The reason for this being that changing anonymous endpoint access will break a lot of deployments and monitoring configurations. It's quite possible that someone would want to keep anonymous access on endpoints enabled, but still want to lock down anonymous reads/writes to streams.

I'm still unsure about what the default for anonymous stream access should be. On the one hand, the fact that anonymous stream access is enabled by default is surprising to a lot of users, on the other, changing this default is a breaking change.

I would lean towards setting it to `true` by default, with the intentions of changing it to `false` with the release after next (and advertising this change appropriately before then)

tambeau:
Concur with Hayley on setting AllowAnonymousStreamAccess to True.

docs/security.md
```diff
@@ -75,6 +75,33 @@ For this to work, you can use the `DefaultOpsPassword` option:
...
 :::
 
+### Allow anonymous access to some endpoints and streams
+
+The default authorization policy allows access to some endpoints and streams. This was the intended behaviour as EventStoreDB was meant to be running in the secure environment.
+The streams that require higher levels of security should have ACL's set on them to restrict the read/write access. We are adding a new feature where the user
+can expose the EventStoreDB on the public internet by setting the following options in their EventStoreDB configuration. The higher level of security access to streams should work in the same way
+by setting the ACL's on stream to restrict read/write access.
+
+For this to work, you can use the `AllowAnonymousEndpointAccess` option:
```
timothycoleman, line 85: lets consider the wording here

timothycoleman:
we should probably record in the description here some justification for the anonymously accessible endpoints (ping, gossip read etc) in case we wonder why we decided this way later. are we sure about gossip?

src/EventStore.Core/Authorization/LegacyStreamPermissionAssertion.cs
```diff
@@ -25,6 +30,20 @@ public class
 
 		public ValueTask<bool> Evaluate(ClaimsPrincipal cp, Operation operation, PolicyInformation policy,
 			EvaluationContext context) {
+			var startedAt = _sw.Elapsed;
+
+			if (!_allowAnonymousStreamAccess) {
+				var claims = cp.Claims;
+				
+				foreach (var claim in claims) {
+					if (claim.Type == ClaimTypes.Anonymous) {
```
timothycoleman, line 39: is it a bit odd to be checking for the existence of an anonymous claim rather than the absense of some other claim indicating that aren't anonymous

timothycoleman (changes requested):
looks good, questions and sugguestions inline ^

src/EventStore.Core/Authorization/LegacyStreamPermissionAssertion.cs
```diff
@@ -25,6 +30,26 @@ public class
...
+				var claims = cp.Claims;
+				
+				foreach (var claim in claims) {
+					if (claim.Type != ClaimTypes.Name) continue;
+					authorize = true;
+					break;
+				}
+
+				if (!authorize) {
+					Log.Warning(
```
timothycoleman, line 45: im not sure we need the warning and stopwatch in here, the ~same message will be logged in the PolicyAuthorizationProvider i think?

timothycoleman: actually scratch that, i dont think we should be changing this class at all, i'll push what i mean

src/EventStore.Core/Authorization/LegacyAuthorizationProviderFactory.cs
```diff
@@ -19,29 +21,40 @@ public class LegacyAuthorizationProviderFactory : IAuthorizationProviderFactory
...
 			var policy = new Policy("Legacy", 1, DateTimeOffset.MinValue);
-			var streamAssertion = new LegacyStreamPermissionAssertion(_mainQueue);
+			var legacyStreamAssertion = new LegacyStreamPermissionAssertion(_mainQueue);
 
-			policy.AllowAnonymous(Operations.Node.Redirect);
-			policy.AllowAnonymous(Operations.Node.StaticContent);
 			policy.AllowAnonymous(Operations.Node.Ping);
-			policy.AllowAnonymous(Operations.Node.Options);
-
+			policy.AllowAnonymous(Operations.Node.Gossip.Read);
```
hayley-jean, line 35: `Operations.Node.Gossip.ClientRead` should be used for gossip requests from external clients, and `Operations.Node.Gossip.Read` should be used for node-to-node gossip.

As such, I think only `Gossip.ClientRead` should allow anonymous access. And perhaps the `Gossip.Read` operation should be further locked down to requiring system access?

That being said, the HTTP GossipController is registering `/gossip` as a `Gossip.Read` operation. This should really be changed to `Gossip.ClientRead` ([see here](https://github.com/EventStore/EventStore/blob/master/src/EventStore.Core/Services/Transport/Http/Controllers/GossipController.cs#L37))

hayley-jean (changes requested):

src/EventStore.Core/Authorization/LegacyAuthorizationProviderFactory.cs
hayley-jean: In summary, changes needed:
- `Gossip.ClientRead` should use the `allow-anonymous-endpoints` option
- `Gossip.Read` should require a system user
- `/gossip` in the HTTP gossipController should be changed to use `Gossip.ClientRead`

src/EventStore.Core/Authorization/LegacyAuthorizationProviderFactory.cs
```diff
@@ -19,29 +21,42 @@ public class LegacyAuthorizationProviderFactory : IAuthorizationProviderFactory
...
+			var legacyStreamAssertion = new LegacyStreamPermissionAssertion(_mainQueue);
 
-			policy.AllowAnonymous(Operations.Node.Redirect);
-			policy.AllowAnonymous(Operations.Node.StaticContent);
+			// The Gossip.ClientRead is set to allow anonymous because it should be used for gossip requests from external clients. 
+			// The Node.Ping is set to allow anonymous as it does not disclose any secure information.
 			policy.AllowAnonymous(Operations.Node.Ping);
-			policy.AllowAnonymous(Operations.Node.Options);
-
+			policy.AllowAnonymous(Operations.Node.Gossip.ClientRead);
```
timothycoleman, line 37: and this is the one that should be configurable

timothycoleman:
Looks good, I think the tests need updating following the latest change, and the original description in the ticket ^

Actually for the 'added' line perhaps adjust slightly

`Added: Option to restrict EventStoreDB access for an anonymous user`

timothycoleman (approved):
good work :+1:

hayley-jean (approved):
Looks good, thank you 👍

PR #3788, opened 2023-03-17
Update Jint to version 3.0.0-beta-2048

lahma:
Changed: Jint library version to 3.0.0-beta-2048

* required API changes
* instead of removing `eval` from Global, now using new option to prevent dynamic script parsing (you missed that code parsing can also happen via `Function` constructor
* note that Jint's `JsonParser` [has been improved](https://github.com/sebastienros/jint/pull/1485) and it has default maximum recursion of 64
* replaced `IReadOnlyDictionary` field type with concrete to prevent interface dispatch penalty

### DIFF
--- a/src/EventStore.Projections.Core.Tests/Services/Jint/Serialization/when_serializing_state.cs
+++ b/src/EventStore.Projections.Core.Tests/Services/Jint/Serialization/when_serializing_state.cs
@@ -124,7 +124,7 @@ namespace EventStore.Projections.Core.Tests.Services.Jint.Serialization {
 
 		[Test]
 		public void undefined_property() {
-			var instance = new ObjectInstance(_engine);
+			var instance = new JsObject(_engine);
 			instance.Set("foo", JsValue.Undefined);
 			instance.Set("bar", "baz");
 			var serialized = _sut.Serialize(instance);
@@ -153,7 +153,7 @@ namespace EventStore.Projections.Core.Tests.Services.Jint.Serialization {
 				).SingleOrDefault();
 			if(streamName == null) throw new InvalidOperationException($"Could not find {filename}");
 			using var stream = assembly.GetManifestResourceStream(streamName);
-			
+
 			var doc = JsonDocument.Parse(stream);
 			using var ms = new MemoryStream();
 			var writer = new Utf8JsonWriter(ms);
--- a/src/EventStore.Projections.Core/EventStore.Projections.Core.csproj
+++ b/src/EventStore.Projections.Core/EventStore.Projections.Core.csproj
@@ -37,7 +37,7 @@
 		<PackageReference Include="Newtonsoft.Json" Version="13.0.2" />
 	</ItemGroup>
 	<ItemGroup>
-		<PackageReference Include="Jint" Version="3.0.0-beta-2038" />
+		<PackageReference Include="Jint" Version="3.0.0-beta-2048" />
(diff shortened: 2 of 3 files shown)


lahma:
On top of this, I would be happy to hear about possible performance scenarios you are interested in. I've checked the code base and some access patterns might get faster, but I'm unsure what are the hot paths.

pvanbuijtene:
Nice work, thanks for your contribution 👍 

The biggest problem we had I think is with the serializer, we currently have to use our own.
https://github.com/EventStore/EventStore/blob/master/src/EventStore.Projections.Core/Services/Interpreted/JintProjectionStateHandler.cs#L1025

A benchmark for this can be found in the `EventStore.MicroBenchmarks` project.

Noticed some changes were merged recently in JInt regarding this, but haven't been able yet to see how close it will get to ours.

pvanbuijtene (approved):

lahma:
@pvanbuijtene thank you for the feedback and pointer to your implementation. I will check it out!

timothycoleman (approved):
Thanks, that's great! I've added a line to the top of the description which is required by the ci

lahma:
OK I think I know why your serializer is fast... because you cheated 😉 

You use `_iterator = instance.GetOwnProperties().GetEnumerator();` for an object instance, which sounds correct at first, but it's different from JSON serialization spec.

 The correct way would be to use `instance.GetOwnPropertyKeys(Types.String)` which will 

* only return string keys (not symbols) 
* and follow the spec by sorting first by numeric keys as first in the list of list 
* and then return property non-numeric property names in insertion order

I understand that your use case probably never hits numeric object indexes and symbols so this might be something that could be optimized on Jint's side.

lahma:
We've gotten Jint serialization faster between this PR's version and current main (unreleased). Here are results from your benchmark.

## Jint in this PR

|           Method |       Mean |     Error |    StdDev | Ratio |     Gen0 |     Gen1 |     Gen2 |  Allocated | Alloc Ratio |
|----------------- |-----------:|----------:|----------:|------:|---------:|---------:|---------:|-----------:|------------:|
|   JintSerializer | 4,927.9 us | 119.62 us | 350.84 us |  1.00 | 703.1250 | 578.1250 | 492.1875 | 5287.36 KB |        1.00 |
| CustomSerializer |   404.0 us |   7.09 us |   6.29 us |  0.08 | 101.5625 |  91.3086 |  91.3086 |  537.42 KB |        0.10 |


## Jint `main`

|           Method |     Mean |    Error |   StdDev | Ratio |     Gen0 |    Gen1 |    Gen2 | Allocated | Alloc Ratio |
|----------------- |---------:|---------:|---------:|------:|---------:|--------:|--------:|----------:|------------:|
|   JintSerializer | 920.6 us | 14.23 us | 13.31 us |  1.00 | 108.3984 | 98.6328 | 98.6328 | 523.59 KB |        1.00 |
| CustomSerializer | 411.1 us |  7.93 us | 10.58 us |  0.44 | 102.5391 | 92.2852 | 92.2852 | 537.41 KB |        1.03 |


Getting faster from here might be difficult as Jint needs to check more things like object implementing `toJSON` etc, which your implementation omits.

timothycoleman:
Thanks for the PR, comments, and explanation @lahma that's really helpful.

What's the name of the spec that governs serializing objects to json? We'll take a closer look to see what might apply to our use case

lahma:
Here's the official spec: https://tc39.es/ecma262/multipage/structured-data.html#sec-json.stringify

PR #3792, opened 2023-03-20
Added es version info in gossip message

hunter1703:
Added: ES version in gossip message between nodes
Added: Every node will monitor versions of other alive nodes; if version mismatch detected in cluster, nodes will log this
Added: "/gossip" endpoint will also have ES version info


src/EventStore.Core/Cluster/MemberInfo.cs
```diff
@@ -59,15 +61,15 @@ public class MemberInfo : IEquatable<MemberInfo> {
 			int epochNumber,
 			Guid epochId,
 			int nodePriority,
-			bool isReadOnlyReplica) {
+			bool isReadOnlyReplica, string version = null) {
```
hayley-jean, line 64: This name should be kept consistent
```suggestion
			bool isReadOnlyReplica, string esVersion = null) {
```

src/EventStore.Core/Services/Gossip/GossipServiceBase.cs
```diff
@@ -168,10 +168,18 @@ public abstract class GossipServiceBase : IHandle<SystemMessage.SystemInit>,
 
 			message.Envelope.ReplyWith(new GossipMessage.SendGossip(_cluster, _memberInfo.HttpEndPoint));
 
-			if (_cluster.HasChangedSince(oldCluster))
+			if (_cluster.HasChangedSince(oldCluster)) {
```
hayley-jean, line 171: Something seems to have affected the check for whether a cluster has changed.
I see the full `CLUSTER HAS CHANGED` log every time gossip occurs when I've got a node running a different version to the others, even though nothing in the cluster has actually changed.

src/EventStore.Core/Services/Gossip/GossipServiceBase.cs
```diff
@@ -168,10 +168,18 @@ public abstract class GossipServiceBase : IHandle<SystemMessage.SystemInit>,
...
 				LogClusterChange(oldCluster, _cluster, $"gossip received from [{message.Server}]");
+				PublishIfClusterHasMultipleVersions(_cluster);
```
hayley-jean, line 173: I think that this call should be moved out of the "cluster has changed" if check. The different versions warning should be published whenever a gossip comes back with a different version, even if the cluster has not changed.

hayley-jean (changes requested):

src/EventStore.Core/Services/Gossip/GossipServiceBase.cs
hunter1703: Would "spamming" the log be okay? Unless something is changed in cluster, logging again and again doesn't give any extra information.

src/EventStore.Core/Cluster/MemberInfo.cs
hunter1703: changed it

src/EventStore.Core/Services/Gossip/GossipServiceBase.cs
hunter1703: Made changes so that version will be updated only when communicating with the peer directly i.e. when a node receives a gossip, it will only update version of the gossip sender node.
Since version info is not critically important and mostly static, this should be fine.

hunter1703: Changed the behavior : 

1. If gossip is from new node, I will update version info from gossip
2. If gossip is from old node, I will retain previous version info

src/EventStore.Core/Services/Gossip/GossipServiceBase.cs
pvanbuijtene: Think spamming in this case is what we want to "encourage" people to upgrade 🤔 unless it turns into flooding 😅

src/EventStore.Core/Services/Gossip/GossipServiceBase.cs
```diff
@@ -383,5 +401,16 @@ public abstract class GossipServiceBase : IHandle<SystemMessage.SystemInit>,
...
 		}
+		
+		private void PublishIfClusterHasMultipleVersions(ClusterInfo cluster) {
+			List<MemberInfo> aliveMembers = cluster.Members.Where(memberInfo => memberInfo.IsAlive).ToList();
+			int numDistinctVersions = aliveMembers.Select(memberInfo => memberInfo.ESVersion).Where(esVersion => !VersionInfo.UnknownVersion.Equals(esVersion)).Distinct().Count();
+
+			if (numDistinctVersions > 1) {
+				Dictionary<EndPoint, string> ipAndVersion = aliveMembers.ToDictionary(memberInfo => memberInfo.HttpEndPoint,
+					memberInfo => memberInfo.ESVersion);
+				_bus.Publish(new ClusterStateMessage.MultipleVersionsOnNodes(ipAndVersion));
```
pvanbuijtene, line 412: It looks like this is done for testing, would it be an idea to have a separate class for this logic and use/subscribe to already available messages?

src/EventStore.Core/Cluster/ClientClusterInfo.cs
```diff
@@ -108,7 +112,7 @@ public class ClientMemberInfo {
 					$"HttpEndPointIp: {HttpEndPointIp}, HttpEndPointPort: {HttpEndPointPort}, " +
 					$"LastCommitPosition: {LastCommitPosition}, WriterCheckpoint: {WriterCheckpoint}, ChaserCheckpoint: {ChaserCheckpoint}, " +
 					$"EpochPosition: {EpochPosition}, EpochNumber: {EpochNumber}, EpochId: {EpochId:B}, NodePriority: {NodePriority}, " +
-					$"IsReadOnlyReplica: {IsReadOnlyReplica}";
+					$"IsReadOnlyReplica: {IsReadOnlyReplica}, $ESVersion: ${ESVersion}";
```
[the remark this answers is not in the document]
hunter1703: my bad. fixed.

src/EventStore.Core.Tests/Services/GossipService/NodeGossipServiceTests.cs
```diff
@@ -338,7 +348,463 @@ public class when_gossip_received_with_older_timestamp_about_peer_node :
...
+				{ _currentNode.HttpEndPoint, VersionInfo.DefaultVersion }
+			};
+			return endpointVsVersion;
+		}
+
+		[Test]
+		public void should_detect_version_mismatch() {
+			ExpectMessages(new GossipMessage.GossipUpdated(GetExpectedClusterInfo()));
+			
+			Dictionary<EndPoint, string> ipAddressVsVersion = ClusterStateChangeListener.GetIPAddressVsVersion(GetExpectedClusterInfo(), out int numDistinctKnownVersions);
```
pvanbuijtene, line 476: Nothing mayor, but would be nice to have these in a separate file same.

pvanbuijtene (commented):
I see what you mean with the logging, every second is a bit much😅
Maybe should limit a bit, will ask during standup.

src/EventStore.Common/Log/ThrottledLog.cs
```diff
@@ -0,0 +1,61 @@
...
+	
+	public bool Fatal(string message) {
+		bool canLog = CanLog();
+		if (canLog) _log.Warning(message);
+		return canLog;
+	}
+	
+	public bool Information(string message) {
+		bool canLog = CanLog();
+		if (canLog) _log.Warning(message);
```
hayley-jean, line 33:
```suggestion
		if (canLog) _log.Information(message);
```
The other log levels need to be changed as well

src/EventStore.Common/Log/ThrottledLog.cs
```diff
@@ -0,0 +1,61 @@
...
+		return canLog;
+	}
+
+	private bool CanLog() {
+		var currentTime = DateTime.Now.Ticks;
+		bool canLog = false;
+
+		// double-checked locking
+		if (currentTime - _lastLogged >= _duration) {
+			lock (_log) {
```
hayley-jean, line 49: Should the lock on `_log` be taken outside of the above `if` statement? `_lastLogged` is used in the condition but may change between checking it, taking the lock, and then updating it later

hayley-jean (changes requested):

src/EventStore.Common/Log/ThrottledLog.cs
hunter1703: Oops. Fixed it. My bad.

src/EventStore.Core.Tests/Services/GossipService/NodeGossipServiceTests.cs
hunter1703: Makes sense. I have created a separate test file for these tests.

src/EventStore.Core/Services/Gossip/ClusterStateChangeListener.cs
```diff
@@ -0,0 +1,35 @@
...
+using EventStore.Common.Log;
+using EventStore.Common.Utils;
+using EventStore.Core.Bus;
+using EventStore.Core.Cluster;
+using EventStore.Core.Messages;
+
+namespace EventStore.Core.Services.Gossip;
+
+public class ClusterStateChangeListener : IHandle<GossipMessage.GossipUpdated> {
+	private static readonly ThrottledLog<ClusterStateChangeListener> Log = new(TimeSpan.FromSeconds(1));
```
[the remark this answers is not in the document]
hunter1703: Changed the default to 1 minute

src/EventStore.Common/Log/ThrottledLog.cs
hayley-jean: Ah sorry, I missed that second check there, thanks!

hayley-jean (approved):

pvanbuijtene (approved):

Issue #3534, opened 2022-07-25
Stop ESDB from starting when --config specified but not found


ylorph:
**Is your feature request related to a problem? Please describe.**
When ESDB is started with --config specified if the file is not found it starts with all defaults on.
 and stops because the default is secure and no config is found 
 `Invalid Configuration: A certificate is required unless insecure mode (--insecure) is set`

**Describe the solution you'd like**
The server should not start and tell "config file not found"
since the error given is misleading

**Describe alternatives you've considered**

**Additional context**

jamesmintram:
I took a look at this and see that a solution is not obvious. 

By the time it comes to loading the config yaml, all of the configuration sources have been merged (Default, Command Line and Env) The Default configuration contains an entry for Config. 

So a simple "The config file does not exist" check is not enough to determine an error due to a default being present.

One approach to this would be to check if the value of Config was equal to the Default value of config, and if not - treat a missing file as an error.

Alternatively, one could build another config containing all user/environment supplied config options and use that to determine if Config had been explicitly specified.

I prefer the second option, so I'll look at creating a fix based on that - unless there are any objections? :)

PR #3794, opened 2023-03-21
Add check for config file not found at EventStoreDB start up

dhingrak:
Changed: Ensure the config file is found on startup if it has been specified

Fixed: #3534 

This PR added a check for the config file path at EventStoreDB start up.

### DIFF
--- a/src/EventStore.Common.Tests/Configuration/ConfigurationRootExtensionsTest.cs
+++ b/src/EventStore.Common.Tests/Configuration/ConfigurationRootExtensionsTest.cs
@@ -1,10 +1,13 @@
+using System.Collections;
+using EventStore.Common.Configuration;
 using Microsoft.Extensions.Configuration;
 
 namespace EventStore.Common.Tests.Configuration;
 
 public class ConfigurationRootExtensionsTest {
-	private static string GOSSIP_SEED = "GossipSeed";
-
+	private const string GOSSIP_SEED = "GossipSeed";
+	private const string CONFIG_FILE_KEY = "Config";
+	
 	[Fact]
 	public void successful_comma_separated_value() {
 		var config = new Dictionary<string, string> {
@@ -46,5 +49,46 @@ public class ConfigurationRootExtensionsTest {
 		Assert.Throws<ArgumentException>(() =>
 			EventStore.Common.Configuration.ConfigurationRootExtensions.GetCommaSeparatedValueAsArray(
 				configuration, "GossipSeed"));
+	}
+
+	[Fact]
+	public void user_specified_config_file_through_environment_variables_returns_true() {
+		IDictionary environmentVariables = new Dictionary<string, string>();
+		environmentVariables.Add("EVENTSTORE_CONFIG", "pathToConfigFileOnMachine");
+
+		var configurationRoot = new ConfigurationBuilder()
+			.Add(new EnvironmentVariablesSource(environmentVariables))
+			.Build();
+
+		var result = configurationRoot.IsUserSpecified(CONFIG_FILE_KEY);
+
+		Assert.True(result);
+	}
+	
+	[Fact]
+	public void user_specified_config_file_through_command_line_returns_true() {
+		var args = new string[] {
+			"--config=pathToConfigFileOnMachine"
+		};
+
+		var configurationRoot = new ConfigurationBuilder()
+			.Add(new CommandLineSource(args))
+			.Build();
+
+		var result = configurationRoot.IsUserSpecified(CONFIG_FILE_KEY);
+
+		Assert.True(result);
+	}
+
+	[Fact]
+	public void user_did_not_specified_config_file_returns_false() {
+		var configurationRoot = new ConfigurationBuilder()
+			.Add(new DefaultSource(new Dictionary<string, object> {
+			}))
+			.Build();
+
+		var result = configurationRoot.IsUserSpecified(CONFIG_FILE_KEY);
+
+		Assert.False(result);
 	}
 }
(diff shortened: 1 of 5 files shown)


timothycoleman (changes requested):
we could consider what happens if the file exists but cannot be read due to permissions, or exists when we checked it but then is deleted before we read the configuration from it

i think really it would be better to have the configuration read in the normal way, but throw an exception if this fails. the 'Optional' setting on the yamlSource could be useful, there is also a OnLoadException callback we could specify but im not sure we'll need to

src/EventStore.Common/Configuration/ConfigurationBuilderExtensions.cs
```diff
@@ -16,15 +17,25 @@ public static class ConfigurationBuilderExtensions {
...
 			yamlSource.ResolveFileProvider();
 			builder.Sources.Insert(1, yamlSource);
 
 			return builder;
 		}
+
+		private static bool CompareConfigFilePath(IEnumerable<KeyValuePair<string, object?>> defaultValues, string? configurationPath) {
+			var defaultValuesDictionary = defaultValues.ToDictionary(x => x.Key, x => x.Value);
+
+			if (defaultValuesDictionary.TryGetValue("Config", out var defaultConfigurationPath)) {
```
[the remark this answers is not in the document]
timothycoleman: (although i do see that this was also the case before this pr)

src/EventStore.Common/Configuration/ConfigurationBuilderExtensions.cs
```diff
@@ -16,15 +17,25 @@ public static class ConfigurationBuilderExtensions {
...
+				return string.Equals(configurationPath, defaultConfigurationPath?.ToString());
```
timothycoleman, line 36: i think the ReferenceEquals that we had before would be preferable (perhaps with a comment), so that if the user happens to specify a config path that is the same as the default then they still get the advantage of this existence check

timothycoleman (changes requested):
looks good, couple of comments inline, and it would be nice to see a unit test that covers this too :+1:

src/EventStore.Common/Configuration/ConfigurationBuilderExtensions.cs
timothycoleman: actually, come to think of it, we have `CheckProvidersForEnvironmentVariables` in ConfigurationRootExtensions. Maybe we could add a method in there that checks if a value has been overridden from the default. that would be less cryptic than ReferenceEquals

src/EventStore.Common/Configuration/ConfigurationBuilderExtensions.cs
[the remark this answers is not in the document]
dhingrak: The ClusterVNodeOptions.ApplicationOptions.Config exits in a different project (EventStore.Core) . Usually we can access the public objects through using namesapce and a reference but we already have a reference of EventStore.Common project in EventStore.Core. So for this reason we cannot access Applicationoptions.Config in ConfigurationBuilderExtensions. 

Please let me know if I am getting this correct.

src/EventStore.Common/Configuration/ConfigurationBuilderExtensions.cs
dhingrak: > i think the ReferenceEquals that we had before would be preferable (perhaps with a comment), so that if the user happens to specify a config path that is the same as the default then they still get the advantage of this existence check

The  default configuration ReferenceEquals check was failing while running the tests in Ubuntu. The following check supposed to return true but it's returning false and as per Logic I was seeing "Config not found" in the logs. 
ReferenceEquals(configurationPath, defaultConfigurationPath)

timothycoleman:
hey i pushed another commit to a branch based on this one and opened a pr here https://github.com/EventStore/EventStore/pull/3805 it deals with the comments i made above in what i think is a reasonable way. treat it as a suggestion

after that i think we just need a test or two

dhingrak:
> hey i pushed another commit to a branch based on this one and opened a pr here #3805 it deals with the comments i made above in what i think is a reasonable way. treat it as a suggestion
> 
> after that i think we just need a test or two

Thank you so much Tim. I will check the suggestions.

src/EventStore.Common.Tests/Configuration/ConfigurationRootExtensionsTest.cs
```diff
@@ -47,4 +52,47 @@ public class ConfigurationRootExtensionsTest {
...
+
+		var args = new string[] {
+			"--config=pathToConfigFileOnMachine"
+		};
+
+		_configurationRoot = new ConfigurationBuilder()
+			.Add(new CommandLineSource(args))
+			.Build();
+
+		var result = ConfigurationRootExtensions.IsUserSpecified(_configurationRoot, _configFileKey);
```
timothycoleman, line 82:
```suggestion
		var result = _configurationRoot.ConfigurationRootExtensions.IsUserSpecified(_configFileKey);
```

just cause this is the way we're expecting the method to really be used (same in the other tests too)

src/EventStore.Common.Tests/Configuration/ConfigurationRootExtensionsTest.cs
```diff
@@ -1,10 +1,15 @@
...
+using EventStore.Common.Configuration;
 using Microsoft.Extensions.Configuration;
+using ConfigurationRootExtensions = EventStore.Common.Configuration.ConfigurationRootExtensions;
 
 namespace EventStore.Common.Tests.Configuration;
 
 public class ConfigurationRootExtensionsTest {
 	private static string GOSSIP_SEED = "GossipSeed";
-
+	private IConfigurationRoot? _configurationRoot;
```
timothycoleman, line 10: this can just be a local variable

timothycoleman (approved):
thanks, good work :+1:

i made the small tidying up changes that i suggested above so that i can approve it before going on annual leave

dhingrak:
> thanks, good work 👍
> 
> i made the small tidying up changes that i suggested above so that i can approve it before going on annual leave

Thanks a lot Tim.

hayley-jean (approved):

PR #3796, opened 2023-03-24
Implicit transaction bug fixes (DB-65)

shaan1337:
Fixed: Implicit transaction bugs on index rebuild

On index rebuild at startup implicit transaction events were being treated as single events due to the `HasAnyOf` instead of the `HasAllOf` flag check method (`SingleWrite` is an aggregate flag containing `Data`, `TransactionBegin` and `TransactionEnd`). So the `else` block was executed only when the prepare had the `StreamDelete` flag (hard deleting a stream in a transaction, i.e almost never)

Given that the writer checkpoint is usually flushed only after writing all the implicit transaction events, this bug has an effect only at chunk boundaries because the writer checkpoint is flushed when a chunk is closed. Furthermore a crash would need to occur at that precise moment (or slightly after but before the last event)

The fix also clears `committedPrepares` whenever a new transaction begins to prevent incomplete transactions from being merged into a transaction adjacent to it.

### DIFF
--- a/src/EventStore.Core/Services/Storage/ReaderIndex/IndexCommitter.cs
+++ b/src/EventStore.Core/Services/Storage/ReaderIndex/IndexCommitter.cs
@@ -133,7 +133,7 @@ namespace EventStore.Core.Services.Storage.ReaderIndex {
 				var fullRebuild = startPosition == 0;
 				reader.Reposition(startPosition);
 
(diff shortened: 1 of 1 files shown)


shaan1337:
Will be superseded by part 3 of https://github.com/EventStore/EventStore/pull/3808

PR #3799, opened 2023-03-28
Prevent torn implicit transactions (DB-64)

shaan1337:
Fixed: Prevent torn implicit transactions

- Move writer checkpoint across transaction boundaries, not across each log record
- Do not flush writer checkpoint when completing a chunk


hayley-jean:
Closing this in favour of https://github.com/EventStore/EventStore/pull/3808

PR #3803, opened 2023-03-29
Throw error on trying to emit events to deleted stream

hunter1703:
fixed: wrong error being detected and thrown

### DIFF
--- a/src/EventStore.Projections.Core/Services/Processing/EmittedStream.cs
+++ b/src/EventStore.Projections.Core/Services/Processing/EmittedStream.cs
@@ -270,7 +270,8 @@ namespace EventStore.Projections.Core.Services.Processing {
 			if (message.CorrelationId != _pendingRequestCorrelationId)
 				return;
 			if (message.Result == ReadStreamResult.StreamDeleted) {
-				throw new InvalidOperationException($"Stream : {_streamId} is deleted. Cannot emit events to it");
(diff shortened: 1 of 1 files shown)


src/EventStore.Projections.Core/Services/Processing/EmittedStream.cs
```diff
@@ -269,6 +269,10 @@ public class StreamMetadata {
 				return;
 			if (message.CorrelationId != _pendingRequestCorrelationId)
 				return;
+			if (message.Result == ReadStreamResult.StreamDeleted) {
+				throw new InvalidOperationException($"Stream : {_streamId} is deleted. Cannot emit events to it");
```
hayley-jean, line 273: The error message looks much better, but the projection is no longer stopping because of it.
This can be dangerous because while the projection will continue projecting to other streams, it will no longer emit events to the deleted stream, and the user may never be notified of this.

If you call `Failed` here like the other errors below, the projection will correctly report the error and stop:

```suggestion
				Failed($"Stream : {_streamId} is deleted. Cannot emit events to it");
				return;
```

hayley-jean (changes requested):

src/EventStore.Projections.Core/Services/Processing/EmittedStream.cs
hunter1703: Makes sense. I have updated it.

pvanbuijtene (approved):

Issue #3736, opened 2023-02-14
EventStore server crashes when failing to set attributes on PTable file


itdevops-channelcapital:
**Describe the bug**
The EventStore server crashes when creating a PTable once the MaxMemTableSize has been exceeded.

**To Reproduce**
1. Create a PVC (Persistent Volume and Claim in AKS connected to an Azure File Share in an Azure Storage Account. The PVC should have the following mount options - file_mode=0777, dir_mode=0777, mfsymlinks, uid=0, gid=0, actimeo=30, cache=strict, nosharesock.
3. Install EventStore in AKS (Azure Kubernetes Services) using the PVC created above.
4. Create eventstreams until the MaxMemeTableSize is exceeded.

**Expected behavior**
The PTable is created, whether file attributes can be set or not.

**Actual behavior**
The EventStore server crashes when trying to set file attributes on the PTable index file with an UnauthorizedAccessException.

**Config/Logs/Screenshots**
We are using the default MaxMemTableSize of 1000000.

**EventStore details**
 - EventStore server version: [22.10.0-bionic](https://hub.docker.com/layers/eventstore/eventstore/22.10.0-bionic/images/sha256-f8e27176956427e6be00bb634526c32fa6b32c8b856cecad6b8c790a6aceb28b?context=explore)

 - Operating system: AKS - Node (Ubuntu 18.04).

 - EventStore client version (if applicable):

**Additional context**
We have narrowed down the error to this file: https://github.com/EventStore/EventStore/blob/master/src/EventStore.Core/Index/PTable.cs
Line 131.
Based on the logs, the exception is happening when trying to set the file attributes. This is not possible when using a PVC as the mount options dictate the read/write settings of files and as such a file cannot be set to read-only.

Not sure of the impact, but we thought of using the `Helper.EatException(() => {...});` function as in https://github.com/EventStore/EventStore/blob/master/src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs on line 513.

As a temporary workaround, we have increased the MaxMemTableSize.

PR #3807, opened 2023-03-31
Ignore exceptions when failing to change PTable index file permissions

itdevops-channelcapital:
Fixed: Ignore exceptions when failing to change PTable index file permissions

This change will ignore file permission exceptions when trying to change the file attributes on a Unix system with a user that does not have access to change file permissions.

### DIFF
--- a/src/EventStore.Core/Index/PTable.cs
+++ b/src/EventStore.Core/Index/PTable.cs
@@ -128,7 +128,10 @@ namespace EventStore.Core.Index {
 			var sw = Stopwatch.StartNew();
 			_size = new FileInfo(_filename).Length;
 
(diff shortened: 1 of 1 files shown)


itdevops-channelcapital:
Relates to issue #3736

shaan1337 (approved):
looks good to me :+1: thanks for your contribution!

if I get it correctly, it happens when using an SMB/CIFS server that doesn't have POSIX compatibility and thus it's not possible to adjust file permissions/attributes?

hayley-jean (approved):
Thanks @itdevops-channelcapital !

For future reference, the chunks already ignore exceptions when setting the readonly attribute. See [here](https://github.com/EventStore/EventStore/blob/d975952efef48f0ece97de32e734eb789c062c1f/src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs#L513)

PR #3808, opened 2023-04-05
Torn transactions (1/3): Prevent at chunk boundaries on the leader node (DB-64)

shaan1337:
Fixed: Prevent risk of implicit transactions being partially written when crossing a chunk boundary on the leader node

Parts 2 & 3 are planned to be as follows:
2. Prevent torn transactions when replicating
3. Handle existing torn transactions during index rebuild at startup

### DIFF
--- a/src/EventStore.Core/Services/Storage/StorageWriterService.cs
+++ b/src/EventStore.Core/Services/Storage/StorageWriterService.cs
@@ -241,6 +241,9 @@ namespace EventStore.Core.Services.Storage {
 				_vnodeState != VNodeState.PreReadOnlyReplica)
 				throw new Exception(string.Format("{0} appeared in {1} state.", message.GetType().Name, _vnodeState));
 
+			if (Writer.HasOpenTransaction())
+				throw new InvalidOperationException("Writer has an open transaction.");
+
 			if (Writer.FlushedPosition != Writer.Position) {
 				Writer.Flush();
 				Bus.Publish(new ReplicationTrackingMessage.WriterCheckpointFlushed());
@@ -293,13 +296,15 @@ namespace EventStore.Core.Services.Storage {
 
 				var prepares = new List<IPrepareLogRecord<TStreamId>>();
 				if (msg.Events.Length > 0) {
+					OpenTransaction();
+
 					// get/write the eventtypes first because they go in a different stream
 					var eventTypes = new TStreamId[msg.Events.Length]; // todo: pool
 					for (int i = 0; i < msg.Events.Length; ++i) {
 						var evnt = msg.Events[i];
-						eventTypes[i] = GetOrWriteEventType(evnt.EventType, ref logPosition);
+						eventTypes[i] = GetOrWriteEventType(evnt.EventType, ref logPosition, inTransaction: true);
 					}
-					
+
 					var transactionPosition = logPosition;
 					for (int i = 0; i < msg.Events.Length; ++i) {
 						var evnt = msg.Events[i];
@@ -316,13 +321,16 @@ namespace EventStore.Core.Services.Storage {
 						var res = WritePrepareWithRetry(
 							LogRecord.Prepare(_recordFactory, logPosition, msg.CorrelationId, evnt.EventId,
 								transactionPosition, i, streamId,
-								expectedVersion, flags, eventTypes[i], evnt.Data, evnt.Metadata));
+								expectedVersion, flags, eventTypes[i], evnt.Data, evnt.Metadata),
+							inTransaction: true);
 						logPosition = res.NewPos;
 						if (i == 0)
 							transactionPosition = res.WrittenPos;
 						// transaction position could be changed due to switching to new chunk
 						prepares.Add(res.Prepare);
 					}
+
+					CommitTransaction();
 				} else {
 					WritePrepareWithRetry(
 						LogRecord.Prepare(_recordFactory, logPosition, msg.CorrelationId, Guid.NewGuid(), logPosition, -1,
@@ -348,7 +356,7 @@ namespace EventStore.Core.Services.Storage {
 			}
 		}
 
-		private TStreamId GetOrWriteEventType(string eventType, ref long logPosition) {
+		private TStreamId GetOrWriteEventType(string eventType, ref long logPosition, bool inTransaction = false) {
 			_eventTypeIndex.GetOrReserveEventType(
 				recordFactory: _recordFactory,
 				eventType: eventType,
@@ -358,7 +366,7 @@ namespace EventStore.Core.Services.Storage {
 
 			if (eventTypeRecord != null)
 			{
-				var result = WritePrepareWithRetry(eventTypeRecord);
+				var result = WritePrepareWithRetry(eventTypeRecord, inTransaction);
 				logPosition = result.NewPos;
 			}
 			
@@ -657,11 +665,25 @@ namespace EventStore.Core.Services.Storage {
 			}
 		}
 
-		private WriteResult WritePrepareWithRetry(IPrepareLogRecord<TStreamId> prepare) {
+		private bool WritePrepare(ILogRecord prepare, bool inTransaction, out long newPos) {
+			return inTransaction ? Writer.WriteToTransaction(prepare, out newPos) : Writer.Write(prepare, out newPos);
+		}
+
+		private void CompleteChunk(bool inTransaction, out long newPos) {
+			if (inTransaction)
+				Writer.CompleteChunkInTransaction();
+			else
+				Writer.CompleteChunk();
+
+			newPos = Writer.Position;
+		}
+
+		private WriteResult WritePrepareWithRetry(IPrepareLogRecord<TStreamId> prepare, bool inTransaction = false) {
 			long writtenPos = prepare.LogPosition;
 			long newPos;
 			var record = prepare;
-			if (!Writer.Write(prepare, out newPos)) {
+
+			if (!WritePrepare(prepare, inTransaction, out newPos)) {
 				var transactionPos = prepare.TransactionPosition == prepare.LogPosition
 					? newPos
 					: prepare.TransactionPosition;
@@ -671,7 +693,7 @@ namespace EventStore.Core.Services.Storage {
 					transactionPosition: transactionPos);
 
 				writtenPos = newPos;
-				if (!Writer.Write(record, out newPos)) {
+				if (!WritePrepare(record, inTransaction, out newPos)) {
 					throw new Exception(
 						string.Format("Second write try failed when first writing prepare at {0}, then at {1}.",
 							prepare.LogPosition,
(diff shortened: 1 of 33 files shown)


src/EventStore.Core/TransactionLog/ITransactionFileWriter.cs
timothycoleman, line 1: can remove `using EventStore.Core.TransactionLog.Checkpoint;` now

src/EventStore.Core/TransactionLog/Chunks/TFChunkWriter.cs
```diff
@@ -43,35 +45,45 @@ public class TFChunkWriter : ITransactionFileWriter {
 		public bool Write(ILogRecord record, out long newPos) {
 			var result = _currentChunk.TryAppend(record);
 			if (result.Success)
-				_writerCheckpoint.Write(result.NewPosition + _currentChunk.ChunkHeader.ChunkStartPosition);
+				_nextRecordPosition = result.NewPosition + _currentChunk.ChunkHeader.ChunkStartPosition;
 			else
 				CompleteChunk(); // complete updates checkpoint internally
```
timothycoleman, line 50: i think the comment here is out of date now

src/EventStore.Core/TransactionLog/Chunks/TFChunkWriter.cs
```diff
@@ -43,35 +45,45 @@ public class TFChunkWriter : ITransactionFileWriter {
...
+			newPos = _nextRecordPosition;
 			return result.Success;
 		}
 
-		public void CompleteChunk() {
+		public void Commit() {
+			_writerCheckpoint.Write(_nextRecordPosition);
+		}
+
+		public void CompleteChunk(bool commitAndFlush = false) {
```
timothycoleman, line 59: hesitant about the default parameter value here and in CompleteReplicatedRawChunk, i think better to be explicit?

src/EventStore.Core/TransactionLog/ITransactionFileWriter.cs
```diff
@@ -6,9 +6,12 @@ namespace EventStore.Core.TransactionLog {
 	public interface ITransactionFileWriter : IDisposable {
 		void Open();
 		bool Write(ILogRecord record, out long newPos);
+		void Commit();
```
timothycoleman, line 9: We could consider having something like a Begin that just checks that there aren't already uncommitted writes (that might therefore end up being committed accidentally as part of a separate transaction)

src/EventStore.Core/TransactionLog/ITransactionFileWriter.cs
shaan1337: fixed :+1:

src/EventStore.Core/TransactionLog/Chunks/TFChunkWriter.cs
shaan1337: fixed :+1:

src/EventStore.Core/TransactionLog/Chunks/TFChunkWriter.cs
shaan1337: fixed, it's no longer necessary now with Begin-Commit, we can commit inside CompleteChunk() if there is no ongoing transaction

src/EventStore.Core/TransactionLog/ITransactionFileWriter.cs
shaan1337: great idea :+1: i've pushed the required changes

src/EventStore.Core/TransactionLog/Chunks/TFChunkWriter.cs
```diff
@@ -40,24 +43,50 @@ public class TFChunkWriter : ITransactionFileWriter {
...
 
+		public void Commit() {
+			if (!_inTransaction)
+				throw new Exception("Attempt to commit while there is no ongoing transaction.");
+
+			_inTransaction = false;
+			CommitInternal();
+		}
+
+		private void CommitInternal(bool flush = false) {
```
timothycoleman, line 75: personally i think this would be better if the parameter was non-optional, and be explicit about the intention on every call (theres only four of them anyway)

src/EventStore.Core/Services/Storage/StorageWriterService.cs
```diff
@@ -236,7 +236,12 @@ public class StorageWriterService<TStreamId> : IHandle<SystemMessage.SystemInit>
 				_vnodeState != VNodeState.PreReadOnlyReplica)
 				throw new Exception(string.Format("{0} appeared in {1} state.", message.GetType().Name, _vnodeState));
 
-			if (Writer.Checkpoint.Read() != Writer.Checkpoint.ReadNonFlushed()) {
+			if (Writer.CommittedPosition != Writer.NextRecordPosition)
+				throw new Exception("Invariant violation: " +
```
timothycoleman, line 240: probably an invalidoperationexception

i'd be tempted to move this check into the Writer itself and add a hint about why they might be out of sync 'There may be an open transaction' or something like that

src/EventStore.Core/TransactionLog/Chunks/TFChunkWriter.cs
```diff
@@ -40,24 +43,50 @@ public class TFChunkWriter : ITransactionFileWriter {
...
+
 		public bool Write(ILogRecord record, out long newPos) {
 			var result = _currentChunk.TryAppend(record);
 			if (result.Success)
-				_writerCheckpoint.Write(result.NewPosition + _currentChunk.ChunkHeader.ChunkStartPosition);
+				_nextRecordPosition = result.NewPosition + _currentChunk.ChunkHeader.ChunkStartPosition;
 			else
-				CompleteChunk(); // complete updates checkpoint internally
-			newPos = _writerCheckpoint.ReadNonFlushed();
+				CompleteChunk();
```
timothycoleman, line 58:
```suggestion
				CompleteChunk(); // complete updates _nextRecordPosition internally
```

on more careful thought it should read like this actually

src/EventStore.Core/TransactionLog/Chunks/TFChunkWriter.cs
```diff
@@ -40,24 +43,50 @@ public class TFChunkWriter : ITransactionFileWriter {
...
+
+			if (!_inTransaction)
```
timothycoleman, line 60: it took me a while to put my finger on what was making me uncertain about this, but:   i think if we want to give the user of TFChunkWriter a way to use it without having to Open and Commit transactions, then we should make a separate write method for that.

Or more likely have a separate method for writing to a transaction and then have the regular Write method call OpenTransaction, WriteToTransaction and CommitTransaction.

The risk with the way that it is here currently is that if something has somehow left a transaction open, then the EpochManager and PartitionManager will just call Write and Flush, expecting the data to then be safely on disk, but in fact it will just be added to the open transaction and not flushed (and no problem detected)

src/EventStore.Core/TransactionLog/Chunks/TFChunkWriter.cs
```diff
@@ -40,24 +43,50 @@ public class TFChunkWriter : ITransactionFileWriter {
...
+			_writerCheckpoint.Write(_nextRecordPosition);
+			if (flush)
+				_writerCheckpoint.Flush();
+		}
+
 		public void CompleteChunk() {
 			var chunk = _currentChunk;
 			_currentChunk = null; // in case creation of new chunk fails, we shouldn't use completed chunk for write
 
 			chunk.Complete();
```
timothycoleman, line 85: is there a problem in the end with the chunk becomming readonly with the writercheckpoint still in it (in case we end up not advancing the writer checkpoint)?

src/EventStore.Core/TransactionLog/Chunks/TFChunkWriter.cs
```diff
@@ -40,24 +43,50 @@ public class TFChunkWriter : ITransactionFileWriter {
 			// DO NOTHING
 		}
 
+		public void Begin() {
+			if (_inTransaction)
+				throw new Exception("Attempt to begin a new transaction while there is an ongoing transaction.");
```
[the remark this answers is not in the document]
shaan1337: fixed :+1:

src/EventStore.Core/TransactionLog/Chunks/TFChunkWriter.cs
shaan1337: fixed

src/EventStore.Core/TransactionLog/Chunks/TFChunkWriter.cs
shaan1337: great point, fixed :+1:

src/EventStore.Core/TransactionLog/Chunks/TFChunkWriter.cs
shaan1337: if we end up not advancing the writer checkpoint, it won't be a problem because on startup, the last chunk is opened as an ongoing chunk file which is not read-only (even if the chunk already has a footer)

src/EventStore.Core/Services/Storage/StorageWriterService.cs
shaan1337: fixed: added a method named: `HasOpenTransaction()` to `ITransactionFileWriter`

src/EventStore.Core/TransactionLog/Chunks/TFChunkWriter.cs
```diff
@@ -41,43 +43,80 @@ public class TFChunkWriter : ITransactionFileWriter {
 		}
 
 		public bool Write(ILogRecord record, out long newPos) {
+			OpenTransaction();
+			bool wasWritten = WriteToTransaction(record, out newPos);
+			CommitTransaction(flush: !wasWritten);
```
timothycoleman, line 48: i dont think we need to flush here. this is for the case where we try to write but the chunk is full. before this PR that would cause a new chunk to be written and the writer checkpoint to be flushed at the start of the new chunk. now though we can leave it unflushed because we'll land in the same case a partially completed transaction: with the writer checkpoint in a previous chunk and the startup routine detecting and removing the new chunk(i write all this to explain it to future tim rather than to explain it to you)

then we can take the Flush call out of Write and just let the two places that want to flush call flush themselves and keep them separate

src/EventStore.Core/Services/Storage/StorageWriterService.cs
```diff
@@ -293,13 +296,18 @@ public class StorageWriterService<TStreamId> : IHandle<SystemMessage.SystemInit>
 
 				var prepares = new List<IPrepareLogRecord<TStreamId>>();
 				if (msg.Events.Length > 0) {
+					var doTransaction = msg.Events.Length > 1;
```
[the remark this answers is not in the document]
shaan1337: fixed :+1:

src/EventStore.Core/TransactionLog/Chunks/TFChunkWriter.cs
shaan1337: fixed :+1: i've also adjusted the commit message to describe the change in behaviour (although I've not added the details)

timothycoleman (approved):
great work :+1:

shaan1337:
> great work +1

thanks for the thorough review and great suggestions!

hayley-jean (approved):
Looks good👍

PR #3812, opened 2023-04-06
Fixed errors while deleting projections without substreams (DB-22)

hunter1703:
Fixed: race condition in ManagedProjection code when deleting a projection
Fixed: projection code used to throw error if projection substreams (e.g. emitted streams, checkpoint streams etc.) did not exist, when deleting projection

### DIFF
--- a/src/EventStore.Projections.Core/Services/Management/ManagedProjection.cs
+++ b/src/EventStore.Projections.Core/Services/Management/ManagedProjection.cs
@@ -515,8 +515,7 @@ namespace EventStore.Projections.Core.Services.Management {
 		}
 
 		private void DeleteIfConditionsAreMet() {
-			Interlocked.Decrement(ref PersistedProjectionState.NumberOfPrequisitesMetForDeletion);
-			if (PersistedProjectionState.NumberOfPrequisitesMetForDeletion <= 0) {
+			if (Interlocked.Decrement(ref PersistedProjectionState.NumberOfPrequisitesMetForDeletion) <= 0) {
 				Deleted = true;
 				Deleting = false;
 				Reply();
@@ -723,6 +722,7 @@ namespace EventStore.Projections.Core.Services.Management {
 			if (Mode == ProjectionMode.Transient) {
 				//TODO: move to common completion procedure
 				_lastWrittenVersion = PersistedProjectionState.Version ?? -1;
+				_pendingWritePersistedState = false;
 				StartOrLoadStopped();
 				return;
 			}
@@ -781,6 +781,14 @@ namespace EventStore.Projections.Core.Services.Management {
 
 		private void DeleteStreamCompleted(ClientMessage.DeleteStreamCompleted message, string streamId,
 			Action completed) {
+			// this piece of code works because, DeleteStream request is always made with ExpectedVersion.Any
+			// this is a workaround to maintain contract and compatibility with TCP/GRPC/Web clients who expect WrongExpectedVersion in response to deleting non-existing streams
+			if (message.Result == OperationResult.WrongExpectedVersion) {
+				// stream was never created
+				_logger.Information("PROJECTIONS: Projection Stream '{stream}' was not deleted since it does not exist", streamId);
+				completed();
+				return;
+			}
 			if (message.Result == OperationResult.Success || message.Result == OperationResult.StreamDeleted) {
 				_logger.Information("PROJECTIONS: Projection Stream '{stream}' deleted", streamId);
 				completed();
@@ -1022,7 +1030,7 @@ namespace EventStore.Projections.Core.Services.Management {
 			if (_lastReplyEnvelope != null)
 				_lastReplyEnvelope.ReplyWith(new ProjectionManagementMessage.Updated(_name));
 			_lastReplyEnvelope = null;
-			if (Deleted) {
+			if (Deleted && !_pendingWritePersistedState) {
 				DisposeCoreProjection();
 				_output.Publish(new ProjectionManagementMessage.Internal.Deleted(_name, Id));
 			}
--- a/src/EventStore.Projections.Core/Services/Management/ManagedProjection.cs
+++ b/src/EventStore.Projections.Core/Services/Management/ManagedProjection.cs
@@ -781,8 +781,9 @@ namespace EventStore.Projections.Core.Services.Management {
 
 		private void DeleteStreamCompleted(ClientMessage.DeleteStreamCompleted message, string streamId,
 			Action completed) {
-			// this piece of code works because, DeleteStream request is always made with ExpectedVersion.Any
-			// this is a workaround to maintain contract and compatibility with TCP/GRPC/Web clients who expect WrongExpectedVersion in response to deleting non-existing streams
+			// currently, WrongExpectedVersion is returned when deleting non-existing streams, even when specifying ExpectedVersion.Any.
+			// it is not too intuitive but changing the response would break the contract and compatibility with TCP/gRPC/web clients or require adding a new error code to all clients.
+			// note: we don't need to check if CurrentVersion == -1 here to make sure it's a non-existing stream since the deletion is done with ExpectedVersion.Any
 			if (message.Result == OperationResult.WrongExpectedVersion) {
 				// stream was never created
 				_logger.Information("PROJECTIONS: Projection Stream '{stream}' was not deleted since it does not exist", streamId);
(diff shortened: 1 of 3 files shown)


src/EventStore.Core/Services/Storage/StorageWriterService.cs
```diff
@@ -447,10 +447,15 @@ public class StorageWriterService<TStreamId> : IHandle<SystemMessage.SystemInit>
 				} else {
 					// SOFT DELETE
 					var metastreamId = _systemStreams.MetaStreamOf(streamId);
-					var expectedVersion = _indexWriter.GetStreamLastEventNumber(metastreamId);
-
-					if (_indexWriter.GetStreamLastEventNumber(streamId) < 0 && expectedVersion < 0) {
-						var result = new CommitCheckResult<TStreamId>(CommitDecision.WrongExpectedVersion, streamId,
+					var metadataLastEventNumber = _indexWriter.GetStreamLastEventNumber(metastreamId);
```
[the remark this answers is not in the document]
hunter1703: @shaan1337 Thanks for pointing out the breaking change. Please let me know what you think of my alternate solution : 

- Instead of messing with projection subsystem code to handle this "special" case (one solution is, as you pointed : check currentversion), what if I intercept reply to client and transform `StreamDeletedException`  to `WrongExpectedVersionException`; something along the lines of following 

```var httpEnvelope = new SendToHttpEnvelope(_networkSendQueue, manager, Format.DeleteStreamCompleted, ConfigureResponse);
var envelope = new CallbackEnvelope(message => {
	if (message is ClientMessage.DeleteStreamCompleted completed && completed.Result == OperationResult.StreamDeleted) {
		httpEnvelope.ReplyWith(new ClientMessage.DeleteStreamCompleted(completed.CorrelationId, OperationResult.WrongExpectedVersion, "Stream is already deleted.", completed.CurrentVersion, completed.PreparePosition, completed.CommitPosition));
	} else {
		httpEnvelope.ReplyWith(message);
	}
});```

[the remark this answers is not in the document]
Comment (hunter1703) [reply src/EventStore.Core/Services/Storage/StorageWriterService.cs:450]: Making changes to projection subsytem to handle "special" case and my solution, both, to me, seems a little "hacky".
But my solution has advantage that projection code remains clean and logical.
Please let me know.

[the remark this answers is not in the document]
Comment (timothycoleman) [reply src/EventStore.Core/Services/Storage/StorageWriterService.cs:450]: > but i think the reason was to keep things simple and to avoid having to update all clients with a new error code. can you please confirm?

yes i think that's right. it would be weird if soft deleting a stream that doesn't exist actually _created_ a stream (the metadatastream) so we reject that append, and i think wrongexpectedversion is the only existing error code that made some sense - but i agree that it is confusing in the ExpectedVersion.Any case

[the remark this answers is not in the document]
Comment (hunter1703) [reply src/EventStore.Core/Services/Storage/StorageWriterService.cs]: As per @shaan1337 's suggestion, I have made changes to projection subsystem instead, to maintain contract with TCP/GRPC/Web clients

File: src/EventStore.Projections.Core/Services/Management/ManagedProjection.cs [hunk 2]
```diff
@@ -1022,7 +1030,7 @@ private readonly
 			if (_lastReplyEnvelope != null)
 				_lastReplyEnvelope.ReplyWith(new ProjectionManagementMessage.Updated(_name));
 			_lastReplyEnvelope = null;
-			if (Deleted) {
+			if (Deleted && !_pendingWritePersistedState) {
 				DisposeCoreProjection();
 				_output.Publish(new ProjectionManagementMessage.Internal.Deleted(_name, Id));
```
shaan1337, line 1035: is it true that if `_pendingWritePersistedState = true` here and we don't publish the `ProjectionManagementMessage.Internal.Deleted`  message, then we will always publish it later?

it doesn't seem to be necessarily the case when entering `StopUnlessPreparedOrLoaded()` in `Handle(ProjectionManagementMessage.Command.Delete message)`:
- for example, if the projection was already stopped and was in the `Prepared` or `Loaded` state, it seems that it'll just persist the state but never publish the `ProjectionManagementMessage.Internal.Deleted`  message.
- but if the projection was running, then it seems that it'll stop it, then do the deletion again here: https://github.com/EventStore/EventStore/blob/d3b31d85f7b8ba33be444141f01a2fe8973a18e8/src/EventStore.Projections.Core/Services/Management/ManagedProjection.cs#L564 which will eventually publish the event.

src/EventStore.Projections.Core/Services/Management/ManagedProjection.cs
```diff
@@ -781,6 +781,14 @@ private readonly
 
 		private void DeleteStreamCompleted(ClientMessage.DeleteStreamCompleted message, string streamId,
 			Action completed) {
+			// this piece of code works because, DeleteStream request is always made with ExpectedVersion.Any
+			// this is a workaround to maintain contract and compatibility with TCP/GRPC/Web clients who expect WrongExpectedVersion in response to deleting non-existing streams
```
shaan1337, line 785:
```suggestion
			// currently, WrongExpectedVersion is returned when deleting non-existing streams.
			// it is not too intuitive but changing the response would break the contract and compatibility with TCP/gRPC/web clients or require adding a new error code to all clients.
			// note: we don't need to check if CurrentVersion == -1 here to make sure it's a non-existing stream since the deletion is always done with ExpectedVersion.Any
```

hunter1703: Done. Thanks.

src/EventStore.Projections.Core/Services/Management/ManagedProjection.cs
hunter1703: So, [this statement](https://github.com/EventStore/EventStore/blob/2de22db4f53b48eb5ee2e7b4382615eed172c1c9/src/EventStore.Projections.Core/Services/Management/ManagedProjection.cs#L422) ensures that projection state will always be Stopped or Faulted
[Delete](https://github.com/EventStore/EventStore/blob/2de22db4f53b48eb5ee2e7b4382615eed172c1c9/src/EventStore.Projections.Core/Services/Management/ManagedProjection.cs#L445) ensures that _pendingWritePersistedState will be true
From step 1 and 2, it means [WritePersistedState](https://github.com/EventStore/EventStore/blob/2de22db4f53b48eb5ee2e7b4382615eed172c1c9/src/EventStore.Projections.Core/Services/Management/ManagedProjection.cs#L721) will always be called
Hence, _pendingWritePersistedState will always be set to false when write completes

shaan1337: @hunter1703 thanks for the clarifications, i get it now: `Reply()` is always being called twice: once from `DeleteIfConditionsAreMet()` and once from `StartOrLoadStopped()`.

The check: `Deleted && !_pendingWritePersistedState` ensures that the `ProjectionManagementMessage.Internal.Deleted` message will be published only once from whichever process (i.e the stream deletions or the persisted state write) finishes last.

shaan1337: another thought (not directly linked to the problem this PR is solving)

there are currently many possible states on startup due to the parallelism of the stream deletions/write of the persisted state and it makes it difficult to ensure we're actually handling all these cases correctly (list may not be exhaustive):
<Deleting = true, Deleted = false, projection deleted from $projections-$all>
<Deleting = false, Deleted = true, projection deleted from $projections-$all>
<Deleting = true, Deleted = false, projection NOT deleted from $projections-$all> (partial deletion)
<Deleting = false, Deleted = true, projection NOT deleted from $projections-$all> (partial deletion)
<Deleting = false, Deleted = false, projection NOT deleted from $projections-$all> (e.g some projection streams deleted)

in the future, we could consider serializing the operations as follows so that it's simpler to reason about it / process partial deletions on startup:
1. write persisted state with `Deleting = true, Deleted = false`
2. delete the projection streams
3. write persisted state with `Deleting = false, Deleted = true`
4. delete the projection from $projections-$all

hunter1703: Ah. okay. I can look into this as part of a separate PR. For now, I am resolving this thread. Thanks for the comments.

hayley-jean (approved):

PR #3813, opened 2023-04-12
Scavenge: Handle events that have been deleted from (now empty) chunks but not from the index

shaan1337:
Fixed: Handle events that have been deleted from (now empty) chunks but not from the index

Previously this would result in the scavenge stopping with an error like "Could not get TimeStamp range for chunk <chunk number>"

But this is a possible state if the old scavenger had scavenged the chunks but not the index. This fix correctly handles this scenario so that the scavenge process can continue

### DIFF
--- a/src/EventStore.Core/TransactionLog/Scavenging/Data/DiscardDecision.cs
+++ b/src/EventStore.Core/TransactionLog/Scavenging/Data/DiscardDecision.cs
@@ -7,5 +7,6 @@
 		// index will just keep such indexentries.
 		MaybeDiscard,
 		Keep,
+		AlreadyDiscarded,
 	}
 }
--- a/src/EventStore.Core.XUnit.Tests/Scavenge/EmptyScavengedChunksTests.cs
+++ b/src/EventStore.Core.XUnit.Tests/Scavenge/EmptyScavengedChunksTests.cs
@@ -1,0 +1,291 @@
+﻿using System;
+using System.Threading.Tasks;
+using EventStore.Core.Data;
+using EventStore.Core.Tests;
+using EventStore.Core.Tests.TransactionLog.Scavenging.Helpers;
+using EventStore.Core.TransactionLog.Scavenging;
+using EventStore.Core.XUnit.Tests.Scavenge.Sqlite;
+using Xunit;
+using static EventStore.Core.XUnit.Tests.Scavenge.StreamMetadatas;
+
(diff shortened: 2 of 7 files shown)


shaan1337:
please hold on the review - i'm improving the test cases

src/EventStore.Core/TransactionLog/Scavenging/Data/DiscardDecision.cs
```diff
@@ -7,5 +7,6 @@ public enum DiscardDecision {
 		// index will just keep such indexentries.
 		MaybeDiscard,
 		Keep,
+		DiscardIfAllPreviousDiscarded,
```
timothycoleman, line 10: What do you think about calling this `PreviouslyDiscarded`

shaan1337: yup, that sounds better :+1: i had previously used `Discarded` but since it wasn't a 'decision' about what to do with the event, i'd renamed it to: `DiscardIfAllPreviousDiscarded` but i think it's ok to have an exception for this one.

shaan1337: following our discussion, i've renamed it to `AlreadyDiscarded`

timothycoleman:
great, nice work :+1:   squash the changes and i'll approve

timothycoleman:
for posterity, shaan and i identified that the scenario where the AlreadyDiscarded check kicks in is quite narrow:

1. we have an event E in the index that is not in the chunk
2. the logical chunk that it would have been in, in fact, has no prepares left in it at all
3. there is an earlier event for the same stream, the stream has a maxage, and we MaybeDiscard it in the calculator
4. but we do, in fact, want to keep that previous event D based on the current metadata (which means that E was removed in error or because of an earlier metadata)

the net effect is that without this check, D would be removed.

note that a stream in this state is hard to read. in order to read D the read must not also try to include E

timothycoleman (approved):
:+1:

shaan1337:
> great, nice work +1 squash the changes and i'll approve

thanks for the thorough review!

timothycoleman:
> > great, nice work +1 squash the changes and i'll approve
> 
> thanks for the thorough review!

thanks for the careful code!

hayley-jean (approved):

PR #3823, opened 2023-04-24
Removes handlerType from UpdateQuery when we modify projection using the gRPC client (DB-53)

### LINKED ISSUE #3751: System projections should not accept updates to the handler type
**Describe the bug**
See [https://github.com/EventStore/EventStore-Client-Dotnet/issues/243](https://github.com/EventStore/EventStore-Client-Dotnet/issues/243)

When updating the `$by-correlationId` projection through gRPC, the `$by-correlationId` was changed from a native projection to a JS projection. This corrupted the projection.

**To Reproduce**

1. Using the dotnet gRPC client, update the `$by-correlationId` with the method  `EventStoreProjectionManagementClient.UpdateAsync`

**Expected behavior**
The update is rejected, and the projection is unchanged.

**Actual behavior**
The update is accepted, and the projection fails.

**Config/Logs/Screenshots**
If applicable, please attach your node configuration, logs or any screenshots.

**EventStore details**

* EventStore server version: 22.10.1
* Operating system: N/A
* EventStore client version (if applicable): N/A

**Additional context**

<sub>[DB-53](https://linear.app/eventstore/issue/DB-53/system-projections-should-not-accept-updates-to-the-handler-type)</sub>

dhingrak:
Fixed: Do not change the handlerType of a projection when it is updated

Fixes #3751

Currently the system projections results in faulted when update using gRPC client. The goal of this PR is to remove the handlerType from the update projection so the query can use the existing handler type.

### DIFF
--- a/src/EventStore.Projections.Core.Tests/Integration/system_projections/when_changing_categorization_projection_configurations.cs
+++ b/src/EventStore.Projections.Core.Tests/Integration/system_projections/when_changing_categorization_projection_configurations.cs
@@ -22,11 +22,11 @@ namespace EventStore.Projections.Core.Tests.Integration.system_projections {
 			yield return
 				new ProjectionManagementMessage.Command.UpdateQuery(
 					Envelope, ProjectionNamesBuilder.StandardProjections.StreamByCategoryStandardProjection,
-					ProjectionManagementMessage.RunAs.System, handlerType: null, query: query, emitEnabled: null);
+					ProjectionManagementMessage.RunAs.System, query: query, emitEnabled: null);
 			yield return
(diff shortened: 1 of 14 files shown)


hayley-jean (changes requested):

hayley-jean (approved):
Looks good 👍

It would be great if you could also add a test to the dotnet gRPC client that does an update of a system projection once this PR is merged.

For posterity - this PR also removes the ability to update a projection's handler type. So you cannot update a projection from a `native:` type to a `js` type and vice versa. This should be fine as updating the handler type will probably break the projection anyway

timothycoleman (approved):
great :+1:

PR #3828, opened 2023-04-26
Add support for reading from $all with gRPC to `testclient`

pvanbuijtene:
Fixed: Only call `context.Success()` once to prevent early exit of command.
Added: Read $all command `RDALLGRPC`.
Removed: Some logging to prevent flooding of console/logs.

### DIFF
--- a/src/EventStore.TestClient/GrpcCommands/ReadAllProcessor.cs
+++ b/src/EventStore.TestClient/GrpcCommands/ReadAllProcessor.cs
@@ -1,0 +1,134 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Threading;
+using System.Threading.Tasks;
+using EventStore.Client;
+using Serilog;
+
+namespace EventStore.TestClient.GrpcCommands {
+	internal class ReadAllProcessor : ICmdProcessor {
+		public string Usage {
+			get { return "RDALLGRPC [[F|B] [clients] [<commit pos> <prepare pos>]]"; }
+		}
+
+		public string Keyword {
+			get { return "RDALLGRPC"; }
+		}
+
+		public bool Execute(CommandProcessorContext context, string[] args) {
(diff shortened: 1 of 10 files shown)


src/EventStore.TestClient/GrpcCommands/ReadAllProcessor.cs
```diff
@@ -0,0 +1,134 @@
...
+				else
+					return false;
+
+				if (args.Length > 1) {
+					clientCount = MetricPrefixValue.ParseInt(args[1]);
+				}
+
+				if (args.Length >= 3) {
+					posOverriden = true;
+					if (!ulong.TryParse(args[1], out var commitPos) || !ulong.TryParse(args[2], out var preparePos))
```
dhingrak, line 43: commitPos is taking the value of args[1] rather than args[2]

if (!ulong.TryParse(args[2], out var commitPos) || !ulong.TryParse(args[3], out var preparePos))

pvanbuijtene: Fixed.

dhingrak (approved):

src/EventStore.TestClient/GrpcCommands/ReadAllProcessor.cs
```diff
@@ -0,0 +1,134 @@
...
+
+		public bool Execute(CommandProcessorContext context, string[] args) {
+			Direction direction = Direction.Forwards;
+			Position position = Position.Start;
+			bool forward = true;
+			bool posOverriden = false;
+			int clientCount = 1;
+
+			if (args.Length > 0) {
+				if (args.Length != 1 && args.Length != 2 && args.Length != 3 && args.Length != 4)
```
hunter1703, line 27: maybe, `args.Length > 4` ?

hunter1703 (approved):

dhingrak (approved):

PR #3832, opened 2023-05-03
Log error message when the SSL handshakes fail [DB-32]

### LINKED ISSUE #3765: Improve error messaging when SSL handshakes fail
When there’s an error with SSL handshake (e.g. invalid certificate, untrusted certificate), the stack trace is dumped with an internal gRPC error. These messages are also logged as information-level logs.

The messages around these failures should be clearer and guide the user more.

1. Increase the log level for these warnings (possibly only for internal connections)
2. Do not dump the stack trace
3. Show a human-readable error message that the connection attempt failed to to SSL issues and suggest possible reasons

<sub>[DB-32](https://linear.app/eventstore/issue/DB-32/improve-error-messaging-when-ssl-handshakes-fail)</sub>

dhingrak:
Fixed: Log more readable error message in the server logs when the SSL handshakes fail.

Fixes #3765 

The goal of this PR to add more readable error messages rather than dumping the error status when the SSL handshakes fail.

### DIFF
--- a/src/EventStore.Core/ClusterVNode.cs
+++ b/src/EventStore.Core/ClusterVNode.cs
@@ -1752,7 +1752,7 @@ namespace EventStore.Core {
 				}
 			}
 
-			var chainStatus = CertificateUtils.BuildChain(certificate, intermediates, trustedRootCertsSelector());
+			var chainStatus = CertificateUtils.BuildChain(certificate, intermediates, trustedRootCertsSelector(), out var chainStatusInformation);
 			if (chainStatus == X509ChainStatusFlags.NoError)
 				sslPolicyErrors &= ~SslPolicyErrors.RemoteCertificateChainErrors; //clear the RemoteCertificateChainErrors flag
 			else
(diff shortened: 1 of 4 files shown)


src/EventStore.Core/ClusterVNode.cs
```diff
@@ -1752,7 +1752,7 @@ public class ClusterVNode<TStreamId> :
 				}
 			}
 
-			var chainStatus = CertificateUtils.BuildChain(certificate, intermediates, trustedRootCertsSelector());
+			var chainStatus = CertificateUtils.BuildChain(certificate, intermediates, trustedRootCertsSelector(), out _);
```
hayley-jean, line 1755: The `chainStatusInformation` returned from this call to `BuildChain` should also be captured and logged to give more detailed information about why other nodes' certificates are invalid

hayley-jean (changes requested):

hayley-jean (approved):

timothycoleman (approved):

PR #3834, opened 2023-05-09
Migrated IO metrics from Counter to ObservableCounter instrument (DB-85)

hunter1703:
Changed : IO metrics using Counter instrument to ObservableCounter instrument

### DIFF
--- a/src/EventStore.Core/MetricsBootstrapper.cs
+++ b/src/EventStore.Core/MetricsBootstrapper.cs
@@ -81,8 +81,8 @@ public static class MetricsBootstrapper {
 		var gossipProcessingMetric = new DurationMetric(coreMeter, "eventstore-gossip-processing-duration");
 		var queueQueueingDurationMaxMetric = new DurationMaxMetric(coreMeter, "eventstore-queue-queueing-duration-max");
 		var queueProcessingDurationMetric = new DurationMetric(coreMeter, "eventstore-queue-processing-duration");
-		var byteMetric = coreMeter.CreateCounter<long>("eventstore-io", unit: "bytes");
-		var eventMetric = coreMeter.CreateCounter<long>("eventstore-io", unit: "events");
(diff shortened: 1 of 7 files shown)


src/EventStore.Core/MetricsBootstrapper.cs
```diff
@@ -84,15 +82,15 @@ public static class MetricsBootstrapper {
 		if (conf.Events.TryGetValue(Conf.EventTracker.Read, out var readEnabled) && readEnabled) {
 			var readTag = new KeyValuePair<string, object>("activity", "read");
 			trackers.TransactionFileTracker = new TFChunkTracker(
-				readBytes: new CounterSubMetric<long>(byteMetric, readTag),
-				readEvents: new CounterSubMetric<long>(eventMetric, readTag));
+				readBytes: new FastCounterMetric(coreMeter, "eventstore-io", readTag,  "bytes"),
+				readEvents: new FastCounterMetric(coreMeter, "eventstore-io", readTag, "events"));
```
timothycoleman, line 86: im slightly suspicious of creating two separate instruments that output to the same name (`eventstore-io-events`), unless you saw some doc that indicates that this is definitely ok? otherwise i expect we need a `CounterMetric` class and a `CounterSubMetric` class

hunter1703: Functionally, this won't cause any issues. `coreMeter` will have two observable counters with name `eventstore-io-events` and both counters will correctly report `read` and `write` tagged counts respectively. But, I think having submetric might help in code readability.

hunter1703: I have changed the implementation.

timothycoleman:
looks good :+1: can be squashed

YoEight (approved):

timothycoleman (approved):
:+1:

PR #3838, opened 2023-05-11
Add paged bytes, virtual bytes, and thread pool tasks queue length metrics.

YoEight:
Added: Add paged bytes, virtual bytes, and thread pool tasks queue length metrics


src/EventStore.Common/Configuration/TelemetryConfiguration.cs
```diff
@@ -108,7 +108,10 @@ public enum ProcessTracker {
 			UpTime = 1,
 			Cpu,
 			MemWorkingSet,
+			MemPagedBytes,
+			MemVirtualBytes,
 			ThreadCount,
+			ThreadQueueLength,
```
timothycoleman, line 114: can we call this `ThreadPoolPendingWorkItemCount`  it's a bit of a mouthful but at it's close to the name of what it really is on the platform so it'll be more obvious

timothycoleman (approved):
:+1:

PR #3839, opened 2023-05-15
DB-33: Add a check to verify if something is blocking the data transfer between nodes

### LINKED ISSUE #3766: Detect and warn when something is blocking data transfer between nodes
When one node cannot reach another node due to, for example, a firewall configuration that blocks the necessary port, the server produces debug-level log messages (which aren’t enabled by default).

These log messages look the same as other connection lost errors, except the number of sent and received bytes are always `0`.

This makes it difficult to diagnose issues where two nodes in a cluster can't communicate with each other.

EventStoreDB could detect when it's talking to another node (as opposed to an external client) and not receiving/sending any data before the connection is closed.

If it detects such a case, it can escalate that as an error stating that the connection is potentially blocked.

<sub>[DB-33](https://linear.app/eventstore/issue/DB-33/detect-and-warn-when-something-is-blocking-data-transfer-between-nodes)</sub>

dhingrak:
Fixed: Improved log message for connectivity problem between nodes

Fixes #3766 

Currently we are logging the message as connection lost even the node cannot reach to another node due to some reasons. For example, a firewall configuration blocks the necessary port. The goal of this PR is to log a warning to differentiate whether the connection is lost or the connection is blocked for some reasons.

### DIFF
--- a/src/EventStore.Core/Services/Transport/Tcp/TcpConnectionManager.cs
+++ b/src/EventStore.Core/Services/Transport/Tcp/TcpConnectionManager.cs
@@ -209,6 +209,13 @@ namespace EventStore.Core.Services.Transport.Tcp {
 				"Connection '{connectionName}{clientConnectionName}' [{remoteEndPoint}, {connectionId:B}] closed: {e}.",
 				ConnectionName, ClientConnectionName.IsEmptyString() ? string.Empty : ":" + ClientConnectionName,
 				connection.RemoteEndPoint, ConnectionId, socketError);
+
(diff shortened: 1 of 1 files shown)


hayley-jean (approved):
The changes look good

As a note, this should also log the warning for external client connections that close without sending/receiving data

hayley-jean (changes requested):
After discussing during standup, it would be better to only log this warning if the connection was opened by the server (as opposed to having been opened by another node or client). This would prevent the warning from being logged for connections that were opened by an external client and then the client immediately dropping it

timothycoleman (approved):

hayley-jean (approved):

PR #3842, opened 2023-05-15
Add GC max Execution Engine Suspension duration.

YoEight:
Added: Add CG max Execution Engine Suspension duration.

```
# TYPE eventstore_gc_max_pause_duration_max_seconds gauge
# UNIT eventstore_gc_max_pause_duration_max_seconds seconds
eventstore_gc_max_pause_duration_max_seconds{range="16-20 seconds"} 0.000253622 1684188262384
```

image

The execution engine is suspended for the duration of any blocking GC (gen0, gen1, and blocking gen2 collections). For a background gen2 collection the execution engine is suspended twice at points during the collection but not for the duration of the collection.

The idea here is to be able to determine if execution engine pauses due to GC are the cause of any 'slow queue' messages

### DIFF
--- a/src/EventStore.Core/MetricsBootstrapper.cs
+++ b/src/EventStore.Core/MetricsBootstrapper.cs
@@ -235,7 +235,7 @@ public static class MetricsBootstrapper {
 		});
 
 		// process
-		var processMetrics = new ProcessMetrics(coreMeter, timeout, conf.Process);
+		var processMetrics = new ProcessMetrics(coreMeter, timeout, conf.ExpectedScrapeIntervalSeconds, conf.Process);
 		processMetrics.CreateObservableMetrics(new() {
 			{ Conf.ProcessTracker.UpTime, "eventstore-proc-up-time" },
 			{ Conf.ProcessTracker.Cpu, "eventstore-proc-cpu" },
@@ -246,6 +246,7 @@ public static class MetricsBootstrapper {
 			{ Conf.ProcessTracker.HeapSize, "eventstore-gc-heap-size" },
 			{ Conf.ProcessTracker.HeapFragmentation, "eventstore-gc-heap-fragmentation" },
 			{ Conf.ProcessTracker.TotalAllocatedBytes, "eventstore-gc-total-allocated" },
+			{ Conf.ProcessTracker.GcPauseDuration, "eventstore-gc-pause-duration-max" },
 		});
 
 		processMetrics.CreateMemoryMetric("eventstore-proc-mem", new() {
--- a/src/EventStore.Core/Telemetry/GCSuspensionMetric.cs
+++ b/src/EventStore.Core/Telemetry/GCSuspensionMetric.cs
@@ -1,0 +1,53 @@
+using System;
+using System.Diagnostics.Tracing;
+
+namespace EventStore.Core.Telemetry;
+
+public class GcSuspensionMetric : EventListener {
+	private const int GcKeyword = 0x0000001;
+	private const int GCSuspendEEBegin = 9;
+	private const int GCRestartEEEnd = 3;
+	private const uint SuspendForGc = 0x1;
+	private const uint SuspendForGcPrep = 0x6;
+	private readonly DurationMaxTracker _tracker;
+	private DateTime? _started;
+
+	public GcSuspensionMetric(DurationMaxTracker tracker) {
+		_tracker = tracker;
+	}
+
+	protected override void OnEventSourceCreated(EventSource eventSource) {
+		if (eventSource.Name.Equals("Microsoft-Windows-DotNETRuntime")) {
+			EnableEvents(eventSource, EventLevel.Informational, (EventKeywords)GcKeyword);
+		}
+	}
+
+	protected override void OnEventWritten(EventWrittenEventArgs eventData) {
+		if (_tracker == null)
+			return;
+
+		switch (eventData.EventId) {
+			case GCSuspendEEBegin: {
+				var idx = eventData.PayloadNames!.IndexOf("Reason");
+				var value = (uint)eventData.Payload![idx]!;
+
+				// We only track suspensions that are meant for garbage collection.
+				// See https://learn.microsoft.com/en-us/dotnet/fundamentals/diagnostics/runtime-garbage-collection-events
+				if (value is SuspendForGc or SuspendForGcPrep)
+					_started = eventData.TimeStamp;
+
+				break;
+			}
+
+			case GCRestartEEEnd: {
+				// Means that the suspension end event comes from a suspension that was not due to garbage collection.
+				if (!_started.HasValue)
+					return;
+
+				_tracker.RecordNow(eventData.TimeStamp.Subtract(_started.Value));
+				_started = null;
+				break;
+			}
+		}
+	}
+}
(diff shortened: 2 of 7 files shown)


src/EventStore.Core/Telemetry/GCSuspensionMetric.cs
```diff
@@ -0,0 +1,54 @@
...
+
+		switch (eventData.EventName) {
+			case "GCSuspendEEBegin_V1": {
+				var idx = eventData.PayloadNames!.IndexOf("Reason");
+				var value = (uint)eventData.Payload![idx]!;
+
+				// We only track suspensions that are meant for garbage collection.
+				// See https://learn.microsoft.com/en-us/dotnet/fundamentals/diagnostics/runtime-garbage-collection-events
+				if (value == 0x1)
+					_started = _clock.Now;
```
timothycoleman, line 39: i think we'd better use `eventData.TimeStamp` for accuracy

we could add an method to DurationMaxTracker like

```
	public void Record(TimeSpan duration) {
		_recentMax.Record(_clock.Now, duration.TotalSeconds);
	}
```

src/EventStore.Core/Telemetry/GCSuspensionMetric.cs
```diff
@@ -0,0 +1,54 @@
...
+	private Instant? _started;
+
+	public GcSuspensionMetric(IClock clock, DurationMaxTracker tracker) {
+		_tracker = tracker;
+		_clock = clock;
+	}
+
+	protected override void OnEventSourceCreated(EventSource eventSource) {
+		if (eventSource.Name.Equals("Microsoft-Windows-DotNETRuntime")) {
+			EnableEvents(eventSource, EventLevel.Verbose, (EventKeywords)GcKeyword);
```
timothycoleman, line 20: i think information might be enough for the events we need

src/EventStore.Core/Telemetry/GCSuspensionMetric.cs
```diff
@@ -0,0 +1,54 @@
...
+		}
+	}
+
+	protected override void OnEventWritten(EventWrittenEventArgs eventData) {
+		if (_tracker == null || _clock == null)
+			return;
+
+		if (eventData.EventName is not "GCSuspendEEBegin_V1" && eventData.EventName is not "GCSuspendEEEnd_V1")
```
timothycoleman, line 28: probably add a default case to the switch instead of the additional check up here

src/EventStore.Core/Telemetry/GCSuspensionMetric.cs
```diff
@@ -0,0 +1,54 @@
...
+			return;
+
+		switch (eventData.EventName) {
+			case "GCSuspendEEBegin_V1": {
```
timothycoleman, line 32: im suspicious of the V1 and what will happen when they change to V2. e.g GCAllocationTick is documented as V3 but the runtime seems to emit V4 these days

rather switch on eventData.EventId and have some defined constants (bonus, it'll be cheaper than strings)

src/EventStore.Core/Telemetry/GCSuspensionMetric.cs
```diff
@@ -0,0 +1,54 @@
...
+
+				// We only track suspensions that are meant for garbage collection.
+				// See https://learn.microsoft.com/en-us/dotnet/fundamentals/diagnostics/runtime-garbage-collection-events
+				if (value == 0x1)
+					_started = _clock.Now;
+
+				return;
+			}
+
+			case "GCSuspendEEEnd_V1":
```
timothycoleman, line 44: this is only the end of the process of suspending the EE, i think we need to wait for the end of when the EE was restarted

src/EventStore.Core/Telemetry/GCSuspensionMetric.cs
```diff
@@ -0,0 +1,54 @@
...
+			return;
+
+		switch (eventData.EventName) {
+			case "GCSuspendEEBegin_V1": {
+				var idx = eventData.PayloadNames!.IndexOf("Reason");
+				var value = (uint)eventData.Payload![idx]!;
+
+				// We only track suspensions that are meant for garbage collection.
+				// See https://learn.microsoft.com/en-us/dotnet/fundamentals/diagnostics/runtime-garbage-collection-events
+				if (value == 0x1)
```
timothycoleman, line 38: maybe 0x6 as well.   and best have them as constants

src/EventStore.Core/Telemetry/GCSuspensionMetric.cs
```diff
@@ -0,0 +1,54 @@
...
+					_started = _clock.Now;
+
+				return;
```
timothycoleman, line 41:
```suggestion
				break;
```

for consistency

src/EventStore.Core/MetricsBootstrapper.cs
```diff
@@ -246,6 +247,7 @@ public static class MetricsBootstrapper {
 			{ Conf.ProcessTracker.HeapSize, "eventstore-gc-heap-size" },
 			{ Conf.ProcessTracker.HeapFragmentation, "eventstore-gc-heap-fragmentation" },
 			{ Conf.ProcessTracker.TotalAllocatedBytes, "eventstore-gc-total-allocated" },
+			{ Conf.ProcessTracker.GcMaxDurationPause, "eventstore-gc-max-pause-duration-max" },
```
timothycoleman, line 250: "eventstore-gc-pause-duration-max"

timothycoleman (approved):
:+1: great thanks!

for posterity, the loop in the can_detect_gc_pauses test is there because although GC.Collect runs synchronously, the associated events dont necessarily arrive immediately

hunter1703:
Some comments, please feel free to ignore if you have already thought about it :

1. Why are metrics only collected for large object heap and not small object heap?
2. Should we be collecting metrics for number of gcs? frequent but small gc pauses might be an issue which would not be detected if we are only collecting metrics for gc pause length (I understand this might not be within scope of this PR)

timothycoleman:
> Some comments, please feel free to ignore if you have already thought about it :
> 
> 1. Why are metrics only collected for large object heap and not small object heap? I understand g0 gc would be frequent and hence, calculating recent max might impact performance, but I think there should still be an option to collect g0 gc metrics for debugging purposes
> 2. Should we be collecting metrics for number of gcs? frequent but small gc pauses might be an issue which would not be detected if we are only collecting metrics for gc pause length (I understand this might not be within scope of this PR)

we're measuring the pauses here for for any GC (well, unless i've misunderstood something). i'll add some extra description at the top to try and make the idea clearer

yep we are already counting the number of gen 0/1/2 gcs

hunter1703:
> > Some comments, please feel free to ignore if you have already thought about it :
> > 
> > 1. Why are metrics only collected for large object heap and not small object heap? I understand g0 gc would be frequent and hence, calculating recent max might impact performance, but I think there should still be an option to collect g0 gc metrics for debugging purposes
> > 2. Should we be collecting metrics for number of gcs? frequent but small gc pauses might be an issue which would not be detected if we are only collecting metrics for gc pause length (I understand this might not be within scope of this PR)
> 
> we're measuring the pauses here for for any GC not just loh (well, unless i've misunderstood something). i'll add some extra description at the top to try and make the idea clearer
> 
> yep we are already counting the number of gen 0/1/2 gcs

Okay, my bad. I misunderstood small and large object heaps as g0/g2.

PR #3851, opened 2023-05-22
Added support for multiple private key file format

hunter1703:
Added : Based on header of the private key file, ES will now accept encrypted and unencrypted PKCS8 private key files

### DIFF
--- a/src/EventStore.Core/Certificates/CertificateUtils.cs
+++ b/src/EventStore.Core/Certificates/CertificateUtils.cs
@@ -6,6 +6,7 @@ using System.Security.Cryptography;
 using System.Security.Cryptography.Pkcs;
 using System.Security.Cryptography.X509Certificates;
 using EventStore.Common.Utils;
+using System.Text;
 using EventStore.Core.Exceptions;
 
 namespace EventStore.Core {
@@ -77,7 +78,7 @@ namespace EventStore.Core {
 		public static (X509Certificate2 certificate, X509Certificate2Collection intermediates) LoadFromFile(
 			string certificatePath,
 			string privateKeyPath,
-			string password) {
+			string certificatePassword, string certificatePrivateKeyPassword = null) {
 
 			if (string.IsNullOrEmpty(privateKeyPath)) {
 				// no private key file - assume PKCS12 format
@@ -86,7 +87,7 @@ namespace EventStore.Core {
 				try {
 					if (TryReadPkcs12CertificateBundle(
 						certificatePath,
-						password,
+						certificatePassword,
 						out var nodeCertificate,
 						out var intermediateCertificates)) {
 						return (nodeCertificate, intermediateCertificates);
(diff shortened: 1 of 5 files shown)


hayley-jean (dismissed):

src/EventStore.Core/ClusterVNodeOptions.cs
```diff
@@ -270,9 +270,14 @@ public record CertificateFileOptions {
 			[Description("The password to the certificate if a PKCS #12 (.p12/.pfx) certificate file is provided."),
 			 Sensitive]
 			public string? CertificatePassword { get; init; }
+			
+			[Description("The password to the certificate private key if an encrypted PKCS #8 private key file is provided."),
+			 Sensitive]
+			public string? PrivateKeyPassword { get; init; }
```
shaan1337, line 276:
```suggestion
			public string? CertificatePrivateKeyPassword { get; init; }
```

suggestion: for consistency with other certificate options, we can rename it to `CertificatePrivateKeyPassword`

src/EventStore.Core/ClusterVNodeOptions.cs
```diff
@@ -270,9 +270,14 @@ public record CertificateFileOptions {
 			[Description("The password to the certificate if a PKCS #12 (.p12/.pfx) certificate file is provided."),
 			 Sensitive]
 			public string? CertificatePassword { get; init; }
+			
+			[Description("The password to the certificate private key if an encrypted PKCS #8 private key file is provided."),
```
shaan1337, line 274:
```suggestion
			[Description("The password to the certificate private key file if an encrypted PKCS #8 private key file is provided."),
```

src/EventStore.Core/Certificates/CertificateUtils.cs
```diff
@@ -96,18 +97,36 @@ private static bool TryReadPkcs12CertificateBundle
...
+					break;
+				case "BEGIN ENCRYPTED PRIVATE KEY":
+					if (privateKeyPassword is null) {
+						throw new ArgumentException(
+							"Need encryption password to read encrypted node private key");
+					}
+					rsa.ImportEncryptedPkcs8PrivateKey(new ReadOnlySpan<byte>(Encoding.UTF8.GetBytes(privateKeyPassword)), new ReadOnlySpan<byte>(privateKey), out _);
+					break;
+				default:
+					throw new InvalidFileException("Node private key file format is unrecognizable");
```
shaan1337, line 126:
```suggestion
					throw new NotSupportedException($"Unsupported private key file format: {header}");
```

suggestion: throw a NotSupportedException instead of InvalidFileException

src/EventStore.Core/Certificates/CertificateUtils.cs
```diff
@@ -96,18 +97,36 @@ private static bool TryReadPkcs12CertificateBundle
...
+					break;
+				case "BEGIN RSA PRIVATE KEY":
+					rsa.ImportRSAPrivateKey(new ReadOnlySpan<byte>(privateKey), out _);
+					break;
+				case "BEGIN ENCRYPTED PRIVATE KEY":
+					if (privateKeyPassword is null) {
+						throw new ArgumentException(
+							"Need encryption password to read encrypted node private key");
+					}
+					rsa.ImportEncryptedPkcs8PrivateKey(new ReadOnlySpan<byte>(Encoding.UTF8.GetBytes(privateKeyPassword)), new ReadOnlySpan<byte>(privateKey), out _);
```
shaan1337, line 123: suggestion: we can wrap this in a `try-catch (System.Security.Cryptography.CryptographicException)` and throw an argument exception: `Failed to read the certificate's private key file. The password may be incorrect.`

src/EventStore.Core/Certificates/CertificateUtils.cs
```diff
@@ -96,18 +97,36 @@ private static bool TryReadPkcs12CertificateBundle
...
+				case "BEGIN PRIVATE KEY":
+					rsa.ImportPkcs8PrivateKey(new ReadOnlySpan<byte>(privateKey), out _);
+					break;
+				case "BEGIN RSA PRIVATE KEY":
+					rsa.ImportRSAPrivateKey(new ReadOnlySpan<byte>(privateKey), out _);
+					break;
+				case "BEGIN ENCRYPTED PRIVATE KEY":
+					if (privateKeyPassword is null) {
+						throw new ArgumentException(
+							"Need encryption password to read encrypted node private key");
```
shaan1337, line 121:
```suggestion
							"A password is required to read the certificate's private key file.");
```

src/EventStore.Core/Certificates/CertificateUtils.cs
```diff
@@ -96,18 +97,36 @@ private static bool TryReadPkcs12CertificateBundle
...
+					}
+					rsa.ImportEncryptedPkcs8PrivateKey(new ReadOnlySpan<byte>(Encoding.UTF8.GetBytes(privateKeyPassword)), new ReadOnlySpan<byte>(privateKey), out _);
+					break;
+				default:
+					throw new InvalidFileException("Node private key file format is unrecognizable");
+
```
shaan1337, line 127: suggestion: after loading the key, check if it is an RSA key and throw a `NotSupportedException` if it's not.

Some time back, i had tested elliptical curve keys and it was not stable causing native code crashes. i don't know if the state has changed, but for the scope of this PR, we can just prevent use of non-RSA keys.

cc: @hunter1703 @alexeyzimarev @hayley-jean

shaan1337: EDIT: ignore the above. just noticed we use rsa.Import*, so if we're loading an EC key, it'll throw an exception

shaan1337 (commented):
tested the changes and it works as expected :+1:  i've just made a few cosmetic suggestions.

src/EventStore.Core/Certificates/CertificateUtils.cs
hunter1703: Yes. done.

src/EventStore.Core/Certificates/CertificateUtils.cs
```diff
@@ -96,18 +97,41 @@ private static bool TryReadPkcs12CertificateBundle
...
+				case "BEGIN ENCRYPTED PRIVATE KEY":
+					if (certificatePrivateKeyPassword is null) {
+						throw new ArgumentException(
+							"A password is required to read the certificate's private key file.");
+					}
+					try {
+						rsa.ImportEncryptedPkcs8PrivateKey(new ReadOnlySpan<byte>(Encoding.UTF8.GetBytes(certificatePrivateKeyPassword)), new ReadOnlySpan<byte>(privateKey), out _);
+					} catch (CryptographicException e) {
+						throw new ArgumentException(
+							"Failed to read the certificate's private key file. The password may be incorrect.", e);
```
shaan1337, line 127:
```suggestion
							"Failed to read the certificate's private key file. The password may be incorrect or the private key may not be an RSA key.", e);
```

shaan1337: @hunter1703 i've tested with an encrypted EC key and it throws the same type of exception. which may be misleading if we don't mention we're also expecting an RSA key.

hunter1703: Done. Thanks for the comments

shaan1337 (approved):

src/EventStore.Core/Certificates/CertificateUtils.cs
shaan1337: @hunter1703 thanks!

hayley-jean (approved):

PR #3854, opened 2023-05-25
Update log level to Verbose for NACK logs for persistent subscriptions

dhingrak:
Fixed: Lower the log level of NACK logs for persistent subscriptions

Fixes https://linear.app/eventstore/issue/DB-140/remove-or-lower-the-log-level-of-nack-logs-for-persistent

Currently we are logging the NACK messages for persistent subscriptions at DEBUG level. The goal of this PR is lower the log level to VERBOSE to make server logs more clean. The user can modify the LogLevel in the config file as per their requirements.


timothycoleman (approved):

Lougarou (approved):
Should be good, since the retrying messages are still logged on the server side, we can use that for investigating issues

PR #3855, opened 2023-05-30
Add a warning for node certificate expiring within 30 days (DB-138)

dhingrak:
Fixed: Log a warning when certificates are close to expiry.

Fixes https://linear.app/eventstore/issue/DB-138/log-a-warning-when-certificates-are-close-to-expiry

The goal of this PR is to add a warning daily if the node certificate is going to expire within 30 days.

### DIFF
--- a/src/EventStore.Core/Certificates/CertificateExpiryMonitor.cs
+++ b/src/EventStore.Core/Certificates/CertificateExpiryMonitor.cs
@@ -1,0 +1,62 @@
+﻿using System;
+using System.Security.Cryptography.X509Certificates;
+using EventStore.Common.Utils;
+using EventStore.Core.Bus;
+using EventStore.Core.Messages;
+using EventStore.Core.Messaging;
+using EventStore.Core.Services.TimerService;
+using Serilog;
+
+namespace EventStore.Core.Certificates;
+
+public class CertificateExpiryMonitor :
+	IHandle<SystemMessage.SystemStart>,
+	IHandle<MonitoringMessage.CheckCertificateExpiry> {
+
+	private static readonly TimeSpan _warningThreshold = TimeSpan.FromDays(30);
+	private static readonly TimeSpan _interval = TimeSpan.FromDays(1);
+
+	private readonly Func<X509Certificate2> _getCertificate;
+	private readonly IPublisher _publisher;
+	private readonly TimerMessage.Schedule _nodeCertificateExpirySchedule;
+	private readonly ILogger _logger;
+
+	public CertificateExpiryMonitor(
+		IPublisher publisher,
+		Func<X509Certificate2> getCertificate,
+		ILogger logger) {
+
+		Ensure.NotNull(publisher, nameof(publisher));
+		Ensure.NotNull(getCertificate, nameof(getCertificate));
+		Ensure.NotNull(logger, nameof(logger));
+
+		_publisher = publisher;
+		_getCertificate = getCertificate;
+		_logger = logger;
+		_nodeCertificateExpirySchedule = TimerMessage.Schedule.Create(
+			_interval,
+			new PublishEnvelope(publisher),
+			new MonitoringMessage.CheckCertificateExpiry());
+	}
+
+	public void Handle(SystemMessage.SystemStart message) {
+		_publisher.Publish(new MonitoringMessage.CheckCertificateExpiry());
+	}
+
+	public void Handle(MonitoringMessage.CheckCertificateExpiry message) {
+		var certificate = _getCertificate();
+
+		if (certificate != null) {
+			var certExpiryDate = certificate.NotAfter;
+			var timeUntilExpiry = certExpiryDate - DateTime.Now;
+
+			if (timeUntilExpiry <= _warningThreshold) {
+				_logger.Warning(
+					"Certificates are going to expire in {daysUntilExpiry:N1} days",
+					timeUntilExpiry.TotalDays);
+			}
+		}
+
+		_publisher.Publish(_nodeCertificateExpirySchedule);
+	}
+}
--- a/src/EventStore.Core/ClusterVNode.cs
+++ b/src/EventStore.Core/ClusterVNode.cs
@@ -193,7 +193,7 @@ namespace EventStore.Core {
 		private readonly CertificateDelegates.ClientCertificateValidator _externalClientCertificateValidator;
 		private readonly CertificateDelegates.ServerCertificateValidator _externalServerCertificateValidator;
 		private readonly CertificateProvider _certificateProvider;
-
+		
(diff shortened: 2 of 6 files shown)


src/EventStore.Core/ClusterVNode.cs
```diff
@@ -1679,9 +1681,29 @@ public class ClusterVNode<TStreamId> :
...
+					_mainQueue.Publish(TimerMessage.Schedule.Create(TimeSpan.FromHours(24),
+						new PublishEnvelope(_mainQueue),
+						new SystemMessage.VerifyNodeCertificateExpiry()));
+
 				}
 			});
 		}
+		
+		public void Handle(SystemMessage.VerifyNodeCertificateExpiry message) {
+			var (certificate, intermediates) = _options.LoadNodeCertificate();
```
hayley-jean, line 1693: Using `LoadNodeCertificate()` here causes an error to be thrown when running in either `--dev` or `--insecure` mode because there aren't any certificate options set for either of those cases:

```
[12880,15,13:32:06.971,ERR] Error while processing message "EventStore.Core.Messages.SystemMessage+VerifyNodeCertificateExpiry" in queued handler '"MainQueue"'.
EventStore.Common.Exceptions.InvalidConfigurationException: A certificate is required unless insecure mode (--insecure) is set.
   at EventStore.Core.ClusterVNodeOptionsExtensions.LoadNodeCertificate(ClusterVNodeOptions options) in C:\Users\hayle\code\EventStore\src\EventStore.Core\ClusterVNodeOptionsExtensions.cs:line 243
   at EventStore.Core.ClusterVNode`1.Handle(VerifyNodeCertificateExpiry message) in C:\Users\hayle\code\EventStore\src\EventStore.Core\ClusterVNode.cs:line 1693
   at EventStore.Core.Bus.MessageHandler`1.TryHandle(Message message) in C:\Users\hayle\code\EventStore\src\EventStore.Core\Bus\MessageHandler.cs:line 30
   at EventStore.Core.Bus.InMemoryBus.Publish(Message message) in C:\Users\hayle\code\EventStore\src\EventStore.Core\Bus\InMemoryBus.cs:line 290
   at EventStore.Core.Bus.QueuedHandlerMRES.ReadFromQueue(Object o) in C:\Users\hayle\code\EventStore\src\EventStore.Core\Bus\QueuedHandlerMRES.cs:line 124
[12880,25,13:33:49.018,INF] Exiting with exit code: 0.
```

Instead, it would be better to use the `_certificateSelector` to get the certificate, as this will try and get the cert from the correct provider (either from options, or from dev certificates)

```suggestion
			var certificate = _certificateSelector();
```

You will need to do a null check on the certificate after getting it though, because it will always be null in `--insecure` mode.

src/EventStore.Core/ClusterVNode.cs
```diff
@@ -1679,9 +1681,29 @@ public class ClusterVNode<TStreamId> :
...
+			
+			var certExpiryDate = certificate.NotAfter;
+			var addMonthToCurrentDate = DateTime.Now.AddDays(30);
+
+			if (addMonthToCurrentDate >= certExpiryDate) {
+				Log.Warning(
+					$"Certificates are going to expire within 30 days. Current Date: {DateTime.Now}, Expiry Date: {certificate.NotAfter}");
```
hayley-jean, line 1700: You should use the positional formatting for Serilog rather than string interpolation here:
```suggestion
					"Certificates are going to expire within 30 days. Current Date: {currentDate}, Expiry Date: {certificateExpiryDate}", DateTime.Now, certificate.NotAfter);
```

src/EventStore.Core/Messages/SystemMessage.cs
```diff
@@ -287,5 +287,12 @@ public partial class EpochWritten : Message {
 				Epoch = epoch;
 			}
 		}
+		
+		[DerivedMessage(CoreMessage.System)]
+		public partial class VerifyNodeCertificateExpiry : Message {
+			public VerifyNodeCertificateExpiry() {
+			}
+		}
+
```
timothycoleman, line 296:
```suggestion
```

timothycoleman (changes requested):
I think it'd be nice to pass _certificateSelector into a new class that is responsible for monitoring the expiry rather than adding the logic to ClusterVNode. _certificateSelector is already passed elsewhere

We can store the ScheduleMessage returned by TimerMessage.Schedule.Create in a member variable and reuse it each time to save on these allocations that will end up in gen2 because they're long lived

timothycoleman (changes requested):
great thats look good

it would be cool if schedule message construction could be left to the monitor itself as well, then the whole implementation of how it works would be encapsulated away from the ClusterVNode. Then we'd be free to change the implementation of the monitor without it having a knock on effect elsewhere

it would likely make sense to have the Monitor Handle SystemMessage.SystemStart

and then we'll be in a good position to write some unit tests for the monitor

one other thing, the monitor can receive the mainbus as a IPublisher rather than IQueuedHandler (all we are going to do with it is publish), and the member variable needn't be called `_mainQueue` because the monitor doesn't mind what queue it really is (which is nice because it means we can change the details of the ClusterVNode if we want without having a knock on effect on the monitor) so we can just call it `_publisher`

timothycoleman:
come to think of it the easiest thing to do would probably be to pass a

`Func<DateTime?> getCertificateExpiry` instead of `Func<X509Certificate2> certificateSelector`

then the tests will be easy even using DateTIme.now

timothycoleman (approved):
Great, thanks!

src/EventStore.Core/Certificates/CertificateExpiryMonitor.cs
```diff
@@ -0,0 +1,62 @@
...
+	public void Handle(MonitoringMessage.CheckCertificateExpiry message) {
+		var certificate = _getCertificate();
+
+		if (certificate != null) {
+			var certExpiryDate = certificate.NotAfter;
+			var timeUntilExpiry = certExpiryDate - DateTime.Now;
+
+			if (timeUntilExpiry <= _warningThreshold) {
+				_logger.Warning(
+					"Certificates are going to expire in {daysUntilExpiry:N1} days",
```
hayley-jean, line 55: Looks good, I'm happy to merge as-is but I think it would be better to only show the remaining days as whole numbers:

```suggestion
					"Certificates are going to expire in {daysUntilExpiry:N0} days",
```

hayley-jean (approved):

src/EventStore.Core/Certificates/CertificateExpiryMonitor.cs
timothycoleman: (for posterity, it can go either way but seeing `0.4 days` is slightly more informative than `0 days`)

PR #3856, opened 2023-05-30
[POC] Implement accessing logs as a stream subscription.

YoEight:
Logs are not accessible through the `$logs` stream.  No logs are actually saved on the transaction log (duh)


timothycoleman:
changed it to draft so we dont somehow accidentally merge (since its a poc)

PR #3857, opened 2023-05-31
Lazy heading event reader init

hunter1703:
Changed : Lazy initialization and eager disposal of heading event reader


hunter1703:
This PR has multiple commits :

1. Oldest commit has some variable renames and simple code cleanup. I had trouble understanding code as it is which, I thought, was due to "misnamed" variables. Please let me know if my commit is inappropriate and I will remove it from the PR
2. Rest of the commits are actual changes for the issue

timothycoleman:
Changing this to draft, we need to discuss and evaluate the approach here against some other possibilities (described in linear ticket) and decide how to proceed. thanks for the poc!

timothycoleman:
this POC of automatically setting the projection thread count is complete

PR #3858, opened 2023-05-31
Removing Commit Count and Prepare Count from DB Settings

### LINKED ISSUE #3750: Want to remove PrepareCount and CommitCount db settings and docs
They don't seem to do anything and the docs are confusing

<sub>[DB-57](https://linear.app/eventstore/issue/DB-57/want-to-remove-preparecount-and-commitcount-db-settings-and-docs)</sub>

lakshdeepsingheventstore:
Fixed: Want to remove PrepareCount and CommitCount db settings and docs

Fixes #3750

These DB configuration settings seem to be redundant and also sometimes tend to be confusing, hence have been removed.


gregoryyoung (commented):
There are some situations where you actually want to use these.

alexeyzimarev:
@gregoryyoung we desperately want to kill explicit transactions, so prepares will disappear anyway. The commit acknowledgement setting is legit (if it still works), but it's not great to have it as a database setting. It's better to have it as an append option set by the client, something we are discussing.

src/EventStore.Core.Tests/Common/ClusterNodeOptionsTests/when_building/with_cluster_node_and_custom_settings.cs
```diff
@@ -85,68 +85,15 @@ public class with_dns_discovery_disabled_and_gossip_seeds_defined<TLogFormat, TS
...
-			};
-
-		[Test]
-		public void should_set_prepare_count_to_the_quorum_size() {
-			Assert.AreEqual(_quorumSize, _options.Cluster.PrepareAckCount);
-		}
-	}
 
 	[TestFixture(typeof(LogFormat.V2), typeof(string))]
 	[TestFixture(typeof(LogFormat.V3), typeof(uint))]
```
hayley-jean, line 96: These `TestFixture` attributes should also be removed with the deleted tests

timothycoleman (changes requested):
looks good :+1: just needs to squash to one commit

timothycoleman:
added breaking change label because ES will not start if these settings are present -  but these settings don't actually do anything (i haven't checked since when exactly)

src/EventStore.Core.Tests/Common/ClusterNodeOptionsTests/when_building/with_cluster_node_and_custom_settings.cs
lakshdeepsingheventstore: TestFixture attributes removed.

hayley-jean (approved):

timothycoleman (approved):
thanks :+1:

PR #3861, opened 2023-06-05
[POC] Implement faster and more memory compact memtable. (DB-165)

YoEight:
Changed: Implement faster and more memory compact memtable.

### DIFF
--- a/src/EventStore.Core.Tests/Index/IndexV1/ptable_midpoint_cache_should.cs
+++ b/src/EventStore.Core.Tests/Index/IndexV1/ptable_midpoint_cache_should.cs
@@ -48,8 +48,6 @@ namespace EventStore.Core.Tests.Index.IndexV1 {
 				memTable.Add((uint)rnd.Next(), rnd.Next(0, 1 << 20), Math.Abs(rnd.Next() * rnd.Next()));
 			}
 
-			memTable.Dump();
-
 			var ptable = PTable.FromMemtable(memTable, file, Constants.PTableInitialReaderCount, Constants.PTableMaxReaderCountDefault, depth, skipIndexVerify: _skipIndexVerify);
 			return ptable;
 		}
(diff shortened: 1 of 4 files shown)


gregoryyoung (commented):
There is a reason lock() {} was not used and instead it was done explicitly with timeouts associated.

Can you assure me that no deadlock will ever appear in any of this code?

YoEight:
@gregoryyoung Not yet!

I get rid of the concurrent dictionary and the inner monitored access to the sorted list to have a clearer view of what can be done.

the memtable purpose is to increase our write output, that's why I want to look at other concurrent data structures. I'm looking into using a ReadWriterLock for example: https://learn.microsoft.com/en-us/dotnet/api/system.threading.readerwriterlock?view=net-7.0

It guarantees one single concurrent write access but multiple reads, which captures the initial intent better imho.

gregoryyoung (commented):
lock (_hash) {
				if (!_hash.TryGetValue(hash, out var block))
					return false;

It previously locked only the list. I would expect locking the entire hash to result in worse performance.

YoEight:
@gregoryyoung  It is indeed!

After I'm done fixing all the failed tests, I'm addressing the concurrent part. Currently it's about putting  the concurrent dictionary back (which get rid of the fat lock the in-between implementation has) and switching over ReadWriteLock at the table level, instead of using Monitor. I'm expecting better performance on read intensive scenarios with with no impact on write intensive ones.

YoEight:
Without any optimization done, here's how the new implementation performs on worst case scenario (where we have to re-arrange the mem table because the new entry we are adding is smaller than the last entry we've done):

100 insertions with random event revisions each time. Position remains monotonic.
```
|     Method |     Mean |    Error |   StdDev | Allocated |
|----------- |---------:|---------:|---------:|----------:|
| Insertions | 6.575 ms | 2.646 ms | 2.475 ms |   7.88 KB |

```

This is how the current implementation performs under the same conditions:

```
|     Method |     Mean |    Error |   StdDev |   Gen0 | Allocated |
|----------- |---------:|---------:|---------:|-------:|----------:|
| Insertions | 7.106 ms | 2.799 ms | 2.618 ms | 1.9531 |  25.07 KB |


```

gregoryyoung:
That allocated is a touch worrying.

Have you done longer tests to see how GC comes into play?

YoEight:
@gregoryyoung The second table is the current implementation we have on the `master` branch not the rewrite I'm currently doing. Right now, I'm conducting deeper benchmarking and better profiling. I even had to setup a Windows box because .NET profiling tools are much better there 😅

timothycoleman:
spring cleaning: closing this pr for now since its not on the agenda, keeping the branch

PR #3865, opened 2023-06-08
Read common name from node certificate

hunter1703:
Added : If common name is not specified in config (in secure mode), read from node certificate

### DIFF
--- a/src/EventStore.Core.Tests/options.cs
+++ b/src/EventStore.Core.Tests/options.cs
@@ -67,9 +67,12 @@ namespace EventStore.Core.Tests {
 				nameof(ClusterVNodeOptions.Unknown.Options),
 			};
 			var actual = new List<string>();
-			ClusterVNodeOptions.FromConfiguration(new FakeConfigurationRoot(
-				new ConfigurationBuilder().Add(new DefaultSource(ClusterVNodeOptions.DefaultValues)).Build(),
-				actual.Add));
+			var specificValues = new[]
+				{ new KeyValuePair<string, object>(nameof(ClusterVNodeOptions.ApplicationOptions.Insecure), true) };
+			var defaultValues = SpecificOrDefaultValues(new(specificValues));
+			ClusterVNodeOptions.FromConfiguration(
+				new FakeConfigurationRoot(new ConfigurationBuilder().Add(new DefaultSource(defaultValues)).Build(),
+					actual.Add));
 
 			var expected = typeof(ClusterVNodeOptions).GetProperties().Where(x => !excluded.Contains(x.Name))
 				.SelectMany(property => property.PropertyType.GetProperties().Where(x => !excluded.Contains(x.Name)))
@@ -77,6 +80,13 @@ namespace EventStore.Core.Tests {
 				.Select(property => property.Name);
 
 			CollectionAssert.AreEquivalent(expected, actual);
+		}
+
+		private static IEnumerable<KeyValuePair<string, object>> SpecificOrDefaultValues(List<KeyValuePair<string, object>> specificValues) {
+			var defaultValues = ClusterVNodeOptions.DefaultValues;
+			if (specificValues is null || specificValues.IsEmpty()) return defaultValues;
+			var specifiedKeys = specificValues.Select(kvp => kvp.Key).ToList();
+			return defaultValues.Where(kvp => !specifiedKeys.Contains(kvp.Key)).Concat(specificValues);
 		}
 
 		public static IEnumerable<PrecedenceCase> PrecedenceCases() {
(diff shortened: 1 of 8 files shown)


dhingrak:
Discussed with Rahul and requested a change regarding the current defaultValues behaviour.

hunter1703:
> Discussed with Rahul and requested a change regarding the current defaultValues behaviour.
I have pushed theses changes.

dhingrak (dismissed):

hayley-jean (changes requested):
I think this PR should wait until https://github.com/EventStore/EventStore/pull/3868 is merged, because we need to test these changes with reloading a certificate as well. The [ClientCertificateAuthenticationProvider](https://github.com/EventStore/EventStore/blob/master/src/EventStore.Core/ClusterVNode.cs#L1001) will need to be updated to reflect changes to the common name when new certificates are loaded

I'm also getting an error when running this in `--dev` mode:

```
Invalid Configuration: A certificate is required unless insecure mode (--insecure) is set.
```

because `--dev` doesn't use the options to load the certificate, so no certificate is found

hunter1703:
> I think this PR should wait until #3868 is merged, because we need to test these changes with reloading a certificate as well. The [ClientCertificateAuthenticationProvider](https://github.com/EventStore/EventStore/blob/master/src/EventStore.Core/ClusterVNode.cs#L1001) will need to be updated to reflect changes to the common name when new certificates are loaded
> 
> I'm also getting an error when running this in `--dev` mode:
> 
> ```
> Invalid Configuration: A certificate is required unless insecure mode (--insecure) is set.
> ```
> 
> because `--dev` doesn't use the options to load the certificate, so no certificate is found

I have addressed issue with `--dev` mode

hayley-jean (changes requested):
I think  a better approach here would be to add a `CertificateReservedCommonName` property to the `CertificateProvider` class and use the provider to get the common name when needed.

That way, you can have the logic to conditionally set the common name when `LoadCertificates` is called in `OptionsCertificateProvider` and there would be no need to call `Sanitize` on the options or change the existing tests

hunter1703:
Makes sense. I have pushed those changes

src/EventStore.Core/ClusterVNodeOptionsExtensions.cs
```diff
@@ -89,6 +89,11 @@ public static class ClusterVNodeOptionsExtensions {
 					ClusterDns = string.Empty
 				}
 			};
+		
+		public static ClusterVNodeOptions WithDevMode(this ClusterVNodeOptions options, bool dev, bool removeDevCerts = false) =>
```
hayley-jean, line 93: Could you add a summary to this extension method please

src/EventStore.Core.Tests/options.cs
```diff
@@ -79,33 +82,40 @@ class FakeConfigurationRoot : IConfigurationRoot {
...
+			return defaultValues.Where(kvp => !specifiedKeys.Contains(kvp.Key)).Concat(specificValues);
+		}
+
 		public static IEnumerable<PrecedenceCase> PrecedenceCases() {
 			yield return new(
 				new Dictionary<string, object> {[nameof(ClusterVNodeOptions.Application.StatsPeriodSec)] = 1},
 				string.Empty,
 				new Dictionary<string, string>(),
-				Array.Empty<string>(),
+				new[] { "--insecure" },
```
hayley-jean, line 97: Is it still necessary to set the insecure flag in these tests after the most recent changes?

src/EventStore.Core.XUnit.Tests/ClusterVNodeOptionsTests.cs
```diff
@@ -9,7 +9,7 @@ public class ClusterVNodeOptionsTests {
 
 	[Fact]
 	public void confirm_suggested_option() {
-		var c = BuildConfiguration("--cluster-sze 3");
+		var c = BuildConfiguration("--cluster-sze 3 --insecure");
```
hayley-jean, line 12: See above

hayley-jean (changes requested):
Thanks, the changes look good 👍 

I just have some small comments on the tests and extension method

src/EventStore.Core/ClusterVNodeOptionsExtensions.cs
hunter1703: Yup. Done

src/EventStore.Core.Tests/options.cs
hunter1703: Removed it.

dhingrak (approved):

timothycoleman:
This behaviour was merged in this PR https://github.com/EventStore/EventStore/pull/3966

PR #3867, opened 2023-06-12
[POC] Implement opt-out Segment telemetry tracking.

YoEight:
Added: Implement opt-out Segment telemetry tracking.

Screenshot 2023-06-12 155831

### DIFF
--- a/src/EventStore.Core/ClusterVNode.cs
+++ b/src/EventStore.Core/ClusterVNode.cs
@@ -62,8 +62,6 @@ using EventStore.Plugins.Authorization;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.Data.Sqlite;
 using Mono.Unix.Native;
-using Segment.Analytics;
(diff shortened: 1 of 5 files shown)


src/EventStore.Core/ClusterVNode.cs
```diff
@@ -1526,14 +1528,15 @@ public class ClusterVNode<TStreamId> :
 				}
 			}
 
+			var segmentConf = new Configuration(options.Application.SegmentWriteKey, flushAt:1, flushInterval:10, storageProvider: new InMemoryStorageProvider());
```
timothycoleman, line 1531: there are two places we're constructing a Configuration object like this, do we need them both?

YoEight: I remember now. Initially, I intended to pass that configuration through dependency injection but it wasn't use by the info controller.

timothycoleman:
Thanks, I'll declare this POC done :+1:

PR #3868, opened 2023-06-12
Loading new certs on calling the admin/reloadconfig endpoint

lakshdeepsingheventstore:
Fixed : Calling the admin/reloadconfig endpoint only reloaded/updated the LogLevel and the certificates were not updated on reloading the config. 

Fixes : https://linear.app/eventstore/issue/DB-166/new-certs-are-not-loaded-on-calling-the-adminreloadconfig-endpoint

As of now, as per the codebase, we only allow reloading of certificates or LogLevel to be readjusted on reloading the config file by calling the admin/reloadconfig endpoint.

### DIFF
--- a/src/EventStore.Core/Certificates/OptionsCertificateProvider.cs
+++ b/src/EventStore.Core/Certificates/OptionsCertificateProvider.cs
@@ -5,22 +5,16 @@ using Serilog;
 namespace EventStore.Core.Certificates {
 	public class OptionsCertificateProvider: CertificateProvider {
 		private static readonly ILogger Log = Serilog.Log.ForContext<ClusterVNode>();
-		private ClusterVNodeOptions _options;
-
-		public OptionsCertificateProvider(ClusterVNodeOptions options) {
-			_options = options;
-		}
-
-		public override LoadCertificateResult LoadCertificates() {
-			if (_options.Application.Insecure) {
+		public override LoadCertificateResult LoadCertificates(ClusterVNodeOptions options) {
+			if (options.Application.Insecure) {
 				Log.Information("Skipping reload of certificates since TLS is disabled.");
 				return LoadCertificateResult.Skipped;
 			}
 
-			var (certificate, intermediates) = _options.LoadNodeCertificate();
+			var (certificate, intermediates) = options.LoadNodeCertificate();
 
 			var certificateCN = certificate.GetCommonName();
-			var reservedNodeCN = _options.Certificate.CertificateReservedNodeCommonName;
+			var reservedNodeCN = options.Certificate.CertificateReservedNodeCommonName;
 
 			if (certificateCN != reservedNodeCN) {
 				Log.Error(
@@ -33,14 +27,14 @@ namespace EventStore.Core.Certificates {
 			var newThumbprint = certificate.Thumbprint;
 			Log.Information("Loading the node's certificate. Subject: {subject}, Previous thumbprint: {previousThumbprint}, New thumbprint: {newThumbprint}",
 				certificate.SubjectName.Name, previousThumbprint, newThumbprint);
-
+			
 			if (intermediates != null) {
 				foreach (var intermediateCert in intermediates) {
 					Log.Information("Loading intermediate certificate. Subject: {subject}, Thumbprint: {thumbprint}", intermediateCert.SubjectName.Name, intermediateCert.Thumbprint);
 				}
 			}
 
-			var trustedRootCerts = _options.LoadTrustedRootCertificates();
+			var trustedRootCerts = options.LoadTrustedRootCertificates();
 
 			foreach (var trustedRootCert in trustedRootCerts) {
 				Log.Information("Loading trusted root certificate. Subject: {subject}, Thumbprint: {thumbprint}", trustedRootCert.SubjectName.Name, trustedRootCert.Thumbprint);
@@ -58,7 +52,7 @@ namespace EventStore.Core.Certificates {
 			Log.Information("All certificates successfully loaded.");
 			return LoadCertificateResult.Success;
 		}
-		
+
 		private static bool VerifyCertificates(X509Certificate2 nodeCertificate, X509Certificate2Collection intermediates, X509Certificate2Collection trustedRoots) {
 			bool error = false;
 
--- a/src/EventStore.Core/ClusterVNode.cs
+++ b/src/EventStore.Core/ClusterVNode.cs
@@ -193,7 +193,6 @@ namespace EventStore.Core {
 		private readonly CertificateDelegates.ClientCertificateValidator _externalClientCertificateValidator;
 		private readonly CertificateDelegates.ServerCertificateValidator _externalServerCertificateValidator;
 		private readonly CertificateProvider _certificateProvider;
-
 		private readonly ClusterVNodeStartup<TStreamId> _startup;
 		private readonly EventStoreClusterClientCache _eventStoreClusterClientCache;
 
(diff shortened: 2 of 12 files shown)


timothycoleman:
Do we know if this has been broken for long time? perhaps we broke it recently somehow

Would be good to see a test if possible

lakshdeepsingheventstore:
> Do we know if this has been broken for long time? perhaps we broke it recently somehow 
> Would be good to see a test if possible

@timothycoleman I tested this on v22.10.2, and observed the same behavior. The steps I followed were as mentioned [here](https://linear.app/eventstore/issue/DB-166/new-certs-are-not-loaded-on-calling-the-adminreloadconfig-endpoint)

hayley-jean:
Can you also check whether this is broken in 21.10 @lakshdeepsingheventstore ? If it is, it would be better to fix and merge it before we cut the release

src/EventStore.Core/Certificates/OptionsCertificateProvider.cs
```diff
@@ -11,7 +11,8 @@ public class OptionsCertificateProvider: CertificateProvider {
 			_options = options;
 		}
 
-		public override LoadCertificateResult LoadCertificates() {
+		//NEW METHOD IMPLEMENTATION OF LoadCertificates() method TO TRACK PREVIOUS THUMBPRINT
+		public override LoadCertificateResult LoadCertificates(string _previousThumbprint) {
```
pvanbuijtene, line 15: Don't think this comment is needed :)

src/EventStore.Core/ClusterVNode.cs
```diff
@@ -62,6 +62,7 @@
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.Data.Sqlite;
 using Mono.Unix.Native;
+using Serilog.Enrichers;
```
pvanbuijtene, line 65: Looks like this is not used.

src/EventStore.Core/ClusterVNode.cs
```diff
@@ -1816,10 +1816,14 @@ public class ClusterVNode<TStreamId> :
 				Log.Information("Skipping reload of certificates since TLS is disabled.");
 				return;
 			}
-
-			if (_certificateProvider?.LoadCertificates() == LoadCertificateResult.VerificationFailed){
+			
+			var prevthumbprint = _certificateProvider.Certificate?.Thumbprint;
+			var temp = new OptionsCertificateProvider(options);
+			if (temp?.LoadCertificates(prevthumbprint) == LoadCertificateResult.VerificationFailed){
```
pvanbuijtene, line 1822: Have you considered passing the options to the `LoadCertificates` call?

src/EventStore.Core/ClusterVNodeOptionsValidator.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.IO;
 using EventStore.Common.Exceptions;
+using EventStore.Core;
```
pvanbuijtene, line 4: Also not used.

src/EventStore.Core/ClusterVNode.cs
lakshdeepsingheventstore: I think passing options to LoadCertificates() call would not help. The only reason to come up with a new implementation of the LoadCertificates() method was because the previous thumbprint value was lost on loading the new certs, so came up with this new implementation. Before 22.10 and before DevCertificates, LoadCertificates() was not implemented in a separate class, so there was no need of tracking the old certificate thumbprint through a parameter.

src/EventStore.Core/ClusterVNodeOptionsValidator.cs
lakshdeepsingheventstore: This is used. We need this because in Line 13, we make use of class ClusterVNodeOptions, for which, we need this.
image

src/EventStore.Core/ClusterVNode.cs
lakshdeepsingheventstore: Removed

src/EventStore.Core/Certificates/OptionsCertificateProvider.cs
lakshdeepsingheventstore: Removed

src/EventStore.Core/ClusterVNodeOptionsValidator.cs
pvanbuijtene: hmm weird, it's already in the `EventStore.Core` namespace 🤔

src/EventStore.Core/ClusterVNode.cs
lakshdeepsingheventstore: Implemented the LoadCertificates() method by passing options as parameter along with the previous thumbprint.

src/EventStore.Core/ClusterVNodeOptionsValidator.cs
lakshdeepsingheventstore: Fixed in the latest commit.

src/EventStore.Core/Certificates/OptionsCertificateProvider.cs
```diff
@@ -6,12 +6,8 @@ namespace EventStore.Core.Certificates {
 	public class OptionsCertificateProvider: CertificateProvider {
 		private static readonly ILogger Log = Serilog.Log.ForContext<ClusterVNode>();
 		private ClusterVNodeOptions _options;
```
pvanbuijtene, line 8: One last change, don't think there's a need for this anymore 🤔 

Besides that it looks good 👍

pvanbuijtene (changes requested):

src/EventStore.Core/Certificates/OptionsCertificateProvider.cs
lakshdeepsingheventstore: Updated

pvanbuijtene (approved):

src/EventStore.Core/ClusterVNodeOptionsValidator.cs
timothycoleman, line 1: do we need the change to this file? it adds a lot to the diff but im not sure it actually does anything

timothycoleman (changes requested):
would it be possible to have a unit test that checks that the reload works correctly now, or is that too awkward?
looks pretty good. i added a comment above to simplify reviewing of this pr 

these commits can be squashed into one commit. It's nice if the commit summary message completes the sentence "When applied, this commit will..." and then the body of the commit message can extra explantion if necessary

here the commit message could be "Use reloaded options when reloading certificates". to me that makes it easy to understand why the dboptions have been moved from the constructor to the method call

src/EventStore.Core/ClusterVNodeOptionsValidator.cs
lakshdeepsingheventstore: The changes that were made here were indentation changes. If you see line number 10, it says "namespace EventStore.Core;", which did not make any sense, since we still had to import EventStore.Core. So we added a brace after "namespace EventStore.Core;"
All the changes you see in the file are indentation changes, automatically implemented by Rider on getting an additional pair of braces.

timothycoleman:
Adding a test of the sort I asked for above is a bit fiddly, I've added a separate ticket to cover this off as part of the other runtime db config changes project https://linear.app/eventstore/issue/DB-237/improve-tests-around-live-config-changes

timothycoleman (approved):
great, thanks :+1:

PR #3869, opened 2023-06-13
Server can return incorrect address in NotHandled message (DB-173)

huynhhuuloc129:
Fixed: Server now always returns a valid address when replying with a NotHandled.NotLeader response

Fixed: https://github.com/EventStore/EventStore/issues/2840
Test all passed.

### DIFF
--- a/src/EventStore.Core/Services/VNode/ClusterVNodeController.cs
+++ b/src/EventStore.Core/Services/VNode/ClusterVNodeController.cs
@@ -840,7 +840,8 @@ namespace EventStore.Core.Services.VNode {
 		}
 
 		private void DenyRequestBecauseNotLeader(Guid correlationId, IEnvelope envelope) {
-			var endpoints = GetLeaderInfoEndPoints();
+			LeaderInfoProvider leaderInfoProvider = new LeaderInfoProvider(_node.GossipAdvertiseInfo, _leader);
+			var endpoints = leaderInfoProvider.GetLeaderInfoEndPoints();
 			envelope.ReplyWith(
 				new ClientMessage.NotHandled(correlationId,
 					ClientMessage.NotHandled.Types.NotHandledReason.NotLeader,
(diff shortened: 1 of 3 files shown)


dhingrak (dismissed):

src/EventStore.Core/Services/VNode/ClusterVNodeController.cs
```diff
@@ -865,7 +865,16 @@ public class ClusterVNodeController<TStreamId> : ClusterVNodeController, IHandle
 					AdvertiseHost: "",
 					AdvertiseHttpPort: 0,
 					AdvertiseTcpPort: 0);
-
+			if (endpoints.TcpEndPoint != null && (_nodeInfo.ExternalTcp ?? _nodeInfo.ExternalSecureTcp).Address.Equals(IPAddress.Any)) {
+				endpoints = (
```
pvanbuijtene, line 869: Are you sure you want to overwrite the `endpoints` value incase it's set when there's `_leader` information available?
There's also a null reference warning which I think should addressed.

huynhhuuloc129: Thank you, I fixed.
I checked the leader is null and `endpoints.TcpEndPoint != null` so  `( _nodeInfo.ExternalTcp ?? _nodeInfo.ExternalSecureTcp)` always != null

huynhhuuloc129: You mean extract the `GetLeaderInfoEndPoints()` to different class?

pvanbuijtene: Don't be afraid to break something, make sure the right tests are in place so you can do the change in a safer way.

The code which is currently added might also break something but you're unaware of it.

To change this, we could extract the method and write tests for the current behavior and add 1 or more tests for the bug you're fixing.

I'm available on Slack if needed.

pvanbuijtene (dismissed):

src/EventStore.Core/Services/VNode/LeaderInfoProvider.cs
```diff
@@ -0,0 +1,72 @@
...
+				var externalTcp = _nodeInfo.ExternalTcp ?? _nodeInfo.ExternalSecureTcp;
+				if (externalTcp != null && externalTcp.Address.Equals(IPAddress.Any)) {
+					endpoints = (
+						TcpEndPoint: _gossipInfo.ExternalTcp ?? _gossipInfo.ExternalSecureTcp,
+						IsTcpEndPointSecure:_gossipInfo.ExternalSecureTcp != null,
+						HttpEndPoint: _gossipInfo.HttpEndPoint,
+						AdvertiseHost:_gossipInfo.AdvertiseHostToClientAs,
+						AdvertiseHttpPort: _gossipInfo.AdvertiseHttpPortToClientAs,
+						AdvertiseTcpPort: _gossipInfo.AdvertiseTcpPortToClientAs);
+				} else {
```
hayley-jean, line 45: This `else` block should be removed. If there is no leader info then we should always return the endpoints from `gossipInfo`.

Otherwise, the node returns the incorrect address if it has an non `0.0.0.0` IP address and an advertised address configured. For example the configuration:

```
IntIp: 127.0.0.1
ExtIp: 127.0.0.1
ClusterSize: 3
DiscoverViaDns: false
GossipSeed: 127.0.0.1:1111,127.0.0.1:2222 # These nodes must not be running
Insecure: true
EnableExternalTcp: true
ReadOnlyReplica: true
AdvertiseHostToClientAs: 192.168.0.7
```

Should return `192.168.0.7` as the leader endpoint even though the `ExtIp` is not `0.0.0.0`

hayley-jean (changes requested):

pvanbuijtene (approved):

PR #3870, opened 2023-06-13
Authenticate client certificateCN with client hostname (DB-113)

dhingrak:
Added: Add check to verify the certificate CN matches with the client host name instead of using the reserved certificate common name.

Fixes https://linear.app/eventstore/issue/DB-113/reserved-common-name-for-gossip-seed

### DIFF
--- a/src/EventStore.Core/Services/Transport/Http/Authentication/ClientCertificateAuthenticationProvider.cs
+++ b/src/EventStore.Core/Services/Transport/Http/Authentication/ClientCertificateAuthenticationProvider.cs
@@ -1,30 +1,47 @@
 ﻿using System;
 using System.Linq;
+using System.Net;
 using System.Security.Cryptography;
 using System.Security.Cryptography.X509Certificates;
 using EventStore.Common.Utils;
+using EventStore.Core.Cluster;
+using EventStore.Core.Services.Gossip;
 using EventStore.Core.Services.UserManagement;
 using Microsoft.AspNetCore.Http;
 using Serilog;
 
 namespace EventStore.Core.Services.Transport.Http.Authentication {
 	public class ClientCertificateAuthenticationProvider : IHttpAuthenticationProvider {
-		private readonly string _certificateReservedNodeCommonName;
+		private string _certificateReservedNodeCommonName;
+		private EndPoint[] _gossipSeed;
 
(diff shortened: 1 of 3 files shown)


src/EventStore.Core/Services/Transport/Http/Authentication/ClientCertificateAuthenticationProvider.cs
```diff
@@ -20,11 +21,16 @@ public class ClientCertificateAuthenticationProvider : IHttpAuthenticationProvid
...
+			if (remoteIpAddress is null) return false;
+
 			bool hasReservedNodeCN;
 			string clientCertificateCN;
+			
 			try {
+				var clientHostName = Dns.GetHostEntry(remoteIpAddress).HostName;
 				clientCertificateCN = clientCertificate.GetCommonName();
-				hasReservedNodeCN = clientCertificateCN == _certificateReservedNodeCommonName;
+				hasReservedNodeCN = VerifyCertificateReservedNodeCommonName(clientCertificateCN, _certificateReservedNodeCommonName, clientHostName);
```
alexeyzimarev, line 33: What would be the value of `clientHostName` here?

dhingrak: This will provide us the host name of the client machine. On my local machine (3 nodes cluster) I am getting the following value for IP: 127.0.0.1 :- 
"kubernetes.docker.internal"

alexeyzimarev: The idea was to use the list of hostnames provided in the gossip seed setting, not the locally resolved hostname.

Here's what I wrote in the issue:
> When node2 connects to node1 could it check if the certificate CN matches with one of the host names in the gossip seed instead of using the reserved certificate common name?

hayley-jean: @alexeyzimarev if we check the certificate CN against the gossip seed, then it means that another node with a different address could not join the cluster like it can today. I think that this is a pretty rare case, but is that acceptable?

dhingrak:
Will resolve conflicts when the other PR merge https://github.com/EventStore/EventStore/pull/3865

PR #3875, opened 2023-06-15
Fix IsExportable check for dev cert on Windows

hayley-jean:
Fixed: Running EventStore with `--dev` on Windows

This started failing on master and looking into it, it appears that the certificates being generated have both `AllowExport` and `AllowPlaintextExport` flags set, which was causing the IsExportable check to fail


shaan1337 (dismissed):
looks good :+1: 

can also consider using HasFlag:
`cngPrivateKey.Key.ExportPolicy.HasFlag(CngExportPolicies.AllowExport)`

hayley-jean:
> looks good 👍
> 
> can also consider using HasFlag: `cngPrivateKey.Key.ExportPolicy.HasFlag(CngExportPolicies.AllowExport)`

Thanks, that looks much better to me!

shaan1337 (approved):

pvanbuijtene (approved):

PR #3883, opened 2023-06-29
Update scavenging documentation DB-183

huynhhuuloc129:
Write documentation: https://linear.app/eventstore/issue/DB-183/write-scavenging-documentation

### DIFF
--- a/docs/operations.md
+++ b/docs/operations.md
@@ -8,24 +8,22 @@ EventStoreDB requires regular maintenance with two operational concerns:
 You might also be interested learning about EventStoreDB [diagnostics](diagnostics.md)
 and [indexes](./indexes.md), which might require some Ops attention.
 
-## Scavenging events
+## Scavenging
 
-When you delete events or streams in EventStoreDB, they aren't removed immediately. To permanently delete
-these events you need to run a 'scavenge' on your database.
+In EventStoreDB, events are no longer present in stream reads or subscriptions after they have been deleted, or they have expired according to the metadata of the stream.
 
-A scavenge operation reclaims disk space by rewriting your database chunks, minus the events to delete, and
-then deleting the old chunks. The scavenged events are also removed from the index.
+The events are, however, still present in the database and will be visible in reads and subscriptions to `$all`.
 
-Running a scavenge causes the active chunk to be completed so that it can be scavenged.
+To remove these events from the database, which may be necessary for GDPR, you need to run a 'scavenge' on each of your nodes.
+
+A scavenge operation removes events and reclaims disk space by creating a copy of the relevant chunk, minus those events, and
+then deleting the old chunk. The scavenged events are also removed from the index.
 
 ::: warning 
 Scavenging is destructive. Once a scavenge has run, you cannot recover any deleted events except from a backup.
 :::
 
 ### Starting a scavenge
-
-EventStoreDB doesn't run Scavenges automatically. We recommend that you set up a scheduled task, for example
-using cron or Windows Scheduler, to trigger a scavenge as often as you need.
 
 You start a scavenge by issuing an empty `POST` request to the HTTP API with the credentials of an `admin`
 or `ops` user:
@@ -38,8 +36,16 @@ You can also start scavenges from the _Admin_ page of the Admin UI.
 ![Start a scavenge in the Admin UI](./images/admin-scavenge.png)
 :::
 
-Each node in a cluster has its own independent database. As such, when you run a scavenge, you need to issue a
+Each node in a cluster has its own independent copy of the database. As such, when you run a scavenge, you need to issue a
 scavenge request to each node. The scavenges can be run concurrently, but can also be run in series to spread the load.
+
+### Getting the current scavenge ID
+
+Get the ID of the currently running scavenge, if there is one, by issuing a `GET` with an ID of `current`
+
+```bash:no-line-numbers
+curl -i -X GET http://127.0.0.1:2113/admin/scavenge/current -u "admin:changeit"
+```
 
 ### Stopping a scavenge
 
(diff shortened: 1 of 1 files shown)


docs/operations.md
```diff
@@ -18,10 +18,54 @@ then deleting the old chunks. The scavenged events are also removed from the ind
...
 Running a scavenge causes the active chunk to be completed so that it can be scavenged.
 
-::: warning 
+::: warning
 Scavenging is destructive. Once a scavenge has run, you cannot recover any deleted events except from a backup.
 :::
 
+### Mechanism of the scavenging
+
+Running scavenge for the first time will take longer than usual because it needs to create a "scavenge point", then execute all phases, create folders and SQL database files, and finally mark the current chunk as completed before removing it.
```
Lougarou, line 27: Running scavenge for the first time will take longer than usual because it needs to create a "scavenge point", then execute all phases, create folders and SQL database files, and finally mark the current chunk as completed before removing it.

->

Running scavenge for the first time will take longer than subsequent scavenges because it needs to create a "scavenge point" as part of the first phase i.e. accumulation phase. This is followed by two more phases at the end of which events are removed from the database, including from the current chunk. More information on the phases can be found below.

docs/operations.md
```diff
@@ -18,10 +18,54 @@ then deleting the old chunks. The scavenged events are also removed from the ind
...
+
+Whenever a scavenging process starts on a node, it checks if there are any existing unused scavenge point on that. If there are, it utilizes one of them; otherwise, it creates a new scavenge point and replicates it to another node.
+
+_Most of the time, the scavenging process does not run on the leader node to reduce the workload on the leader. It starts on one of those nodes and send the information back to the leader._
```
Lougarou, line 31: Remove "_Most of the time, the scavenging process does not run on the leader node to reduce the workload on the leader. It starts on one of those nodes and send the information back to the leader._" and add a best practice section to advise running scavenging on Follower nodes and details there.

docs/operations.md
```diff
@@ -123,6 +165,110 @@ stays in the database:
...
+
+_Thread must be 1 or above_
+
+```
+Ex: POST to http://127.0.0.1:2113/admin/scavenge?thread=2
+```
+
+### Threshold
+
+By default, when running a scavenge, all chunks related to the events that need to be scavenged are removed, and new chunks are created. Setting this option allows you to optimize the chunk weight threshold for scavenging. This optimization helps in resource and time efficiency.
```
Lougarou, line 226: add:

If you find that scavenge is spending a long time scavenging chunks without removing much information from them, you can provide a threshold greater than zero. The weights of the chunks that are being scavenged or skipped can be found in the log files by searching for "with weight". The weight of a chunk is approximately double the number of records that can be removed from that chunk.

docs/operations.md
```diff
@@ -18,10 +18,52 @@ then deleting the old chunks. The scavenged events are also removed from the ind
...
-::: warning 
+::: warning
 Scavenging is destructive. Once a scavenge has run, you cannot recover any deleted events except from a backup.
 :::
 
+### Mechanism of the scavenging
+
+Running scavenge for the first time will take longer compared to subsequent scavenges because it needs to create a "scavenge point" as part of the first phase, known as the accumulation phase. This is followed by two more phases at the end of which events are removed from the database, including from the current chunk. More information on these phases can be found below.
+
+Whenever a scavenging process starts on a node, it checks if there are any existing unused scavenge point on that. If there are, it utilizes one of them; otherwise, it creates a new scavenge point and replicates it to another node.
```
Lougarou, line 29: - Can you we also mention that every time scavenging starts, the current active chunk is closed? 
- Can we also mention that the scavenge point is replicated to other nodes. So you can run scavenging up to the same scavenging point on another nodes using the SyncOnly parameter.

docs/operations.md
```diff
@@ -123,6 +165,110 @@ stays in the database:
...
+* If the latest scavenge point exists and has not been used yet,the scavenge process runs on that scavenge point.
+
+```
+Ex: POST to http://127.0.0.1:2113/admin/scavenge?syncOnly=true
+```
+
+### Best Practices
+
+1. _Plan and schedule_: Create a plan for when and how frequently you will run the scavenging process. Consider factors such as system load and usage pattern to determine the optimal schedule for scavenging. It is generally advisable to perform scavenging during periods of low activity to minimize potential disruptions, especially when running it for the first time.
+2. _Run scavenge on follower node_: Running scavenging on a follower node helps reduce the workload on the leader. After the process, information will be sent back to the leader.
```
Lougarou, line 268: Running scavenging on a follower node helps reduce the workload on the leader. After the process, information will be sent back to the leader.

->

We recommend running scavenging on only one node at a time and to run scavenging on the Follower nodes to avoid adding load on the Leader. Once scavenging is done on the Follower nodes, you can resign the Leader node and perform scavenging on that node.

docs/operations.md
```diff
@@ -123,6 +165,110 @@ stays in the database:
...
+3. _Back up data_: It is important to have a backup before running scavenging in case any unforeseen issues or data loss occurs during the process.
```
Lougarou, line 269: It is important to have a backup before running scavenging in case any unforeseen issues or data loss occurs during the process.

->

Do not take backups while scavenging is running. You can use the /admin/scavenge/current endpoint to find out if scavenging is running.

Lougarou:
Can you also document the /admin/scavenge/current endpoint?

HTTP GET /admin/scavenge/current
This returns the id of the currently running scavenge if any (and link for stopping it)

HTTP DELETE /admin/scavenge/current
Stops the current scavenge regardless of id. Waits for the scavenge to stop before sending the http response.

build.sh
```diff
@@ -5,7 +5,6 @@ PRODUCTNAME="Event Store Open Source"
 COMPANYNAME="Event Store Ltd"
 COPYRIGHT="Copyright 2021 Event Store Ltd. All rights reserved."
 
-
```
timothycoleman, line 8: probably better to remove this change since its unrelated to the pr

docs/operations.md
```diff
@@ -285,4 +451,4 @@ cluster is available in case of a disaster on the primary cluster.
 
 If you are using this strategy, we recommend you only support manual fail over from primary to secondary as
 automated strategies risk causing a [split brain](http://en.wikipedia.org/wiki/Split-brain_%28computing%29)
-problem.
+problem.
```
timothycoleman, line 454: keep the newline at the end of the file please (https://stackoverflow.com/a/729795/866066)

docs/operations.md
```diff
@@ -18,10 +18,54 @@ then deleting the old chunks. The scavenged events are also removed from the ind
 
 Running a scavenge causes the active chunk to be completed so that it can be scavenged.
 
-::: warning 
+::: warning
 Scavenging is destructive. Once a scavenge has run, you cannot recover any deleted events except from a backup.
 :::
 
+### Mechanism of the scavenging
```
timothycoleman, line 25: I think it would probably make sense to have the section explaining the mechanism after the general information about how to run it

docs/operations.md
```diff
@@ -123,6 +167,128 @@ stays in the database:
 
 **Default**: `30` (days)
 
+## Database option
```
timothycoleman, line 170: at the moment this is its own section on the same level as 'Scavenging' itself and also 'Backup and restore', it should probably be part of the scavenging section

there are already some dabase options described above

docs/operations.md
```diff
@@ -123,6 +167,128 @@ stays in the database:
...
+
+### Scavenge backend page size
+
+Specify the page size of the scavenge database.The default value is 16KiB.
+
+| Format               | Syntax                         |
+|:---------------------|:-------------------------------|
+| Command line         | `--scavenge-backend-page-size` |
+| YAML                 | `ScavengeBackendPageSize`      |
+| Environment variable | N/A                            |
```
timothycoleman, line 190: are the environment variables really not applicable? if so we probably made a mistake

docs/operations.md
```diff
@@ -18,10 +18,54 @@ then deleting the old chunks. The scavenged events are also removed from the ind
...
+
+#### 3. Execution
+
+The final phase involves removing all the events and index that need to be scavenged base on the data obtained from the calculation phase. Additionally, small chunks are merged together.
+
+This phase includes the following sub-phases:
+* Chunk Execution: removal of all related chunks.
+* Chunk merging: merging of small chunks.
+* Index execution: removal of all related indexes.
+* Cleaning: cleaning of checkpoint and other data.
```
timothycoleman, line 65: cleaning is its own phase really. it removes data from the scavenge database that is no longer needed

docs/operations.md
```diff
@@ -123,6 +171,128 @@ stays in the database:
 
 **Default**: `30` (days)
 
+### Scavenge database option
```
timothycoleman, line 174: there are already database options described above this section, would probably sensible to merge them together. it's a bit odd at the moment 
image

huynhhuuloc129: Thank you I fixed

docs/operations.md
```diff
@@ -123,6 +171,128 @@ stays in the database:
...
+
+Specify the number of stream hashes to remember when checking for collisions.
+
+| Format               | Syntax                                          |
+|:---------------------|:------------------------------------------------|
+| Command line         | `--scavenge-hash-users-cache-capacity`          |
+| YAML                 | `ScavengeHashUsersCacheCapacity`                |
+| Environment variable | `EVENTSTORE_SCAVENGE_HASH_USERS_CACHE_CAPACITY` |
+
+## HTTP option
```
timothycoleman, line 214: some of the things in this section are options but some aren't. it would probably be nice to just keep this to options that you can provide when starting a scavenge (because thats the only one with lots of options) and explain the other bits elsewhere.

for example the 'Stop current scavenge' section below isn't an option, but there is a section above on stopping a scavenge that we could add the information to instead

docs/operations.md
```diff
@@ -123,6 +171,128 @@ stays in the database:
...
+
+| Format               | Syntax                             |
+|:---------------------|:-----------------------------------|
+| Command line         | `--always-keep-scavenged`          |
+| YAML                 | `AlwaysKeepScavenged`              |
+| Environment variable | `EVENTSTORE_ALWAYS_KEEP_SCAVENGED` |
+
+#### Scavenge backend page size
+
+Specify the page size of the scavenge database.The default value is 16KiB.
```
timothycoleman, line 186:
```suggestion
Specify the page size of the scavenge database. The default value is 16KiB.
```

docs/operations.md
```diff
@@ -123,6 +242,53 @@ stays in the database:
...
+
+Specify the number of stream hashes to remember when checking for collisions.
+
+| Format               | Syntax                                          |
+|:---------------------|:------------------------------------------------|
+| Command line         | `--scavenge-hash-users-cache-capacity`          |
+| YAML                 | `ScavengeHashUsersCacheCapacity`                |
+| Environment variable | `EVENTSTORE_SCAVENGE_HASH_USERS_CACHE_CAPACITY` |
+
+### Best Practices
```
timothycoleman, line 285: This is a great set of best practices. They're in the 'options' section at the moment though and should probably be somewhere else.

Some of the best practices are also already written elsewhere in this same document but I think the wording here is better so we can remove the old

docs/operations.md
```diff
@@ -123,6 +242,53 @@ stays in the database:
...
+
+1. _Plan and schedule_: Create a plan for when and how frequently you will run the scavenging process. Consider factors such as system load and usage pattern to determine the optimal schedule for scavenging. It is generally advisable to perform scavenging during periods of low activity to minimize potential disruptions, especially when running it for the first time.
+2. _Run scavenge on follower node_: We recommend running scavenging on only one node at a time and to run scavenging on the Follower nodes to avoid adding load on the Leader. Once scavenging is done on the Follower nodes, you can resign the Leader node and perform scavenging on that node.
+3. _Back up data_: Do not take backups while scavenging is running. You can use the /admin/scavenge/current endpoint to find out if scavenging is running.
```
timothycoleman, line 289:
```suggestion
3. _Back up data_: Do not take file-copy backups while scavenging is running. You can use the /admin/scavenge/current endpoint to find out if scavenging is running. Disk snapshot backups can be taken while scavenge is running.
```

docs/operations.md
```diff
@@ -123,6 +242,53 @@ stays in the database:
 
 **Default**: `30` (days)
 
+### Always keep scavenged (_Deprecation Notice_)
```
timothycoleman, line 245: what's _Deprecation Notice_ ?

docs/operations.md
```diff
@@ -50,6 +115,12 @@ an `admin` or `ops` user and the ID of the scavenge you want to stop:
 curl -i -X DELETE http://localhost:2113/admin/scavenge/{scavengeId} -u "admin:changeit"
 ```
 
+Or stopping the currently running scavenge regardless of ID.
+
+```
+Ex: DELETE to http://127.0.0.1:2113/admin/scavenge/current
```
timothycoleman, line 121: it would be neat if these samples matched the other samples in the document

Lougarou (approved):

ylorph (approved):

timothycoleman (approved):

PR #3884, opened 2023-06-30
Metrics documentation (DB-185)

hunter1703:
Added documentation for new metrics

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -1,0 +1,467 @@
+# Metrics
+
+EventStoreDB collects metrics in [prometheus format](https://prometheus.io/docs/instrumenting/exposition_formats/#text-based-format), available on the `/metrics` endpoint. Promethus can be configured to scrape this endpoint directly. The metrics are configured in `telemetryconfig.json`. 
+
+::: note
+`/metrics` does not yet contain metrics for Projections and Persistent Subscriptions. To view these in Prometheus it is still recommended to use the [Prometheus exporter](./diagnostics.md#prometheus)
+:::
+
+## Metrics Reference
+
+### Caches
+
+#### Cache hits and misses
+EventStoreDB tracks cache hits/misses metrics for `stream-info` and `chunk` caches.
+
+| Time Series | Type       | Description |
+| :---------- | :--------- | :---------- |
+| `eventstore_cache_hits_misses{cache=<CACHE_NAME>,kind=<"hits"\|"misses">}` | [Counter](#common-types) | Total hits/misses on _CACHE_NAME_ cache   |
+
+Example Configuration:
+```
+"CacheHitsMisses": {
+	"StreamInfo": true,
+	"Chunk": false
+}
+```
+
+Example Output:
+```
+# TYPE eventstore_cache_hits_misses counter
+eventstore_cache_hits_misses{cache="stream-info",kind="hits"} 104329 1688157489545
+eventstore_cache_hits_misses{cache="stream-info",kind="misses"} 117 1688157489545
+```
+
+#### Dynamic Cache Resources
+
+Certain caches that EventStoreDB uses are dynamic in nature i.e. their capacity scales up/down during their lifetime. EventStoreDB records metrics for resources being used by each such dynamic cache.
+
+| Time Series | Type       | Description |
+| :---------- | :--------- | :---------- |
+| `eventstore_cache_resources_bytes{cache=<CACHE_NAME>,kind=<"capacity"\|"size">}` | [Gauge](#common-types) | Current capacity/size of _CACHE_NAME_ cache in bytes |
+| `eventstore_cache_resources_entries{cache=<CACHE_NAME>,kind="count"}`            | [Gauge](#common-types) | Current number of entries in _CACHE_NAME_ cache |
+
+Example Configuration:
+```
+"CacheResources": true
+```
+
+Example Output:
+```
+# TYPE eventstore_cache_resources_bytes gauge
+# UNIT eventstore_cache_resources_bytes bytes
+eventstore_cache_resources_bytes{cache="LastEventNumber",kind="capacity"} 50000000 1688157491029
+eventstore_cache_resources_bytes{cache="LastEventNumber",kind="size"} 15804 1688157491029
+
+# TYPE eventstore_cache_resources_entries gauge
+# UNIT eventstore_cache_resources_entries entries
+eventstore_cache_resources_entries{cache="LastEventNumber",kind="count"} 75 1688157491029
+```
+
+### Checkpoints
+
+| Time Series | Type       | Description |
+| :---------- | :--------- | :---------- |
+| `eventstore_checkpoints{name=<CHECKPOINT_NAME>,read="non-flushed"}` | [Gauge](#common-types) | Value for _CHECKPOINT_NAME_ checkpoint|
+
+Example Configuration:
+```
+"Checkpoints": {
+	"Replication": true,
+	"Chaser": false,
+	"Epoch": false,
+	"Index": false,
+	"Proposal": false,
+	"Truncate": false,
+	"Writer": false,
+	"StreamExistenceFilter": false
+}
+```
+
+Example Output:
+```
+# TYPE eventstore_checkpoints gauge
+eventstore_checkpoints{name="replication",read="non-flushed"} 613363 1688054162478
+```
+
+### Events
+
+These metrics track events written to and read from the server, including reads from caches.
+
+| Time Series | Type       | Description |
+| :---------- | :--------- | :---------- |
+| `eventstore_io_bytes{activity="read"}`               | [Counter](#common-types) | Event bytes read |
+| `eventstore_io_events{activity=<"read"\|"written">}` | [Counter](#common-types) | Events read/written |
+
+Example Configuration:
+```
+"Events": {
+	"Read": false,
+	"Written": true
+}
+```
+
+Example Output:
+```
+# TYPE eventstore_io_events counter
+# UNIT eventstore_io_events events
+eventstore_io_events{activity="written"} 320 1687963622074
+```
+
+### Gossip
+
+Measures the round trip latency and processing time of gossip.
+Usually a node pushes new gossip to other nodes periodically or when its view of the cluster changes. Sometimes nodes pull gossip from each other if there is a suspected network problem. 
+
+#### Gossip Latency
+
+| Time Series | Type       | Description |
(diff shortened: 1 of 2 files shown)


timothycoleman (changes requested):
thanks this looks great :+1:

i've pushed some suggestions to `timothycoleman/metrics-documentation-suggestions` that you could look at and consider encorporating

and some general thoughts

- the metrics endpoint is in prometheus text format that we can probably just link to rather than explaining it https://github.com/prometheus/docs/blob/main/content/docs/instrumenting/exposition_formats.md#text-based-format
- we probably don't need to explain about what counter/histogram etc are explicitly, but if you like we could link to some prometheus or open telemetry documentation
- (however keep the explanations about recentmax and status, they're not different metric types but they do need a bit of explanation)
- perhaps sort the sections alphabetically

timothycoleman (changes requested):
Fab, I've pushed some suggestions to the same suggestions branch as before, let me know what you think. Apart from that I dont think a 'Duration' metric is differen to a Histogram, I'd say we can remove the Duration section and label the duration metrics as histograms, but happy to hear your thoughts on this too

timothycoleman (dismissed):
Great, thanks! :+1:

(i adjusted the commit slightly for some white space and the commit message)

docs/metrics.md
```diff
@@ -0,0 +1,466 @@
+# Metrics
+
+Since EventStoreDB version `v23.6.0`, some metrics in [prometheus format](https://github.com/prometheus/docs/blob/main/content/docs/instrumenting/exposition_formats.md#text-based-format) are available on the `/metrics` endpoint. Promethus can be configured to scrape this endpoint directly. The metrics are configured in `telemetryconfig.json`. In the near future we hope to publish a sample grafana dashboard.
```
ylorph, line 3:
```suggestion
EventStoreDB collects metrics in [prometheus format](https://prometheus.io/docs/instrumenting/exposition_formats/#exposition-formats) , available on the `/metrics` endpoint. Promethus can be configured to scrape this endpoint directly. The metrics are configured in `telemetryconfig.json`. 
```
Resoning for the change: 
- the doc are versionend, so this will appear in 23.6 version and the following one. no need to mention "since"
- prefer to link to the online doc of prometheus ( so that _they_ can track that we use their format ) 
- remove "in the near future" : because it's a roadmap promise in the documentation .

docs/metrics.md
```diff
@@ -0,0 +1,466 @@
...
+## Metrics Reference
+
+### Caches
+
+#### Cache hits and misses
+EventStoreDB tracks cache hits/misses metrics for `stream-info` and `chunk` caches.
+
+| Time Series | Type       | Description |
+| :---------- | :--------- | :---------- |
+| `eventstore_cache_hits_misses{cache=<CACHE_NAME>,kind=<"hits"\|"misses">}` | [Counter](#common-types) | Total hits/misses on _CACHE_NAME_ cache   |
```
ylorph, line 18:
```suggestion
| `eventstore_cache_hits_misses{cache=<CACHE_NAME>,kind=<"hits"\|"misses">}` | [Counter](https://prometheus.io/docs/tutorials/understanding_metric_types/) | Total hits/misses on _CACHE_NAME_ cache   |
```
Reasoning: either link directly to the source doc in prometheus, or copy some information in that section .
the text when following that link is " go to prometheus doc" without a link to it .

docs/metrics.md
```diff
@@ -0,0 +1,466 @@
...
+Example Output:
+```
+# TYPE eventstore_io_events counter
+# UNIT eventstore_io_events events
+eventstore_io_events{activity="written"} 320 1687963622074
+```
+
+### Gossip
+
+Usually a node pushes new gossip to other nodes periodically or when its view of the cluster changes. Sometimes nodes pull gossip from each other if there is a suspected network problem. We measure round trip latency and processing time.
```
ylorph, line 113:
```suggestion
Measures the round trip latency and processing time of gossip.
Usually a node pushes new gossip to other nodes periodically or when its view of the cluster changes.  Sometimes nodes pull gossip from each other if there is a suspected network problem. 
```
reasoning: tell fisrt _what_ is measured, then why .

docs/metrics.md
```diff
@@ -0,0 +1,466 @@
...
+Example Output:
+```
+# TYPE eventstore_current_incoming_grpc_calls gauge
+eventstore_current_incoming_grpc_calls 1 1687963622074
+
+# TYPE eventstore_incoming_grpc_calls counter
+eventstore_incoming_grpc_calls{kind="failed"} 1 1687962877623
+```
+
+#### gRPC methods
```
ylorph, line 180:
```suggestion
#### Client protocol gRPC methods
```

docs/metrics.md
```diff
@@ -0,0 +1,466 @@
...
+
+In addition, EventStoreDB also records metrics for each of the gRPC methods : `StreamRead`, `StreamAppend`, `StreamBatchAppend`, `StreamDelete` and `StreamTombstone`. They are grouped together according to the mapping defined in the configuration.
```
ylorph, line 182:
```suggestion
In addition, EventStoreDB also records metrics for each of client protocol gRPC methods : `StreamRead`, `StreamAppend`, `StreamBatchAppend`, `StreamDelete` and `StreamTombstone`. They are grouped together according to the mapping defined in the configuration.
```

docs/metrics.md
```diff
@@ -0,0 +1,466 @@
...
+eventstore_grpc_method_duration_seconds_bucket{activity="append",status="successful",le="0.0001"} 129 1688157491029
+eventstore_grpc_method_duration_seconds_bucket{activity="append",status="successful",le="0.001"} 143 1688157491029
+eventstore_grpc_method_duration_seconds_bucket{activity="append",status="successful",le="0.01"} 168 1688157491029
+eventstore_grpc_method_duration_seconds_bucket{activity="append",status="successful",le="0.1"} 169 1688157491029
+eventstore_grpc_method_duration_seconds_bucket{activity="append",status="successful",le="1"} 169 1688157491029
+eventstore_grpc_method_duration_seconds_bucket{activity="append",status="successful",le="10"} 169 1688157491029
+eventstore_grpc_method_duration_seconds_bucket{activity="append",status="successful",le="+Inf"} 169 1688157491029
+```
+
+### Kestrel
```
ylorph, line 215: maybe a link to what Kesterl is migh be good ?

docs/metrics.md
```diff
@@ -0,0 +1,466 @@
...
+
+Example Output:
+```
+# TYPE eventstore_kestrel_connections gauge
+eventstore_kestrel_connections 1 1688070655500
+```
+
+### Process
+
+EventStoreDB collects large range of metrics for its own process.
```
ylorph, line 236:
```suggestion
EventStoreDB collects large range of metrics about the running process.
```

docs/metrics.md
```diff
@@ -0,0 +1,466 @@
...
+
+# TYPE eventstore_queue_queueing_duration_max_seconds gauge
+# UNIT eventstore_queue_queueing_duration_max_seconds seconds
+eventstore_queue_queueing_duration_max_seconds{name="Readers",range="16-20 seconds"} 0.06434454 1688157489545
+eventstore_queue_queueing_duration_max_seconds{name="Others",range="16-20 seconds"} 0 1688157489545
+```
+
+### Status
+
+EventStoreDB can track the current status of the `Node`, the `Index`, and the `Scavenge`
```
ylorph, line 352:
```suggestion
EventStoreDB tracks the current status of the `Node` role  as well as the `Index`, and the `Scavenge` processes.
```
I'd put that one on top , this is probably yhe first one we want people to actually track

docs/metrics.md
```diff
@@ -0,0 +1,466 @@
...
+`Scavenge` can have one of the following statuses:
+- Accumulation
+- Calculation
+- Chunk Execution
+- Chunk Merging
+- Index Execution
+- Cleaning
+- Idle
+
+`Node` can have one of the node statuses: `Leader`, `Follower`, etc.
```
ylorph, line 377:
```suggestion
Read more about the node status [here](cluster.html#node-roles)
```

docs/metrics.md
```diff
@@ -0,0 +1,466 @@
...
+}
+```
+
+Example Output:
+```
+# TYPE eventstore_writer_flush_size_max gauge
+eventstore_writer_flush_size_max{range="16-20 seconds"} 410 1688056823193
+```
+
+### System
```
ylorph, line 415: put this one near the `Process` as they are closely related

ylorph (changes requested):
Looks good .


I'd add a section explaining the config file and a with a sample one 
( a bit like the 'what if' here : https://developers.eventstore.com/server/v22.10/configuration.html#testing-the-configuration
with the "label" and regex detailed 
^this does not need to be in this pr ;-) , I'll add an issue .

docs/metrics.md
timothycoleman: thanks Yves, most of those are my doing :)
- I agree about the link, thanks!
- I see what you mean about the promise, do you think it would be ok to say 'in the future'? im happy if you'd prefer to leave it out entirely
- I mentioned 23.6 because there is sort of some overlap with the older prometheus exporter and I wanted to make it clear somehow that this one is the new one (the old one is still documented).

docs/metrics.md
timothycoleman: currently its alphabetical, i'd probably keep it that way but open to persuasion

docs/metrics.md
ylorph: >  see what you mean about the promise, do you think it would be ok to say 'in the future'? im happy if you'd prefer to leave it out entirely

I'd leave it out .

> 23.6 : 
Let's amend the prometheus doc then ?
add a link to this , and make it clear it's about the Stats/gossip endopint

docs/metrics.md
ylorph: is ok alphabetical

docs/metrics.md
hunter1703: Currently, metrics sections are arranged alphabetically. Please let me know if you think some other order is better. (I personally prefer alphabetical order.)

hunter1703: Other then that, I have rephrased the statement according to your suggestion.

docs/metrics.md
hunter1703: Yup. Added it!

ylorph (approved):

timothycoleman:
(squashed into one commit)

timothycoleman (approved):

PR #3887, opened 2023-07-05
fix - out-of-order-error-for-progress-changed

dhingrak:
Fixed: Projections throw an out of order error for Progress changed

Fixes https://linear.app/eventstore/issue/DB-181/projections-throw-an-out-of-order-error-for-progresschanged

Currently we are getting an out of order error message for the progress report. The root cause is that we are publishing the ReportProgress message on a different thread which leads to locking issue.


lakshdeepsingheventstore (approved):
The changes look good to me. I have tested the PR extensively, on a fresh database as well as on an already existing database with 10 chunks. In both the cases, I have been successfully able to append and process upto 200,000 events through custom projections, and did not encounter the Out of order error, or the projection faulted error on any occassion.

hayley-jean (approved):

PR #3890, opened 2023-07-07
Remove reliance on UseNet5CompatFileStream

YoEight:
Changed: Always read the active chunk from memory instead of from the FileStream.

We used Net5CompatFileStream until now because in net6+ if one filestream writes to a file, another filestream that is reading from the same file cannot necessarily tell that anything has changed.

This change makes it so that FileStreams are not instantiated for reading until the chunk is readonly. Instead, the active chunk is always cached, and reads are served from the memstream.

### DIFF
--- a/src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
+++ b/src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
@@ -33,7 +33,8 @@ namespace EventStore.Core.TransactionLog.Chunks.TFChunk {
 		private static readonly ILogger Log = Serilog.Log.ForContext<TFChunk>();
 
 		public bool IsReadOnly {
-			get { return _isReadOnly; }
+			get { return Interlocked.CompareExchange(ref _isReadOnly, 0, 0) == 1; }
+			set { Interlocked.Exchange(ref _isReadOnly, value ? 1 : 0); }
 		}
 
 		public bool IsCached {
@@ -80,7 +81,7 @@ namespace EventStore.Core.TransactionLog.Chunks.TFChunk {
 		private readonly bool _inMem;
 		private readonly string _filename;
 		private int _fileSize;
-		private volatile bool _isReadOnly;
+		private int _isReadOnly;
 		private ChunkHeader _chunkHeader;
 		private ChunkFooter _chunkFooter;
 
@@ -249,7 +250,7 @@ namespace EventStore.Core.TransactionLog.Chunks.TFChunk {
 				throw new CorruptDatabaseException(new ChunkNotFoundException(_filename));
 
 			_fileSize = (int)fileInfo.Length;
-			_isReadOnly = true;
+			IsReadOnly = true;
 			SetAttributes(_filename, true);
 			CreateReaderStreams();
 
@@ -303,7 +304,7 @@ namespace EventStore.Core.TransactionLog.Chunks.TFChunk {
 			Ensure.Positive(fileSize, "fileSize");
 
 			_fileSize = fileSize;
-			_isReadOnly = false;
+			IsReadOnly = false;
 			_chunkHeader = chunkHeader;
 			_physicalDataSize = 0;
 			_logicalDataSize = 0;
@@ -313,12 +314,17 @@ namespace EventStore.Core.TransactionLog.Chunks.TFChunk {
 			else {
 				CreateWriterWorkItemForNewChunk(chunkHeader, fileSize);
 				SetAttributes(_filename, false);
-				CreateReaderStreams();
 			}
 
 			_readSide = chunkHeader.IsScavenged
 				? (IChunkReadSide)new TFChunkReadSideScavenged(this, false, tracker)
 				: new TFChunkReadSideUnscavenged(this, tracker);
+
+			// Always cache the active chunk
+			// If the chunk is scavenged we will definitely mark it readonly before we are done writing to it.
+			if (!chunkHeader.IsScavenged) {
+				CacheInMemory();
+			}
 		}
 
 		private void InitOngoing(int writePosition, bool checkSize, ITransactionFileTracker tracker) {
@@ -328,7 +334,7 @@ namespace EventStore.Core.TransactionLog.Chunks.TFChunk {
 				throw new CorruptDatabaseException(new ChunkNotFoundException(_filename));
 
 			_fileSize = (int)fileInfo.Length;
-			_isReadOnly = false;
+			IsReadOnly = false;
 			_physicalDataSize = writePosition;
 			_logicalDataSize = writePosition;
 
@@ -339,7 +345,6 @@ namespace EventStore.Core.TransactionLog.Chunks.TFChunk {
 			    _chunkHeader.Version != (byte)ChunkVersions.Unaligned)
 				throw new CorruptDatabaseException(new WrongFileVersionException(_filename, _chunkHeader.Version,
 					CurrentChunkVersion));
-			CreateReaderStreams();
 
 			if (checkSize) {
 				var expectedFileSize = _chunkHeader.ChunkSize + ChunkHeader.Size + ChunkFooter.Size;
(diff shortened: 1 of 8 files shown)


src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
```diff
@@ -99,7 +100,7 @@ public enum ChunkVersions : byte {
 		// the lock protects all three parts of the caching process:
 		// CacheInMemory, UnCacheFromMemory and TryDestructMemStreams
 		// so that the variables _cacheStatus, _cachedData, and _cachedLength are synchronized by the lock,
-		// and so is the creation and removal of the mem readers.
+		// and so is the creation and removal of the pooled mem readers.
```
shaan1337, line 103: Is this an intended change? looks like it may be an old change from the base branch: `timothycoleman/chunk-lock`

timothycoleman: ah yeah we can remove the word pooled here, the lock guards them all (pooled or not). i'll do that

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
```diff
@@ -1165,7 +1204,71 @@ private enum CacheStatus {
...
+		// (a) doesn't block if a file reader would be acceptable instead
+		//     (we might be in the middle of caching which could take a while)
+		// (b) _does_ throw if we can't get a memstream and a filestream is not acceptable
+		private bool TryAcquireBulkMemReader(out TFChunkBulkReader reader) {
+			reader = null;
+
+			if (IsReadOnly) {
+				// chunk is definitely readonly and will remain so, so a filestream would be acceptable.
+				// we might be able to get a memstream but we don't want to wait for the lock incase we
+				// are currently performancing a slow operation with it such as caching.
```
shaan1337, line 1217:
```suggestion
				// chunk is definitely read-only and will remain so, so a filestream would be acceptable.
				// we might be able to get a memstream but we don't want to wait for the lock in case we
				// are currently performing a slow operation with it such as caching.
```

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
```diff
@@ -1165,7 +1204,71 @@ private enum CacheStatus {
...
+			if (TryCreateBulkMemReader(out reader))
+				return true;
+
+			// we couldn't get a memreader, maybe we just became readonly and got uncached.
+			if (IsReadOnly) {
+				// we did become readonly, it is acceptable to fall back to filestream.
+				return false;
+			} else {
+				// we are still readonly, we shouldn't have failed to get a memstream and we
+				// cannot fall back to file stream.
```
shaan1337, line 1239:
```suggestion
				// we are not yet readonly, we shouldn't have failed to get a memstream and we
				// cannot fall back to file stream.
```

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
```diff
@@ -673,13 +686,13 @@ private enum CacheStatus {
 		}
 
 		private void BuildCacheArray() {
-			var workItem = GetReaderWorkItem();
+			var workItem = AcquireFileReader();
```
shaan1337, line 689: suggestion: `using var reader = AcquireFileReader()` and remove try-finally clause (+ also rename `workItem` to `reader`)

timothycoleman: i think it'd rather do this in another pr (although i could be talked into doing it in this pr in a separate commit) for the sake of the diff being as clear as possible because these changes are quite delicate

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
timothycoleman: ohyeah :)

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
shaan1337: i'm also ok with the change as-is, i mentioned it only for consistency reasons to make it similar to the way AcquireReader() is called in `VerifyFileHash()`

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
```diff
@@ -1155,7 +1194,7 @@ private enum CacheStatus {
 
 			// if we get here, then we reserved TFChunk for sure so no one should dispose of chunk file
 			// until client returns dedicated reader
-			return new TFChunkBulkReader(this, GetSequentialReaderFileStream());
+			return new TFChunkBulkReader(this, GetSequentialReaderFileStream(), isMemory: false);
```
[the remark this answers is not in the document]
timothycoleman: yeah we wondered about doing something like that, but since (as you say) GetSequentialReaderStream always incremented the filestreamcount even for inmem, i just decided to look at the memdb case as operating at a higher level of abstraction and tried to interfere with it as little as possible here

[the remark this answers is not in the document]
timothycoleman: in the medium to long term i wonder if it would be better to somehow separate the memdb implementation (if we need to keep it at all) it's (sometimes) a bit confusing to have it interleaved in tfchunk

timothycoleman:
cheers for the review, i've encorporated some of the suggestions and replied to others, and squashed back into 1 commit

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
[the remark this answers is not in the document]
shaan1337: > yeah we wondered about doing something like that, but since (as you say) GetSequentialReaderStream always incremented the filestreamcount even for inmem, i just decided to look at the memdb case as operating at a higher level of abstraction and tried to interfere with it as little as possible here

ok sounds good :+1:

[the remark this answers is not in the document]
shaan1337: > in the medium to long term i wonder if it would be better to somehow separate the memdb implementation (if we need to keep it at all) it's (sometimes) a bit confusing to have it interleaved in tfchunk

yup i agree that it's sometimes confusing (and probably error-prone too), maybe abstracting the whole file-based stuff could work

shaan1337:
thanks @timothycoleman

i have a couple of additional comments about --chunks-cache-size:
- on startup, should we validate: --cached-chunks to be least 1 and --chunks-cache-size to be at least 256MB (the chunk size)? if someone sets these config values to 0, they might be under the impression that no memory will be used for caching but it's not true.

there's also this change in behaviour that i'm pondering on
- even if someone sets the --chunks-cache-size to be 256MB, the system might actually use a minimum of 512MB for chunk caching: when completing an active chunk, a new active chunk is automatically created which will also be cached. the previous active chunk's cached memory is not released, so we're actually using 512MB of memory until the background caching process runs again and uncaches the previous chunk if necessary. It's not a lot of additional memory but it might be a problem in  environments with memory constraints (and also we're not honoring --chunks-cache-size)

we could potentially fix that by passing a boolean flag: `uncacheOnCompletion` to TFChunk which would be true if --chunks-cache-size < 2 * ChunkSize

timothycoleman:
hi @shaan1337 

i wondered about the first suggestion, and i could be persuaded, but i decided not to do that for now because i tend to think of the chunk cache from the users perspective as a separate thing (...even though it is the same thing in the chunk). the user can specify some chunk cache, but regardless of the chunk cache we use some memory for the general operation of the database

for the second one, it's already the case before this pr i think

shaan1337:
> hi @shaan1337
> 
> i wondered about the first suggestion, and i could be persuaded, but i decided not to do that for now because i tend to think of the chunk cache from the users perspective as a separate thing (...even though it is the same thing in the chunk). the user can specify some chunk cache, but regardless of the chunk cache we use some memory for the general operation of the database

yes, i think it's fair enough. having to specify the minimum cache size possible would also be a bit harder/inconvenient for the user if it's a non-zero value.

> for the second one, it's already the case before this pr i think

my bad, i mistakenly thought that the active chunk is cached as part of the background process (which would uncache chunks first, then cache chunks), but it is indeed immediately cached.

shaan1337 (dismissed):
great work :+1: please don't merge yet, as i still need to review/approve the base branch PR

shaan1337:
>  the user can specify some chunk cache, but regardless of the chunk cache we use some memory for the general operation of the database

just had some additional thoughts about this:

we could also do the following changes to make it unambiguous for the user:

- add `completed` in the description:
```
[Description("The number of *completed* chunks to cache in unmanaged memory.")]
public int CachedChunks { get; init; } = -1;

[Description("The amount of unmanaged memory to use for caching *completed* chunks in bytes.")]
public long ChunksCacheSize { get; init; } = TFConsts.ChunksCacheSize;
```

- exclude the active chunk from the `totalSize` in CacheUncacheReadOnlyChunks()

the only downside is that it will cause an additional memory usage of 256MB based on a user's previous config

timothycoleman:
i discussed the above with shaan and we concluded it's a bit simpler as it currently is

timothycoleman (approved):

ylorph:
Forced Merge

PR #3891, opened 2023-07-07
Make StreamExistenceFilter 10k by default in tests

timothycoleman:
Changed: reduced memory usage in tests

There are some tests that are constructing the MiniNode without setting a smaller value. But it would be very very unusual for a test to need more than 10k


timothycoleman:
covered by https://github.com/EventStore/EventStore/pull/3901

PR #3896, opened 2023-07-11
Prevent torn transactions during replication (DB-66)

shaan1337:
Fixed: Prevent torn transactions during replication

Torn transactions can currently happen on replicas for the following reasons:
- replication is done in 8kb chunks (DataChunkBulk) and thus some events of an implicit transaction may be in one DataChunkBulk while others may be in the next DataChunkBulks.
- during replication, log records are unframed from the DataChunkBulks one at a time. The writer checkpoint is moved after writing each log record, and the writer is flushed at an arbitrary time (when the queue is empty or after a minimum delay)
- based on the above two points, if replication is randomly interrupted, a replica may have its writer checkpoint moved/flushed in the middle of a transaction. thus its log may contain only "half" of a transaction if viewed at an arbitrary point in time.

This is usually not a problem because when the replica reconnects to the leader, it'll replicate the rest of the transaction. but if the replica becomes the leader, then the log will contain only part of the transaction. Furthermore, the events of an implicit transaction are not committed to the index as long as the whole transaction has not been received. This causes the problems explained in detail here: https://github.com/EventStore/EventStore/pull/3896#issuecomment-1630579265

This PR builds upon https://github.com/EventStore/EventStore/pull/3918 to prevent torn transactions on replicas. https://github.com/EventStore/EventStore/pull/3918 prevents transactions from spanning more than one chunk which makes the problem much simpler:
- during replication of a "data" chunk, we can hold a transaction in memory, then write it to disk only when we receive the whole transaction. To achieve this, the PR introduces a new framer called the "TransactionFramer" which unframes data transaction-wise instead of logrecord-wise.
- replication of a "raw" (scavenged chunk) is done in the same way as before - its first written to a temporary file, then switched in. Given that the chunk should end with a complete transaction, we can safely move the writer checkpoint to the end of the scavenged chunk. Also given that only a "prefix" of a stream can be scavenged, it means that a scavenged chunk will always end with a log record that represents the end of a transaction (which is a nice property to have)

Most of the code in this PR introduces the necessary infrastructure for advanced replication testing - the actual fix is relatively small (~160 lines).


src/EventStore.Core/Helpers/LengthPrefixSuffixFramer.cs
```diff
@@ -96,7 +97,7 @@ public class LengthPrefixSuffixFramer {
 						_memStream.SetLength(_packageLength - PrefixLength); // remove suffix length
 						_memStream.Position = 0;
 
-						_packageHandler(_binaryReader);
+						_packageHandler?.Invoke(_binaryReader);
```
timothycoleman, line 100: if `_packageHandler` had a default action that was a noop then we could avoid the null check here (and in the two new framers)

src/EventStore.Core/Services/Replication/TransactionFramer.cs
```diff
@@ -0,0 +1,54 @@
...
+
+	public void Reset() {
+		_records.Clear();
+		_inner.Reset();
+	}
+
+	private void OnLogRecordUnframed(ILogRecord record) {
+		_records.Add(record);
+		if (record.IsTransactionBoundary()) {
+			_handler?.Invoke(_records.ToArray());
```
timothycoleman, line 33: are the ToArray() calls here and below important? it's quite a lot of allocations we could possibly avoid

src/EventStore.Core/Services/ClusterStorageWriterService.cs
```diff
@@ -259,6 +263,7 @@ public class ClusterStorageWriterService<TStreamId> : StorageWriterService<TStre
 				Log.Debug("Completing raw chunk {chunkStartNumber}-{chunkEndNumber}...", message.ChunkStartNumber,
 					message.ChunkEndNumber);
 				Writer.CompleteReplicatedRawChunk(_activeChunk);
+				Flush();
```
timothycoleman, line 266: is this just for neatness or is it important for correctness?

src/EventStore.Core/Services/ClusterStorageWriterService.cs
```diff
@@ -333,23 +343,11 @@ public class ClusterStorageWriterService<TStreamId> : StorageWriterService<TStre
...
-			long newPos;
-			if (!Writer.Write(record, out newPos))
-				ReplicationFail(
-					"First write failed when writing replicated record: {0}.",
-					"First write failed when writing replicated record: {record}.",
-					record);
+		private void OnTransactionUnframed(ILogRecord[] records) {
+			Writer.OpenTransaction();
+			foreach (var record in records)
+				Writer.WriteToTransaction(record, out _);
```
timothycoleman, line 349: should we call the TryWriteToTransaction instead and call ReplicationFail if it returns false?

timothycoleman (changes requested):
Looks great, and its come out very neatly too :+1:

questions inline

src/EventStore.Core/TransactionLog/Chunks/TFChunkDb.cs
```diff
@@ -11,6 +11,7 @@ namespace EventStore.Core.TransactionLog.Chunks {
 	public class TFChunkDb : IDisposable {
 		public readonly TFChunkDbConfig Config;
 		public readonly TFChunkManager Manager;
+		public ITransactionFileTracker Tracker => _tracker;
```
timothycoleman, line 14: in the test that calls this you can probably just pass `new TFChunkTracker.NoOp()` instead

src/EventStore.Core/Helpers/LengthPrefixSuffixFramer.cs
shaan1337: fixed :+1:

src/EventStore.Core/Services/ClusterStorageWriterService.cs
shaan1337: fixed :+1:

src/EventStore.Core/TransactionLog/Chunks/TFChunkDb.cs
shaan1337: fixed :+1:

src/EventStore.Core/Services/Replication/TransactionFramer.cs
shaan1337: well spotted, fixed :+1:

src/EventStore.Core/TransactionLog/LogRecords/LogRecordExtensions.cs
```diff
@@ -0,0 +1,16 @@
...
+
+public static class LogRecordExtensions {
+	public static bool IsTransactionBoundary(this ILogRecord rec) {
+		return rec.RecordType switch {
+			LogRecordType.Stream => false, // log v3 stream record
+			LogRecordType.EventType => false, // log v3 event type record
+			LogRecordType.Prepare => rec is IPrepareLogRecord prepare &&
+			                         (!prepare.Flags.HasFlag(PrepareFlags.IsCommitted) || // explicit transaction
+			                         prepare.Flags.HasFlag(PrepareFlags.TransactionEnd)), // last prepare in implicit transaction
+			_ => true
```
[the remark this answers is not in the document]
shaan1337: @YoEight it might not be safe if we add a new log record type that's not atomic. would it be preferable to enumerate all the log record types? (the performance impact should be negligible i think since apparently the compiler builds a jump/hash table if there are enough options)

timothycoleman:
rebased to latest master because the nunit commit has already been encorporated into master so is no longer needed in this pr

timothycoleman (dismissed):

YoEight (approved):

timothycoleman (approved):

PR #3899, opened 2023-07-12
Added CaughtUp control message for subscriptions

hunter1703:
Added: "CaughtUp" gRPC control message for subscriptions when a subscription becomes live


src/EventStore.Core/Services/Transport/Grpc/Enumerators.StreamSubscription.cs
```diff
@@ -222,6 +222,12 @@ public class StreamSubscription<TStreamId> : StreamSubscription {
 
 				async Task PumpLiveMessages() {
 					await caughtUpSource.Task.ConfigureAwait(false);
+					
+					await _channel.Writer.WriteAsync(new ReadResp {
+						LiveCaughtUp = true
+					}, _cancellationToken).ConfigureAwait(false);
+					
+					Console.WriteLine("caught up to live");
```
timothycoleman, line 230: this left in accidentally perhaps

timothycoleman:
This looks cool, thanks!

I wonder about the name `LiveCaughtUp`, perhaps `IsLive`?

Unit test would be great as well of course

timothycoleman:
@thefringeninja will the existing clients ignore the new message?

thefringeninja:
@timothycoleman the dotnet one will. Anything that it doesn't recognize is turned into an `Unknown` message.

As for the boolean, what about `CaughtUp`? Alternatively, a message describing each condition might be more appropriate. Like

```
Empty CaughtUp = 8;
Empty FellBehind = 9;
```

or even

```
AllStreamPosition CaughtUp = 8;
Empty FellBehind = 9;
```

timothycoleman:
@thefringeninja cool, `CaughtUp` is nice. I don't think I have a preference between keeping it a bool or having a separate 'FellBehind' message, is there an advantage either way? Former is in danger of being a property-sourcingy I suppose

thefringeninja:
@timothycoleman only that the latter allows us the possibility to enrich it later. For example, knowing the position you got up to as you went back into catch up mode.

timothycoleman:
I'd be happy with bespoke messages if you prefer @thefringeninja. that does give us options for the future and fits with a bunch of the other fields in the oneof:

```
		SubscriptionConfirmation confirmation = 2;
		Checkpoint checkpoint = 3;
		StreamNotFound stream_not_found = 4;
```

so we could have something like:

```
...
		CaughtUp caught_up = 8;
		FellBehind fell_behind = 9;
	}

	message CaughtUp {
	}

	message FellBehind {
	}
```

if that works for everyone?

@hunter1703 after that i think we just need some test coverage in this repo and its looking great

although, i wonder if we might also want the same thing for $all subscriptions (in a separate pr)

hunter1703:
@timothycoleman I have already made changes for both $all and filtered $all subscriptions in addition to stream subscriptions. As for your other suggestions, I have added new message types and tests.

src/EventStore.Core/Services/Transport/Grpc/Enumerators.AllSubscriptionFiltered.cs
```diff
@@ -4,6 +4,7 @@
 using System.Threading;
 using System.Threading.Channels;
 using System.Threading.Tasks;
+using EventStore.Client;
```
timothycoleman, line 7:
```suggestion
```

src/EventStore.Core/Services/Transport/Grpc/Enumerators.StreamSubscription.cs
```diff
@@ -4,6 +4,7 @@
 using System.Threading;
 using System.Threading.Channels;
 using System.Threading.Tasks;
+using EventStore.Client;
```
timothycoleman, line 7:
```suggestion
```

src/EventStore.Core/Services/Transport/Grpc/Enumerators.AllSubscription.cs
```diff
@@ -4,6 +4,7 @@
 using System.Threading;
 using System.Threading.Channels;
 using System.Threading.Tasks;
+using EventStore.Client;
```
timothycoleman, line 7:
```suggestion
```

src/EventStore.Core/Services/Transport/Grpc/Enumerators.AllSubscription.cs
```diff
@@ -217,13 +215,10 @@ public class AllSubscription : IAsyncEnumerator<ReadResp> {
...
 				async Task PumpLiveMessages() {
-					var position = await caughtUpSource.Task.ConfigureAwait(false);
+					await caughtUpSource.Task.ConfigureAwait(false);
 
 					await _channel.Writer.WriteAsync(new ReadResp {
-						Checkpoint = new ReadResp.Types.Checkpoint {
-							CommitPosition = position.CommitPosition,
-							PreparePosition = position.PreparePosition
-						}
+						CaughtUp = new ReadResp.Types.CaughtUp()
```
timothycoleman, line 221: Do we definitely want to send the CaughtUp instead of the Checkpoint? it might be better to send both because the checkpoint was presumably there for a reason, but I am not 100% certain

timothycoleman (changes requested):
great, thanks! a couple of suggestions and questions inline.

im also quite suspicious about adding things to the production code for the purpose of supporting the tests

- i'd rather have the test enumerator stuff in the tests themselves and make the grpc classes public (@thefringeninja thoughts on this? im not sure if there is a strong reason to keep them internal)
- or alternatively use the GrpcSpecification like when_subscribing_to_all does, but this comes with the downside of having to run a whole mininode

src/EventStore.Core/Services/Transport/Grpc/Enumerators.AllSubscription.cs
hunter1703: So, I had a conversation with @thefringeninja.
`i forget who mentioned it but there is a bug there where we send an extra checkpoint once the user has caught up live. that behavior should be removed`
`yes because if you subscribe from that checkpoint you will miss something. so right now it's at most once delivery which is extremely wrong`

hunter1703:
Just my thoughts :

src/EventStore.Core/Services/Transport/Grpc/Enumerators.AllSubscription.cs
timothycoleman: ahh right, thanks!

timothycoleman (approved):
Thanks! I adjusted the tests a bit and I've approved it.

@thefringeninja would you mind reviewing as well?

> Production code should be testable by itself (in most cases) and unit testing framework should not have to do convoluted stuff (e.g. we try to avoid reflection in unit testing as long as we can) just to test it.

Yeah I agree

> Making classes public would precipitate all the way down to making protobuf classes public which are internal sealed. Do we want to make them public?

I went for this approach for now

thefringeninja (approved):
Looks good. If we decide to add the position on caughtup / fellbehind we can always do so later

PR #3902, opened 2023-07-15
Adjust _fileStreamCount properly when failing to create a ReaderWorkItem (DB-208)

timothycoleman:
Changed: No longer log misleading error about 'Max internal streams limit' when running out of file handles

Previously we left _fileStreamCount and _internalStreamsCount incremented, eventually causing the error "Unable to acquire reader work item. Max internal streams limit reached." to be logged and causing it to be impossible to successfully close the chunk because it will be waiting for more readers than exist.

We also here adjust the error message because 'streams' sounds like it means event store streams but it doesn't.

The filestream 'lights out' mechanism has been simplified by making it idempotent


YoEight (approved):

shaan1337 (approved):

timothycoleman:
im pretty sure this change is safe... want to cut 23.6 before merging this _just in case_

PR #3918, opened 2023-08-01
Prevent implicit transactions from spanning multiple chunks (DB-66)

shaan1337:
Added: Prevent implicit transactions from spanning multiple chunks

This PR prevents implicit transactions from going past chunk boundaries: it ensures that a whole batch of events is written to the next chunk if there is not enough space in the current chunk. Explicit transactions still work the same way as before and can go past a chunk boundary.

This PR also effectively restricts the max size of a transaction to the chunk size (i.e 256MB) . It was previously unbounded.

The PR does not also cover the replication part - it will be dealt with separately.

Thanks to @timothycoleman for the great suggestions/ideas.

## Motivation
This PR is closely linked to: https://github.com/EventStore/EventStore/pull/3808 which still allows transactions to go past chunk boundaries but the writer checkpoint is moved only after the whole transaction has been written (prior to that, the writer checkpoint was moved after every event write). This effectively prevents torn transactions from occurring on the _leader_ node. 

However, when doing the work to prevent torn transactions on _replica_ nodes, it has become apparent that allowing transactions past chunk boundaries increases the complexity of the problem by many times.

<details>
<summary>See details</summary>
<blockquote>
Consider the following scenario:
- a transaction that contains 3 events A, B and C is torn across 2 chunks: A is in chunk-1, B & C are in chunk-2.
- the transaction has been written & committed on the leader and is being replicated to a replica.

### Writing partial transactions directly to the disk
Say the replica has received events A and B but not yet C. It writes events A and B to chunk-1 and chunk-2 respectively but then the leader crashes and this replica becomes the leader. The replica will need to rollback the partial transaction from its log. This involves doing 3 things:
- delete chunk-2 on the replica
- "uncomplete" chunk-1 by resetting its read-only status to false, re-create its work item etc.
- truncate chunk-1 to the beginning of the transaction (this also requires recomputing the MD5 hash up to the beginning of the transaction)

This is feasible (i have working code for it) but adds complexity.

###  Writing partial transactions to memory first
To avoid the need to rollback, say we write the whole transaction to memory first then to disk. This is not as easy as it seems due to scavenged chunks. Say there is a partial transaction at the end of a non-scavenged chunk and the next chunk is a scavenged one.

We don't have any guarantee that the scavenged chunk being replicated ends with a complete transaction. So we might need to wait for an arbitrary amount of scavenged chunks to be replicated until we have one that ends with a complete transaction before we can do a commit (i.e move the writer checkpoint). This is a problem because many chunks may need to be re-replicated in case the replica crashes.

### Moving the writer checkpoint to the last complete transaction of a scavenged chunk
So instead of doing the above, say we identify the last complete transaction in the scavenged chunk and move the writer checkpoint there to be sure we only have "complete transactions" in the log. This causes the following problem (which is also present when we write the transaction directly to the disk instead of to memory first):

Say the replica crashes after replicating a scavenged chunk and the writer checkpoint is in the middle (i.e not at the end) of the scavenged chunk:

This state is a bit inconsistent because:
- When the replica starts up, it needs to resume replication from the end of the scavenged chunk while keeping the writer checkpoint at the current position. It will presumably rejoin the leader and continue replicating.
- If a backup is taken in this state and a restore done from it, there'll be a torn transaction in the log.

All of the above problems are solvable but again, they add a lot of complexity.

### The proposed solution
Preventing partial transactions at chunk boundaries simplifies the problem a lot during replication as:
- When we've replicated a chunk, whether it's scavenged or not, we can simply move the writer checkpoint to the end of the chunk. We don't need to keep track of multiple chunks at any point in time. Of course, existing torn transactions in the log will still be present but they have always been there previously so it's not a big issue.
- The replica can then easily hold the transaction in memory and write it to disk after the whole transaction is received from the leader.
</blockquote>
</details>

The proposed solution (i.e Preventing transactions to go past chunk boundaries) is also not as straightforward as it seems. The main difficulty lies in knowing the size of a transaction before we write it to a chunk.

<details>
<summary>See details</summary>
<blockquote>
We could:

i) simply write the log records to a memory stream first to know the size
- this is not efficient as it adds more buffers (we also already write the log records again to a temporary buffer before writing to the chunks)

ii) calculate an approximate upper bound for the size of the transaction
- there are some parts of a log record that don't have a predictable size (e.g stream id, event type because they're UTF-8 encoded), so we could cheaply calculate an approximate upper bound for its size. Unfortunately, we cannot easily create a static method to calculate the approximate log record sizes as Log V3 has some parts that are protobuf-encoded. So say we create the log record to do the calculation of the upper bound: The main issue is that we will not be able to set the exact `LogPosition` property of all the log records in the transaction, so we will need to re-create the log records a second time which causes extra allocations.

iii) calculate the exact size of the log record
- this approach is more efficient as it does not cause any extra buffering or extra allocations and is the adopted method in this PR. Its complexities are detailed below.
</blockquote>
</details>

### Complexities behind calculating the exact size of a log record
As mentioned above, there are some parts of a log record that don't have a predictable size: the stream id and event type because they're UTF-8 encoded. To calculate the exact size of a prepare log record, we thus need to re-encode the stream id/event type strings to UTF-8 (by calling Encoding.UTF8.GetByteCount()).

Given that all the strings we receive over the wire via TCP/gRPC are protobuf-encoded, it means they are already encoded to UTF-8 so we could reuse this information to avoid doing the extra encoding work. However after weighing the pros and cons, it's not worth doing this optimization (atleast for now) as:
- it doesn't have significant performance benefits (less than 1% regression on the write throughput if we don't apply the optimization)
- it will add complexity to sharing the proto files across multiple projects
- C# support for native UTF-8 strings is not mature enough

Draft PR for UTF-8 optimization:
https://github.com/EventStore/EventStore/pull/3939


timothycoleman:
> It will require some minor code changes on all our clients to support the new proto.

Do they have to support the new proto at all, it might be neater to have the client protos as strings and only change it to bytes in some server-only copy

src/Directory.Build.props
```diff
@@ -33,4 +33,7 @@
 	<ItemGroup>
 		<RuntimeHostConfigurationOption Include="System.IO.UseNet5CompatFileStream" Value="true" />
 	</ItemGroup>
+	<PropertyGroup>
+		<ErrorOnDuplicatePublishOutputFiles>false</ErrorOnDuplicatePublishOutputFiles>
+	</PropertyGroup>
```
timothycoleman, line 38:
```suggestion
```

it seems to build/publish ok for me locally without this change (haven't tried on ci though)

src/EventStore.Projections.Core/Services/Processing/EmittedStream.cs
```diff
@@ -142,7 +144,9 @@ public class StreamMetadata {
 			if (ioDispatcher == null) throw new ArgumentNullException("ioDispatcher");
 			if (readyHandler == null) throw new ArgumentNullException("readyHandler");
 			_streamId = streamId;
+			_streamIdSize = Helper.UTF8NoBom.GetBytes(_streamId).Length;
```
timothycoleman, line 147:
```suggestion
			_streamIdSize = Helper.UTF8NoBom.GetBytes(_streamId).Length;
			_streamIdSize = Helper.UTF8NoBom.GetByteCount(_streamId);
```

when we just want the length it might be cheaper to do it this way

YoEight (changes requested):
I am not convinced that the optimization that leverages protobuf encoded data worth the hassle of both changing the definitions and threading of the computed size down to the tfchunk code.

UTF-16 to UTF-8 conversion is a O(n) process where n is the number of code points in the string. All readable characters, regardless of the languages (roman or japanese to name a few) have single code points.

We are not dealing with very long stream names either. If we were constantly dealing with 1_000 long stream names, we would indeed add an extra 0.001ms latency on every invocation, which would be amortized the larger the event batch is.

What is really sad is prepare logs encode their flags before the transaction log position. If that wasn't the case, all type of records would encode their transaction log position before their specific data. It would be part of the record header, along the type and version fields. We would have been able to serialize record payload ahead of time and only have to use the size of the byte writer to know how much space we would need. The fact that prepare logs are "aware" of their transaction log position means that optimization is out of reach if we are not willing to change prepare log binary layout.

src/Directory.Build.props
shaan1337: fixed (removed after PR was split into 2) (occurs only on .NET 7.0)

src/EventStore.Projections.Core/Services/Processing/EmittedStream.cs
shaan1337: thanks, that's better :+1:

shaan1337:
@timothycoleman @YoEight thanks for your reviews!
i think you've both convinced me about removing the UTF-8 optimization - my initial thoughts were to add the optimization just because we're already getting it for free and to prevent any regression on the write throughput but the extra complexity is not worth it. Based on @timothycoleman's tests, `Encoding.UTF8.GetByteCount()` takes less than 40 ns for a 100 character UTF-8 string (with non ASCII characters) and it's going to affect the write throughput by less than 1% in the worst case.

i've force pushed the changes (thanks for splitting up the PR @timothycoleman)

timothycoleman (approved):
Great work, thanks!

YoEight (approved):

PR #3920, opened 2023-08-03
Add an override option for anonymous gossip

hayley-jean:
Added: New 'OverrideAnonymousGossipAccess' option which will override the 'AllowAnonymousEndpiontAccess' setting for the gossip endpoint when set to true.

This is set to true by default, but we intend on changing the default after we are sure that the clients can support it.


timothycoleman (approved):
:+1:

PR #3927, opened 2023-08-08
Fix chunk caching/uncaching race conditions

timothycoleman:
_isCached used to just have two values: 0 meaning uncached and 1 meaning cached, set at the start of successful calls to CacheInMemory and UnCacheFromMemory respectively.

This meant it was dangerous to call CacheInMemory and UnCacheFromMemory simultaneously (this was well advertised and didnt happen in practice) But it also made it dangerous (in fairly rare circumstances), to call CacheInMemory and ReturnReaderWorkItem simultaneously, which is common.

Consider the scenario where a chunk has been cached, then uncached, but there is still an outstanding memreader so _cachedData has not been cleared. This scenario is uncommon, but possible. Say we are chunk 1, we've been cached and then uncached as part of normal operations, and then a scavenge runs that happens not to affect us but sufficiently reduces the size of chunks 2 and 3 such that we, chunk 1, fit in the chunk cache with those chunks, then we may be cached again.

In the first place, if we are cached again, a new assignment will be made to _cachedData, rendering it impossible to free the old _cachedData resulting in a leak.

Furthermore, if thread A calls CacheInMemory simultaneously with thread B calling ReturnReaderWorkItem, the following interleaving is possible:

- B returns the reader to the pool in ReturnReaderWorkItem
- B checks the _isCached flag, is it is 0 so B decides to call TryDestructMemStreams()

- A sets _isCached to 1
- A calls BuildCacheArray(), which assigns to _cachedData

- B checks _memStreamCount, it is 1
- B removes the last mem reader from the pool and decrements _memStreamCount to 0
- B therefore calls FreeCachedData(), freeing the memory pointed to by _cachedData and setting the pointer to 0

- A Adds n on to _memStreamCount (taking it from 0 to n)
- A instantiates n readerworkitems that read from _cachedData

The readerworkitems just created be A may be on null or possibly even wild _cacheData.

Another interleaving that looks possible is for A to create the new readerworkitems, then for B to remove them, resulting in a 'cached' chunk with no mem readers.

In this PR we extend the _isCached flag to be a richer _cacheStatus with four states instead of the previous two.

- Uncached
- Caching
- Cached
- Uncaching

CacheInMemory is guarded by a successful transition from Uncached to Caching UnCacheFromMemory is guarded by a successful transition from Caching to Uncaching

Therefore calls can simultaneously be made to both safely.

We transition from Caching to Cached when successfully completing the call to CacheInMemory We transition from Uncaching to Uncached only after the last reader is returned

Therefore calls to CacheInMemory in the scenario above will not do anything, until after the last reader is returned.

//qqqq explain this last bit better
FreeCachedData after returning a workitem is guarded by a successful transition from Uncaching to Uncached We transition from Uncaching to Uncached when the last reader is returned and the _cachedData is released

### DIFF
--- a/src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
+++ b/src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
@@ -37,7 +37,7 @@ namespace EventStore.Core.TransactionLog.Chunks.TFChunk {
 		}
 
 		public bool IsCached {
-			get { return _isCached != 0; }
+			get { return GetCacheStatus() == CacheStatus.Cached; }
 		}
 
 		// the logical size of data (could be > PhysicalDataSize if scavenged chunk)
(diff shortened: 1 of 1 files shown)


timothycoleman:
there were still some uncovered cases here, and it's too fiddly. doing a lock based approach instead https://github.com/EventStore/EventStore/pull/3930

PR #3928, opened 2023-08-09
Simplify TFChunk by removing unused paths (DB-321)

timothycoleman:
Changed: Removed unused code paths

Alignv2File was never called

ResizeMemStream was only called when scavenging an in-memory chunk, which isn't supported on new or old scavenges Added an error log to make this slightly more obvious


shaan1337 (approved):
looks good :+1: 

while testing, i just noticed something not directly linked to this PR: the chunks switch lock is not released (called in `HandleCleanupWhenFinished`) when an exception is thrown in the scavenger factory.

timothycoleman:
thanks, i filed DB-329 about the lock. would be nice to sort out but not too urgent because it just means redaction wont be able to run after attempting a scavenge on an inmem database, which is unlikely to be important

PR #3930, opened 2023-08-09
Make an explicit lock for chunk Caching/Uncaching/TryDestructMemStreams (DB-321)

timothycoleman:
Fixed: race conditions when caching/uncaching chunks

(recommended to review commit at a time)

CacheInMemory and UnCacheFromMemory were correctly synchronized, but UnCacheFromMemory doesn't do the whole uncaching.
The uncaching continues in TryDestructMemStreams each time a mem reader is returned, and eventually the _cachedData is freed.
These continuations of uncaching can be concurrent with a subsequent call to CacheInMemory resulting in a variety of potential problems (elaborated below)

This PR makes an explicit lock around all three parts of the caching process: CacheInMemory, UnCacheFromMemory and TryDestructMemStreams
so that the variables _cacheStatus, _cachedData, and _cachedLength are synchronized by the lock, and so is the creation and removal of the mem readers.

CacheInMemory and UnCacheFromMemory are called relatively infrequently, and won't contend with each other. We aren't worried about performance implications of the lock for those.
However, ReturnReaderWorkItem is called frequently on the read path. To help make this call quick we don't acquire the lock every time we return a memory work item,
instead we use a lock-free check to determine if we might need to acquire the lock first.
Then, if we do need to acquire the lock, it would be a shame if we got blocked waiting for a slow CacheInMemory call, but to minimise the risk of this, we make _isCached into _cacheStatus and give it three states. The chunk will only be recached after it has finished uncaching to returning mem readers should rarely, if ever, occur while caching.

ObjectPool suffered from a similar problem, which we fix here also.

### Interleaving problems elaborated

Consider the scenario where a chunk has been cached, then uncached, but there is still an outstanding memreader so _cachedData has not been cleared. This scenario is uncommon, but possible. Say we are chunk 1, we've been cached and then uncached as part of normal operations, and then a scavenge runs that happens not to affect us but sufficiently reduces the size of chunks 2 and 3 such that we, chunk 1, fit in the chunk cache with those chunks, then we may be cached again.

In the first place, if we are cached again, a new assignment will be made to _cachedData, rendering it impossible to free the old _cachedData resulting in a leak.

Furthermore, if thread A calls CacheInMemory simultaneously with thread B calling ReturnReaderWorkItem, the following interleaving is possible:

- B returns the reader to the pool in ReturnReaderWorkItem
- B checks the _isCached flag, is it is 0 so B decides to call TryDestructMemStreams()

- A sets _isCached to 1
- A calls BuildCacheArray(), which assigns to _cachedData

- B checks _memStreamCount, it is 1
- B removes the last mem reader from the pool and decrements _memStreamCount to 0
- B therefore calls FreeCachedData(), freeing the memory pointed to by _cachedData and setting the pointer to 0

- A Adds n on to _memStreamCount (taking it from 0 to n)
- A instantiates n readerworkitems that read from _cachedData

The readerworkitems just created be A may be on null or possibly even wild _cacheData.

Another interleaving that looks possible is for A to create the new readerworkitems, then for B to remove them, resulting in a 'cached' chunk with no mem readers.

### DIFF
--- a/src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
+++ b/src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
@@ -639,9 +639,9 @@ namespace EventStore.Core.TransactionLog.Chunks.TFChunk {
 					return;
 				}
 
-				bool wasZero = Interlocked.Add(ref _memStreamCount, _maxReaderCount) - _maxReaderCount == 0;
+				Interlocked.Add(ref _memStreamCount, _maxReaderCount);
 				if (_selfdestructin54321) {
-					if (Interlocked.Add(ref _memStreamCount, -_maxReaderCount) == 0 && !wasZero)
+					if (Interlocked.Add(ref _memStreamCount, -_maxReaderCount) == 0)
 						FreeCachedData();
 					Log.Debug("CACHING ABORTED for TFChunk {chunk} as TFChunk was probably marked for deletion.", this);
 					return;
@@ -1021,7 +1021,6 @@ namespace EventStore.Core.TransactionLog.Chunks.TFChunk {
 					writerWorkItem.DisposeMemStream();
 
 				int memStreamCount = Interlocked.CompareExchange(ref _memStreamCount, 0, 0);
-				var wasZero = memStreamCount == 0;
 
 				ReaderWorkItem workItem;
 				while (_memStreams.TryDequeue(out workItem)) {
@@ -1030,13 +1029,10 @@ namespace EventStore.Core.TransactionLog.Chunks.TFChunk {
 
 				if (memStreamCount < 0)
 					throw new Exception("Count of memory streams reduced below zero.");
-				if (memStreamCount == 0)
-				{
-					if (!wasZero) {
-						// we are the last who should "turn the light off" for memory streams
-						FreeCachedData();
-					}
 
+				if (memStreamCount == 0) {
+					// make sure "the light is off" for memory streams
+					FreeCachedData();
 					return true;
 				}
 


src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
```diff
@@ -374,7 +398,7 @@ public enum ChunkVersions : byte {
 			var md5 = MD5.Create();
 
 			// ALLOCATE MEM
-			Interlocked.Exchange(ref _isCached, 1);
+			_cacheStatus = CacheStatus.Cached;
 			_cachedLength = fileSize;
 			_cachedData = Marshal.AllocHGlobal(_cachedLength);
```
shaan1337, line 403: do we also need to lock the CreateInMemChunk contents with a _cachedDataLock?

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
```diff
@@ -979,38 +1012,44 @@ public enum ChunkVersions : byte {
...
-			return false;
 		}
 
 		private void FreeCachedData() {
-			var cachedData = Interlocked.Exchange(ref _cachedData, IntPtr.Zero);
-			if (cachedData != IntPtr.Zero) {
-				Marshal.FreeHGlobal(cachedData);
-				GC.RemoveMemoryPressure(_cachedLength);
+			lock (_cachedDataLock) {
+				var cachedData = _cachedData;
```
shaan1337, line 1045: can we get rid of the `cachedData` variable, now that it's within a lock?

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
```diff
@@ -1066,8 +1105,36 @@ public enum ChunkVersions : byte {
...
+				// 2. we don't call TryDestructMemStreams when we should.
+				//    the memory barrier, corresponding to the barrier in UnCacheFromMemory, protects
+				//    against this by guaranteeing ordering as emphasied in _italics_ below.
+				//    the case protected against is that another thread might call UnCacheFromMemory just as
+				//    we are returning an item here and neither thread tidys up the item we are returning.
+				//    but consider: UnCacheFromMemory sets the state to Uncaching _and then_ tidys up the
+				//    items in the pool. if it does not tidy up our item, it is because our item wasn't
+				//    in the pool. which means we hadn't enqueued it yet, which means that we will
+				//    enqueue it and then _after that_ read _cacheStatus and find it is Uncaching,
+				//    which means we will tidy it up here.
```
shaan1337, line 1128: do we also need to argue about the case where we don't tidy it up the memstream on this thread? :

```
if we don't tidy it here, it  means that _cacheStatus was not yet CacheStatus.Uncaching (it must have been CacheStatus.Cached). we can be sure that UnCacheFromMemory() will tidy us up as we've already crossed our barrier and thus UnCacheFromMemory will see our item in the queue and tidy it up.
```

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
timothycoleman: i dont _think_ so, because inmem doesn't do any uncaching/recaching, it just caches when creating the chunk and keeps it cached

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
timothycoleman: yeah we could

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
timothycoleman: i don't think we need to explain about it because its the contrapositive

we explained that `!tidied_there => tidied_here`
therefore we get `!tidied_here => tidied_there` (the contrapositive) for free

and its already quite a long comment 😅

shaan1337: nice :smile:  i hadn't realized that

src/EventStore.Core/DataStructures/ObjectPool.cs
```diff
@@ -104,6 +105,7 @@ public class ObjectPool<T> : IDisposable {
 
 		public void Return(T item) {
 			_queue.Enqueue(item);
+			Thread.MemoryBarrier();
 			if (_disposing)
```
shaan1337, line 109: there's something i'm not sure about:
is there any guarantee that if `_disposing` is false here, the other thread will be able to dequeue the enqueued item? or TryDequeue might also return false?

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
```diff
@@ -1066,8 +1105,36 @@ public enum ChunkVersions : byte {
...
+				//    we protect against this by checking the condition again inside the lock
+				// 2. we don't call TryDestructMemStreams when we should.
+				//    the memory barrier, corresponding to the barrier in UnCacheFromMemory, protects
+				//    against this by guaranteeing ordering as emphasied in _italics_ below.
+				//    the case protected against is that another thread might call UnCacheFromMemory just as
+				//    we are returning an item here and neither thread tidys up the item we are returning.
+				//    but consider: UnCacheFromMemory sets the state to Uncaching _and then_ tidys up the
+				//    items in the pool. if it does not tidy up our item, it is because our item wasn't
+				//    in the pool. which means we hadn't enqueued it yet, which means that we will
+				//    enqueue it and then _after that_ read _cacheStatus and find it is Uncaching,
```
shaan1337, line 1127: what guarantees do we have that the read of _cacheStatus will return `Uncaching` here? it appears that it may still be `Cached` but I may have gotten something wrong.

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
shaan1337: fair enough :+1: 

i was just thinking if it would make sense if we move the `_cacheStatus = CacheStatus.Cached` assignment further below in the call after we have created the memstreams just to be consistent with acquire-release semantics. i.e if later on we make a change where another thread reads _cacheStatus = Cached (say outside the _cachedDataLock for some reason), then we can be sure that it has been cached.

src/EventStore.Core/DataStructures/ObjectPool.cs
shaan1337: resolving this as after discussions with @timothycoleman i realized the question wasn't completely valid.

shaan1337 (approved):

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
shaan1337: after discussions with @timothycoleman and some research: a barrier should enforce cache coherence, so the above read should return the latest value given that we know it's already been written by another CPU core.

timothycoleman:
To comment a bit more carefully on the MemoryPool, the shape of the code is like this

Thread A, calling Dispose:

```
_disposing = true
<barrier>
empty the pool and dispose each item
```

Thread B, calling Return:

```
enqueue the item in the pool
<barrier>
if _disposing then empty the pool and dispose each item
```

The barriers have been added in this PR. They are necesssary because we do need A to set _disposing before emptying the pool, and we do need B to enqueue before checking _disposing. _disposing is volatile but that does not prevent the reorderings (setting it to true has write-release semantics which means it can't be reordered with earlier operations but it can be reordered with later ones, and checking it has read-acquire semantics which means it cant be reordered with later operations but can be reordered with earlier ones), so these are the opposite barriers to the ones we need. The concurrent queue itself may well contain full barriers in the dequeue/enqueue implementations (it uses interlocked operations which act as a full barrier) but we dont want to rely on this implementation detail, and in any case it seems likely that the barriers have to also be in the right place (at the start of dequeueing and the end of enqueueing respectively)

So does it work now? What we need to avoid is A and B interleaving in such a way that neither of them dispose the item that B is returning.

Case (i): If B returns the item to the pool before A empties the pool, then A will dispose the item. No problem. (we _are_ relying on the concurrent queue being able to dequeue an item as soon as it has been completely enqueued. if the queue only made the item available after, say, 1 second, then this approach would not work... but also neither would explicit locking around enqueue/dequeue)

Case (ii): If B returns the item to the pool just too late for A to dispose it, then it must be that A has at least already started emptying the pool, so it must have _already_ (because of the barrier) set _disposing to true. Therefore when B checks disposing _after_ returning the item (because of the other barrier), it must find it to be true.

I thought for a while here that we are relying on cache coherency of the memory (which is a guarantee given by the .net platform (https://github.com/dotnet/runtime/blob/main/docs/design/specs/Memory-model.md), and I was worried about the possible complexities that multi-copy atomicity might introduce, but on reflection I _think_ we are only relying on the memory barriers and not directly on cache coherence (though the implementation of the memory barriers themselves will presumably rely on cache coherence). Consider case (ii) again. There are 4 operations, and we have a total order on all of them

1. A sets _disposing before starting to empty the pool
2. A starts to empty the pool before B enqueues the item in the pool (both of these intrincially require reading/writing memory)
3. B enqueues the item in the pool before reading _disposing

These are each guaranteed by memory barriers. Therefore, if the barriers do indeed preserve order, B must find _disposing set to true. 

However I'm not convinced that I have the best way of thinking about this, happy to be corrected

src/EventStore.Core/DataStructures/ObjectPool.cs
timothycoleman: i tried to write something coherent (tee hee) about this below https://github.com/EventStore/EventStore/pull/3930#issuecomment-1686216888

PR #3934, opened 2023-08-16
[DBX-34] Update API to get current database settings

dhingrak:
Added: Add more information to the current database settings API response

Fixes: https://linear.app/eventstore/issue/DB-15/new-api-to-get-current-database-settings

The goal of this PR is to provide more information regarding the EventStoreDB database setting so the user who does not have access to the system can see how the database is configured. Attached the current API response and the new API response.

[oldAPIresponse.txt](https://github.com/EventStore/EventStore/files/15313455/oldAPIresponse.txt)
[newAPIresponse.txt](https://github.com/EventStore/EventStore/files/15420207/newAPIresponse.txt)

### DIFF
--- a/src/EventStore.Common/Configuration/UnitAttribute.cs
+++ b/src/EventStore.Common/Configuration/UnitAttribute.cs
@@ -1,0 +1,11 @@
+﻿using System;
+
+namespace EventStore.Common.Configuration;
+
+[AttributeUsage(AttributeTargets.Property)]
+public class UnitAttribute(string unit) : Attribute {
+	public string Unit { get; } = unit;
+}
+
+
+
--- a/src/EventStore.Core/Services/Transport/Http/Controllers/InfoController.cs
+++ b/src/EventStore.Core/Services/Transport/Http/Controllers/InfoController.cs
@@ -10,6 +10,7 @@ using EventStore.Core.Data;
 using EventStore.Core.Messages;
 using EventStore.Plugins.Authentication;
 using EventStore.Plugins.Authorization;
+using Newtonsoft.Json.Linq;
 using ILogger = Serilog.ILogger;
 
 namespace EventStore.Core.Services.Transport.Http.Controllers {
@@ -80,8 +81,10 @@ namespace EventStore.Core.Services.Transport.Http.Controllers {
 						Name = x.Metadata.Name,
 						Description = x.Metadata.Description,
 						Group = x.Metadata.SectionMetadata.SectionType.Name,
-						PossibleValues = x.Metadata.AllowedValues.Length > 0 ? x.Metadata.AllowedValues : null,
 						Value = x.DisplayValue,
+						ConfigurationSource = x.SourceDisplayName,
+						DeprecationMessage = x.Metadata.DeprecationMessage,
+						Schema = x.Metadata.OptionSchema
 					}
 				);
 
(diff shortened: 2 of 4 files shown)


src/EventStore.Core/ClusterVNodeOptions.cs
```diff
@@ -138,7 +138,7 @@ public record ApplicationOptions {
 
 			[Description("Disable HTTP caching.")] public bool DisableHttpCaching { get; init; } = false;
 
-			[Description("The number of seconds between statistics gathers.")]
+			[Description("The number of seconds between statistics gathers."), OptionTypeForTime("Seconds")]
```
pvanbuijtene, line 141: It's a bit more easier to read when these are on new lines :)

src/EventStore.Core/Services/Transport/Http/Controllers/InfoController.cs
```diff
@@ -95,7 +97,10 @@ public class OptionStructure {
 			public string Description { get; set; }
 			public string Group { get; set; }
 			public string Value { get; set; }
-			public string[] PossibleValues { get; set; }
+			public string[] AllowedValues { get; set; }
+			public string OptionType { get; set; }
```
pvanbuijtene, line 101: Not sure if we should expose the dotnet types over JSON 🤔 could look into the types supported by https://json-schema.org/

And for the duration settings it's maybe an idea to expose those as an `ISO 8601 ABNF` duration, this way we could have a more consistent way of communicating durations which are also standardized and part of json schema.

src/EventStore.Core/ClusterVNodeOptions.cs
```diff
@@ -779,17 +779,17 @@ public record ProjectionOptions {
...
 			[Description("Fault the projection if the Event number that was expected in the stream differs " +
 			             "from what is received. This may happen if events have been deleted or expired.")]
 			public bool FaultOutOfOrderProjections { get; init; } = false;
 
-			[Description("The time in milliseconds allowed for the compilation phase of user projections")]
+			[Description("The time in milliseconds allowed for the compilation phase of user projections"), OptionTypeForTime("Milliseconds")]
 			public int ProjectionCompilationTimeout { get; set; } = 500;
 
-			[Description("The maximum execution time in milliseconds for executing a handler in a user projection. It can be overridden for a specific projection by setting ProjectionExecutionTimeout config for that projection")]
+			[Description("The maximum execution time in milliseconds for executing a handler in a user projection. It can be overridden for a specific projection by setting ProjectionExecutionTimeout config for that projection"), OptionTypeForTime("Milliseconds")]
```
pvanbuijtene, line 792: There are some gossip & tcp heart beat settings missing.

pvanbuijtene (commented):
A few settings are missing and maybe need to decide how we want to expose the settings.

src/EventStore.Core/Services/Transport/Http/Controllers/InfoController.cs
dhingrak: yeah, mapping the dotnet types is still pending. I will work on this and mark the linear issue to InProgress

src/EventStore.Common/Configuration/UnitAttribute.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+
+namespace EventStore.Common.Configuration;
+
+[AttributeUsage(AttributeTargets.Property)]
+public class UnitAttribute(string message) : Attribute {
+	public string Message { get; } = message;
```
hayley-jean, line 7: `Unit` would be a better name for this property

```suggestion
	public string Unit { get; } = unit;
```

hayley-jean (changes requested):
I think we should add units (`d`) for  `ScavengeHistoryMaxAge` since that's similar to the other time-based units

hayley-jean (approved):

PR #3941, opened 2023-08-29
Send checkpoints on correct interval after catching-up (DB-348)

pvanbuijtene:
Fixed: Checkpoints of filtered $all subscription not always send on correct interval.


shaan1337 (approved):
looks good :+1:

dhingrak (approved):
changes looks good

hayley-jean (approved):

PR #3947, opened 2023-09-05
Call home from the database.

YoEight:
Added: Call home database telemetry.

- First payload is sent after the server is up for 1h.
- Payloads are sent once per 24h after that.
- Notice and optout message is logged on startup:
```
Telemetry
---------
EventStoreDB collects usage data in order to improve your experience. The data is anonymous and collected by EventStore Ltd.
You can opt out of sending telemetry by setting the EVENTSTORE_TELEMETRY_OPTOUT environment variable to true.
For more information visit https://eventstore.com/telemetry
```

- Payload contents:
```json
{
  "version": "23.6.1.0",
  "tag": "oss-v23.6.0-91-g2091bb932",
  "uptime": "00:00:30.2391235",
  "cluster": {
    "leaderId": "b4950b0e-abf1-4036-961e-878812cf9b0c",
    "nodeId": "b4950b0e-abf1-4036-961e-878812cf9b0c",
    "nodeState": "Leader"
  },
  "configuration": {
    "clusterSize": 3,
    "enableAtomPubOverHttp": true,
    "enableExternalTcp": true,
    "insecure": false,
    "runProjections": "All"
  },
  "database": {
    "epochNumber": 26,
    "firstEpochId": "783e96d6-a595-4127-8d7c-03a084b21ec6",
    "activeChunkNumber": 0
  },
  "environment": {
    "coreCount": 24,
    "isContainer": false,
    "isKubernetes": false,
    "processorArchitecture": "X64",
    "totalDiskSpace": 1000186310656,
    "totalMemory": 68631052288
  },
  "gossip": {
    "b8be0b16-ddc1-4eb5-a869-9fd189b5dcf5": "ReadOnlyReplica",
    "a9d7ef4d-c63a-494c-af8e-ddfd2862aa3a": "Follower",
    "8c3dd2b6-a5bb-4b44-a4fd-07d90f2c7e16": "Follower",
    "b4950b0e-abf1-4036-961e-878812cf9b0c": "Leader"
  },
  "projections": {
    "customProjectionCount": 1,
    "standardProjectionCount": 4,
    "customProjectionRunningCount": 1,
    "standardProjectionRunningCount": 0,
    "totalCount": 5,
    "totalRunningCount": 1,
  },
  "persistentSubscriptions": {
    "count": 0
  }
}
```

### DIFF
--- a/src/EventStore.Core/Telemetry/ContainerInfo.cs
+++ b/src/EventStore.Core/Telemetry/ContainerInfo.cs
@@ -1,0 +1,29 @@
+using System;
+using System.IO;
+using EventStore.Common.Utils;
+
+namespace EventStore.Core.Telemetry; 
+
+public class ContainerInfo {
+	public static ContainerInfo Instance { get; } = Collect();
+	public bool IsContainer { get; init; }
+	public bool IsKubernetes { get; private set; }
+
+	private static ContainerInfo Collect() {
+		var info = new ContainerInfo {
+			 IsContainer = Environment.GetEnvironmentVariable("DOTNET_RUNNING_IN_CONTAINER") != null,
+		};
+
+		if (OS.OsFlavor != OsFlavor.Linux || !File.Exists("/proc/self/cgroup"))
+			return info;
+
+		try {
+			string cgroup = File.ReadAllText("/proc/self/cgroup");
+			info.IsKubernetes = cgroup.Contains("kubepods");
+		} catch (Exception) {
+			// ignored
+		}
+
+		return info;
+	}
+}
--- a/src/EventStore.Projections.Core/Services/Management/ProjectionManager.cs
+++ b/src/EventStore.Projections.Core/Services/Management/ProjectionManager.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text.Json.Nodes;
 using EventStore.Common.Options;
 using EventStore.Common.Utils;
 using EventStore.Core;
@@ -12,6 +13,7 @@ using EventStore.Core.Services.TimerService;
 using EventStore.Core.Services.UserManagement;
 using EventStore.Projections.Core.Messages;
 using EventStore.Core.Helpers;
+using EventStore.Core.Telemetry;
 using EventStore.Projections.Core.Services.Processing;
 using EventStore.Projections.Core.Standard;
 using EventStore.Projections.Core.Common;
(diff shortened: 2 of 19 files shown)


alexeyzimarev:
I'd like to change a couple of things:
* Change the serialiser options to use camel case
* Rename `UpTime` to `uptime`
* Report the number of custom projections separately
* Report eh number of running projections. If possible, the number of running standard and custom projections separately.

timothycoleman (approved):

src/EventStore.Projections.Core/Services/Management/ProjectionManager.cs
```diff
@@ -1304,5 +1307,40 @@ public PendingProjection(long projectionId, ProjectionManagementMessage.Command.
...
 		}
+
+		public void Handle(TelemetryMessage.Request message) {
+			var customProjectionCount = 0;
+			var customProjectionRunningCount = 0;
+			var standardProjectionCount = 0;
+			var standardProjectionRunningCount = 0;
+
+			foreach (var proj in _projections.Values) {
+				if (proj.Mode != ProjectionMode.Continuous)
```
hayley-jean, line 1318: Are we not interested in whether users are running queries or one-time projections?

YoEight: My rationals were they would be quite hard to capture live anyway because we only query the manager once day. Also, they cause to ramp up artificially the `totalCount` telemetry too.

src/EventStore.Core/Telemetry/ContainerInfo.cs
lakshdeepsingheventstore, line 1:
```
if (OS.OsFlavor != OsFlavor.Linux || !File.Exists("/proc/self/cgroup"))
			return info;

		try {
			string cgroup = File.ReadAllText("/proc/self/cgroup");
			info.IsKubernetes = cgroup.Contains("kubepods");
		} catch (Exception) {
			// ignored
		}
```
Shouldn't we consider having the try block start before the if statement, as there might be access restrictions to cgroup file, which might throw an exception?

hayley-jean (approved):

src/EventStore.Core/Telemetry/ContainerInfo.cs
timothycoleman: doing what you suggested would make it more obviously correct, but it so happens that File.Exists doesn't throw anyway and returns false if it can't figure out if a file exists

PR #3957, opened 2023-09-12
Ensure not override the filter when updating a persistent subscription (DB-315)

huynhhuuloc129:
Fixed: Updating a persistent subscription clears the filter


timothycoleman (commented):
thanks!

- would you mind removing the extra whitespace changes that have been introduced on those empty lines
- there is a test failing at the moment
- an extra test that check that this fix works would be fab

timothycoleman (changes requested):
thanks - i suspect something isn't quite right with the test because it passes even without the fix

huynhhuuloc129:
> thanks - i suspect something isn't quite right with the test because it passes even without the fix
thank you, I'll check it

timothycoleman (approved):
great, thanks :+1:

hayley-jean (approved):

PR #3966, opened 2023-09-25
Configure `CertificateReservedNodeCommonName` to be the node's CN by default

shaan1337:
Changed: Configure `CertificateReservedNodeCommonName` to be the node's CN by default
Changed: Make `CertificateReservedNodeCommonName` dynamically reloadable
Fixed: Cache client certificate authentication results for better performance/to make sure already established TLS connections continue to work properly.

This PR makes the following main change:
- read the CN from the node's certificate and use it as the `CertificateReservedNodeCommonName` (as proposed by @alexeyzimarev). The rationale is that in most cases, the CN is exactly the same on all nodes so this makes the user experience better by not having to specify an additional configuration option. In cases where the CN doesn't match on all nodes, a wildcard FQDN can be manually specified (https://github.com/EventStore/EventStore/pull/3960)

Additionally, the following changes are included:
- `CertificateReservedNodeCommonName` is now dynamically reloadable when reloading certificates. This means that new certificates don't necessarily need to match with the CN of the previous set of certificates. In production, it is reasonable to assume that cases will exist where certificates are rotated after a long period of time (e.g a few years) and within this time there may have been PKI changes within an organization enforcing new constraints on certificates. So this change also makes the user experience better as no restart is required.
- To make the above change possible, an additional optimization is required. Prior to this PR, all HTTP requests were re-authenticated by the client certificate authentication provider even if they belonged to the same TLS connection (for e.g due to HTTP/2 multiplexing). The authentication result is now cached for better performance / to make sure that already established TLS connections continue to work properly.

### DIFF
--- a/src/EventStore.Core.Tests/Services/Transport/Http/Authentication/client_certificate_authentication_provider.cs
+++ b/src/EventStore.Core.Tests/Services/Transport/Http/Authentication/client_certificate_authentication_provider.cs
@@ -6,7 +6,6 @@ using System.Security.Cryptography.X509Certificates;
 using System.Threading.Tasks;
 using EventStore.Core.Services.Transport.Http.Authentication;
 using EventStore.Core.Services.UserManagement;
-using EventStore.Core.Util;
 using Microsoft.AspNetCore.Http;
 using NUnit.Framework;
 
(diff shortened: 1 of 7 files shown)


timothycoleman (approved):
Shaan and I are a bit suspicious about the locking the dictionary in ClientCertificateAuthenticationProvider.cs

It will warrent further investigation - is it possible that two requests for the same connection are handled simultaneously and present the same dictionary? if so how are they supposed to be synchronised?

For now locks are present 'just in case'

PR #3974, opened 2023-09-29
Decode Base64 usernames/passwords as UTF-8 strings (DB-421)

shaan1337:
Fixed: Allow users to login with usernames/passwords having unicode characters

Corresponding UI PR: https://github.com/EventStore/EventStore.UI/pull/364

The password change tests have been marked as `Explicit` and must be run manually as password changes take a long time to be processed just after startup (since the password change notification reader's stream does not initially exist and it waits for 10 seconds before the next retry)

### DIFF
--- a/src/EventStore.Core.Tests/ClientAPI/UserManagement/creating_a_user.cs
+++ b/src/EventStore.Core.Tests/ClientAPI/UserManagement/creating_a_user.cs
@@ -57,5 +57,18 @@ namespace EventStore.Core.Tests.ClientAPI.UserManagement {
 			Assert.AreEqual("foo", d.Groups[0]);
 			Assert.AreEqual("bar", d.Groups[1]);
 		}
+
+		[Test]
+		public async System.Threading.Tasks.Task creating_a_user_with_unicode_chars_can_be_readAsync() {
+			UserDetails d = null;
+			await _manager.CreateUserAsync("码ou£ro码", "ourofull", new[] { "foo", "bar" }, "ou码码ro",
+				new UserCredentials("admin", "changeit"));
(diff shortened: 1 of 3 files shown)


YoEight (approved):

timothycoleman (approved):

shaan1337:
For posterity, assuming that the encoding is UTF-8:

- is fine (extract from [RFC 7617](https://datatracker.ietf.org/doc/html/rfc7617))
```
   The original definition of this authentication scheme failed to
   specify the character encoding scheme used to convert the user-pass
   into an octet sequence.  In practice, most implementations chose
   either a locale-specific encoding such as ISO-8859-1 ([ISO-8859-1]),
   or UTF-8 ([RFC3629]).  For backwards compatibility reasons, this
   specification continues to leave the default encoding undefined, as
   long as it is compatible with US-ASCII (mapping any US-ASCII
   character to a single octet matching the US-ASCII character code).
```

- is still compatible with [RFC 7617](https://datatracker.ietf.org/doc/html/rfc7617) as the `charset` parameter is optional.

```
   In challenges, servers can use the 'charset' authentication parameter
   to indicate the character encoding scheme they expect the user agent
   to use when generating "user-pass" (a sequence of octets).  This
   information is purely advisory.

   The only allowed value is "UTF-8"; it is to be matched case-
   insensitively (see [RFC2978], Section 2.3).  It indicates that the
   server expects character data to be converted to Unicode
   Normalization Form C ("NFC"; see Section 3 of [RFC5198]) and to be
   encoded into octets using the UTF-8 character encoding scheme
   ([RFC3629]).
```

It would be good though if all our clients add `charset="UTF-8"` parameter to be more explicit/more compatible with the RFC.

PR #3985, opened 2023-10-05
Implement $mem-node-state in-memory stream.

YoEight:
Added: Implement $mem-node-state in-memory stream.

Payload would be as follow:

```json
{ "state": "Leader" }
```

### DIFF
--- a/src/EventStore.Core.XUnit.Tests/Services/Storage/InMemoryStreamReaderTests.cs
+++ b/src/EventStore.Core.XUnit.Tests/Services/Storage/InMemoryStreamReaderTests.cs
@@ -1,0 +1,285 @@
+using System;
+using System.Collections.Generic;
+using System.Security.Claims;
+using System.Threading.Channels;
+using EventStore.Core.Data;
+using EventStore.Core.Messages;
+using EventStore.Core.Messaging;
+using EventStore.Core.Services;
+using EventStore.Core.Services.Storage;
+using Xunit;
+
+namespace EventStore.Core.XUnit.Tests.Services.Storage;
+
+public class InMemoryStreamReaderTests {
+	private readonly InMemoryStreamReader _sut;
+	private readonly NodeStateListenerService _listener;
+
+	public InMemoryStreamReaderTests() {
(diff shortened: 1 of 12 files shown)


src/EventStore.Core.XUnit.Tests/Services/Storage/InMemoryStreamReaderTests.cs
```diff
@@ -0,0 +1,149 @@
...
+			Assert.Equal(10, result.MaxCount);
+			Assert.Equal(-1, result.NextEventNumber);
+			Assert.Equal(-1, result.LastEventNumber);
+			Assert.Empty(result.Events);
+		}
+	}
+
+	public class ReadBackwardsTests : InMemoryStreamReaderTests {
+		[Fact]
+		public void read_backwards() {
```
hayley-jean, line 119: Would it be worth adding tests for reading forward/backward from a position greater/less than the last node status event position?

src/EventStore.Core/Services/NodeStateListenerService.cs
```diff
@@ -0,0 +1,139 @@
...
+
+namespace EventStore.Core.Services;
+
+// threading: we expect to handle one StateChangeMessage at a time, but Reads can happen concurrently
+// with those handlings and with other reads.
+public class NodeStateListenerService :
+	IInMemoryStreamReader,
+	IHandle<SystemMessage.StateChangeMessage> {
+	private readonly IPublisher _publisher;
+	public const string EventType = "NodeStateChanged";
```
hayley-jean, line 17: Should this be an admin event type?
```suggestion
	public const string EventType = "$NodeStateChanged";
```

hayley-jean (changes requested):

src/EventStore.Core.XUnit.Tests/Services/Storage/InMemoryStreamReaderTests.cs
YoEight: Done

timothycoleman (approved):

hayley-jean (approved):

PR #3986, opened 2023-10-05
Add support for pluggable subsystems (DB-427)

timothycoleman:
Added: Support for pluggable subsystems

- Use the ISubsystem/ISubsystemFactory/ISubsystemsPlugin interfaces from EventStore.Plugins
- The ClusterVNodeOptions contain subsystem factories now instead of subsystems, and ISubsystem no longer has a Register method. The upshot is that instead of constructing the subsystems upfront and register()ing them later with the standard components, we construct the factories up front and use them to produce the subsystems later using the standard components.
- For minimal impact, the projections subsystem is its own factory.

Intended use:
These plugable subsystems that the server loads directly are are _tightly coupled_ to the implementation of the server, with all the advantages and disadvantages that that brings. They literally subscribe to the bus and publish to the main queue. It is anticipated that there will be very few of these, and users should never extend the system in this way. The immediate practical purpose of this is to host the forthcoming ConnectedSubsystems plugin (which will come out after the 23.10 LTS). The ConnectedSubsystems plugin will provide a more stable and decoupled API (currently: IClient) that other subsystems can implement.


timothycoleman:
(draft so that we merge the other pr that this is merging into first)

YoEight (dismissed):

hayley-jean (approved):

YoEight (approved):

PR #3988, opened 2023-10-09
[POC] Add RPCs to support authorization from external subsystems (DB-424)

shaan1337:
Added: RPCs to support authorization from external subsystems

Plugin changes: https://github.com/EventStore/EventStore.Plugins/pull/22
.NET client changes: https://github.com/EventStore/EventStore-Client-Dotnet/pull/268


timothycoleman:
we went a different route with this

PR #3998, opened 2023-10-16
Refactor the gRPC enumerators

dhingrak:
Added: Refactor gRPC enumerators 

Fixes https://linear.app/eventstore/issue/DB-392/refactor-the-grpc-enumerators

Decouple all the enumerators from gRPC so they can work directly with connectors.


src/EventStore.Core/Services/Transport/Grpc/Enumerators.AllSubscription.cs
```diff
@@ -44,26 +42,16 @@ public class AllSubscription : IAsyncEnumerator<ReadResp> {
...
-				}
-
-				if (readIndex == null) {
-					throw new ArgumentNullException(nameof(readIndex));
-				}
-
 				_expiryStrategy = expiryStrategy;
 				_subscriptionId = Guid.NewGuid();
-				_bus = bus;
+				_bus = bus ?? throw new ArgumentNullException(nameof(bus));
```
timothycoleman, line 48: i could be persuaded, but i think i'd be tempted to leave these tidyups for another PR, so that this one can be as minimal and obviously correct as possible

src/EventStore.Core/Services/Transport/Grpc/Enumerators.AllSubscription.cs
```diff
@@ -384,28 +390,15 @@ await foreach (var @event in liveEvents.Reader.ReadAllAsync(_cancellationToken)
...
 				? new ValueTask(Task.CompletedTask)
-				: _channel.Writer.WriteAsync(new ReadResp {
-					Confirmation = new ReadResp.Types.SubscriptionConfirmation {
-						SubscriptionId = SubscriptionId
-					}
-				}, _cancellationToken);
-
-			private void Fail(Exception exception) {
-				Interlocked.Exchange(ref _subscriptionStarted, 1);
-				_channel.Writer.TryComplete(exception);
```
timothycoleman, line 407: this closes the channel with an exception but most of that places that used to call this now close the channel cleanly or dont close it at all, is that an intentional change?

src/EventStore.Core/Services/Transport/Grpc/ReadResponse.cs
```diff
@@ -0,0 +1,65 @@
+﻿using EventStore.Core.Data;
+
+namespace EventStore.Core.Services.Transport.Grpc; 
+
+public abstract class ReadResponse {
+	
+}
+
+public class ResolvedEventWrapper: ReadResponse {
```
timothycoleman, line 9: it might be nice if these were nested in the abstract base class, its a pattern we often do with the message classes

src/EventStore.Core/Services/Transport/Grpc/Enumerators.AllSubscription.cs
dhingrak: sure I will update this.

src/EventStore.Core/Services/Transport/Grpc/ReadResponse.cs
dhingrak: Okay

timothycoleman (approved):
thanks! :+1:

shaan1337 (approved):

PR #4033, opened 2023-11-01
[DOC-7] Fix Scavenging Link

rsowald:
We had the wrong anchor tag in the scavenging doc link.

### DIFF
--- a/docs/streams.md
+++ b/docs/streams.md
@@ -2,10 +2,11 @@
 
 EventStoreDB is purpose-built for event storage. Unlike traditional state-based databases, which retain only the most recent entity state, EventStoreDB allows you to store each state alteration as an independent event.
 
-These **events** are logically organized into **streams**, typically only one stream per entity. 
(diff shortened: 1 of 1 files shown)


rsowald:
I run Prettier on files by default, so there are several spacing changes. The only content that changed was the link on line 75 (line 68 after extra spaces were removed).

cdevarenne (approved):

timothycoleman (approved):
Just something to consider, it looks like the only important thing to backport here is the link itself, but the formatting changes will be cherry picked with it and make the chances of a merge conflict higher. For changes that need to be backported it can be worth making them as small as possible (and then fixing the formatting in another non-backported pr)

rsowald:
@timothycoleman Good point. Going forward, we're going to be doing formatting/style/copyedit changes on the most recent version, so we shouldn't need to cherry pick in general. I'll turn off my auto-run of Prettier for ones I know we'll be backporting.

PR #4045, opened 2023-11-08
Remove TCP client API. (DB-623)

YoEight:
Changed: Remove TCP client API.


ylorph:
FYI ; I added a linear issue to look at the EventStore.TestClient , as it is ising the tcp API for ot's work .

PR #4057, opened 2023-11-16
[DB-550] Don't use ReadIndex when subscribing

hayley-jean:
Changed: Don't require ReadIndex in the enumerators when subscribing from $all

Instead, read from the provided position and skip the first event.


shaan1337 (approved):
changes look good :+1:

timothycoleman (approved):

PR #4063, opened 2023-11-17
[DOC-75] Anchor link fixes and Maintenance copyedit v23.6

rsowald:


cdevarenne (approved):

docs/operations.md
```diff
@@ -231,25 +237,25 @@ Below you can find some options that change the way how scavenging works on the
...
-| Environment variable | `EVENTSTORE_DISABLE_SCAVENGE_MERGING` | 
+| Environment variable | `EVENTSTORE_DISABLE_SCAVENGE_MERGING` |
 
 **Default**: `false`, small scavenged chunks are merged together.
 
 ### Scavenge history
 
-Each scavenge operation gets an id and creates a stream. You might want to look at these streams to see the
-scavenge history, see how much time each operation took and how much disk space was reclaimed. However, you
+Each scavenge operation gets an ID and creates a stream. You might want to look at these streams to see the
```
cdevarenne, line 249:
```suggestion
Each scavenge operation is assigned a unique ID and creates a corresponding stream. These streams provide valuable insights into the scavenge history, including the duration of each operation and the amount of disk space reclaimed. However, retaining this history indefinitely may not be necessary. To manage this, you can limit the duration for which scavenge history is stored in the database using the option below:```

File: docs/operations.md [hunk 2]
```diff
@@ -231,25 +237,25 @@ Below you can find some options that change the way how scavenging works on the
...
+scavenge history, how much time each operation took, and how much disk space was reclaimed. However, you
```
cdevarenne, line 250:
```suggestion
```

docs/operations.md
```diff
@@ -231,25 +237,25 @@ Below you can find some options that change the way how scavenging works on the
...
 might not want to keep this history forever. Use the following option to limit how long the scavenge history
```
cdevarenne, line 251:
```suggestion
```

cdevarenne (dismissed):

cdevarenne (dismissed):

cdevarenne:
> @rsowald @cdevarenne
> 
> For section below "Scavenge history"
> 
> Perhaps we can revise this to improve the readability and clarify the purpose of the scavenge history a bit more. What about something like:
> 
> Each scavenge operation is assigned a unique ID and creates a corresponding stream. These streams provide valuable insights into the scavenge history, including the duration of each operation and the amount of disk space reclaimed. However, retaining this history indefinitely may not be necessary. To manage this, you can limit the duration for which scavenge history is stored in the database using the option below:

OK I have committed this now.

rsowald:
> OK I have committed this now.

3 issues: 

- [ ] The first is that the commit on this branch introduced a typo. 
- [ ] The second is that this change also needs to be added to the other 2 PRs for the other versions. 
- [ ] And the 3rd is that this change will also need to be added to master in another PR.

cdevarenne (approved):

rmalik-es (approved):

PR #4093, opened 2023-12-04
[DB-393] gRPC stream subscriptions with smooth transitions between live and catchup

shaan1337:
Added: gRPC stream subscriptions with smooth transitions between live and catchup. Subscriptions no longer drop with "consumer too slow" reason.

Fixes: https://github.com/EventStore/EventStore/issues/4089

This PR restores the ability of gRPC stream subscriptions to smoothly switch between live and catch-up. It also stops using the ContinuationEnvelope which internally uses a semaphore which could block critical threads like the subscription service thread.

Thanks to @timothycoleman for contributing many of the main ideas in this PR!

The main changes in this PR are:
- As before we use a bounded channel to store live events but the `FullMode` has been changed to `DropOldest`. When the live buffer becomes full and we want to add a new event, the oldest event is dropped to create space for the new event. We thus keep a "window" of the last 32 live events. Thus, the subscription service now never blocks when writing events to the live channel.
- To detect gaps in the live channel due to dropped events, we use a sequence number when adding items to the live channel. When we're live and a gap is detected, we switch back to catch-up.
- Many of the methods like `Catchup`, `GoLive`  now return Tasks, which allows us to put the core logic of the subscription in a loop (`MainLoop`) and there's no longer the need to do any "manual" synchronization between the catch-up and live threads.
- The live subscription also runs permanently now as this makes the algorithm simpler and more efficient:
  - we can switch to live directly when the subscription starts if the checkpoint matches the last live event number
  - if the subscription is switching back and forth between live/catch-up, we don't need to unsubscribe/re-subscribe internally each time

## Fast producer / Slow consumer
Note that the main channel that the subscription reads from (`_channel`) is still a bounded channel with `FullMode` = `Wait`

### Live
If we're live and events are written to the live channel/database at a rate that's faster than they're being read by the client, then the live channel will eventually become full and events will be dropped. Dropped events will be detected (with sequence numbers) and we will switch to catch-up mode.

In parallel, it's also possible that the output channel (`_channel`) will become full. Since `_channel` uses `FullMode` = `Wait`, then the live task will wait and write events as space is made on _channel.

### Catch-up
If we're in catch-up mode and the rate at which events are read from the disk is higher than the rate they're being read by the client, then the output channel (`_channel`) will eventually become full when writing events in the `ReadPage` callback. Just like when we're live, the task will wait till space is made on _channel.

## Performance
Some basic tests done on my machine:

### Catch-up
| Branch  | Catch-up time for 1M events (events / sec) |
|---------|--------------------------------------------|
| This PR | ~76000                                      |
| master  | ~76000                                      |

### Live
| Branch  | Live event rate without switching to catch-up/consumer too slow (events / sec)               |
|---------|----------------------------------------------------------------------------------------------|
| This PR | > 1000
| master  | 465 (gets consumer too slow after this point)

### DIFF
--- a/src/EventStore.Core/Services/Transport/Enumerators/Enumerator.StreamSubscription.cs
+++ b/src/EventStore.Core/Services/Transport/Enumerators/Enumerator.StreamSubscription.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Security.Claims;
 using System.Threading;
 using System.Threading.Channels;
@@ -30,11 +31,10 @@ namespace EventStore.Core.Services.Transport.Enumerators {
 			private readonly bool _resolveLinks;
 			private readonly ClaimsPrincipal _user;
 			private readonly bool _requiresLeader;
-			private readonly CancellationToken _cancellationToken;
+			private readonly CancellationTokenSource _cts;
 			private readonly Channel<ReadResponse> _channel;
-			private readonly SemaphoreSlim _semaphore;
+			private readonly Channel<(ulong SequenceNumber, ResolvedEvent ResolvedEvent)> _liveEvents;
 
-			private int _subscriptionStarted;
 			private ReadResponse _current;
 			private bool _disposed;
 			private StreamRevision? _currentRevision;
@@ -46,18 +46,13 @@ namespace EventStore.Core.Services.Transport.Enumerators {
 				IPublisher bus,
 				IExpiryStrategy expiryStrategy,
 				string streamName,
-				StreamRevision? startRevision,
+				StreamRevision? checkpoint,
 				bool resolveLinks,
 				ClaimsPrincipal user,
 				bool requiresLeader,
 				CancellationToken cancellationToken) {
-				if (bus == null) {
-					throw new ArgumentNullException(nameof(bus));
-				}
-
-				if (streamName == null) {
-					throw new ArgumentNullException(nameof(streamName));
-				}
+				ArgumentNullException.ThrowIfNull(bus);
+				ArgumentNullException.ThrowIfNull(streamName);
 
 				_expiryStrategy = expiryStrategy;
 				_subscriptionId = Guid.NewGuid();
@@ -66,50 +61,49 @@ namespace EventStore.Core.Services.Transport.Enumerators {
 				_resolveLinks = resolveLinks;
 				_user = user;
 				_requiresLeader = requiresLeader;
-				_cancellationToken = cancellationToken;
-				_subscriptionStarted = 0;
+				_cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
 				_channel = Channel.CreateBounded<ReadResponse>(BoundedChannelOptions);
-				_semaphore = new SemaphoreSlim(1, 1);
+				_liveEvents = Channel.CreateBounded<(ulong, ResolvedEvent)>(LiveChannelOptions);
 				_currentRevision = null;
 
 				SubscriptionId = _subscriptionId.ToString();
 
-				Subscribe(startRevision);
+				Subscribe(checkpoint, _cts.Token);
 			}
 
 			public override ValueTask DisposeAsync() {
 				if (_disposed) {
-					return new ValueTask(Task.CompletedTask);
+					return ValueTask.CompletedTask;
 				}
 
-				Log.Information("Live subscription {subscriptionId} to {streamName} disposed.", _subscriptionId,
+				Log.Debug("Subscription {subscriptionId} to {streamName} disposed.", _subscriptionId,
 					_streamName);
 				_disposed = true;
-				_channel.Writer.TryComplete();
+
 				Unsubscribe();
-				return new ValueTask(Task.CompletedTask);
+
+				_cts.Cancel();
+				_cts.Dispose();
+
+				return ValueTask.CompletedTask;
 			}
 
 			public override async ValueTask<bool> MoveNextAsync() {
 				ReadLoop:
 
-				if (!await _channel.Reader.WaitToReadAsync(_cancellationToken)) {
+				if (!await _channel.Reader.WaitToReadAsync(_cts.Token)) {
 					return false;
 				}
 
-				var readResponse = await _channel.Reader.ReadAsync(_cancellationToken);
+				var readResponse = await _channel.Reader.ReadAsync(_cts.Token);
 
 				if (readResponse is ReadResponse.EventReceived eventReceived) {
 					var @event = eventReceived.Event;
-
-					if (@event.OriginalEvent.EventType == SystemEventTypes.StreamDeleted) {
-						Fail(new ReadResponseException.StreamDeleted(_streamName));
-					}
-
 					var streamRevision = StreamRevision.FromInt64(@event.OriginalEventNumber);
 
 					if (_currentRevision.HasValue && streamRevision <= _currentRevision.Value) {
-						Log.Verbose(
+						// this should no longer happen
+						Log.Warning(
 							"Subscription {subscriptionId} to {streamName} skipping event {streamRevision} as it is less than {currentRevision}.",
 							_subscriptionId, _streamName, streamRevision, _currentRevision);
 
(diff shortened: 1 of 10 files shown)


src/EventStore.Core/Services/Transport/Enumerators/Enumerator.StreamSubscription.cs
```diff
@@ -171,22 +273,29 @@ public class StreamSubscription<TStreamId> : StreamSubscription {
...
-								return;
+								if (completed.Events.Length > 0) {
+									catchupCompletionTcs.TrySetResult(completed.Events[^1].OriginalEventNumber);
+									return;
+								}
+
+								// the read returned an empty result - we don't have the last event number.
+								// we could derive it from `completed.NextEventNumber` but it's safer to stay in catch-up
+								// and retry the read as it's an edge case that shouldn't happen often.
+								await Task.Delay(TimeSpan.FromMilliseconds(500), ct);
```
hayley-jean, line 284: I'm not sure if it's a problem necessarily, but if you have a stream that isn't deleted but has no events (for example, with `$maxAge` or `$tb`) then the reads will continue to be retried until a new event is written (if ever)

src/EventStore.Core/Messaging/ContinuationEnvelope.cs
```diff
@@ -36,4 +36,22 @@ public class ContinuationEnvelope : IEnvelope {
...
+		public AsyncTaskEnvelope(Func<Message, CancellationToken, Task> onMessage, CancellationToken cancellationToken) {
+			_onMessage = onMessage;
+			_cancellationToken = cancellationToken;
+		}
+
+		public void ReplyWith<T>(T message) where T : Message {
+			try {
+				_onMessage(message, _cancellationToken);
+			}
+			catch (ObjectDisposedException) {}
```
timothycoleman, line 53: probably this catch was for the semaphore, we might not need it here?

src/EventStore.Core/Messaging/ContinuationEnvelope.cs
```diff
@@ -36,4 +36,22 @@ public class ContinuationEnvelope : IEnvelope {
...
+		private readonly CancellationToken _cancellationToken;
+
+		public AsyncTaskEnvelope(Func<Message, CancellationToken, Task> onMessage, CancellationToken cancellationToken) {
+			_onMessage = onMessage;
+			_cancellationToken = cancellationToken;
+		}
+
+		public void ReplyWith<T>(T message) where T : Message {
+			try {
+				_onMessage(message, _cancellationToken);
```
timothycoleman, line 51: depending on the implementation of _onMessage, the OperationCancelled exception might be caught here or it might end up on the Task (according to whether it is thrown synchronously or not). might be worth mentioned that in a comment so that it doesn't look like its something that might have been missed

src/EventStore.Core/Messaging/ContinuationEnvelope.cs
```diff
@@ -36,4 +36,22 @@ public class ContinuationEnvelope : IEnvelope {
 			catch (OperationCanceledException) {}
 		}
 	}
+
+	public class AsyncTaskEnvelope : IEnvelope {
```
timothycoleman, line 40: probably in its own file

and probably worth a comment at the top of this that it needs to be used carefully because the callback is called initially on the thread that is replying to the message so must be quick

src/EventStore.Core/Services/Transport/Enumerators/Enumerator.StreamSubscription.cs
```diff
@@ -371,29 +379,6 @@ public class StreamSubscription<TStreamId> : StreamSubscription {
...
-					Fail(new ReadResponseException.Timeout(msg));
-				}
-
-				void Fail(Exception exception) {
-					this.Fail(exception);
-					caughtUpSource.TrySetException(exception);
-				}
 			}
 
 			private ValueTask ConfirmSubscription() => Interlocked.CompareExchange(ref _subscriptionStarted, 1, 0) != 0
```
timothycoleman, line 384: i think we might be able to simplify this now too. `SubscribeToLive` only completes when the subscription is confirmed by the subscription service, and the `MainLoop` is awaiting `SubscribeToLive`. so I think the MainLoop can just call `ConfirmSubscription` after `SubscribeToLive` completes.

I think that would remove the need for `OnSubscriptionMessage` and `OnMessage` to call `ConfirmSubscription`, and then since the call to ConfirmSubscription is only coming from 1 place and at a well known moment that ought to remove the need for the `_subscriptionStarted` variable and the compare-exchange

src/EventStore.Core/Services/Transport/Enumerators/Enumerator.StreamSubscription.cs
```diff
@@ -129,21 +132,120 @@ public class StreamSubscription<TStreamId> : StreamSubscription {
...
+				return lastLiveEventInfo.EventNumber == lastCatchupEventNumber;
+			}
+
+			private async ValueTask<(long lastSubscriptionEventNumber, LiveEventInfo lastLiveEventInfo)> GoLive(LiveEventInfo lastLiveEventInfo) {
+				var lastSubscriptionEventNumber = lastLiveEventInfo.EventNumber;
+				await NotifyCaughtUp(lastSubscriptionEventNumber);
+
+				await foreach (var liveEvent in _liveEvents.Reader.ReadAllAsync(_cancellationToken)) {
+					bool sequenceCorrect = liveEvent.sequenceNumber == lastLiveEventInfo.SequenceNumber + 1;
+					lastLiveEventInfo = new LiveEventInfo(liveEvent.resolvedEvent.OriginalEventNumber, liveEvent.sequenceNumber);
```
timothycoleman, line 223: prefer to avoid the allocation on each live event here. it might be sensible to make LiveEventInfo a `record struct` and remove  the explicit get/set (and just construct a new one in TrySyncLiveWithCatchup)

src/EventStore.Core/Services/Transport/Enumerators/Enumerator.StreamSubscription.cs
shaan1337: hey, thanks, i hadn't thought about these cases. There is an additional case where this can happen:
- if you subscribe to a checkpoint that's ahead of the last event number

I'll try to adjust things so that we can switch directly to live in these cases.

src/EventStore.Core/Messaging/ContinuationEnvelope.cs
shaan1337: fixed :+1:

src/EventStore.Core/Messaging/ContinuationEnvelope.cs
shaan1337: added, thanks

src/EventStore.Core/Messaging/ContinuationEnvelope.cs
shaan1337: fixed, comment added :+1:

src/EventStore.Core/Services/Transport/Enumerators/Enumerator.StreamSubscription.cs
shaan1337: great idea, thanks, i've made the changes accordingly

src/EventStore.Core/Services/Transport/Enumerators/Enumerator.StreamSubscription.cs
shaan1337: fixed :+1: turns out the compiler doesn't complain about passing a `ref record struct` as parameter to `TrySyncLiveWithCatchup`, so i've not made any changes there. i think that's because the lastLiveEventInfo is on the stack and thus the stack address of the struct can be passed in.

src/EventStore.Core/Services/Transport/Enumerators/Enumerator.StreamSubscription.cs
shaan1337: fixed

shaan1337:
@hayley-jean @timothycoleman thanks for your feedback, the PR is up for review again. note that the base branch has been changed to `db-393-additional-tests` so, it's better to wait for https://github.com/EventStore/EventStore/pull/4108 and https://github.com/EventStore/EventStore/pull/4104 to be merged first.

timothycoleman (approved):
Nice!

timothycoleman:
For posterity: there is a 'FellBehind' message into the proto but it is not clear that it would be sufficiently useful to include it, so we have left out sending it to the client for now.

YoEight (approved):

shaan1337:
> Nice!

thanks for you invaluable help on this PR!

PR #4095, opened 2023-12-12
Fix stream (hard) deletion when gRPC subscription is live [regression] [DB-573]

shaan1337:
Fixed: Stream (hard) deletion when gRPC subscription is live [regression]

A stream deleted exception was not being thrown on the gRPC client side when a gRPC stream subscription was live and the stream was hard deleted. That's because the exception was set only on the channel and nothing reads from the channel again after returning false in MoveNextAsync().

This is a regression introduced by: 0923c5632e3341adfe603d8d6a6b2e4bc04d3b69. The `return false` wasn't previously present. The regression isn't present in any release (we would have caught it during QA as the .NET client tests fail)


timothycoleman (approved):
(this is picked up by a failing dotnet client test, longer term we could consider bringing those into the CI for this repository)

hayley-jean (approved):

PR #4101, opened 2023-12-18
Add section headings to projections HTTP API docs

hayley-jean:
The HTTP API docs were missing headings for each operation, which makes it difficult to find what you're looking for

### DIFF
--- a/docs/http-api/projections.md
+++ b/docs/http-api/projections.md
@@ -157,15 +157,33 @@ The server then returns the state for the partition:
 |:---------------------------------------------------|:-----------------------------|:----------|
 | `/projections/any`                                 | Returns all known projections.             | GET       |
 | `/projections/all-non-transient`                   | Returns all known non ad-hoc projections.  | GET       |
-| `/projections/transient`                           | Returns all known ad-hoc projections.      | GET       |
-| `/projections/onetime`                             | Returns all known one-time projections.    | GET       |
-| `/projections/continuous`                          | Returns all known continuous projections.  | GET       |
 
-### Create an Ad-Hoc Projection
+
+### Manage Continuous Projections
+
+Continuous projections will continue to run until they are disabled or until they encounter an unrecoverable error. These are the most typical projections.
 
 | URI                                                | Description                  | HTTP verb |
 |:---------------------------------------------------|:-----------------------------|:----------|
-| `/projections/transient?name={name}&type={type}&enabled={enabled}` | Create an ad-hoc projection (or query). This type of projection runs until completion and automatically deleted afterwards. | POST      |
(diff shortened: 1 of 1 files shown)


docs/http-api/projections.md
```diff
@@ -209,6 +221,8 @@ The server then returns the state for the partition:
 * `emit`: Allow the projection to write to streams (true/false)
 * `trackemittedstreams`: Write the name of the streams the projection is managing to a separate stream. $projections-{projection-name}-emittedstreams (true/false)
 
+### Delete a Projection
+
 | URI                                                                                                                                                    | Description                                                                                     | HTTP verb |
 |:-------------------------------------------------------------------------------------------------------------------------------------------------------|:------------------------------------------------------------------------------------------------|:----------|
 | `/projection/{name}`                                                                                                                                   | Returns information for a projection.                                                           | GET       |
```
cdevarenne, line 228: Shouldn't the verb be delete?


```suggestion
| `/projection/{name}`                                                                                                                                   | Returns information for a projection.                                                           | DELETE       |
```

docs/http-api/projections.md
```diff
@@ -221,6 +235,8 @@ The server then returns the state for the partition:
 * `deleteCheckpointStream`: Delete the checkpoint stream (true/false)
```
cdevarenne, line 235: Should the arguments be shown in the example?

cdevarenne (changes requested):

docs/http-api/projections.md
rsowald: from the other examples, it looks like just including it in the URI is setting it true. Omitting is false. This is unclear in the documentation though. If that's the case, I think we should just remove the (true/false) from the descriptions. So, it's also unclear if any/all of these parameters are required.

docs/http-api/projections.md
hayley-jean: Ah there were actually two different actions in the table. I think the original formatting was meant to be closer to what's in the most recent commit. It's up to you whether that's more readable than different sections per action @cdevarenne @rsowald 🙂

docs/http-api/projections.md
hayley-jean: > it looks like just including it in the URI is setting it true. Omitting is false.

@rsowald This isn't always the case. Omitting it is leaving it as the default (which we should list in the parameters), but if you want to set it you should use either `deleteCheckpointStream=true` or `deleteCheckpointStream=false`

> Should the arguments be shown in the example?

@cdevarenne Do you mean something like this?

```
/projections/continuous?name={name}&type={ js | native }&enabled={ true | false }&emit={ true | false }&trackemittedstreams={ true | false }
```

alexeyzimarev (approved):

PR #4102, opened 2023-12-21
Avoid hanging calls to stats/replication endpoint for a single node

lakshdeepsingheventstore:
Fixed: Calls to /stats/replication on a single node cluster would hang forever.
Fixes : https://linear.app/eventstore/issue/DB-556/get-statsreplication-hangs-for-single-node-cluster


src/EventStore.Core/ClusterVNode.cs
```diff
@@ -1472,6 +1472,9 @@ public class ClusterVNode<TStreamId> :
 				_mainBus.Subscribe<ReplicationMessage.AckLogPosition>(replicaService);
 				_mainBus.Subscribe<ClientMessage.TcpForwardMessage>(replicaService);
 			}
+			else {
+				monitoringInnerBus.Subscribe<ReplicationMessage.GetReplicationStats>(leaderReplicationService);
```
hayley-jean, line 1476: It might be worth adding a comment here about the LeaderReplicationService only running on a single node to provide stats, which is why it's not subscribed to the other message types (e.g. SystemStart and StateChangeMessage)

hayley-jean (approved):
Looks good, with this approach the `LeaderReplicationService` will always be running, but this shouldn't be a problem because:
- the replication TCP interface isn't running
- the `LeaderReplicationService` is not subscribed to the `SystemStart` messages, so will not start the main loop or queue stats collector
- the `LeaderReplicationService` runs on other node types anyway (e.g. followers and ReadOnly Replicas)

I was going to comment that the [expected number of service shutdowns](https://github.com/EventStore/EventStore/blob/master/src/EventStore.Core/Services/VNode/ClusterVNodeController.cs#L97) in `ClusterVNodeController` should be updated, but since the main loop doesn't get started, there's nothing to dispose

In future, it might be a good idea to fully support running replication on a single node so that we can do replication from a single node to a Read Only Replica, but that's not part of this PR

src/EventStore.Core/ClusterVNode.cs
lakshdeepsingheventstore: Done.

hayley-jean (approved):

timothycoleman (changes requested):
squash to a single commit please, thanks!

timothycoleman (approved):

PR #4103, opened 2023-12-21
Do not autosize when EventStoreDB runs in containerized environment

dhingrak:
Changed: Do not autosize thread count and streaminfocache size in containerized environments

Fixes: https://linear.app/eventstore/issue/DB-408/do-not-autosize-when-containerised

### DIFF
--- a/src/EventStore.Common/Utils/ContainerizedEnvironment.cs
+++ b/src/EventStore.Common/Utils/ContainerizedEnvironment.cs
@@ -1,0 +1,20 @@
+﻿using System;
+
+namespace EventStore.Common.Utils;
+
+public static class ContainerizedEnvironment {
+
+	public const int ReaderThreadCount = 4;
+	public const int WorkerThreadCount = 5;
+	public const int StreamInfoCacheCapacity = 100000;
+
+	private static readonly string _dotnetContainerEnv;
+
+	static ContainerizedEnvironment() {
+		_dotnetContainerEnv = Environment.GetEnvironmentVariable("DOTNET_RUNNING_IN_CONTAINER");
+	}
+
+	public static bool IsRunningInContainer() {
+		return !string.IsNullOrEmpty(_dotnetContainerEnv);
+	}
+}
--- a/src/EventStore.Core.Tests/Settings/ReaderThreadCountCalculatorTests.cs
+++ b/src/EventStore.Core.Tests/Settings/ReaderThreadCountCalculatorTests.cs
@@ -5,24 +5,30 @@ namespace EventStore.Core.Tests.Settings {
 	[TestFixture]
 	public class ReaderThreadCountCalculatorTests {
 		[TestCase]
-		public void configured_takes_precedence() => Test(configuredCount: 1, processorCount: 4, expected: 1);
+		public void configured_takes_precedence() => Test(configuredCount: 1, processorCount: 4, isRunningInContainer: false, expected: 1);
 
 		[TestCase]
-		public void enforces_minimum() => Test(configuredCount: 0, processorCount: 1, expected: 4);
+		public void enforces_minimum() => Test(configuredCount: 0, processorCount: 1, isRunningInContainer: false, expected: 4);
 
 		[TestCase]
-		public void enforces_maximum() => Test(configuredCount: 0, processorCount: 20, expected: 16);
+		public void enforces_maximum() => Test(configuredCount: 0, processorCount: 20, isRunningInContainer: false, expected: 16);
 
 		[TestCase]
-		public void at_3_cores() => Test(configuredCount: 0, processorCount: 3, expected: 6);
+		public void at_3_cores() => Test(configuredCount: 0, processorCount: 3, isRunningInContainer: false, expected: 6);
 
 		[TestCase]
-		public void at_4_cores() => Test(configuredCount: 0, processorCount: 4, expected: 8);
+		public void at_4_cores() => Test(configuredCount: 0, processorCount: 4, isRunningInContainer: false, expected: 8);
 
 		[TestCase]
-		public void at_6_cores() => Test(configuredCount: 0, processorCount: 6, expected: 12);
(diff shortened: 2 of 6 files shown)


src/EventStore.Common/Utils/ContainerizedEnvironment.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+
+namespace EventStore.Common.Utils;
+
+public static class ContainerizedEnvironment {
+
+	public const int ContainerReaderThreadCount = 4;
```
timothycoleman, line 7: no need to reapeat the word 'Container' in these names since it's included in the context (ContainerizedEnvironment class)

src/EventStore.Core.Tests/Settings/ReaderThreadCountCalculatorTests.cs
```diff
@@ -5,24 +5,30 @@ namespace EventStore.Core.Tests.Settings {
 	[TestFixture]
 	public class ReaderThreadCountCalculatorTests {
 		[TestCase]
-		public void configured_takes_precedence() => Test(configuredCount: 1, processorCount: 4, expected: 1);
+		public void configured_takes_precedence() => Test(configuredCount: 1, processorCount: 4, false, expected: 1);
```
timothycoleman, line 8: i think it would be nice here to include the parameter name to be consistent with the others (and otherwise the meaning of 'false' isn't immediately obvious)

src/EventStore.Core.Tests/Settings/WorkerThreadCountCalculatorTests.cs
```diff
@@ -5,22 +5,28 @@ namespace EventStore.Core.Tests.Settings {
...
-		public void enforces_minimum() => Test(configuredCount: 0, readerCount: 1, expected: 5);
+		public void enforces_minimum() => Test(configuredCount: 0, readerCount: 1, false, expected: 5);
 
 		[TestCase]
-		public void enforces_maximum() => Test(configuredCount: 0, readerCount: 1000, expected: 10);
+		public void enforces_maximum() => Test(configuredCount: 0, readerCount: 1000, false, expected: 10);
 
 		[TestCase]
-		public void at_4_reader_threads() => Test(configuredCount: 0, 4, expected: 5);
+		public void at_4_reader_threads() => Test(configuredCount: 0, readerCount:4, false, expected: 5);
```
timothycoleman, line 17:
```suggestion
		public void at_4_reader_threads() => Test(configuredCount: 0, readerCount: 4, false, expected: 5);
```

src/EventStore.Core/Settings/ThreadCountCalculator.cs
```diff
@@ -1,23 +1,28 @@
...
 		private const int ReaderThreadCountFloor = 4;
 
-		public static int CalculateReaderThreadCount(int configuredCount, int processorCount) {
+		public static int CalculateReaderThreadCount(int configuredCount, int processorCount,
+			bool isRunningInContainer) {
 			if (configuredCount > 0)
 				return configuredCount;
 
-			var readerCount = Math.Clamp(processorCount * 2, ReaderThreadCountFloor, 16);
+			if (isRunningInContainer) return ContainerizedEnvironment.ContainerReaderThreadCount;
```
timothycoleman, line 14: maybe a bit picky, but i think it's best to have the return on its own line to be consistent with the rest of the method

src/EventStore.Core/ClusterVNode.cs
```diff
@@ -610,13 +607,15 @@ public class ClusterVNode<TStreamId> :
 					out streamLastEventNumberCache,
 					out streamMetadataCache,
 					out streamInfoCacheResizer);
-			else
+			else {
```
timothycoleman, line 610: better not to add in the curly braces here since there aren't any in the 'if'. it's a small thing but things like this add up to make the code easier/harder to read

src/EventStore.Common/Utils/ContainerizedEnvironment.cs
```diff
@@ -0,0 +1,22 @@
...
+
+	private static bool _isContainer;
+
+	public static bool IsRunningInContainer() {
+		string dotnetContainerEnv = Environment.GetEnvironmentVariable("DOTNET_RUNNING_IN_CONTAINER");
+		_isContainer = !string.IsNullOrEmpty(dotnetContainerEnv);
+		return _isContainer;
+	}
+
+	public static bool IsContainer() {
```
timothycoleman, line 19: there is a danger here that if `IsContainer` is called before `IsRunningInContainer` then `_isContainer` won't be populated

src/EventStore.Core/Settings/ThreadCountCalculator.cs
```diff
@@ -1,24 +1,55 @@
...
 
-		public static int CalculateReaderThreadCount(int configuredCount, int processorCount) {
+		public static int CalculateReaderThreadCount(int configuredCount, int processorCount,
+			bool isRunningInContainer) {
 			if (configuredCount > 0)
 				return configuredCount;
 
-			var readerCount = Math.Clamp(processorCount * 2, ReaderThreadCountFloor, 16);
+			if (isRunningInContainer) {
+				Log.Information(
```
timothycoleman, line 16: we used to log what value we are calculating in the case that the `configuredCount > 0` but we aren't any more, we should add that back please

src/EventStore.Core/ClusterVNode.cs
dhingrak: sure, will make the changes

src/EventStore.Core/Settings/ThreadCountCalculator.cs
dhingrak: oh, I forgot to add that to the configured count > 0. I will add this.

shaan1337 (approved):

timothycoleman (approved):

PR #4104, opened 2023-12-26
[DB-590] Live subscriptions: Re-authorize when a stream's metadata changes

shaan1337:
Changed: Re-authorize stream access in live subscriptions when stream metadata changes

Note that:
i)  some live events can still go through before the AccessDenied message is sent to the subscription.
ii) user group changes are not taken into consideration. this is OK as with claims-based authorization, existing access is maintained until the claims expire.

Note: .NET client tests pass against this branch.

### DIFF
--- a/src/EventStore.Core/Services/SubscriptionsService.cs
+++ b/src/EventStore.Core/Services/SubscriptionsService.cs
@@ -1,5 +1,9 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
+using System.Security.Claims;
+using System.Threading;
+using System.Threading.Tasks;
 using EventStore.Common.Utils;
 using EventStore.Core.Bus;
 using EventStore.Core.Data;
@@ -7,8 +11,7 @@ using EventStore.Core.Messages;
 using EventStore.Core.Messaging;
 using EventStore.Core.Services.Storage.ReaderIndex;
 using EventStore.Core.Services.TimerService;
-using System.Linq;
-using EventStore.Core.Util;
+using EventStore.Plugins.Authorization;
 using ILogger = Serilog.ILogger;
 
 namespace EventStore.Core.Services {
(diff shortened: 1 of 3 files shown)


src/EventStore.Core/Services/SubscriptionsService.cs
```diff
@@ -335,6 +350,69 @@ public class SubscriptionsService<TStreamId> :
 			return resolvedEvent;
 		}
 
+		private void ProcessStreamMetadataChanges(string eventStreamId) {
+			if (eventStreamId == AllStreamsSubscriptionId)
```
hayley-jean, line 354: It might be worth moving these conditions to ProcessEventCommitted so that this method isn't entered for every event

shaan1337: i've moved `ProcessStreamMetadataChanges()` to `Handle(StorageMessage.EventCommitted message)` / `Handle(StorageMessage.InMemoryEventCommitted message)`. this gets rid of the unnecessary `AllStreamsSubscriptionId` empty string check

for the other checks (`SystemStreams.IsMetastream()` / `_subscriptionTopics.TryGetValue()`), i've kept them in the ProcessStreamMetadataChanges(). even though it's on a hot path, i don't think that it should cause any performance penalty as entering/leaving the method is negligible and the method will also be inlined automatically by the JIT compiler if it's called frequently.

shaan1337: it also raises the question as to whether or not we should process stream metadata changes for ephemeral streams. i've tested adding stream metadata to the $mem-node-state ephemeral stream and it's working fine.

timothycoleman:
i think we can use 'Changed' rather than 'Fixed' in the description (probaly dont worry about the branch name though)

src/EventStore.Core/Services/SubscriptionsService.cs
timothycoleman: the memstream thing is fine by me

(moving the conditions was my idea. i was thinking the target method might be on the large side for inlining and and the checks will almost always return false it seemed worth it,   but now that we're calling it from two places, and you're right its probably not going to make much practical diference, lets leave it as is)

timothycoleman:
i pushed some small changes, after this is rebased im happy to approve

YoEight (approved):

YoEight:
How difficult would it be to have an unit test for this?

shaan1337:
@YoEight i've added some tests for it later in https://github.com/EventStore/EventStore/pull/4108 (`subscribe * then revoke access`). I just squashed & rebased this PR to master, then rebased #4108 and launched the new tests - they're passing.

timothycoleman (approved):

PR #4109, opened 2024-01-03
Extend timeout

timothycoleman:
These seem to be taking longer lately, possibly since 'configureawait(false)' was removed, but I'm not sure why that would be related. As far as I know XUnit doesn't run the tests with a synchronization context set


shaan1337 (approved):
it appears that the issue is not with the stream existence filter itself but with the test fixture / xunit as the issue happens when running the whole xunit test project.

it's also not windows-specific as the delay is also visible on linux (although it's still less than 5s)

YoEight (approved):

PR #4110, opened 2024-01-04
Add the 5 seconds shutdown timeout for HostOptions to match with the ESDB shutdown timeout

dhingrak:
Changed: Explicitly set the shutdown timeout to 5s, which was default in previous dotnet versions. Behaviour unchanged since previous release

Fixes https://linear.app/eventstore/issue/DB-588/investigate-esdb-not-shutting-down-quickly-with-slow-subscriptions

Currently the default shutdown timeout for HostOptions is 30 seconds. The goal of this PR is to set the timeout to 5 seconds.


timothycoleman (approved):

shaan1337 (approved):
tested, looks good :+1:

PR #4112, opened 2024-01-08
Reduce gRPC log level on the server

lakshdeepsingheventstore:
Fixed : Decrease grpc log level on the server. https://linear.app/eventstore/issue/DB-568/decrease-grpc-log-level-on-the-server

Before this fix/change, when the log level for gRPC was set to Warning (like the others), there were lots of erroneous logs generated in the server logs for client operations that were cancelled before they were completed (For example killing a test client performing gRPC reads and write flood on a cluster node).

### DIFF
--- a/src/EventStore.Core/Services/Transport/Grpc/Streams.BatchAppend.cs
+++ b/src/EventStore.Core/Services/Transport/Grpc/Streams.BatchAppend.cs
@@ -36,8 +36,17 @@ namespace EventStore.Core.Services.Transport.Grpc {
 				requestStream, responseStream,
 				context.GetHttpContext().User, _maxAppendSize, _writeTimeout,
 				GetRequiresLeader(context.RequestHeaders));
-			
-			await worker.Work(context.CancellationToken);
+			try {
+				await worker.Work(context.CancellationToken);
+			} catch (IOException) {
+				// ignored
+			} catch (TaskCanceledException) {
+				//ignored
+			} catch (InvalidOperationException) {
+				//ignored
+			} catch (OperationCanceledException) {
+				//ignored
+			}
 		}
 
 		private class BatchAppendWorker {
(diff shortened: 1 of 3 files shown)


src/EventStore.Core/Services/Transport/Grpc/Streams.BatchAppend.cs
```diff
@@ -36,8 +36,11 @@ partial class Streams<TStreamId> {
 				requestStream, responseStream,
 				context.GetHttpContext().User, _maxAppendSize, _writeTimeout,
 				GetRequiresLeader(context.RequestHeaders));
-			
-			await worker.Work(context.CancellationToken);
+			try {
+				await worker.Work(context.CancellationToken);
+			} catch(Exception) {
```
thefringeninja, line 41: We should only trap the specific exception that's flooding the logs

src/EventStore.Core/Services/Transport/Grpc/Streams.Read.cs
```diff
@@ -76,6 +76,8 @@ internal partial class Streams<TStreamId> {
 					}
 				} catch (ReadResponseException ex) {
 					ConvertReadResponseException(ex);
+				} catch (Exception) {
```
thefringeninja, line 79: We should only trap the specific exception that's flooding the logs

thefringeninja (changes requested):
I wonder if the reason this happens is because we're not forwarding the cancellation token on line 116

src/EventStore.Core/Services/Transport/Grpc/Streams.BatchAppend.cs
lakshdeepsingheventstore: Done

src/EventStore.Core/Services/Transport/Grpc/Streams.Read.cs
lakshdeepsingheventstore: Done

thefringeninja (approved):

thefringeninja (approved):

hayley-jean (approved):

PR #4113, opened 2024-01-09
Guard TCP API behind hidden flag. (DB-129)

YoEight:
Removed: Remove the external TCP API and related configuration options.


timothycoleman:
The first line of the PR description is for the release notes (i.e. its a one liner for the user). So probably it needn't mention the hidden flag. It can probably just say that we removed the external TCP api and configuration options

shared.env
```diff
@@ -4,5 +4,4 @@ EVENTSTORE_INT_TCP_PORT=1112
 EVENTSTORE_HTTP_PORT=2113
 EVENTSTORE_TRUSTED_ROOT_CERTIFICATES_PATH=/etc/eventstore/certs/ca
 EVENTSTORE_DISCOVER_VIA_DNS=false
-EVENTSTORE_ENABLE_EXTERNAL_TCP=true
 EVENTSTORE_ENABLE_ATOM_PUB_OVER_HTTP=true
```
timothycoleman, line 7: we may have some adjustments to make in the dockerfile in this repository and also in train station

src/EventStore.Core/ClusterVNode.cs
```diff
@@ -1867,6 +1871,25 @@ public class ClusterVNode<TStreamId> :
 			}
 		}
 
+		private static bool TryRunningPublicTcpApi(out int port, out int advertisedPort) {
```
timothycoleman, line 1874: in here we should probably log if we are enabling this unsupported feature

src/EventStore.Core/ClusterVNodeOptions.cs
```diff
@@ -687,11 +687,6 @@ public record InterfaceOptions {
 				}
 			}
 
-			[Description("Whether to enable external TCP communication"),
```
timothycoleman, line 690: Just making some notes of TCP related db settings and how this PR affects them:

| Old db setting | Replaced with env var |
|---|---|
| EnableExternalTcp | EVENTSTORE_UNSUPPORTED_TCP_API_ENABLED |
| ExtTcpPort / NodeTcpPort | EVENTSTORE_UNSUPPORTED_TCP_API_PORT |
| ExtHostAdvertiseAs / NodeHostAdvertiseAs | (kept) |
| ExtTcpPortAdvertiseAs / NodeTcpPortAdvertiseAs | (kept) |
| ExtIp / NodeIp | (kept) |
| ExtTcpHeartbeatTimeout / NodeHeartbeatTimeout | (kept) |
| ExtTcpHeartbeatInterval / NodeHeartbeatInterval | (kept) |

`EVENTSTORE_UNSUPPORTED_TCP_API_ADVERTISED_PORT` is a new environment variable but `ExtTcpPortAdvertiseAs` is still present

I think we need to write down something about why we're keeping some but not others. Personally I would git rid of everything that is used only for the ext tcp interface that we are removing

src/EventStore.Core/ClusterVNode.cs
```diff
@@ -70,6 +70,9 @@
 namespace EventStore.Core {
 	public abstract class ClusterVNode {
 		protected static readonly ILogger Log = Serilog.Log.ForContext<ClusterVNode>();
+		public static readonly string TcpApiEnvVar = "EVENTSTORE_UNSUPPORTED_TCP_API_ENABLED";
+		public static readonly string TcpApiPortEnvVar = "EVENTSTORE_UNSUPPORTED_TCP_API_PORT";
+		public static readonly string TcpApiAdvertisedPortEnvVar = "EVENTSTORE_UNSUPPORTED_TCP_API_ADVERTISED_PORT";
```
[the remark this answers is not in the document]
YoEight: What do you mean by we need to add some handling for them in program.cs?

src/EventStore.Core/ClusterVNode.cs
YoEight: Done

src/EventStore.Core/ClusterVNodeOptions.cs
YoEight: Now the only params remaining are `ExtIp / NodeIp` and `ExtHostAdvertiseAs / NodeHostAdvertiseAs` because those are also used by the http interface. Others have their environment variable name equivalent.

shared.env
YoEight: What kind of adjustments are we talking about?

timothycoleman: removing any references to database settings that are being removed

src/EventStore.Core/ClusterVNode.cs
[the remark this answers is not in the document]
timothycoleman: just somehow prevent the program from exiting immediately if the flags are set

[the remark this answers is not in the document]
YoEight: Done.

shared.env
YoEight: done

dhingrak (approved):
looks good after the requested changes.

timothycoleman (approved):

src/EventStore.Core/ClusterVNodeOptions.cs
Lougarou: DisableExternalTcpTls | n/a (this was unused anyway)

hey @timothycoleman @YoEight, this is actually used

PR #4123, opened 2024-01-22
[DB-614] Make gossip available as a mem stream: $mem-gossip

timothycoleman:
Added: $mem-gossip memory stream

Works very similarly to the $mem-node-state stream. Factored out a helper class.

Plugins will then be able to subscribe to the gossip like other streams without needing a special API


timothycoleman:
hm something might not be quite right because the UI keeps requerying immediately

timothycoleman:
Figured out the problem with help from @Lougarou, the TfLastPosition wasn't correct on the read, causing the longpoll from the UI to be responded to by the server immediately

YoEight (approved):

hayley-jean (approved):

PR #4125, opened 2024-01-22
Fix replication service heartbeat interval init.

YoEight:
Fixed: Initialize replication service heartbeat interval with `ReplicationHeartbeatInterval` instead of `NodeHeartbeatInterval `.


timothycoleman (approved):

timothycoleman:
(at first i thought this bug was introduced with the recentish configuration changes, but it goes back further)

dhingrak (approved):

PR #4126, opened 2024-01-23
Use DI for plugin dependencies

pvanbuijtene:
Changed: Updated plugin API

Also fixed, subsystems sometimes started before being fully configured.


src/EventStore.Core/Services/VNode/ClusterVNodeController.cs
```diff
@@ -579,11 +580,11 @@ public class ClusterVNodeController<TStreamId> : ClusterVNodeController, IHandle
 		}
 
 		private void Handle(AuthenticationMessage.AuthenticationProviderInitialized message) {
-			_startSubsystems();
+			_startSubsystems().Wait();
```
timothycoleman, line 583: maybe this can be a continuewith and the continuation do the remainder of the method (if no exception)

src/EventStore.Core/ClusterVNodeStartup.cs
```diff
@@ -150,7 +123,13 @@ public class ClusterVNodeStartup<TStreamId> : IStartup, IHandle<SystemMessage.Sy
 							return Results.BadRequest("Redaction is only available via Unix Sockets");
 						return await next(c).ConfigureAwait(false);
 					});
+				});
+
+			// allow all subsystems to register their legacy controllers before finalizing the legacy pipeline
+			foreach (var subsystem in _subsystems)
+				subsystem.Configure(app);
 
+			app.UseEndpoints(ep => {
```
timothycoleman, line 132: i'd prefer not to have two useendpoints calls because its kind of confusing - the subsystems can probably go after userouting?

timothycoleman (approved):
Nice

YoEight (approved):
LGTM

PR #4128, opened 2024-01-24
[DB-605] gRPC $all subscriptions: when subscribing at an invalid position

shaan1337:
Fixed: $all subscription enumerator returns `InvalidPosition` when subscribing at an invalid position

This is a follow-up PR to https://github.com/EventStore/EventStore/pull/4117. Tests were missing for the case where a user subscribes at an invalid position.

As @timothycoleman found out, when https://github.com/EventStore/EventStore/pull/4057 got rid of the `ReadIndex`, a small regression was introduced: the code still handles the exception `InvalidReadException` but this exception is no longer being thrown. This results in the `InvalidPosition` exception not being sent to the client when subscribing at an invalid position.

This PR fixes this with the following changes:
- Throw `InvalidReadException` in some additional cases in `TFChunkReadSide`
- Add a new `ReadAllResult` enum value: `InvalidPosition` and use/handle it where necessary
- Add a test `subscribe to $all at an invalid position` (note that the test doesn't cover all the different ways of obtaining an `invalid position` error)

Note: Similar changes must be applied to filtered $all in another PR.

### DIFF
--- a/src/EventStore.Core.Tests/Services/Transport/Enumerators/Enumerator.AllSubscription.CombinationTests.cs
+++ b/src/EventStore.Core.Tests/Services/Transport/Enumerators/Enumerator.AllSubscription.CombinationTests.cs
@@ -38,7 +38,8 @@ public partial class EnumeratorTests {
 			AtZero = 2,
 			OneBeforeLast = 3,
 			AtLast = 4,
-			OneByteAfterLast = 5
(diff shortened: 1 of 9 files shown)


timothycoleman (approved):
cheers

YoEight (approved):

PR #4130, opened 2024-01-25
Support loading configuration for plugins (DB-624)

timothycoleman:
Added: General support for plugin configuration

We now load the whole configuration (ClusterVNodeOptions, Metrics, Kestrel,
Logging, and any additional config json files) into a single IConfiguration object that we
add to the DI and will (shortly) make available to the plugins in ConfigureServices

The json files are loaded into the root of the IConfiguration and are expected
to put their values in the right sections. For historical reasons the metrics
config is an exception to this and is loaded into the section EventStore:Metrics

Additional config files are any json files in the following two directories:
- DefaultConfigurationDirectory/config
- ApplicationDirectory/config

We use the `config` subdirectory in each case to ensure that we don't load
other json files that are not rooted config files (*deps.json, metrics etc)

Additional config can also be set via
- command line: --eventstore:plugins:key=value
- environment: EventStore__Plugins__Key

Limitations: ClusterVNodeOptions cannot yet be configured in the additional json files. This is because ClusterVNodeOptions is not yet resolved out of the main IConfiguration object.


src/EventStore.ClusterNode/Program.cs
```diff
@@ -167,8 +167,40 @@ internal static class Program {
...
+						.AddSection(SectionNames.Metrics, metricsSection =>
+							metricsSection.AddConfigFile("metricsconfig.json")))
+
+					// The other config files are added to the root, and must put themselves in the appropriate sections
+					.AddConfigFile("kestrelsettings.json", optional: true, reloadOnChange: true)
+					.AddConfigFile("logconfig.json", optional: true, reloadOnChange: true)
+
+					// Load all json files in the  `config` subdirectory (if it exists) of each configuration
+					// directory. We use the subdirectory to ensure that we only load configuration files.
+					.AddConfigFiles(subdirectory: "config", pattern: "*.json")
```
RagingKore, line 188: If not a single json file is found it breaks because the function sets `optional:false`

src/EventStore.ClusterNode/Program.cs
```diff
@@ -167,8 +167,40 @@ internal static class Program {
...
-				using (var hostedService = new ClusterVNodeHostedService(options, certificateProvider, metricsConfiguration.Get<MetricsConfiguration>())) {
+				// Create a single IConfiguration object that contains the whole configuration, including 
+				// plugin configuration. We will add it to the DI and make it available to the plugins.
+				//
+				// Three json files are loaded explicitly for backwards compatibility
+				// - metricsconfig.json needs loading into the EventStore:Metrics section.
+				// - kestrelsettings.json is not located in a config/ directory
+				// - logconfig.json is not located in a config/ directory
+				var configuration = new ConfigurationBuilder()
+					.AddSection(SectionNames.EventStore, eventStoreSection => eventStoreSection
```
RagingKore, line 178: could we simply use `EventStore:Metrics`  as the section name?

`.AddSection("EventStore:Metrics", x => x.AddConfigFile("metricsconfig.json"))`

src/EventStore.ClusterNode/Program.cs
```diff
@@ -167,8 +167,40 @@ internal static class Program {
...
+					// The other config files are added to the root, and must put themselves in the appropriate sections
+					.AddConfigFile("kestrelsettings.json", optional: true, reloadOnChange: true)
+					.AddConfigFile("logconfig.json", optional: true, reloadOnChange: true)
+
+					// Load all json files in the  `config` subdirectory (if it exists) of each configuration
+					// directory. We use the subdirectory to ensure that we only load configuration files.
+					.AddConfigFiles(subdirectory: "config", pattern: "*.json")
+
+					.AddEnvironmentVariables()
+					.AddCommandLine(args)
```
RagingKore, line 191: I know this is the default priority, yet we should consider making the env vars take precedence.

src/EventStore.Common/Configuration/ConfigurationBuilderExtensions.cs
```diff
@@ -28,5 +33,59 @@ public static class ConfigurationBuilderExtensions {
...
+			
+			self.Add(new SectionSource(sectionName, configure));
+
+		public static IConfigurationBuilder AddConfigFile(
+			this IConfigurationBuilder builder,
+			string configFilePath,
+			bool optional = false,
+			bool reloadOnChange = false) {
+
+			if (!Locations.TryLocateConfigFile(configFilePath, out var directory, out var fileName)) {
```
RagingKore, line 51: The function hides this custom behaviour that tries to locate the files in yet another path.
We should consider passing the path to the function upfront.

src/EventStore.Common/Configuration/ConfigurationBuilderExtensions.cs
```diff
@@ -28,5 +33,59 @@ public static class ConfigurationBuilderExtensions {
...
+
+		// Allows configuration to be mounted inside a specified section
+		public static IConfigurationBuilder AddSection(
+			this IConfigurationBuilder self,
+			string sectionName,
+			Action<IConfigurationBuilder> configure) =>
+			
+			self.Add(new SectionSource(sectionName, configure));
+
+		public static IConfigurationBuilder AddConfigFile(
```
RagingKore, line 45: Consider changing the name of the function to indicate its behaviour is specific to ESDB.

src/EventStore.Common/Configuration/ConfigurationBuilderExtensions.cs
```diff
@@ -28,5 +33,59 @@ public static class ConfigurationBuilderExtensions {
...
+			if (!Locations.TryLocateConfigFile(configFilePath, out var directory, out var fileName)) {
+				if (optional) {
+					return builder;
+				}
+
+				throw new FileNotFoundException(
+					$"Could not find {configFilePath} in the following directories: {string.Join(", ", Locations.GetPotentialConfigurationDirectories())}");
+			}
+
+			builder.AddJsonFile(config => {
```
RagingKore, line 60: Could we instead pass just the `filename`, `optional` and `reloadOnChange` parameters?

src/EventStore.Common/Configuration/ConfigurationBuilderExtensions.cs
```diff
@@ -28,5 +33,59 @@ public static class ConfigurationBuilderExtensions {
...
+				config.FileProvider = new PhysicalFileProvider(directory) {
+					UseActivePolling = reloadOnChange,
+					UsePollingFileWatcher = reloadOnChange
+				};
+				config.OnLoadException = context => Serilog.Log.Error(context.Exception, "err");
+				config.Path = fileName;
+				config.ReloadOnChange = reloadOnChange;
+			});
+
+			Serilog.Log.Logger.Information($"Loaded configuration file {Path.Combine(directory, fileName)}");
```
RagingKore, line 71: Not sure we should add logging inside these helper functions.

src/EventStore.ClusterNode/Program.cs
timothycoleman: it'll only load files that are there, but fair enough i'll set optional true

src/EventStore.Common/Configuration/ConfigurationBuilderExtensions.cs
timothycoleman: im not sure, i've kept it the same way as it was being called before

RagingKore (approved):

pvanbuijtene (approved):

PR #4132, opened 2024-01-26
[DB-606] Add filtered $all subscription enumerator tests

shaan1337:
Added: Filtered $all subscription enumerator tests
Fixed: Use correct checkpoint in test subscribe_filtered_all_from_start

Note: the current tests don't verify whether `filtered $all checkpoints` are received at all. These will be added in a separate PR (https://github.com/EventStore/EventStore/pull/4138)

### DIFF
--- a/src/EventStore.Core.Tests/Services/Transport/Enumerators/Enumerator.AllSubscriptionFiltered.CombinationTests.cs
+++ b/src/EventStore.Core.Tests/Services/Transport/Enumerators/Enumerator.AllSubscriptionFiltered.CombinationTests.cs
@@ -1,0 +1,533 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+using EventStore.ClientAPI;
+using EventStore.ClientAPI.Common;
+using EventStore.ClientAPI.Messages;
+using EventStore.Core.Data;
+using EventStore.Core.Services.Storage.ReaderIndex;
+using EventStore.Core.Services.Transport.Enumerators;
+using EventStore.Core.Services.UserManagement;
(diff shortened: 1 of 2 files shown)


src/EventStore.Core.Tests/Services/Transport/Enumerators/Enumerator.AllSubscriptionFiltered.CombinationTests.cs
```diff
@@ -0,0 +1,533 @@
...
+			await PopulateExistingEvents();
+			_nextEventIndex = CalculateNextEventIndexFromCheckpoint();
+		}
+
+		[Test]
+		public async Task enumeration_is_correct() {
+			var sub = Subscribe();
+
+			if (SubscriptionProps.CheckpointType == CheckpointType.InvalidPosition) {
+				Assert.ThrowsAsync<ReadResponseException.InvalidPosition>(async () => await sub.GetNext());
```
YoEight, line 514: Do we really need an `async` lambda if we don't actually handle the inner value?

YoEight (approved):

src/EventStore.Core.Tests/Services/Transport/Enumerators/Enumerator.AllSubscriptionFiltered.CombinationTests.cs
shaan1337: @YoEight thanks, you're right, it doesn't look like it's needed. i suggest we leave it as is (for this PR) as it's in the tests and the same mistake has been repeated in the other enumerator tests

timothycoleman (approved):

PR #4136, opened 2024-01-29
Update tag-docker-images.yml to only run on main repository

lahma:
Now the nightly tagging happens also in forked repositories and fails every time:  https://github.com/lahma/EventStore/actions/workflows/tag-docker-images.yml

GitHub's guidance is to check for the repository:

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-only-run-job-for-specific-repository


timothycoleman (approved):
Many thanks!

PR #4144, opened 2024-02-01
[DB-627] Automate and Unify Configuration

RagingKore:
An attempt to unify all configuration entry points into a single one and also support the standard configuration conventions.

The configuration is loaded once at start and used by the ClusterVNode, HostBuilder and all components that will require configuration.

Each options group is automatically binded via the config framework, removing the need to manually set all values.

The node configuration and related functionality is also refactored and code moved closer to its logical context in terms of the projects.


timothycoleman:
Thanks, I've pushed a pile of suggestions and fixes, i'm happy with the state it's in now, we can look over them and get it squashed

thefringeninja (dismissed):

thefringeninja (approved):

timothycoleman (approved):
:+1:

PR #4158, opened 2024-02-13
Add check for empty uuidOption while making a gRPC call

dhingrak:
Fixed: Add a check for the empty uuidOption. 

Fixes https://linear.app/eventstore/issue/DB-504/when-uuid-option-is-not-set-grpc-read-request-hangs

Currently the EventStoreDB (master branch with Enumerator changes) sends an Unknown error exception if the uuidOption is null in the gRPC call. The goal of this PR is to  inform the user with a suitable message. 

EventStoreDB 23.10 and previous versions, we are not catching the exception when the uuid is null and the request hangs.
 
Current behaviour on master branch

After

### DIFF
--- a/src/EventStore.Core/Services/Transport/Grpc/RpcExceptions.cs
+++ b/src/EventStore.Core/Services/Transport/Grpc/RpcExceptions.cs
@@ -237,9 +237,12 @@ namespace EventStore.Core.Services.Transport.Grpc {
 		public static RpcException InvalidArgument<T>(T argument) =>
 			new(new Status(StatusCode.InvalidArgument, $"'{argument}' is not a valid {typeof(T)}"));
 
+		public static RpcException RequiredArgument<T>(string name, T argument) =>
+			new(new Status(StatusCode.InvalidArgument, $" `{name}` is a required argument of type {typeof(T)}"));
+
 		public static RpcException InvalidCombination<T>(T combination) where T : ITuple
 			=> new(new Status(StatusCode.InvalidArgument, $"The combination of {combination} is invalid."));
-		
+
 		public static RpcException InvalidPositionException() =>
 			new(new Status(StatusCode.InvalidArgument, $"Trying to read from an invalid position."));
 	}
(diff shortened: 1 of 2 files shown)


src/EventStore.Core/Services/Transport/Grpc/RpcExceptions.cs
```diff
@@ -237,9 +237,12 @@ public static Exception PersistentSubscriptionDropped(string streamName, string
 		public static RpcException InvalidArgument<T>(T argument) =>
 			new(new Status(StatusCode.InvalidArgument, $"'{argument}' is not a valid {typeof(T)}"));
 
+		public static RpcException RequiredOption<T>(string optionName, T argument) =>
+			new(new Status(StatusCode.InvalidArgument, $" `{optionName}` is a required field of type {typeof(T)}"));
```
pvanbuijtene, line 241: Tiny thing, the status code is about arguments maybe `RequiredArgument` would be better here and in the message 
```suggestion
			new(new Status(StatusCode.InvalidArgument, $" `{name}` is a required argument of type {typeof(T)}"));
```

src/EventStore.Core/Services/Transport/Grpc/Streams.Read.cs
```diff
@@ -38,6 +38,11 @@ internal partial class Streams<TStreamId> {
 				var user = context.GetHttpContext().User;
 				var requiresLeader = GetRequiresLeader(context.RequestHeaders);
 
+				var uuidOption = options.UuidOption;
```
hayley-jean, line 41: You could use `uuidOption` on [line 77](https://github.com/EventStore/EventStore/pull/4158/files#diff-b4a144b42979d2bcf4d8d82ddb732874a39d13b3966f28fe363540f0eadf9893R77) instead of the call to `options.UuidOption`

hayley-jean (approved):

src/EventStore.Core/Services/Transport/Grpc/Streams.Read.cs
dhingrak: Done

thefringeninja (changes requested):

src/EventStore.Core/Services/Transport/Grpc/Streams.Read.cs
```diff
@@ -38,6 +38,11 @@ internal partial class Streams<TStreamId> {
...
+				if (uuidOption == null){
```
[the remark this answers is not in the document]
dhingrak: Done

thefringeninja (approved):

hayley-jean (approved):

PR #4174, opened 2024-02-29
DB-272: Use RandomAccess to reduce file handles to one per chunk

sakno:
Changed: Reduced FileHandle usage by 80%. Now 1 per chunk instead of 5+

This PR suggests RandomAccess I/O for TFChunk with the following significant optimization: use only one file handle per chunk regardless a number of readers
Fixed DB-272

### DIFF
--- a/src/EventStore.Core/TransactionLog/Chunks/ChunkFooter.cs
+++ b/src/EventStore.Core/TransactionLog/Chunks/ChunkFooter.cs
@@ -2,9 +2,15 @@ using System;
 using System.IO;
 using EventStore.Common.Utils;
 using EventStore.Core.TransactionLog.Chunks.TFChunk;
+using DotNext.Buffers;
+using DotNext.Buffers.Binary;
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
 
 namespace EventStore.Core.TransactionLog.Chunks {
-	public class ChunkFooter {
+
+	// TODO: Consider struct instead of class
+	public sealed class ChunkFooter : IBinaryFormattable<ChunkFooter> {
 		public const int Size = TFConsts.ChunkFooterSize;
 		public const int ChecksumSize = 16;
 
@@ -18,7 +24,7 @@ namespace EventStore.Core.TransactionLog.Chunks {
 			LogicalDataSize; // the size of a logical data size (after scavenge LogicalDataSize can be > physicalDataSize)
 
 		public readonly int MapSize;
-		public readonly byte[] MD5Hash;
+		public readonly byte[] MD5Hash; // TODO: Allocation can be removed with InlineArray
 
 		public readonly int MapCount; // calculated, not stored
 
@@ -50,38 +56,67 @@ namespace EventStore.Core.TransactionLog.Chunks {
 			MapCount = mapSize / posMapSize;
 		}
 
-		public byte[] AsByteArray() {
-			var array = new byte[Size];
-			using (var memStream = new MemoryStream(array))
-			using (var writer = new BinaryWriter(memStream)) {
-				var flags = (byte)((IsCompleted ? 1 : 0) | (IsMap12Bytes ? 2 : 0));
-				writer.Write(flags);
-				writer.Write(PhysicalDataSize);
-				if (IsMap12Bytes)
-					writer.Write(LogicalDataSize);
-				else
-					writer.Write((int)LogicalDataSize);
-				writer.Write(MapSize);
+		public ChunkFooter(ReadOnlySpan<byte> source) {
+			Debug.Assert(source.Length >= Size);
 
-				memStream.Position = Size - ChecksumSize;
-				writer.Write(MD5Hash);
+			SpanReader<byte> reader = new(source);
+			byte flags = reader.Read();
+
+			IsCompleted = (flags & 1) is not 0;
+			IsMap12Bytes = (flags & 2) is not 0;
+			PhysicalDataSize = reader.ReadLittleEndian<int>();
+			LogicalDataSize = IsMap12Bytes
+				? reader.ReadLittleEndian<long>()
+				: reader.ReadLittleEndian<int>();
+
+			MapSize = reader.ReadLittleEndian<int>();
+			reader.ConsumedCount = Size - ChecksumSize;
+			MD5Hash = reader.ReadToEnd().ToArray();
+
+			var posMapSize = IsMap12Bytes ? PosMap.FullSize : PosMap.DeprecatedSize;
+			if (MapSize % posMapSize is not 0) {
+				throw new Exception(string.Format("Wrong MapSize {0} -- not divisible by PosMap.Size {1}.", MapSize,
+					posMapSize));
 			}
 
+			MapCount = MapSize / posMapSize;
+		}
+
+		static int IBinaryFormattable<ChunkFooter>.Size => Size;
+
+		public void Format(Span<byte> destination) {
+			Debug.Assert(destination.Length >= Size);
+
+			SpanWriter<byte> writer = new(destination);
+			int flags = Unsafe.BitCast<bool, byte>(IsCompleted)
+				| Unsafe.BitCast<bool, byte>(IsMap12Bytes) << 1;
+
+			writer.Add((byte)flags);
+			writer.WriteLittleEndian(PhysicalDataSize);
+			if (IsMap12Bytes)
+				writer.WriteLittleEndian(LogicalDataSize);
+			else
+				writer.WriteLittleEndian((int)LogicalDataSize);
+
+			writer.WriteLittleEndian(MapSize);
+			writer.WrittenCount = Size - ChecksumSize;
+			writer.Write(MD5Hash);
+		}
+
+		static ChunkFooter IBinaryFormattable<ChunkFooter>.Parse(ReadOnlySpan<byte> source)
+			=> new(source);
+
+		public byte[] AsByteArray() {
+			var array = new byte[Size];
+			Format(array);
 			return array;
 		}
 
+		[SkipLocalsInit]
 		public static ChunkFooter FromStream(Stream stream) {
-			var reader = new BinaryReader(stream);
-			var flags = reader.ReadByte();
-			var isCompleted = (flags & 1) != 0;
-			var isMap12Bytes = (flags & 2) != 0;
-			var physicalDataSize = reader.ReadInt32();
-			var logicalDataSize = isMap12Bytes ? reader.ReadInt64() : reader.ReadInt32();
-			var mapSize = reader.ReadInt32();
-			stream.Seek(-ChecksumSize, SeekOrigin.End);
-			var hash = reader.ReadBytes(ChecksumSize);
-
-			return new ChunkFooter(isCompleted, isMap12Bytes, physicalDataSize, logicalDataSize, mapSize, hash);
+			Span<byte> buffer = stackalloc byte[Size];
+			stream.ReadExactly(buffer);
+			return new(buffer);
 		}
 	}
 }
--- a/src/EventStore.Core/TransactionLog/Chunks/ChunkHeader.cs
+++ b/src/EventStore.Core/TransactionLog/Chunks/ChunkHeader.cs
@@ -1,11 +1,17 @@
 using System;
+using System.Diagnostics;
 using System.IO;
+using System.Runtime.CompilerServices;
 using EventStore.Common.Utils;
 using EventStore.Core.Exceptions;
 using EventStore.Core.Index;
+using DotNext.Buffers;
+using DotNext.Buffers.Binary;
 
 namespace EventStore.Core.TransactionLog.Chunks {
-	public class ChunkHeader {
+
+	// TODO: Consider struct instead of class
+	public sealed class ChunkHeader : IBinaryFormattable<ChunkHeader> {
 		public const int Size = TFConsts.ChunkHeaderSize;
 
 		public readonly long ChunkStartPosition; // return ChunkStartNumber * (long)ChunkSize;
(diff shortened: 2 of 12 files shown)


src/EventStore.Core/TransactionLog/Chunks/ChunkHeader.cs
```diff
@@ -39,36 +45,66 @@ public class ChunkHeader {
...
+			SpanReader<byte> reader = new(source);
+
+			if ((FileType)reader.Read() is not FileType.ChunkFile)
+				throw new CorruptDatabaseException(new InvalidFileException());
+
+			Version = reader.Read();
+			Debug.Assert(Version >= 0);
+
+			ChunkSize = reader.ReadLittleEndian<int>();
+			Debug.Assert(ChunkSize > 0);
```
shaan1337, line 60:
```suggestion
			Debug.Assert(ChunkSize >= 0);
```

It's possible for a chunk to have empty data during a scavenge (before the scavenge merge phase)

src/EventStore.Core/Util/MD5Hash.cs
```diff
@@ -1,7 +1,9 @@
 ﻿using System;
 using System.IO;
 using System.Security.Cryptography;
+using DotNext;
 using EventStore.Common.Utils;
+using Microsoft.Win32.SafeHandles;
```
shaan1337, line 6: unused imports

src/EventStore.Core/TransactionLog/Chunks/TFChunk/WriterWorkItem.cs
```diff
@@ -1,90 +1,74 @@
...
-			if (memStream != null) {
-				memStream.Dispose();
-				_memStream = null;
-			}
-		}
-
-		public void EnsureMemStreamLength(long length) {
-			throw new NotSupportedException("Scavenging with in-mem DB is not supported.");
+			_memStream?.Dispose();
+			_memStream = null;
```
shaan1337, line 71: should `_memStream` be set to null _before_ disposing the memory stream? otherwise it could be possible that we use a disposed stream. (i'm not sure if we need to fix it in this PR as this issue was already present before and it might not be straightforward to fix as it involves adding memory barriers)

shaan1337: hmmm, it also doesn't seem to be a problem as we always cache the active chunk now. I wonder if it would be a good idea to get rid of the `DisposeMemStream()` method on `WriterWorkItem`

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
```diff
@@ -381,30 +378,9 @@ private enum CacheStatus {
...
-					FileAccess.Read,
-					FileShare.ReadWrite,
-					ReadBufferSize,
-					_reduceFileCachePressure ? FileOptions.None : FileOptions.RandomAccess);
-			}
+			Debug.Assert(_handle is not null);
 
-			var reader = new BinaryReader(stream);
-			return new ReaderWorkItem(stream, reader, false);
+			return new ReaderWorkItem(_handle, _unbuffered);
```
shaan1337, line 383: This change seems to alter the semantics of the `--unbuffered` configuration option (defined [here](https://github.com/EventStore/EventStore/blob/fb20c2cb507f4ceaf5216d4a43736967de032a80/src/EventStore.Core/ClusterVNodeOptions.cs#L462)) as we no longer bypass the file caches when `unbuffered` is true here. We may consider changing the definition or removing the configuration option and document it as a breaking change.

src/EventStore.Core/TransactionLog/Chunks/TFChunk/WriterWorkItem.cs
```diff
@@ -1,90 +1,74 @@
...
+			_fileStream?.SetLength(fileSize);
+			_memStream?.SetLength(fileSize);
 		}
 
-		public void Dispose() {
-			if (_fileStream != null)
-				_fileStream.Dispose();
+		protected override void Dispose(bool disposing) {
+			if (disposing) {
+				_fileStream?.Dispose();
```
shaan1337, line 57: is it guaranteed that dotNext's `UnbufferedFileStream` will flush the file to disk when being disposed? (the previous code implicitly assumes that it does)

shaan1337 (commented):
@sakno hey, the PR's looking great! i've added a few questions/comments

i've done some rough benchmarks/tests on ubuntu:

| Test with ~200k chunks, 1 kb each, disk AWS gp3     | master                | This PR               | This PR (with --unbuffered) |
|-----------------------------------------------------|-----------------------|-----------------------|-----------------------------|
| time to open chunks - page cache cleared (secs)     | 159 (0.79 ms / chunk) | 151 (0.75 ms / chunk) | 156 (0.78 ms / chunk)       |
| time to open chunks - page cache not cleared (secs) | 18 (0.09 ms / chunk)  | 12 (0.06 ms / chunk)  | 10 (0.05 ms / chunk)        |
| # of file handles (verified under `/proc/<pid>/fd/`)   | ~1M                   | ~200k                 | ~200k                       |

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
sakno: `--unbuffered` and `--writeThrough` semantics still preserved. If `_unbuffered=false`, `BufferedStream` is created to wrap unbuffered stream, otherwise not.

src/EventStore.Core/TransactionLog/Chunks/TFChunk/WriterWorkItem.cs
sakno: > is it guaranteed that dotNext's UnbufferedFileStream will flush the file to disk when being disposed

No, it's not. However, most of the code inside `TFChunk` do `Flush` before disposing of writer item. Redundant call to `FlushToDisk` leads to extra syscall with no effect. Maybe I missed something. If so, I can add `FlushToDisk` explicitly.

src/EventStore.Core/Util/MD5Hash.cs
sakno: Fixed

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
sakno: `--writeThrough` now is working exactly as expected. In the previous version, `FileStream` created for the writer side didn't take into account this flag using `FileOptions.WriteThrough` flag. Now this option is considered when file handle needs to be created.

src/EventStore.Core/TransactionLog/Chunks/TFChunk/WriterWorkItem.cs
shaan1337: the case i'm worried about is when doing a manual shut down of the database:
`bool TryClose()` in TFChunk should return true only if the chunk was properly closed and flushed if it was being written to (otherwise, the writer checkpoint may be flushed but not the chunk which would cause the database to become inconsistent - see this [bug](https://github.com/EventStore/EventStore/pull/3669) for example)

Based on online articles, it appears that .NET's Filestream's automatically flushes when it's being closed, so we previously didn't need to have an explicit Flush() call, but now it appears that we do need it.

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
shaan1337: > `--unbuffered` and `--writeThrough` semantics still preserved. If `_unbuffered=false`, `BufferedStream` is created to wrap unbuffered stream, otherwise not.

When `--unbuffered` is specified, does it still use the `O_DIRECT` flag on linux / `FILE_FLAG_NO_BUFFERING` on windows?

i believe this is what is intended when `Direct I/O` is mentioned in the configuration option description:
`Enables Unbuffered/DirectIO when writing to the file system, this bypasses filesystem caches.`

sakno: > does it still use the O_DIRECT flag on linux 

It's not. `O_SYNC` is actually used when `WriteThrough` applied.

> O_DIRECT alone only promises that the kernel will avoid copying data from user space to kernel space, and will instead write it directly via DMA (Direct memory access; if possible). Data does not go into caches. There is no strict guarantee that the function will return only after all data has been transferred.

In the same time

> O_SYNC guarantees that the call will not return before all data has been transferred to the disk (as far as the OS can tell)

Both still provide weak guarantees. IMHO, `O_SYNC` has stronger guarantees rather than `O_DIRECT`.

src/EventStore.Core/TransactionLog/Chunks/TFChunk/WriterWorkItem.cs
timothycoleman: yeah i dont think we need to change anything here, as you say it matches the previous behaviour and it doesn't seem to be a problem anyway

src/EventStore.Core/TransactionLog/Chunks/ChunkFooter.cs
```diff
@@ -50,38 +56,67 @@ public readonly long
...
+				throw new Exception(string.Format("Wrong MapSize {0} -- not divisible by PosMap.Size {1}.", MapSize,
+					posMapSize));
 			}
 
+			MapCount = MapSize / posMapSize;
+		}
+
+		static int IBinaryFormattable<ChunkFooter>.Size => Size;
+
+		public void Format(Span<byte> destination) {
```
timothycoleman, line 87: this writes less data than `Size` bytes, should we fill the remainder of Size with 0s? or alternatively write Size zeroes to the span first? same for header

src/EventStore.Core/TransactionLog/Chunks/TFChunk/PosMap.cs
```diff
@@ -1,7 +1,11 @@
+using System;
+using System.Diagnostics;
 using System.IO;
+using DotNext.Buffers;
+using DotNext.Buffers.Binary;
 
 namespace EventStore.Core.TransactionLog.Chunks.TFChunk {
-	public struct PosMap {
+	public struct PosMap : IBinaryFormattable<PosMap> {
```
timothycoleman, line 8: i don't think the new code in this file is used, should it be?

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
```diff
@@ -1194,14 +1136,14 @@ private enum CacheStatus {
...
-			return new TFChunkBulkReader(this, GetSequentialReaderFileStream(), isMemory: false);
+			return new TFChunkBulkReader(this, CreateFileStreamForBulkReader(), isMemory: false);
 		}
 
-		private Stream GetSequentialReaderFileStream() {
-			return _inMem
-				? (Stream)new UnmanagedMemoryStream((byte*)_cachedData, _fileSize)
-				: new FileStream(_filename, FileMode.Open, FileAccess.Read, FileShare.ReadWrite, 65536,
-					FileOptions.SequentialScan);
+		private Stream CreateFileStreamForBulkReader() {
```
timothycoleman, line 1142: the old code here was checking _inMem (but not _unbuffered), was this an intentional change?

src/EventStore.Core/TransactionLog/Chunks/TFChunk/WriterWorkItem.cs
```diff
@@ -1,90 +1,74 @@
...
-		public void FlushToDisk() {
-			if (_fileStream == null) return;
-			var fs = _fileStream as FileStream;
-			if (fs != null)
-				fs.FlushToDisk(); //because of broken flush in filestream in 3.5
-			else {
-				_fileStream.Flush();
-			}
-		}
+		public void FlushToDisk() => _fileStream?.Flush();
```
timothycoleman, line 67: the old code used to set `flushToDisk: true` in the eventual filestream flush and call, and on windows also call FlushFileBuffers

does what we have now do the same as the old filestream 'flushToDisk: true` behaviour

is FlushFileBuffers definitely not necessary (we still have a safeFileHandle if we want to call it)

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
```diff
@@ -250,6 +245,13 @@ private enum CacheStatus {
 				throw new CorruptDatabaseException(new ChunkNotFoundException(_filename));
 
 			_fileSize = (int)fileInfo.Length;
+			_handle = File.OpenHandle(
+				_filename,
+				FileMode.Open,
+				FileAccess.Read,
+				FileShare.ReadWrite,
+				_reduceFileCachePressure ? FileOptions.None : FileOptions.RandomAccess | FileOptions.SequentialScan);
```
timothycoleman, line 253: some parenthesis would be handy to make the intention clear
is it weird to be setting both the randomaccess and sequentialscan flags? if its intentional it probably warrants a comment

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
```diff
@@ -968,6 +924,7 @@ private enum CacheStatus {
 			bool closed = true;
 			closed &= TryDestructFileStreams();
 			closed &= TryDestructMemStreams();
+			_handle?.Dispose();
```
timothycoleman, line 927: i think we only want to dispose this if `closed` is true (otherwise we still have some active readers)

src/EventStore.Core/TransactionLog/Chunks/ChunkFooter.cs
sakno: I though about that. Checksum doesn't include these bits. They considered as reserved so no matter how they filled.

src/EventStore.Core/TransactionLog/Chunks/TFChunk/WriterWorkItem.cs
sakno: `Stream` wrapper over file handle does exactly what we expected: https://github.com/dotnet/dotNext/blob/master/src/DotNext.IO/IO/UnbufferedFileStream.cs#L48

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
sakno: It is intentional, old code uses both options on different instances of `FileStream` (sequential for bulk reader and random for reader work item).

sakno: Fixed

src/EventStore.Core/TransactionLog/Chunks/TFChunk/PosMap.cs
sakno: That was my attempt to migrate away from BinaryReader/BinaryWriter using this new abstraction. Now fixed in the following way: overloads with `BinaryReader`/`BinaryWriter` just delegate calls to Span-based versions.

timothycoleman:
For posterity:

This PR causes all the reader work items and the writer work item for a chunk to share the same file handle. Bulk readers still have their own individual file handle but there are very few of these open at any one time in the whole process.

This changes the flags for the writer work item from `SequentialScan` to `None`/`RandomAccess` (according to the `ReduceFileCachePressure` setting), because the handle is also used for (non bulk) reads now. As far as Roman and I can tell, these flags only affect reading and not writing so it is ok to change these flags for the writer.

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
timothycoleman: :+1:

src/EventStore.Core/TransactionLog/Chunks/ChunkFooter.cs
timothycoleman: fair enough, i think it just means if we start using some of those bits later we will tell the footer some version number so it knows whether it should look at them or not, but perhaps we ought to do that in any case

src/EventStore.Core/TransactionLog/Chunks/TFChunk/WriterWorkItem.cs
timothycoleman: yeah good catch :+1:

timothycoleman (approved):
:+1: great!

shaan1337 (approved):
nice work!

PR #4179, opened 2024-03-01
Add the elections counter metric

dhingrak:
Added: Add the elections counter metric so users can set alerts if the number of elections over a certain period of time exceeds some number.

Fixes https://linear.app/eventstore/issue/DB-599/add-the-elections-counter-to-metrics

### DIFF
--- a/src/EventStore.Core/Metrics/ElectionsCounterTracker.cs
+++ b/src/EventStore.Core/Metrics/ElectionsCounterTracker.cs
@@ -1,0 +1,25 @@
+﻿using EventStore.Core.Bus;
+using EventStore.Core.Messages;
+
+
+namespace EventStore.Core.Metrics;
+
+public interface IElectionCounterTracker : IHandle<ElectionMessage.ElectionsDone> {
+}
+
+public class ElectionsCounterTracker : IElectionCounterTracker {
+	private readonly CounterSubMetric _electionsCounter;
+
+	public ElectionsCounterTracker(CounterSubMetric electionsCounter) {
+		_electionsCounter = electionsCounter;
+	}
+
+	public void Handle(ElectionMessage.ElectionsDone message) {
+		_electionsCounter.Add(1);
+	}
+
+	public class NoOp : IElectionCounterTracker {
+		public void Handle(ElectionMessage.ElectionsDone message) {
+		}
+	}
+}
--- a/src/EventStore.ClusterNode/metricsconfig.json
+++ b/src/EventStore.ClusterNode/metricsconfig.json
@@ -12,6 +12,8 @@
 		"Scavenge": true
 	},
 
+	"ElectionsCount": true,
+
 	"Checkpoints": {
 		"Writer": true,
 		"Index": true,
--- a/src/EventStore.Common/Configuration/MetricsConfiguration.cs
+++ b/src/EventStore.Common/Configuration/MetricsConfiguration.cs
@@ -137,7 +137,9 @@ namespace EventStore.Common.Configuration {
 		public Dictionary<EventTracker, bool> Events { get; set; } = new();
 
 		public Dictionary<Cache, bool> CacheHitsMisses { get; set; } = new();
-		
+
+		public bool ElectionsCount { get; set; } = false;
+
 		public bool CacheResources { get; set; } = false;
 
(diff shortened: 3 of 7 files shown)


src/EventStore.Core/Metrics/ElectionsCounterMetric.cs
```diff
@@ -0,0 +1,16 @@
...
+	private Counter<int> ElectionsCounter { get; }
+
+	public ElectionsCounterMetric(Meter meter, string name) {
+		ElectionsCounter = meter.CreateCounter<int>(name);
+	}
+
+	public void IncrementCounter() {
+		ElectionsCounter.Add(1);
+	}
+
```
timothycoleman, line 15:
```suggestion
```

src/EventStore.Core/Metrics/ElectionsCounterMetric.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System.Diagnostics.Metrics;
+
+namespace EventStore.Core.Metrics;
+
+public class ElectionsCounterMetric {
+	private Counter<int> ElectionsCounter { get; }
```
timothycoleman, line 6: generally the codebase has private fields rather than private properties, would you mind changing this to a private field

src/EventStore.Core/Metrics/ElectionsCounterTracker.cs
```diff
@@ -0,0 +1,21 @@
+﻿
+namespace EventStore.Core.Metrics;
+
+public interface IElectionCounterTracker {
+	void IncrementCounter();
+}
+public class ElectionsCounterTracker: IElectionCounterTracker {
```
timothycoleman, line 7:
```suggestion

public class ElectionsCounterTracker: IElectionCounterTracker {
```

src/EventStore.Core/Metrics/ElectionsCounterTracker.cs
```diff
@@ -0,0 +1,21 @@
...
+	private readonly ElectionsCounterMetric _metrics;
+
+	public ElectionsCounterTracker(ElectionsCounterMetric metrics) {
+		_metrics = metrics;
+	}
+	public void IncrementCounter() {
+		_metrics.IncrementCounter();
+	}
+
+	public class NoOp: IElectionCounterTracker {
```
timothycoleman, line 17:
```suggestion
	public class NoOp : IElectionCounterTracker {
```

timothycoleman:
Oh yeah and updating the documentation would be fab too

dhingrak:
@timothycoleman I have handled the Elections.Done message in tracker itself but I am not 100% sure if this is the right to do it.

src/EventStore.Core/Metrics/ElectionsCounterTracker.cs
```diff
@@ -0,0 +1,36 @@
+﻿using EventStore.Core.Bus;
+using EventStore.Core.Messages;
+
+
+namespace EventStore.Core.Metrics;
+
+public interface IElectionCounterTracker {
+	void IncrementCounter();
+}
```
timothycoleman, line 9:
```suggestion
public interface IElectionCounterTracker : IHandle<Messages.ElectionMessage.ElectionsDone> {
}```

ah ok sorry, i meant like this ^

**Comment by dhingrak** [comment]
I have added the IncrementCounter function because of Unit testing. This way the tests are easy to understand. We can test this with sut.Hanlde(null) if we want.

**Comment by timothycoleman** [comment]
cool, thanks

yes we should test it by calling Handle, and remove the IncrementCounter method. That way the tests are using the tracker how it is supposed to be used

you can construct an electionsdone message like this

```
	private readonly Messages.ElectionMessage.ElectionsDone _electionsDoneMessage;

	public ElectionsCounterTrackerTests() {
		var endPoint = new DnsEndPoint("127.0.0.1", 1113);
		var memberInfo = Cluster.MemberInfo.Initial(Guid.Empty, DateTime.UtcNow,
			VNodeState.Unknown, true,
			endPoint, endPoint, endPoint, endPoint, endPoint,
			null, 0, 0, 0, false);
		_electionsDoneMessage = new(1, 1, memberInfo);
	}
```

YoEight (approved):

PR #4209, opened 2024-03-19
[DB-21-1]: Reduce memory allocations

sakno:
Subsequent work started in #4174 . The goal of this PR is to reduce memory allocations:

- [x] Remove multiple instantiations of UnmanagedMemoryStream (at least for read side)
- [x] Remove temp buffers when possible
- [x] Improve data locality for read side
- [x] Remove `InitialReaderCount` configuration property because it's no longer relevant. The chunk shares the same file handle between multiple readers. Thus, there is no overhead on reader when preparing a new reader work item
- [x] Introduce lazy pooling with predictable take/return behavior to maximize reusability of the buffers on read side

Fixed DB-276


src/EventStore.Core/Configuration/ClusterVNodeOptions.cs
```diff
@@ -364,6 +364,8 @@ public record DatabaseOptions {
 			public bool Unbuffered { get; init; } = false;
 
 			[Description("The initial number of readers to start when opening a TFChunk.")]
+			[Deprecated("This configuration property has no effect")]
+			[Obsolete("This configuration property has no effect")]
```
shaan1337, line 368:
```suggestion
			[Description("The initial number of readers to start when opening a TFChunk.")]
			[Deprecated("The ChunkInitialReaderCount parameter has been deprecated as of version 24.6.0 and currently has no effect.")]
```
Just for proper logging and consistency with other deprecated options.
Also, it appears that the `Obsolete` attribute is set only when a parameter has been renamed, so we may not need it here.

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
```diff
@@ -1241,5 +1221,67 @@ private struct Midpoint {
...
+			}
+
+			internal bool Return(ReaderWorkItem item) {
+				int index = item.PositionInPool;
+				if ((uint)index < (uint)_array.Length) {
+					_indicies.Return(index);
+					return true;
+				}
+
+				return false;
```
shaan1337, line 1283:
```suggestion
				if (index < 0)
					return false;

				_indicies.Return(index);
				return true;
```
is there any specific reason for casting to `uint` instead of doing something like the above? (i guess it's performance related)

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
```diff
@@ -1241,5 +1221,67 @@ private struct Midpoint {
...
 		}
+
+		[StructLayout(LayoutKind.Auto)]
+		private struct ReaderWorkItemPool {
+			private readonly ReaderWorkItem[] _array;
+
+			// IndexPool supports up to 64 elements with O(1) take/return time complexity.
+			// It's a thread-safe data structure with no allocations that provide predictability about
+			// the indicies: smallest available index is always preferred.
+			private IndexPool _indicies;
```
shaan1337, line 1232:
```suggestion
			private IndexPool _indices;
```

typo

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
sakno: > i guess it's performance related

It is. `(uint)index < (uint)_array.Length` is a fast alternative to `index > 0 && index < _array.Length`. Cast from `int` to `uint` has zero cost in assembly (it just a hint to use unsigned comparison instruction instead of signed one).

Comparison with `0` provides better readability.

sakno: Fixed.

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
```diff
@@ -1064,24 +1035,23 @@ private enum CacheStatus {
...
-
-			Interlocked.Increment(ref _fileStreamCount);
 			if (_selfdestructin54321) {
 				if (Interlocked.Decrement(ref _fileStreamCount) == 0)
 					CleanUpFileStreamDestruction();
 				throw new FileBeingDeletedException();
 			}
 
-			return CreateInternalReaderWorkItem();
+			Interlocked.Increment(ref _fileStreamCount);
```
shaan1337, line 1047: it looks like this increment operation should be done before `if (_selfdestructin54321) {` just above (as it was in the previous code)

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
```diff
@@ -369,20 +360,8 @@ private enum CacheStatus {
...
-				_fileStreams.Add(CreateInternalReaderWorkItem());
-			}
-		}
-
-		private ReaderWorkItem CreateInternalReaderWorkItem() {
-			Debug.Assert(_handle is not null);
-
-			return new(_handle);
+			_fileStreams = new();
+			Interlocked.Add(ref _fileStreamCount, _fileStreams.Count);
```
shaan1337, line 364: here also, i'm not 100% sure but it appears that we may still want to do the increment operation before creating the pool as it seems possible for `CreateReaderStreams()` to execute concurrently with `TryDestructFileStreams()`

we may consider making the size a constant in `ReaderWorkItemPool` so that we know it in advance

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
sakno: Correct. The actual reason for that is `Interlocked.Decrement` which should be placed after increment.

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
sakno: > we may consider making the size a constant in ReaderWorkItemPool so that we know it in advance

It's a subject for the potential bug in the future. `IndexPool` can be instantiated with any capacity that is less than or equal to `IndexPool.Capacity` property. To change the capacity, we need to change one line of code. Having constant, we need to keep it synchronized. Moreover, `LOCK.XADD` with constant operand doesn't work faster.

sakno:
```csharp
_fileStreams = new();
Interlocked.Add(ref _fileStreamCount, _fileStreams.Count);
```

To make this code a safe when concurrency with `TryDestructFileStreams` occurs, I need to insert `if` statement to check `localReferenceCount` on equality with 0. This check creates a correct happens-before relationship. The same cannot be done if switch those two lines of code. As a result, I prefer to keep the order.

timothycoleman: im satisfied with everything now except for the threading implications of the two Drain calls, which might be ok but i've run out of steam to tell for today

i'm slightly suspicious because the Drain() can be called on _fileStreams and _memStreams when they are not initialised (this is ok i think) and concurrently with them being initialized, which im less sure about, especially with respect to getting the counting right. it might be worth a few more words to explain why its safe

sakno: `Drain` itself is thread safe, because all instance methods of `IndexPool` are thread safe. The first who reaches `Take` will take all available indices and dispose them. The pool is empty after `Take(Span<int>)`.

sakno: The only thing that is not thread safe is initialization of `ReaderWorkItemPool`.

shaan1337: Following the change in 45e4a3029629d152c5345627fd61ebd314c74a05 (the > 0 check), is an interleaving like this possible where `_fileStreams.Drain()` could return `0` causing `CleanUpFileStreamDestruction();` to be called but there might still be active filestreams?

```
1. CreateReaderStreams() :: _fileStreams = new(); // active chunk completed, reader streams being created
2. GetReaderWorkItem() (internally calls _fileStreams.TryTake) // triggered by read
3. TryDestructFileStreams() // triggered by manual shutdown, _fileStreams.Drain() will return 0.
4. CreateReaderStreams() :: Interlocked.Add(ref _fileStreamCount, _fileStreams.Count);
```

i'm not sure if step 2 would be able to create a file stream due to `IsReadOnly` not being set to true yet (i assume that the barriers in Interlocked.Add and _fileStreams.TryTake could prevent that from happening)

> To change the capacity, we need to change one line of code. Having constant, we need to keep it synchronized.

alternatively, we could also add a `capacity` constructor parameter to ReaderWorkItemPool to avoid the above

timothycoleman (approved):

shaan1337 (approved):

PR #4210, opened 2024-03-20
[DB-722] Add TCP API as a plugin for unit tests

YoEight:
Changed: Unit tests use TCP as a plugin


src/EventStore.Core.Tests/Helpers/MiniClusterNode.cs
```diff
@@ -87,8 +90,9 @@ public class MiniClusterNode<TLogFormat, TStreamId> {
 
 			var useHttps = EnableHttps();
 
-			Environment.SetEnvironmentVariable(ClusterVNode.TcpApiEnvVar, "TRUE");
-			Environment.SetEnvironmentVariable(ClusterVNode.TcpApiPortEnvVar, $"{externalTcp.Port}");
+			var len = subsystems?.Length ?? 0;
+			Array.Resize(ref subsystems, len + 1);
+			subsystems[len] = new TcpApiTestPlugin();
```
timothycoleman, line 95:
```suggestion
			subsystems ??= [];
			subsystems = [.. subsystems, new TcpApiTestPlugin()];
```

src/EventStore.Core.Tests/Helpers/MiniClusterNode.cs
```diff
@@ -142,9 +146,16 @@ public class MiniClusterNode<TLogFormat, TStreamId> {
...
-				Subsystems = subsystems ?? Array.Empty<ISubsystem>()
+				Subsystems = subsystems
 			};
 
+			var inMemConf = new ConfigurationBuilder()
+				.AddInMemoryCollection(new KeyValuePair<string, string>[] {
+					new("EventStore:Plugins:Tcp:ExtPort", externalTcp.Port.ToString()),
+					new("EventStore:Plugins:Tcp:ExtHeartbeatInterval", "10000"),
+					new("EventStore:Plugins:Tcp:ExtHeartbeatTimeout", "10000"),
+					new("EventStore:Plugins:Tcp:Insecure", options.Application.Insecure.ToString()),
```
timothycoleman, line 157: Probably we shouldn't have an extra setting for Tcp Insecure, and instead just have the normal server one (otherwise in real life people will have to coordinate the configuration of their tcp plugin with the server configuration)

src/EventStore.Core.XUnit.Tests/Configuration/ClusterNodeOptionsTests/when_building/with_cluster_node_and_custom_settings.cs
```diff
@@ -120,8 +116,6 @@ public class with_custom_external_ip_address_as_advertise_info<TLogFormat, TStre
 
 		[Test]
 		public void should_set_the_custom_advertise_info_for_external() {
-			Assert.AreEqual(new DnsEndPoint("196.168.1.1", 11131),
-				_node.GossipAdvertiseInfo.ExternalTcp);
```
timothycoleman, line 124: was it intentional to remove this?

src/EventStore.Core/ClusterVNode.cs
```diff
@@ -274,7 +269,10 @@ public class ClusterVNode<TStreamId> :
 
 			var disableInternalTcpTls = options.Application.Insecure;
 			var disableExternalTcpTls = options.Application.Insecure;
-			var enableExternalTcp = ExtTcpOptions.TryParse(out var extTcpOptions);
+			var extTcpOptions = configuration.GetSection("EventStore:Plugins:Tcp").Get<ExtTcpOptions>();
+			var enableExternalTcp = extTcpOptions != null;
```
timothycoleman, line 273: Probably having a explicit config option `EnableExternalTcp` would be nice (same name as the old setting)

src/EventStore.Core/ClusterVNode.cs
```diff
@@ -289,21 +287,21 @@ public class ClusterVNode<TStreamId> :
 
 			var extTcp = disableExternalTcpTls && enableExternalTcp
 				? new IPEndPoint(options.Interface.NodeIp,
-					extTcpOptions.Port)
+					extTcpOptions.ExtPort)
```
timothycoleman, line 290: for the sake of making it easier to migrate to the plugin it might be nice to keep the old names `NodeTcpPort` `NodeTcpPortAdvertiseAs`

src/EventStore.Core/ExtTcpOptions.cs
```diff
@@ -1,39 +1,6 @@
-﻿using System;
-
-namespace EventStore.Core;
+﻿namespace EventStore.Core;
 
 public class ExtTcpOptions {
```
timothycoleman, line 3:
```suggestion
// Still needed for gossip
public class ExtTcpOptions {
```

src/etc/tests/tcp/EventStore.Plugins.Tcp.Tests.dll
timothycoleman, line 1: maybe popping it in /lib instead of /src/etc? im open to persuasion though

src/EventStore.Core/ClusterVNode.cs
YoEight: My logic was we deprecated those names I used but those were never actually removed. So I thought keeping the old names was the least likely to break.

src/EventStore.Core/ExtTcpOptions.cs
YoEight: It we want to be thorough, it's also used to gather tcp stats too

src/etc/tests/tcp/EventStore.Plugins.Tcp.Tests.dll
YoEight: Actually, that was my first choice but though maybe we wanted it to be closer to where it was used.

src/EventStore.Core.Tests/Helpers/MiniClusterNode.cs
YoEight: It's only used for testing. The TCP plugin uses the database settings to set this up.

timothycoleman: ah i see, perhaps if we make it `EventStore:Plugins:TestTcp:...`

timothycoleman (approved):

hayley-jean (approved):

PR #4213, opened 2024-03-26
[DB-26-5] Change event type of Persistent Subscription checkpoint to $SubscriptionCheckpoint

lakshdeepsingheventstore:
Fixed : SubscriptionCheckpoint event type not filtered. https://linear.app/eventstore/issue/DB-543/subscriptioncheckpoint-event-type-not-filtered

Before this fix/change, the Persistent Subscription Checkpoint event was of type "SubscriptionCheckpoint". As a result, it was not filtered out, when reading only non-system events through a subscription to $all stream.

Following discussion has more details : https://discuss.eventstore.com/t/subscription-returning-system-events/5037

### DIFF
--- a/src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionCheckpointReader.cs
+++ b/src/EventStore.Core/Services/PersistentSubscription/PersistentSubscriptionCheckpointReader.cs
@@ -30,7 +30,7 @@ namespace EventStore.Core.Services.PersistentSubscription {
 
 			public void LoadStateCompleted(ClientMessage.ReadStreamEventsBackwardCompleted msg) {
 				if (msg.Events.Length > 0) {
-					var checkpoint = msg.Events.Where(v => v.Event.EventType == "SubscriptionCheckpoint")
+					var checkpoint = msg.Events.Where(v => v.Event.EventType == "$SubscriptionCheckpoint")
 						.Select(x => x.Event).FirstOrDefault();
 					if (checkpoint != null) {
 						string lastEvent = checkpoint.Data.ParseJson<string>();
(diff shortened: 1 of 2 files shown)


src/EventStore.Core/Services/Storage/ReaderIndex/EventFilter.cs
```diff
@@ -173,10 +173,18 @@ private sealed class EventTypeRegexStrategy : IEventFilter {
...
 			}
-
-			public bool IsEventAllowed(EventRecord eventRecord) =>
-				(!_isAllStream || DefaultAllFilter.IsEventAllowed(eventRecord)) &&
-				_expectedRegex.Match(eventRecord.EventType).Success;
+			public bool IsEventAllowed(EventRecord eventRecord) {
+					if (_expectedRegex.ToString().Contains("^[^\\$].*")) {
+						return (!_isAllStream || DefaultAllFilter.IsEventAllowed(eventRecord)) &&
+						       _expectedRegex.Match(eventRecord.EventType).Success	&&
+						       (!eventRecord.EventStreamId.StartsWith("$persistentsubscription")) &&
```
hayley-jean, line 180: This logic should be in the `NonSystemStreamStrategy` rather than on the regex strategy

hayley-jean (changes requested):

src/EventStore.Core/Services/Storage/ReaderIndex/EventFilter.cs
lakshdeepsingheventstore: Unfortunately, making changes to NonSystemStreamStrategy won't work as expected. The NonSystemStreamStrategy is not called when using the FilterSystemEvents() method on the client side. The filtering criteria for NonSystemStreamStrategy  is for filtering out all streams starting with '$', whereas the filtering criteria for FilterSystemEvents() is based upon event types (event types starting with '$') instead of stream names.

hayley-jean (approved):

PR #4214, opened 2024-03-28
[DB-26-3] Open scavenged chunks quickly. Build midpoints later on demand

timothycoleman:
Changed: Open the database faster by building the midpoints for scavenged chunks later on demand

Otherwise opening a database with thousands scavenged chunks can become too slow

### DIFF
--- a/src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunkReadSide.cs
+++ b/src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunkReadSide.cs
@@ -11,7 +11,7 @@ using Range = EventStore.Core.Data.Range;
 namespace EventStore.Core.TransactionLog.Chunks.TFChunk {
 	public partial class TFChunk {
 		public interface IChunkReadSide {
-			void Cache();
+			void RequestCaching();
 			void Uncache();
 
 			bool ExistsAt(long logicalPosition);
@@ -30,7 +30,7 @@ namespace EventStore.Core.TransactionLog.Chunks.TFChunk {
 					throw new ArgumentException("Scavenged TFChunk passed into unscavenged chunk read side.");
 			}
 
-			public void Cache() {
+			public void RequestCaching() {
 				// do nothing
 			}
 
@@ -131,6 +131,9 @@ namespace EventStore.Core.TransactionLog.Chunks.TFChunk {
 		}
 
 		private class TFChunkReadSideScavenged : TFChunkReadSide, IChunkReadSide {
+			// must hold _lock to assign to _wantMidpoints and _midpoints
+			private readonly object _lock = new();
+			private bool _wantMidpoints;
 			private Midpoint[] _midpoints;
 			private bool _optimizeCache;
 			private InMemoryBloomFilter _logPositionsBloomFilter;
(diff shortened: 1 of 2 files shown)


sakno:
I have no questions about functional aspects of the proposed changes. From non-functional point of view, there are two improvements possible:

- If lock contention is rare (this is mentioned in the comments inside of the code), the same can be done in lock-free fashion
- `RequestCaching` might start calculation of midpoints as the background task for each chunk (minimal changes are required) and synchronize at its result only on demand.

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunkReadSide.cs
```diff
@@ -147,11 +150,45 @@ public TFChunkReadSideScavenged(TFChunk chunk, bool optimizeCache, ITransactionF
...
 
-			public void Cache() {
-				_midpoints = PopulateMidpoints(Chunk.MidpointsDepth);
+			private Midpoint[] GetOrCreateMidPoints(ReaderWorkItem workItem) {
+				// don't use mipoints when reading from memory (why not?)
+				if (workItem.IsMemory)
+					return null;
+
+				// if we have midpoints we are happy. no synchronization required.
+				// this value may be stale but the midpoints are still be valid
```
shaan1337, line 171:
```suggestion
				// this value may be stale but the midpoints are still valid
```

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunkReadSide.cs
```diff
@@ -147,11 +150,45 @@ public TFChunkReadSideScavenged(TFChunk chunk, bool optimizeCache, ITransactionF
...
+					return midpoints;
+
+				// if we don't want midpoints we are happy. no synchronization required.
+				// this value may be stale but this is rare and worst case we will perform the read
+				// without the midpoints which will still work.
+				if (!_wantMidpoints)
+					return null;
+
+				lock (_lock) {
+					// guaranateed up to date. we dont want to assign to _midpoints if we aren't supposed to
```
shaan1337, line 183:
```suggestion
					// guaranteed up to date. we don't want to assign to _midpoints if we aren't supposed to
```

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunkReadSide.cs
```diff
@@ -147,11 +150,45 @@ public TFChunkReadSideScavenged(TFChunk chunk, bool optimizeCache, ITransactionF
...
+			public void RequestCaching() {
+				lock (_lock) {
+					_wantMidpoints = true;
+				}
 			}
 
-			public void Cache() {
-				_midpoints = PopulateMidpoints(Chunk.MidpointsDepth);
+			private Midpoint[] GetOrCreateMidPoints(ReaderWorkItem workItem) {
+				// don't use mipoints when reading from memory (why not?)
```
shaan1337, line 166: > (why not?)
i assume it's because using a cached scavenged chunk is equivalent to having a midpoint created for every posmap entry in that chunk

shaan1337, line 166:
```suggestion
				// don't use mipoints when reading from memory
```

shaan1337 (commented):
nice catch :+1: changes look good to me, i've just added a couple of minor comments (and 
i also need to do a quick test)

shaan1337:
On my machine, basic performance tests show at least a 10x improvement in the opening speed of scavenged chunks at startup.
Previously, it was taking roughly 40ms, now it takes around 1-4ms !

shaan1337 (approved):

PR #4216, opened 2024-04-02
[DB-744] Implement automated scavenge.

YoEight:
Added: Implement automated scavenge.


YoEight:
Note to myself when I get back to this: I suggest to attach the current epoch id to the scheduled scavenge to properly track the process at the cluster level.

PR #4217, opened 2024-04-03
[ESDB-106-5] Refactoring to support chunk data transformation

shaan1337:
Changed: Refactoring to support chunk data transformation

This PR introduces a data transformation layer to the database. At the moment, only chunk data is transformed but the interfaces have been designed to allow for future expansion (e.g we might also want index data to be transformed later on).

This allows the database to support data transformation operations like encryption, compression, etc. The primary goal of this PR at the moment is to enable the development of an encryption-at-rest plugin.

- chunk headers & footers are not transformed
- only the chunk's data is transformed (scavenge posmaps are also considered as part of the data)
- the transformed data's size does not necessarily need to match the original data's size, it could be larger or smaller.
- the default transform is the `identity` transform which doesn't change anything in the data
- one additional byte is used in the chunk header to store the transform type. this also implies that we can support only up to 256 transforms.
- a new header type, called the transform header is written just after the chunk's header and before the data starts
  - the transform header is specific to the transform and can be of any size
  - the `identity` transform has an empty transform header
  - the transform header is considered to be part of the data, however it is designed to be replicated separately from the data (because the follower needs to know the transform header in advance to be able to create the transform. the created transform is then applied to the data being replicated)
  - at the moment, the transform header is not replicated to the follower, this will be done in a separate PR.
- replication:
  - data chunks are replicated to followers in their untransformed state.
  - raw (scavenged) chunks are replicated in their transformed state.
- caching:
  - the active chunk is cached in its untransformed state in memory (as we would otherwise incur the cost of transformation twice - once when writing to the file stream and once when writing to the memory stream)
  - read-only chunks are cached in their transformed state in memory (as we would otherwise incur the cost of untransforming the whole chunk when loading it)
- checksum:
  - checksum computation is done on transformed data
  - checksum verification is done simply by hashing the whole chunk file except the last 16 bytes (the data doesn't need to be untransformed)
- truncation
  - the position at which a chunk needs to be truncated is transformed before truncating the file
- scavenging
  - scavenged chunks are transformed normally just as non-scavenged chunk. the posmaps are also transformed as they are considered part of the data.

### DIFF
--- a/src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
+++ b/src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
@@ -13,6 +13,8 @@ using DotNext.IO;
 using EventStore.Common.Utils;
 using EventStore.Core.Exceptions;
 using EventStore.Core.TransactionLog.LogRecords;
+using EventStore.Core.Transforms;
+using EventStore.Core.Transforms.Identity;
 using EventStore.Core.Util;
 using Microsoft.Win32.SafeHandles;
 using ILogger = Serilog.ILogger;
@@ -27,6 +29,7 @@ namespace EventStore.Core.TransactionLog.Chunks.TFChunk {
 		}
 
 		public const byte CurrentChunkVersion = 3;
+		private const int AlignmentSize = 4096;
 
 		private static readonly ILogger Log = Serilog.Log.ForContext<TFChunk>();
 
@@ -39,12 +42,12 @@ namespace EventStore.Core.TransactionLog.Chunks.TFChunk {
 			get { return _cacheStatus == CacheStatus.Cached; }
 		}
 
-		// the logical size of data (could be > PhysicalDataSize if scavenged chunk)
+		// the logical size of (untransformed) data (could be > PhysicalDataSize if scavenged chunk)
 		public long LogicalDataSize {
 			get { return Interlocked.Read(ref _logicalDataSize); }
 		}
 
-		// the physical size of data size of data
+		// the physical size of (untransformed) data
 		public int PhysicalDataSize {
 			get { return _physicalDataSize; }
 		}
@@ -63,6 +66,10 @@ namespace EventStore.Core.TransactionLog.Chunks.TFChunk {
 
 		public ChunkFooter ChunkFooter {
 			get { return _chunkFooter; }
+		}
+
+		public ReadOnlyMemory<byte> TransformHeader {
+			get { return _transformHeader; }
 		}
 
 		public readonly int MidpointsDepth;
@@ -107,6 +114,8 @@ namespace EventStore.Core.TransactionLog.Chunks.TFChunk {
 		// potentially causing problems.
 		private readonly object _cachedDataLock = new();
 		private volatile nint _cachedData;
+		// When the chunk is Cached/Uncaching, _cachedDataTransformed indicates whether _cachedData has had the transformation applied
+		private bool _cachedDataTransformed;
 		private int _cachedLength;
 		private volatile CacheStatus _cacheStatus;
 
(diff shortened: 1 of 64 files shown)


src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
```diff
@@ -138,6 +145,10 @@ private enum CacheStatus {
 
 		private IChunkReadSide _readSide;
 
+		private IChunkTransform _transform;
+		private ReadOnlyMemory<byte> _transformHeader;
+		private readonly IChunkTransform _identityTransform = new IdentityChunkTransform();
```
timothycoleman, line 150: we can probably have a static `IdentityChunkReadTransform` and share it among the chunks rather than instantiating the IdentityChunkTransform, IdentityChunkReadTransform and IdentityChunkWriteTransform for each chunk

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
```diff
@@ -833,13 +861,15 @@ private enum CacheStatus {
...
 
-			_chunkFooter = ReadFooter(_writerWorkItem.WorkingStream);
 			IsReadOnly = true;
 
 			_writerWorkItem?.Dispose();
 			_writerWorkItem = null;
 
 			SetAttributes(_filename, true);
+
+			using var stream = _handle.AsUnbufferedStream(FileAccess.Read);
```
timothycoleman, line 871: is this supposed to work for InMem chunks? i _think_ so, but then we can't use _handle
is there a reason we can't use the _writerWorkItem now?

timothycoleman (commented):
this is probably the neatest PR of it's size that i've seen, nice work :+1:

couple of questions attached

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
shaan1337: > is this supposed to work for InMem chunks? i think so, but then we can't use _handle

nice catch! will fix it 

> is there a reason we can't use the _writerWorkItem now?

yes, because `ChunkDataWriteStream` cannot be used for reading (it throws `InvalidOperationException`)

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
shaan1337: great idea :+1:

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
shaan1337: fixed

timothycoleman: ah got it. thanks

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
```diff
@@ -1185,7 +1224,16 @@ static ReaderWorkItem CreateMemoryStreamWorkItem(Stream sharedMemStream, int ind
 
 				Interlocked.Increment(ref _memStreamCount);
 				var stream = new UnmanagedMemoryStream((byte*)_cachedData, _cachedLength);
-				reader = new TFChunkBulkReader(this, stream, isMemory: true);
+
```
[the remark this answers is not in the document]
shaan1337: hmm yup, i think that you're right, i missed that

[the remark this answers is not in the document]
shaan1337: fixed

timothycoleman (approved):

PR #4220, opened 2024-04-06
[DB-9-8] Implement writes, events and bytes metrics.

YoEight:
Added: Implement writes, events and bytes metrics.

A better commit message would be that PR is tracking how many append stream commands, events and bytes the current leader of the cluster is dealing with.

### DIFF
--- a/src/EventStore.Core/MetricsBootstrapper.cs
+++ b/src/EventStore.Core/MetricsBootstrapper.cs
@@ -7,6 +7,8 @@ using EventStore.Core.TransactionLog.Checkpoint;
 using EventStore.Core.Index;
 using EventStore.Core.Messaging;
 using EventStore.Core.Metrics;
+using EventStore.Core.Services.RequestManager;
+using EventStore.Core.Services.Storage;
 using EventStore.Core.Services.VNode;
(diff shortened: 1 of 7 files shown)


cdevarenne:
[DB-747](https://linear.app/eventstore/issue/DB-747)

src/EventStore.Core/MetricsBootstrapper.cs
```diff
@@ -85,6 +88,9 @@ public static class MetricsBootstrapper {
 		var queueBusyMetric = new AverageMetric(coreMeter, "eventstore-queue-busy", "seconds", label => new("queue", label));
 		var byteMetric = new CounterMetric(coreMeter, "eventstore-io", unit: "bytes");
 		var eventMetric = new CounterMetric(coreMeter, "eventstore-io", unit: "events");
+		var writesMetric = new CounterMetric(coreMeter, "eventstore-io", unit: "writes");
+		var writtenBytesMetric = new CounterMetric(coreMeter, "eventstore-io", unit: "bytes");
```
thefringeninja, line 92: is this going to work? it has the same name as two lines above it

PR #4223, opened 2024-04-14
Replace `UnixSignalManager` with `PosixSignalRegistration`

taspeotis:
Hi,

Thanks for EventStoreDB, we use it at work.

When running multiple EventStoreDB containers (single nodes - each of our microservices gets their own independent copy) I have noticed that the CPU usage of the EventStore container hovers between 1-2%, regardless of how much work they have to do.

I ran `dotnet-trace` against an idle container and the most sampled user-code function was `HandleSignals`:

image

I am almost certain this is a false positive because I would expect `WaitAny` to rather efficiently wait for a signal, however the code is written in such a way that it restarts the wait every second. I assume the reason is to incorporate new signals that have been added since the last wait started. Again I am almost certain that doesn't really matter: computers are fast, and doing a quantum of work every 1s is negligible.

The wait-with-1s timeout seems to be a limitation with Mono's API. Microsoft added native support for POSIX signals in .NET 6 that permit adding/removing signal handlers synchronously.

While I think the existing code performs fine, I submit this PR in the interests of (i) saving a wakeup every second and (ii) keeping the code up-to-date with the latest runtime features.


src/EventStore.Core/ClusterVNode.cs
```diff
@@ -1562,12 +1562,11 @@ public class ClusterVNode<TStreamId> :
 			AddTask(perSubscrQueue.Start());
 			AddTask(redactionQueue.Start());
 
-			if (Runtime.IsUnixOrMac) {
```
taspeotis, line 1565: [`PosixSignalRegistration` on Windows emulates`SIGHUP`](https://github.com/dotnet/runtime/blob/2e8f818677e55ef7570b9cd3b21890fa9ac93b0f/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/PosixSignalRegistration.Windows.cs#L87C1-L89C27) so I removed this check.

RagingKore:
Hi @taspeotis thank you for you contribution. 
We have a new PR that was already tackling the issue. Please feel free to provide feedback, but essentially we still keep our UnixSignalManager to allow us to subscribe to multiple signals and for some added dev experience.

taspeotis:
Cool, that would supersede my work. Thanks for letting me know.

RagingKore:
I wanted to try and merge it but the differences were substantial. I will be paying even more attention to your contributions.

PR #4224, opened 2024-04-14
DB-26-18 Make `DoTiming` wait longer if it is able to

taspeotis:
Fixed: Optimize CPU usage of the timer service when the database is idle

Hi,

This is a continuation of my previous tracing ([here](https://github.com/EventStore/EventStore/pull/4223)) in an effort to make EventStoreDB at idle use as little CPU as possible.

`ThreadBasedScheduler.DoTiming` is responsible for a lot of CPU usage when the database is not being used.

image

It rate limits itself with a call to `Thread.Sleep(10)` which means it is executing quite a bit, even if there are no scheduled tasks to be dispatched.

I have added a calculation for the next due scheduled task, and to delay until then, but no more than 30 seconds. The delay is interrupted by any scheduled task being enqueued, because the calculation may now be wrong.
 
Note that originally this loop was rate limited somewhat by a `Thread.Sleep(1)` and the method (among other things) was changed to use `Thread.Sleep(10)` to reduce idle CPU. You can see the history here: 95be10a0e40c85e1a32fbb760abbaea3ca4f7949.

Any feedback is welcome.

### DIFF
--- a/src/EventStore.Core/Services/TimerService/ThreadBasedScheduler.cs
+++ b/src/EventStore.Core/Services/TimerService/ThreadBasedScheduler.cs
@@ -10,25 +10,27 @@ using EventStore.Core.Metrics;
 using Serilog;
 
 namespace EventStore.Core.Services.TimerService {
-	public class ThreadBasedScheduler : IMonitoredQueue, IScheduler, IDisposable {
+	public sealed class ThreadBasedScheduler : IMonitoredQueue, IScheduler {
 		private static readonly ILogger Log = Serilog.Log.ForContext<ThreadBasedScheduler>();
 		public string Name {
 			get { return _queueStats.Name; }
 		}
 
 		private readonly ConcurrentQueueWrapper<ScheduledTask> _pending = new ConcurrentQueueWrapper<ScheduledTask>();
+		private readonly ManualResetEventSlim _pendingEvent = new ManualResetEventSlim(false);
 
 		private readonly PairingHeap<ScheduledTask> _tasks =
 			new PairingHeap<ScheduledTask>((x, y) => x.DueTime < y.DueTime);
 
 		private readonly IClock _timeProvider;
 
-		private readonly Thread _timerThread;
 		private volatile bool _stop;
 
 		private readonly QueueStatsCollector _queueStats;
 		private readonly QueueTracker _tracker;
 		private readonly TaskCompletionSource<object> _tcs = new TaskCompletionSource<object>();
+
+		private long _nextWakeupTimeTicks = long.MinValue;
 
 		public Task Task {
 			get { return _tcs.Task; }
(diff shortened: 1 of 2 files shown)


src/EventStore.Core/Services/TimerService/ThreadBasedScheduler.cs
```diff
@@ -90,9 +89,23 @@ public class ThreadBasedScheduler : IMonitoredQueue, IScheduler, IDisposable {
...
 					if (processed == 0) {
-						_queueStats.EnterIdle();
+						var timeout = TimeSpan.FromSeconds(30);
+
+						if (_tasks.Count > 0) {
+							var ticksUntilNextDueTime = _tasks.FindMin().DueTime.ElapsedTicksSince(_timeProvider.Now);
+							var secondsUntilNextDueTime = (double)ticksUntilNextDueTime / Instant.TicksPerSecond;
+
+							if (secondsUntilNextDueTime <= 0)
+								continue;
```
shaan1337, line 99:
```suggestion
							var secondsUntilNextDueTime = (double)ticksUntilNextDueTime / Instant.TicksPerSecond;

							if (secondsUntilNextDueTime <= 0)
								continue;
```

@taspeotis it seems possible to get rid of the `secondsUntilNextDueTime` variable and do the `<= 0` check with `ticksUntilNextDueTime` itself. `dueTimeSpan` can then be built with `TimeSpan.FromTicks()` instead of `TimeSpan.FromSeconds`

shaan1337: this would save us from doing any division operation / any approximation

shaan1337: sorry, i missed the fact that we're working with `Instant`s here not `TimeSpan`s, so we won't be able to use `TimeSpan.FromTicks`.  Maybe we could try to add an operation in `Instant` to convert it to a `TimeSpan`

taspeotis: Hi, please give me a day or two to play with it. Thank you for the review and feedback!

Also if you have the time to do it, I think I enabled "contributors can make changes" in GitHub. I've never used it before but I think you can push to my branch.

shaan1337 (commented):
@taspeotis thanks for your contribution ! this looks like a great optimization, 
i've added a few comments

src/EventStore.Core/Services/TimerService/ThreadBasedScheduler.cs
shaan1337: thanks for the quick response @taspeotis yup sure, i'll give it a try, it shouldn't be a big change :+1: 
by any luck, do you also have some numbers/profiler runs after the fix?

shaan1337: i've pushed a commit with what i had in mind, feel free to let me know if you find the change appropriate (i've not yet thoroughly tested it though)

i've also adjusted the max. timeout to 5 seconds instead of 30 seconds as in practice, it looks like the actual timeout is of the order of a few hundred milliseconds (and also just in case there's any bug, we know that the task will be processed within 5 seconds max)

taspeotis: Thanks @shaan1337 I'll try and benchmark this shortly

shaan1337: thanks! feel free to push any additional commits if required

shaan1337 (approved):

PR #4231, opened 2024-04-22
[DB-26-8] Cope with multiple nodes having InstanceId = Guid.Empty when sending telemetry

timothycoleman:
Fixed: Error sending usage telemetry in certain circumstances

### DIFF
--- a/src/EventStore.Core/Telemetry/TelemetryService.cs
+++ b/src/EventStore.Core/Telemetry/TelemetryService.cs
@@ -178,7 +178,11 @@ public sealed class TelemetryService : IDisposable,
 			message.Envelope.ReplyWith(new TelemetryMessage.Response("plugins", key, value));
 
 		foreach (var subsystem in _nodeOptions.PlugableComponents) {
-			subsystem.CollectTelemetry(collect);
+			try {
+				subsystem.CollectTelemetry(collect);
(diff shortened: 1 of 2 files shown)


src/EventStore.Core/Telemetry/TelemetryService.cs
```diff
@@ -178,7 +178,11 @@ public sealed class TelemetryService : IDisposable,
 			message.Envelope.ReplyWith(new TelemetryMessage.Response("plugins", key, value));
 
 		foreach (var subsystem in _nodeOptions.PlugableComponents) {
-			subsystem.CollectTelemetry(collect);
+			try {
+				subsystem.CollectTelemetry(collect);
+			} catch (Exception ex) {
+				_log.Error(ex, "Failed to collect telemetry from a plugable component");
```
thefringeninja, line 184: Do we want to say which component failed? Also, is this an error or a warning?

thefringeninja (changes requested):

src/EventStore.Core/Telemetry/TelemetryService.cs
timothycoleman: fair questions, i've adjusted the code

thefringeninja (approved):

PR #4247, opened 2024-04-26
[DB-26-12] Identify nodes with advertised addresses for Scavenge log

lakshdeepsingheventstore:
Fixed : Use the advertised addresses for identifying nodes in the scavenge log.

https://eventstore.aha.io/develop/requirements/DB-26-12 . Linear issue : https://linear.app/eventstore/issue/DB-706/scavenge-identify-nodes-with-advertised-addresses

### DIFF
--- a/src/EventStore.Core/ClusterVNode.cs
+++ b/src/EventStore.Core/ClusterVNode.cs
@@ -1390,7 +1390,7 @@ namespace EventStore.Core {
 			}
 
 			var scavengerLogManager = new TFChunkScavengerLogManager(
-				nodeEndpoint: _options.Interface.ReplicationHostAdvertiseAs == null ? NodeInfo.HttpEndPoint.ToString() : (_options.Interface.ReplicationHostAdvertiseAs + ":" + NodeInfo.HttpEndPoint.GetPort()),
+				nodeEndpoint: $"{GossipAdvertiseInfo.HttpEndPoint.Host}:{GossipAdvertiseInfo.HttpEndPoint.Port}",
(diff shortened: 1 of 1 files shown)


src/EventStore.Core/ClusterVNode.cs
```diff
@@ -1390,7 +1390,7 @@ public class ClusterVNode<TStreamId> :
 			}
 
 			var scavengerLogManager = new TFChunkScavengerLogManager(
-				nodeEndpoint: NodeInfo.HttpEndPoint.ToString(),
+				nodeEndpoint: _options.Interface.ReplicationHostAdvertiseAs == null ? NodeInfo.HttpEndPoint.ToString() : (_options.Interface.ReplicationHostAdvertiseAs + ":" + NodeInfo.HttpEndPoint.GetPort()),
```
hayley-jean, line 1393: You should use the `GossipAdvertiseInfo` here rather than the `NodeInfo`, since that will already have the best host and port according to configuration:

```suggestion
				nodeEndpoint: $"{GossipAdvertiseInfo.HttpEndPoint.Host}:{GossipAdvertiseInfo.HttpEndPoint.Port}",
```
Don't use `GossipAdvertiseInfo.HttpEndPoint.ToString()` directly though, because that will try and add the address family and format it as `Unspecified/{hostname}:{port}`

hayley-jean (changes requested):

src/EventStore.Core/ClusterVNode.cs
lakshdeepsingheventstore: Thanks for the review, changes done.

hayley-jean (approved):

PR #4249, opened 2024-05-03
[DB-26-16] Log an error when node certificate is not a valid server certificate

hayley-jean:
Added: Log an error when the node certificate is not a valid server certificate

Update the documentation to include the stricter certificate requirements.

If you provide a node certificate with invalid key usages (e.g. missing a required key usage), the server will now log an error:

```
[24448,26,12:47:42.882,ERR] Connection from node: "127.0.0.1" was denied because it is not configured as a valid server certificate: "Missing key usage: Digital Signature"
```

The error is only logged if the provided certificate matches the reserved node common name, and has IP or DNS addresses in the SAN.

### DIFF
--- a/src/EventStore.Core/Services/Transport/Http/Authentication/NodeCertificateAuthenticationProvider.cs
+++ b/src/EventStore.Core/Services/Transport/Http/Authentication/NodeCertificateAuthenticationProvider.cs
@@ -77,8 +77,8 @@ namespace EventStore.Core.Services.Transport.Http.Authentication {
 		}
 
 		private bool AuthenticateUncached(HttpContext context, X509Certificate2 clientCertificate) {
-			if (!clientCertificate.IsServerCertificate(out _))
-				return false;
+			var ip = context.Connection.RemoteIpAddress?.ToString() ?? "<unknown>";
+			var isServerCertificate = clientCertificate.IsServerCertificate(out var serverCertReason);
 
 			var reservedNodeCN = _getCertificateReservedNodeCommonName();
 			bool hasReservedNodeCN;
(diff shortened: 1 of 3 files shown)


src/EventStore.Core/Services/Transport/Http/Authentication/NodeCertificateAuthenticationProvider.cs
```diff
@@ -90,24 +90,32 @@ public class NodeCertificateAuthenticationProvider : IHttpAuthenticationProvider
 				return false;
 			}
 
+			bool hasIpOrDnsSan = clientCertificate.GetSubjectAlternativeNames()
+				.Where(x => x.type is CertificateNameType.DnsName or CertificateNameType.IpAddress)
+				.IsNotEmpty();
+
+			if (!isServerCertificate && !hasReservedNodeCN && !hasIpOrDnsSan) {
+				// We are sure that this is a client certificate,
+				// Not a misconfigured server certificate with incorrect EKUs
```
shaan1337, line 99:
```suggestion
				// We are sure that this is not a misconfigured node certificate with incorrect EKUs, missing SANs, etc. It could be a user certificate.
```

src/EventStore.Core/Services/Transport/Http/Authentication/NodeCertificateAuthenticationProvider.cs
```diff
@@ -90,24 +90,32 @@ public class NodeCertificateAuthenticationProvider : IHttpAuthenticationProvider
...
 				return false;
 			}
+			if (!isServerCertificate) {
+				Log.Error(
+					"Connection from node: {ip} was denied because it is not configured as a server certificate: {failReason}",
+					ip, serverCertReason);
+				return false;
+			}
 
 			return true;
```
shaan1337, line 120:
```suggestion
			return hasReservedNodeCN && hasIpOrDnsSan && isServerCertificate;
```

suggestion: to make it even more user-friendly, we could remove the `return false`s in the above 3 checks but still log the errors. In that way, a user would immediately see all the problems with the node certificate they've generated (instead of having to fix one problem at a time by generating a new certificate and trying again)

shaan1337 (approved):

PR #4251, opened 2024-05-07
[DBX-31] Fix events missing from $all

### LINKED ISSUE #4250: Events missing when from $all but present in stream
**Description**
Events seem to be present when reading from a specific stream but absent when reading from $all. This happens with events which were written using transactions on the TCP client.

**To Reproduce**
See support ticket: [111110](https://eventstore.freshdesk.com/a/tickets/111110)

**Expected behavior**
Events should be present both in $all and in the stream.

**Actual behavior**
Events absent from $all but present in the stream.

**EventStore details**
 - EventStore server version: 23.10

timothycoleman:
Fixed: Events written in explicit transactions via TCP can be missing from $all reads/subscriptions

Fixes #4250

- The bug only affects events written in explicit transactions (which are deprecated and can only be written via the old TCP client)
- The bug appears to affect Filtered all reads when the window is bigger than the maxcount
- The bug can affect regular all reads/subscriptions


thefringeninja (approved):

shaan1337 (approved):
nice catch!

PR #4256, opened 2024-05-08
[DB-26-4] Fix leaderId and epoch number returning empty in telemetry for read only replica and follower node that joins later

lakshdeepsingheventstore:
Fixed : LeaderId and Epoch number would sometimes be empty (Guid.empty) for a follower node or a read only replica that would join a cluster whose leader is already elected. https://eventstore.aha.io/develop/requirements/DB-26-4 . [Linear Issue](https://linear.app/eventstore/issue/DB-611/leaderid-sometimes-guidempty-in-telemetry)


YoEight (changes requested):
LGTM, it only misses some tests.

YoEight (approved):

YoEight (approved):

PR #4257, opened 2024-05-13
[CNN-14] Ensure ESDB can run on OSX ARM64

RagingKore:
Added: new UnixSignalManager using PosixSignalRegistration
Added: EventStore.SystemRuntime project
Removed: EventStore.Common.Utils project and other zombie code
Changed: Runtime and Process stats such as CPU Usage and Disk IO

## Background and Motivation

Currently we are unable to run ESDB on OSX ARM64 because it breaks on 2 different operations:

- _UnixSignalManager_
- Any metrics using [HostStat.NET](https://github.com/EventStore/HostStat.NET) (that depends on [libhoststat](https://github.com/EventStore/libhoststat))

## Solution

- New Implementation of the _UnixSignalManager_ using the [PosixSignalRegistration](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.posixsignalregistration?view=net-8.0)
- New implementation of the breaking metrics removing any dependency on `HostStat.NET`
- Removal of leftover code.

### Bonus
We now have better metrics coverage as the following metrics are available for the indicated platforms:
- `CPU Usage`: Linux*, FreeBSD*, OSX*, Windows
- `CPU Load Averages`: Linux, FreeBSD, OSX
- `Disk IO Stats`: Linux, Windows and OSX* (read bytes and written bytes only)

### Side-quests

The identification of the runtime platform and related operations do not have a single source of truth. 
- We use a [Runtime](https://github.com/EventStore/EventStore/blob/master/src/EventStore.Common.Utils/Runtime.cs) static class that is the single file in the [EventStore.Common.Utils](https://github.com/EventStore/EventStore/tree/master/src/EventStore.Common.Utils) project.
- From the [EventStore.Common](https://github.com/EventStore/EventStore/tree/master/src/EventStore.Common) project we use the [OS](https://github.com/EventStore/EventStore/blob/master/src/EventStore.Common/Utils/OS.cs) static class, the _OsFlavor_  enum and the _Platforms.GetPlatform_
- The _OperatingSystem_ class from System and even the _RuntimeInformation_

Given that, a new `RuntimeInformation` static class providing all the runtime and OS information that was scattered previously, was introduced and it is located in a new Project `EventStore.SystemRuntime` that will also contain all the new diagnostic information to fix the broken metrics.

``` csharp
namespace EventStore.Common.Utils {
	public static class Runtime {
		public static readonly bool IsUnixOrMac = RuntimeInformation.IsOSPlatform(OSPlatform.Linux) 
			| RuntimeInformation.IsOSPlatform(OSPlatform.OSX);

		public static readonly bool IsWindows = !IsUnixOrMac;

		public static readonly bool IsMacOS = RuntimeInformation.IsOSPlatform(OSPlatform.OSX);
	}
}
```

``` csharp
namespace EventStore.Common.Utils {
	public enum OsFlavor { Unknown, Windows, Linux, BSD, MacOS }

	public class OS {
		public static bool IsUnix {
			get {
				//...
			}
		}

		public static string GetHomeFolder() {
			// ...
		}

		public static string GetRuntimeVersion() {
			// ...
		}
        }
}
```

``` csharp
namespace EventStore.Common.Utils {
	public class Platforms {
		public static Platform GetPlatform() {
			// ...
		}
	}

	public enum Platform { Windows, Linux, Mac }
}
```

#### Notes

- The [HostStat.NET](https://github.com/EventStore/HostStat.NET) and [libhoststat](https://github.com/EventStore/libhoststat) repositories can be archived.

#### Related PR's

resolves #4223


YoEight (approved):
LGTM, we should be mindful of potential changes on metrics collection like names and values in case something escaped our review.

PR #4258, opened 2024-05-14
[ESDB-106-14] Add support for chunk data transformation plugins

shaan1337:
Added: Support for chunk data transformation plugins. Breaking Change: The database restarts a second time to complete a truncation operation.

This PR introduces a behavioral change - please see the comment in the code for more details:
- An additional shutdown is now done after truncation of the database


timothycoleman:
I've rebased it now that master has upgraded EventStore.Plugins reference that this requires

timothycoleman (approved):

PR #4263, opened 2024-05-21
Allow an extra chunk at startup

shaan1337:
Changed: Allow an extra chunk at startup

It turns out that the change reverted by EventStore/EventStore#3954 is actually necessary because `CompleteChunkInTransaction()` and `CompleteReplicatedRawChunkInTransaction()` create a new chunk _before_ the writer checkpoint is updated & flushed to disk (in `CommitTransaction()`).

Thus, at startup, it's possible that there is an extraneous chunk - the writer checkpoint would still be pointing to the chunk before it.


timothycoleman (approved):
(this change was in master before anyway)

PR #4268, opened 2024-05-23
[DB-26-6] Configurable status code for /heath/live

### LINKED ISSUE #3468: HTTP 204 returned from /health/live endpoint is unsuitable for Google Kubernetes Cloud healthcheck
**Describe the bug**
Google Cloud does not allow you to configure the behavior of its health checks, other than the endpoint, it requires a 200 response. This means that the "/health/live" endpoint, which seems intended for this sort of use case, can't be used on this cloud provider. The return code should be changed to 200, either by default, or by adding a param to the endpoint.

**Expected behavior**

```
curl -i  --fail --insecure http://10.0.0.75:2113/health/live
HTTP/1.1 200 OK
```

**Actual behavior**

```
curl -i  --fail --insecure http://10.0.0.75:2113/health/live
HTTP/1.1 204 No Content
```

<sub>[DB-299](https://linear.app/eventstore/issue/DB-299/http-204-returned-from-healthlive-endpoint-is-unsuitable-for-google)</sub>

lakshdeepsingheventstore:
Changed: Allow statusCode for live to be provided in query string: `/health/live?liveCode=200`

https://eventstore.aha.io/develop/requirements/DB-26-6
Closes #3468

### DIFF
--- a/src/EventStore.Core/ClusterVNodeStartup.cs
+++ b/src/EventStore.Core/ClusterVNodeStartup.cs
@@ -264,6 +264,7 @@ namespace EventStore.Core {
 
 		private class StatusCheck {
 			private readonly ClusterVNodeStartup<TStreamId> _startup;
+			private readonly int _livecode = 204;
 
 			public StatusCheck(ClusterVNodeStartup<TStreamId> startup) {
 				if (startup == null) {
@@ -279,7 +280,16 @@ namespace EventStore.Core {
 						.MapMiddlewareGet("live", inner => inner.Use(Live)));
 
 			private MidFunc Live => (context, next) => {
-				context.Response.StatusCode = _startup._ready ? 204 : 503;
+				if (_startup._ready) {
+					if (context.Request.Query.TryGetValue("liveCode", out var expected) &&
+						int.TryParse(expected, out var statusCode)) {
+						context.Response.StatusCode = statusCode;
+					} else {
+						context.Response.StatusCode = _livecode;
+					}
+				} else {
+					context.Response.StatusCode = 503;
+				}
 				return Task.CompletedTask;
 			};
 
--- a/src/EventStore.Core.Tests/Http/HealthChecks/when_performing_a_live_check.cs
+++ b/src/EventStore.Core.Tests/Http/HealthChecks/when_performing_a_live_check.cs
@@ -59,6 +59,18 @@ namespace EventStore.Core.Tests.Http.HealthChecks {
 		}
 
 		[TestCaseSource(nameof(MethodAllowedTestCases))]
+		public async Task with_liveCode_parameter_returns_the_same_liveCode(HttpMethod method) {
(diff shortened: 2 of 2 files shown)


hayley-jean (changes requested):
Could you implement this in a non-breaking way, as suggested [here](https://github.com/EventStore/EventStore/issues/3468#issuecomment-1701437283).

Essentially, have the live endpoint accept a parameter that specifies the http code that the requester expects in response. e.g. `https://localhost:2113/health/live?liveCode=200`.

src/EventStore.Core/ClusterVNodeStartup.cs
```diff
@@ -279,7 +281,14 @@ private class StatusCheck {
 						.MapMiddlewareGet("live", inner => inner.Use(Live)));
 
 			private MidFunc Live => (context, next) => {
-				context.Response.StatusCode = _startup._ready ? 204 : 503;
+				if (_startup._ready) {
+					if (context.Request.Query.ContainsKey("liveCode"))
+						context.Response.StatusCode =
+							Int32.Parse(context.Request.Query.ToDictionary().GetValueOrDefault("liveCode")[0]!);
```
hayley-jean, line 287: This needs error handling for if the user sends something that doesn't parse to an int. You can use `int.TryParse` to do this in a cleaner way. Similarly, you can use `TryGetValue` above to get the expected code from the dictionary.
For example:

```
if (context.Request.Query.TryGetValue("liveCode", out var expected)
    && int.TryParse(expected, out var statusCode)) {
        context.Response.StatusCode = statusCode;
}
```

src/EventStore.Core/ClusterVNodeStartup.cs
```diff
@@ -264,13 +264,15 @@ public class ClusterVNodeStartup<TStreamId> : IStartup, IHandle<SystemMessage.Sy
 
 		private class StatusCheck {
 			private readonly ClusterVNodeStartup<TStreamId> _startup;
+			private readonly int _livecode;
```
hayley-jean, line 267: Minor thing, but you can instantiate this here
```suggestion
			private readonly int _liveCode = 204;
```

hayley-jean (changes requested):

src/EventStore.Core/ClusterVNodeStartup.cs
lakshdeepsingheventstore: Done

src/EventStore.Core/ClusterVNodeStartup.cs
lakshdeepsingheventstore: Done

timothycoleman (approved):
Thanks, looks good :+1:

I've
- neatened up the formatting a bit (added extra braces so its consistent)
- PR adjusted the changelog entry and title to reflect the non-breaking implementation

PR #4270, opened 2024-05-26
Support nix shell environment.

YoEight:
This pull request introduces a Nix shell script to set up a dev environment for EventStoreDB. The primary goal of this addition is to simplify the setup process for devs, particularly those using the Nix package manager. It's important to note that this inclusion does not signal a commitment to Nix as our primary package manager. Instead, it serves as an optional tool to ease the lives of devs who prefer or already use the Nix package manager like myself.


timothycoleman:
Spring cleaning

PR #4276, opened 2024-05-30
[DB-26-19] Warn when projection state becomes greater than 8mb

lakshdeepsingheventstore:
Added : A warning to the log when a projection state size becomes greater than 8 MB. https://eventstore.aha.io/develop/requirements/DB-26-19

### DIFF
--- a/src/EventStore.Projections.Core/Services/Processing/CoreProjectionCheckpointManager.cs
+++ b/src/EventStore.Projections.Core/Services/Processing/CoreProjectionCheckpointManager.cs
@@ -2,6 +2,7 @@ using System;
 using System.Diagnostics.Contracts;
 using EventStore.Core.Bus;
 using EventStore.Projections.Core.Messages;
+using Serilog;
 using ILogger = Serilog.ILogger;
 
 namespace EventStore.Projections.Core.Services.Processing {
@@ -34,6 +35,7 @@ namespace EventStore.Projections.Core.Services.Processing {
 		protected bool _stopped;
 
 		private PartitionState _currentProjectionState;
+		private bool _largeStateWarningLogged = false;
 
 		protected CoreProjectionCheckpointManager(
 			IPublisher publisher,
@@ -166,15 +168,33 @@ namespace EventStore.Projections.Core.Services.Processing {
 			if (_stopping)
 				throw new InvalidOperationException("Stopping");
 
+			if (partition == "" && newState.State == null) // ignore non-root partitions and non-changed states
+				throw new NotSupportedException("Internal check");
+
+			CheckStateSize(newState , partition);
 
 			if (_usePersistentCheckpoints && partition != "")
 				CapturePartitionStateUpdated(partition, oldState, newState);
 
-			if (partition == "" && newState.State == null) // ignore non-root partitions and non-changed states
-				throw new NotSupportedException("Internal check");
-
 			if (partition == "")
 				_currentProjectionState = newState;
+		}
+
+		private void CheckStateSize(PartitionState result , string partition) {
+			if (!_largeStateWarningLogged && result.Size >= 8_000_000 && !partition.Equals("")) {
+				Log.Warning(
+					"State size for the Projection {projectionName} for Partition {partitionName} is greater than 8 MB. State size for a projection must be less than 16 MB. Current state size for Partition {partitionName} is {stateSize} MB.",
+					_namingBuilder.EffectiveProjectionName, partition, partition,
+					result.Size / Math.Pow(10, 6));
+				_largeStateWarningLogged = true;
+			}
+			else if (!_largeStateWarningLogged && result.Size >= 8_000_000 && partition.Equals("") ) {
+				Log.Warning(
+					"State size for the Projection {projectionName} is greater than 8 MB. State size for a projection must be less than 16 MB. Current state size for Projection {projectionName} is {stateSize} MB.",
+					_namingBuilder.EffectiveProjectionName, _namingBuilder.EffectiveProjectionName,
+					result.Size / Math.Pow(10, 6));
+				_largeStateWarningLogged = true;
+			}
 		}
 
 		public void EventProcessed(CheckpointTag checkpointTag, float progress) {
--- a/src/EventStore.Projections.Core/Services/Processing/PartitionStateUpdateManager.cs
+++ b/src/EventStore.Projections.Core/Services/Processing/PartitionStateUpdateManager.cs
@@ -46,7 +46,7 @@ namespace EventStore.Projections.Core.Services.Processing {
 				}
 
 				//NOTE: order yb is required to satisfy internal emit events validation
-				// which ensures that events are ordered by causedBy tag.  
+				// which ensures that events are ordered by causedBy tag.
 				// it is too strong check, but ...
 				eventWriter.ValidateOrderAndEmitEvents(list.OrderBy(v => v.Event.CausedByTag).ToArray());
 			}
--- a/src/EventStore.Projections.Core/Services/Processing/CoreProjectionCheckpointWriter.cs
+++ b/src/EventStore.Projections.Core/Services/Processing/CoreProjectionCheckpointWriter.cs
@@ -7,6 +7,7 @@ using EventStore.Core.Messaging;
 using EventStore.Core.Services;
 using EventStore.Core.Services.UserManagement;
 using EventStore.Projections.Core.Messages;
+using Serilog;
 using ILogger = Serilog.ILogger;
 
 namespace EventStore.Projections.Core.Services.Processing {
@@ -27,6 +28,7 @@ namespace EventStore.Projections.Core.Services.Processing {
 		private const int MinAttemptWarnThreshold = 5;
 		private bool _metaStreamWritten;
 		private Random _random = new Random();
+		private bool _largeCheckpointStateWarningLogged = false;
 
 		public CoreProjectionCheckpointWriter(
 			string projectionCheckpointStreamId, IODispatcher ioDispatcher, ProjectionVersion projectionVersion,
@@ -161,10 +163,21 @@ namespace EventStore.Projections.Core.Services.Processing {
 		}
 
 		private void PublishWriteCheckpointEvent() {
+			CheckpointStateSizeCheck();
 			_writeRequestId = _ioDispatcher.WriteEvent(
 				_projectionCheckpointStreamId, _lastWrittenCheckpointEventNumber, _checkpointEventToBePublished,
 				SystemAccounts.System,
 				msg => WriteCheckpointEventCompleted(_projectionCheckpointStreamId, msg.Result, msg.FirstEventNumber));
+		}
+
+		private void CheckpointStateSizeCheck() {
+			if (!_largeCheckpointStateWarningLogged && _checkpointEventToBePublished.Data.Length >= 8_000_000) {
+				Log.Warning(
+					"State size for the Projection {projectionName} is greater than 8 MB. State size for a projection should be less than 16 MB. Current state size for Projection {projectionName} is {stateSize} MB.",
+					_name, _name,
+					_checkpointEventToBePublished.Data.Length / Math.Pow(10, 6));
+				_largeCheckpointStateWarningLogged = true;
+			}
 		}
 
 		public void Initialize() {
--- a/src/EventStore.Projections.Core/Services/Processing/PartitionState.cs
+++ b/src/EventStore.Projections.Core/Services/Processing/PartitionState.cs
@@ -10,7 +10,7 @@ namespace EventStore.Projections.Core.Services.Processing {
 		private static readonly JsonSerializerSettings JsonSettings = new JsonSerializerSettings {
 			DateParseHandling = DateParseHandling.None,
 		};
-		
+
 		public bool IsChanged(PartitionState newState) {
 			return State != newState.State || Result != newState.Result;
 		}
@@ -48,6 +48,7 @@ namespace EventStore.Projections.Core.Services.Processing {
 		private readonly string _state;
 		private readonly string _result;
 		private readonly CheckpointTag _causedBy;
+		private int _size;
 
 		public PartitionState(string state, string result, CheckpointTag causedBy) {
 			if (state == null) throw new ArgumentNullException("state");
@@ -56,6 +57,7 @@ namespace EventStore.Projections.Core.Services.Processing {
 			_state = state;
 			_result = result;
 			_causedBy = causedBy;
+			_size = _state.Length + _result?.Length ?? 0;
 		}
 
 		public string State {
@@ -68,6 +70,10 @@ namespace EventStore.Projections.Core.Services.Processing {
 
 		public string Result {
 			get { return _result; }
+		}
+
+		public int Size {
+			get { return _size; }
 		}
 
 		public string Serialize() {


src/EventStore.Projections.Core/Services/Processing/ResultEventEmitter.cs
```diff
@@ -20,21 +23,31 @@ public class ResultEventEmitter : IResultEventEmitter {
...
 								streamId, Guid.NewGuid(), "Result", true, projectionResult, null, at, null),
 						_resultStreamMetadata);
+				if (!string.IsNullOrEmpty(result.Event.Data) && !warningProjectionList.Contains(_namesBuilder.EffectiveProjectionName)) {
+					if (System.Text.ASCIIEncoding.Unicode.GetByteCount(result.Event.Data) / Math.Pow(10, 6) >= 8) {
+						Log.Warning(
+							"State size for the Projection {p} is greater than 8 MB. State size for a projection should be less than 16 MB. Current state size for Projection {p} is {s} MB",
+							_namesBuilder.EffectiveProjectionName, _namesBuilder.EffectiveProjectionName,
+							System.Text.ASCIIEncoding.Unicode.GetByteCount(result.Event.Data) / Math.Pow(10, 6));
+					}
+					warningProjectionList.Add(_namesBuilder.EffectiveProjectionName);
```
timothycoleman, line 43: isn't this going to add the projection to the list even if the state is small, and then when the state is big it won't log anything?

src/EventStore.Projections.Core/Services/Processing/ResultEventEmitter.cs
```diff
@@ -20,21 +23,31 @@ public class ResultEventEmitter : IResultEventEmitter {
...
-				var result =
+				result =
 					new EmittedEventEnvelope(
 						projectionResult == null
 							? new EmittedDataEvent(
 								streamId, Guid.NewGuid(), "ResultRemoved", true, null, null, at, null)
 							: new EmittedDataEvent(
 								streamId, Guid.NewGuid(), "Result", true, projectionResult, null, at, null),
 						_resultStreamMetadata);
+				if (!string.IsNullOrEmpty(result.Event.Data) && !warningProjectionList.Contains(_namesBuilder.EffectiveProjectionName)) {
```
timothycoleman, line 36: this will do a linear search through the list every time a projection emits

timothycoleman (changes requested):
- in the pr description, this isn't a bug so we don't need to put fixed. we can use `Added:`
- im not sure, but does this apply to projection state specifically or all emitted projection events?

src/EventStore.Projections.Core/Services/Processing/ResultEventEmitter.cs
```diff
@@ -20,21 +23,31 @@ public class ResultEventEmitter : IResultEventEmitter {
 			CheckpointTag at) {
 			var streamId = _namesBuilder.MakePartitionResultStreamName(partition);
 			var allResultsStreamId = _namesBuilder.GetResultStreamName();
+			EmittedEventEnvelope result;
```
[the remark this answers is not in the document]
lakshdeepsingheventstore: Reverted back to previous format. It was required as per some of the changes which I had done locally to test this issue, but later was not needed.

src/EventStore.Projections.Core/Services/Processing/ResultEventEmitter.cs
```diff
@@ -43,6 +56,16 @@ public class ResultEventEmitter : IResultEventEmitter {
...
+				if (!string.IsNullOrEmpty(result.Event.Data) && !warningProjectionList.Contains(_namesBuilder.EffectiveProjectionName)) {
+					if (System.Text.ASCIIEncoding.Unicode.GetByteCount(result.Event.Data) / Math.Pow(10, 6) >= 8) {
+						Log.Warning(
+							"State size for the Projection {p} is greater than 8 MB. State size for a projection should be less than 16 MB. Current state size for Projection {p} is {s} MB",
+							_namesBuilder.EffectiveProjectionName, _namesBuilder.EffectiveProjectionName,
+							System.Text.ASCIIEncoding.Unicode.GetByteCount(result.Event.Data) / Math.Pow(10, 6));
+					}
+					warningProjectionList.Add(_namesBuilder.EffectiveProjectionName);
+				}
+
```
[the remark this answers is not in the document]
lakshdeepsingheventstore: Done

src/EventStore.Projections.Core/Services/Processing/ResultEventEmitter.cs
```diff
@@ -20,21 +23,31 @@ public class ResultEventEmitter : IResultEventEmitter {
...
+				result =
 					new EmittedEventEnvelope(
 						projectionResult == null
 							? new EmittedDataEvent(
 								streamId, Guid.NewGuid(), "ResultRemoved", true, null, null, at, null)
 							: new EmittedDataEvent(
 								streamId, Guid.NewGuid(), "Result", true, projectionResult, null, at, null),
 						_resultStreamMetadata);
+				if (!string.IsNullOrEmpty(result.Event.Data) && !warningProjectionList.Contains(_namesBuilder.EffectiveProjectionName)) {
+					if (System.Text.ASCIIEncoding.Unicode.GetByteCount(result.Event.Data) / Math.Pow(10, 6) >= 8) {
```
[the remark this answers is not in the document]
lakshdeepsingheventstore: Done

src/EventStore.Projections.Core/Services/Processing/ResultEventEmitter.cs
```diff
@@ -1,8 +1,11 @@
 using System;
+using System.Collections.Generic;
+using Serilog;
 
 namespace EventStore.Projections.Core.Services.Processing {
 	public class ResultEventEmitter : IResultEventEmitter {
 		private readonly ProjectionNamesBuilder _namesBuilder;
+		private static List<string> warningProjectionList = new List<string>();
```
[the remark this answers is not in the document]
lakshdeepsingheventstore: It could be a possible case. Changed the variable to a non static one.

src/EventStore.Projections.Core/Services/Processing/ResultEventEmitter.cs
lakshdeepsingheventstore: Yeah it was a mistake on my end, rectified.

src/EventStore.Projections.Core/Services/Processing/ResultEventEmitter.cs
lakshdeepsingheventstore: I have pushed new changes as per which the list would only be searched when an event of event type "Result" is emitted, and not for any other types of emitted events.

src/EventStore.Projections.Core/Services/Processing/ResultEventEmitter.cs
```diff
@@ -1,8 +1,11 @@
 using System;
+using System.Collections.Generic;
+using Serilog;
 
 namespace EventStore.Projections.Core.Services.Processing {
 	public class ResultEventEmitter : IResultEventEmitter {
 		private readonly ProjectionNamesBuilder _namesBuilder;
+		private List<string> WarningProjectionList = new List<string>();
```
timothycoleman, line 8: for non-static member vars it should be _warningProjectionList

is it possible for multiple different names to be added to the list?

timothycoleman (changes requested):
^

src/EventStore.Projections.Core/Services/Processing/ResultEventEmitter.cs
lakshdeepsingheventstore: Based on the investigation and tests so far, it looks like each projection has a different ResultEventEmitter object associated with it. So in the latest changes, I have got ride of the List variable and instead introduced a bool variable to the ResultEventEmitter Class to keep track if the projection's state size is less than 8 MB

src/EventStore.Projections.Core/Services/Processing/ResultEventEmitter.cs
```diff
@@ -29,6 +41,9 @@ public class ResultEventEmitter : IResultEventEmitter {
 							: new EmittedDataEvent(
 								streamId, Guid.NewGuid(), "Result", true, projectionResult, null, at, null),
 						_resultStreamMetadata);
+				if (!string.IsNullOrEmpty(result.Event.Data) && result.Event.EventType.Equals("Result")) {
```
timothycoleman, line 44: i think this is the same check as below isn't it

src/EventStore.Projections.Core/Services/Processing/ResultEventEmitter.cs
```diff
@@ -16,6 +18,16 @@ public class ResultEventEmitter : IResultEventEmitter {
 			return CreateResultUpdatedEvents(partition, result, at);
 		}
 
+		private void CheckStateSize(EmittedEventEnvelope result) {
+			if (System.Text.ASCIIEncoding.Unicode.GetByteCount(result.Event.Data) >= 8_000_000 && !_largeStateWarningLogged) {
```
timothycoleman, line 22: arguments are evaluated from left to right, so it would make sense to check the bool first since it's quicker. counting the bytes is quite expensive. it could even be an early return

src/EventStore.Projections.Core/Services/Processing/ResultEventEmitter.cs
```diff
@@ -1,8 +1,10 @@
 using System;
+using Serilog;
 
 namespace EventStore.Projections.Core.Services.Processing {
 	public class ResultEventEmitter : IResultEventEmitter {
 		private readonly ProjectionNamesBuilder _namesBuilder;
+		private bool _largeStateWarningLogged = false; //This would be set to true as soon as the projection state size becomes greater than 8MB and a warning is logged, preventing flooding of logs with warnings each time an event is emitted.
```
timothycoleman, line 7: i think the comment isn't necessary for this now that the variable name is clear

timothycoleman (changes requested):

src/EventStore.Projections.Core/Services/Processing/ResultEventEmitter.cs
lakshdeepsingheventstore: Yeah it is indeed the same check. The purpose of keeping this check outside of the method CheckStateSize() was to avoid calling the CheckStateSize() method for every emitted event, but rather only call it for the events that determine the size of the projection state (that is events of type "Result").

hayley-jean (changes requested):
Thanks! There's one more place where the state size needs to be checked

What you have here works for the projection result that will be written to the `$projection-{projection_name}-result` stream. This result is only written for Queries, or for Continuous/OneTime projections that call `outputState()` in the projection definition.

The size of the state that makes up the checkpoints written to the `$projection-{projection_name}-checkpoint` stream and `$projection-{projection_name}-{partition_name}-checkpoint` streams also needs to be checked.

Probably the best place to check and log a warning for the state size is when `StateUpdated` is called in the [CoreProjectionCheckpointManager](https://github.com/EventStore/EventStore/blob/8d83ef069dfb1a7ce028f46bb18ce0d5169e0585/src/EventStore.Projections.Core/Services/Processing/CoreProjectionCheckpointManager.cs#L161) (in addition to the check you're doing here in `ResultEventEmitter`)

lakshdeepsingheventstore:
>The size of the state that makes up the checkpoints written to the $projection-{projection_name}-checkpoint stream and $projection-{projection_name}-{partition_name}-checkpoint streams also needs to be checked.
Making changes to [CoreProjectionCheckpointManager](https://github.com/EventStore/EventStore/blob/8d83ef069dfb1a7ce028f46bb18ce0d5169e0585/src/EventStore.Projections.Core/Services/Processing/CoreProjectionCheckpointManager.cs#L161) solves the same purpose as was being sone by the changes in `ResultEventEmitter` file. So in the latest commit/changes, I have removed the changes from `ResultEventEmitter`, and added those in `CoreProjectionCheckpointManager` in the method `StateUpdated`
However, to exclusively check for state sizes for the checkpoint streams, separate checks had to be added to `CoreProjectionCheckpointWriter` and `PartitionStateUpdateManager` files, since the check in `CoreProjectionCheckpointManager` does not log anything when the projection Checkpoint is written.

src/EventStore.Projections.Core/Services/Processing/PartitionStateUpdateManager.cs
```diff
@@ -29,6 +33,16 @@ private class State {
 			}
 		}
 
+		private void CheckPartitionStateSize(PartitionState state) {
+			if (!_largePartitionStateWarningLogged && state.State.Length >= 8_000_000) {
+				Log.Warning(
```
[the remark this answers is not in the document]
lakshdeepsingheventstore: Done

src/EventStore.Projections.Core/Services/Processing/CoreProjectionCheckpointManager.cs
```diff
@@ -175,6 +177,17 @@ public abstract class CoreProjectionCheckpointManager : IProjectionCheckpointMan
 
 			if (partition == "")
 				_currentProjectionState = newState;
+			CheckStateSize(_currentProjectionState);
+		}
+
+		private void CheckStateSize(PartitionState result) {
```
hayley-jean, line 183: You could actually do the partition check here as well instead of in `PartitionStateUpdateManager`. If you move the `if (partition == "" && newState.State == null)` condition, you could check the state size of `newState` before the other two conditions. e.g:

```
if (partition == "" && newState.State == null) // ignore non-root partitions and non-changed states
	throw new NotSupportedException("Internal check");

CheckStateSize(newState, partition);

if (_usePersistentCheckpoints && partition != "")
	CapturePartitionStateUpdated(partition, oldState, newState);

if (partition == "")
	_currentProjectionState = newState;
```

hayley-jean (changes requested):
Sorry about not including this in my earlier review, but I learned new things about how the `result` of a projection interacts with the `state` while testing this. Basically, we need to include both the `state` and `result` in `PartitionState` when checking the size of the projection state.

I previously thought that the result was only written if you had `outputState()` defined on your projection, but it turns out that the [check for that is commented out](https://github.com/EventStore/EventStore/blob/28fa63d6989d8c675cff4f606540394fe6166bec/src/EventStore.Projections.Core/Services/Processing/EventProcessingProjectionProcessingPhase.cs#L260).

Instead, a result is always written for continuous projections (just not always in a result stream). If the projection includes a `transformBy()` function then the result will be the result of that transformation run on the state. Otherwise it [simply duplicates the state into the result](https://github.com/EventStore/EventStore/blob/28fa63d6989d8c675cff4f606540394fe6166bec/src/EventStore.Projections.Core/Services/Processing/EventProcessingProjectionProcessingPhase.cs#L264).

The result is [included with the state](https://github.com/EventStore/EventStore/blob/28fa63d6989d8c675cff4f606540394fe6166bec/src/EventStore.Projections.Core/Services/Processing/PartitionState.cs#L77) when the checkpoint is written, so this can effectively double the state size that is written to disk.

I think the best way to handle this would be to have a `Size` property on `PartitionState` that returns something like `_state.Length + _result?.Length ?? 0;` and use that for the warning log. Otherwise, by the time the warning is logged at 8mb, you are already potentially over the 16mb limit.

Let me know if you want a sample projection to test this case

src/EventStore.Projections.Core/Services/Processing/CoreProjectionCheckpointManager.cs
lakshdeepsingheventstore: Done

PR #4279, opened 2024-05-31
[ESDB-116-4] Add Catchup Subscription Metrics

thefringeninja:
- Added: new metrics for catchup subscriptions


YoEight (approved):
LGTM

src/EventStore.Core/Messages/MonitoringMessage.cs
```diff
@@ -208,13 +209,43 @@ public partial class InternalStatsRequestResponse : Message {
...
+
+		[DerivedMessage(CoreMessage.Monitoring)]
+		public partial class SubscriptionStatsCollected : Message {
+			public ICollection<SubscriptionStats> Stats { get; }
+
+			public SubscriptionStatsCollected(ICollection<SubscriptionStats> stats) {
+				Stats = stats;
+				Ensure.NotNull(stats, nameof(stats));
+			}
+		}
```
timothycoleman, line 233: looks like these aren't used (?)

src/EventStore.Core/Services/SubscriptionTracker.cs
```diff
@@ -0,0 +1,45 @@
...
+public interface ISubscriptionTracker {
+	void AddSubscription(Guid subscriptionId, string? streamName, long endOfStream);
+	void RemoveSubscription(Guid subscriptionId);
+	void RecordEvent(EventRecord @event);
+	void ProcessEvent(ResolvedEvent @event);
+}
+
+public class SubscriptionMetricHandler(ISubscriptionTracker tracker) : IHandle<StorageMessage.EventCommitted> {
+	public void Handle(StorageMessage.EventCommitted message) => tracker.RecordEvent(message.Event);
+}
```
timothycoleman, line 18: i think the tracker could implement IHandle directly, no need for the extra wrapper

src/EventStore.Core/Services/SubscriptionMetric.cs
```diff
@@ -0,0 +1,74 @@
...
+				_subscriptionStats.TryUpdate(stats.SubscriptionId, stats with {
+					EndOfStream = eventPosition
+				}, stats);
+			} else if (stats.StreamName is null) {
+				_subscriptionStats.TryUpdate(stats.SubscriptionId, stats with {
+					EndOfStream = commitPosition
+				}, stats);
+			}
+		}
+	}
```
timothycoleman, line 38: is this quite an expensive operation to do for every event that gets written

src/EventStore.Core/Messages/MonitoringMessage.cs
```diff
@@ -208,13 +209,43 @@ public partial class InternalStatsRequestResponse : Message {
...
+
+		public record struct SubscriptionStats(
+			Guid SubscriptionId,
+			string StreamName,
+			long SubscriptionPosition = 0L,
+			long EndOfStream = 0L) {
+			private const string StreamNameTagKey = "streamName";
```
timothycoleman, line 240:
```suggestion
			private const string StreamNameTagKey = "stream-name";
```

src/EventStore.Core/Messages/MonitoringMessage.cs
```diff
@@ -208,13 +209,43 @@ public partial class InternalStatsRequestResponse : Message {
...
+			private const string SubscriptionIdTagKey = "subscriptionId";
```
timothycoleman, line 241:
```suggestion
			private const string SubscriptionIdTagKey = "subscription-id";
```

src/EventStore.Core/Services/SubscriptionMetric.cs
```diff
@@ -0,0 +1,74 @@
...
+				_subscriptionStats.TryUpdate(stats.SubscriptionId, stats with {
+					SubscriptionPosition = eventPosition
+				}, stats);
+			} else if (stats.StreamName is null) {
+				_subscriptionStats.TryUpdate(stats.SubscriptionId, stats with {
+					SubscriptionPosition = commitPosition
+				}, stats);
+			}
+		}
+	}
```
timothycoleman, line 52: - similarly, quite expensive to do for every event that we send to a subscription?
- also we probably want to update the specific subscription that processed this event or all the subscriptions will have the same position

src/EventStore.Core/Services/Transport/Enumerators/Enumerator.AllSubscriptionFiltered.cs
```diff
@@ -374,6 +381,9 @@ public class AllSubscriptionFiltered : IAsyncEnumerator<ReadResponse> {
 									_subscriptionId, _eventFilter, caughtUp);
 
 								confirmationPositionTcs.TrySetResult(new TFPos(caughtUp, caughtUp));
+
+								_tracker.AddSubscription(_subscriptionId, null, caughtUp);
+
```
timothycoleman, line 386: ive not looked properly, but would it make sense to make these tracker calls in the subscriptionsservice so that the three kinds of subscription (stream, all, allfiltered) are handled together?

src/EventStore.Core/Services/SubscriptionTracker.cs
```diff
@@ -0,0 +1,45 @@
...
+
+public class SubscriptionTracker(SubscriptionMetric metric) : ISubscriptionTracker {
+	public void AddSubscription(Guid subscriptionId, string? streamName, long endOfStream) =>
+		metric.Add(subscriptionId, streamName, endOfStream);
+
+	public void RemoveSubscription(Guid subscriptionId) => metric.Remove(subscriptionId);
+
+	public void RecordEvent(EventRecord @event) =>
+		metric.RecordEvent(@event.EventStreamId, @event.TransactionPosition, @event.EventNumber);
```
timothycoleman, line 27:
```suggestion
		metric.RecordEvent(@event.EventStreamId, @event.LogPosition, @event.EventNumber);
```

src/EventStore.Core/Services/SubscriptionMetric.cs
```diff
@@ -0,0 +1,74 @@
...
+
+namespace EventStore.Core.Services;
+
+public class SubscriptionMetric {
+	private readonly ConcurrentDictionary<Guid, MonitoringMessage.SubscriptionStats> _subscriptionStats;
+
+	public SubscriptionMetric(Meter meter, string name) {
+		_subscriptionStats = new();
+
+		meter.CreateObservableUpDownCounter($"{name}-completed", Completed);
```
timothycoleman, line 16: it might be better to leave this as a calculation to be done in the dashboard, since its just derived from the other metrics

i filed a task to update the grafana dashboard actually, it might be worth doing that to sanity check that it's reasonble to produce the graphs we want from these metrics

timothycoleman (changes requested):
questions/suggestions above

src/EventStore.Core/Services/SubscriptionMetric.cs
```diff
@@ -0,0 +1,74 @@
...
+		meter.CreateObservableUpDownCounter($"{name}-subscription-position", SubscriptionPosition);
+		meter.CreateObservableUpDownCounter($"{name}-stream-position", StreamPosition);
```
timothycoleman, line 18: the ticket talks about having the timestamps of the current vs latest events, is that still in scope?

timothycoleman, line 18: i don't think there is an easy way to tell how many subscriptions there currently are (?) if not, it might be worth adding that as a separate metric

src/EventStore.Core/Services/SubscriptionMetric.cs
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	public void Add(Guid subscriptionId, string? streamName, long endOfStream) =>
+		_subscriptionStats.TryAdd(subscriptionId, new(subscriptionId, streamName, EndOfStream: endOfStream));
+
+	public void Remove(Guid subscriptionId) => _subscriptionStats.TryRemove(subscriptionId, out _);
```
timothycoleman, line 24: could you double check what happens in the dashboard if we stop reporting a particular subscription, does it stop drawing it or does it just keep drawing the last reported values for a while

src/EventStore.Core/Services/SubscriptionMetric.cs
thefringeninja: If we use the read index we could read the last event to get that info, but dunno how expensive that is every `n` seconds. We would have to expose `IAllReader` to do it for the all stream

Or we go back to using `EventCommitted`

Or, we could occasionally update this from the grpc subscription

PR #4284, opened 2024-06-10
[DBX-41] Fix finalizer when AlignedMemory failed to allocate

timothycoleman:
Fixed: Finalizer bug no longer causes process exit if a memory allocation fails.

Before this fix the finalizer's call to `GC.RemoveMemoryPressure(0)` would throw an ArgumentOutOfRangeException, exiting the process.

### DIFF
--- a/src/EventStore.Core/DataStructures/ProbabilisticFilter/AlignedMemory.cs
+++ b/src/EventStore.Core/DataStructures/ProbabilisticFilter/AlignedMemory.cs
@@ -42,6 +42,12 @@ namespace EventStore.Core.DataStructures.ProbabilisticFilter {
 
 			_disposed = true;
 			GC.SuppressFinalize(this);
+
+			if (_intPtr == IntPtr.Zero) {
+				// didn't manage to allocate or add pressure (OOM), so do not free.
+				return;
+			}
+
 			Marshal.FreeHGlobal(_intPtr);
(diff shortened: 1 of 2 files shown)


src/EventStore.Core/DataStructures/ProbabilisticFilter/AlignedMemory.cs
```diff
@@ -42,8 +42,12 @@ public sealed unsafe class AlignedMemory : IDisposable {
 
 			_disposed = true;
 			GC.SuppressFinalize(this);
-			Marshal.FreeHGlobal(_intPtr);
-			GC.RemoveMemoryPressure(_bytesAllocated);
+
+			if (_intPtr != IntPtr.Zero) {
+				// didn't manage to allocate or add pressure (OOM), so do not free.
```
shaan1337, line 47:
```suggestion
				// didn't manage to allocate or add pressure (OOM), so do not free.
```
shouldn't this comment be in the `else` block of this `if`?

timothycoleman: oops, thanks. fixed

shaan1337: thanks!

shaan1337 (approved):

PR #4285, opened 2024-06-10
[DBX-41] Continue index merge even if memory cannot be allocated for the bloom filter

timothycoleman:
Changed: Index merge now continues even if there is not enough memory to store the Bloom filter


shaan1337 (approved):

timothycoleman:
(did the cherry pick manually)

PR #4289, opened 2024-06-11
Bump up the chunk format version to 4 (`Transformed`)

shaan1337:
Added: Chunk version 4 (`Transformed`) file format
Added: Support for forward compatibility in chunks

Thanks to @ylorph for the initial idea to bump up the chunk version and @timothycoleman and @sakno for their feedback.


src/EventStore.Core/Exceptions/UnknownFileVersionException.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+
+namespace EventStore.Core.Exceptions {
+	public class UnknownFileVersionException : Exception {
```
ylorph, line 4: UnsupportedFileVersionException might be more descriptive ?

shaan1337: yup, that's better thanks

timothycoleman:
This PR adds a second version field to the Chunk header. Now there are two:
- MinimumCompatibleVersion (same position as the old Version field)
- Version (new)

This facilitates the creation of forwards-compatible chunk versions and therefore facilitates rolling upgrades when the chunk version is changing.

When a new chunk has no transformation, it will have Version 4 (because it is) and MinVersion 3 (because old versions of ESDB will open it fine)

When a new chunk has a transformation, both versions will be 4. Old versions of ESDB will (correctly) refuse to open it.

The MinimumCompatibleVersion can be thought of as a promise of compatibility from a newer version of ESDB to an older version. In other circumstances this role could be played by a schema registry but that isn't applicable here.

timothycoleman (approved):

PR #4291, opened 2024-06-11
[DBX-40] Fix redacted events break replication on follower

lakshdeepsingheventstore:
Fixed : Redacted events break replication on a follower node

Redacted events break replication on a follower node due to the following [PR](https://github.com/EventStore/EventStore/pull/3793), since the `PrepareLogRecord` returns `NoData` for redacted events, which causes a follower replicating log records to write the record to disk with empty data. The fix includes introducing a new variable `_dataOnDisk` and modifying the `Data` property to return `NoData` if the event is redacted, otherwise return `_dataOnDisk`. We now also use `_dataOnDisk` instead of `Data` when writing the record, copying for retry, etc.

### DIFF
--- a/src/EventStore.Core/TransactionLog/LogRecords/PrepareLogRecord.cs
+++ b/src/EventStore.Core/TransactionLog/LogRecords/PrepareLogRecord.cs
@@ -55,6 +55,7 @@ namespace EventStore.Core.TransactionLog.LogRecords {
 		public string EventType { get; }
 		private int? _eventTypeSize;
 		public ReadOnlyMemory<byte> Data { get; }
+		private readonly ReadOnlyMemory<byte> _dataOnDisk;
 		public ReadOnlyMemory<byte> Metadata { get; }
 
 		private int? _sizeOnDisk;
@@ -73,7 +74,7 @@ namespace EventStore.Core.TransactionLog.LogRecords {
 				       + 16
 				       + 8
 				       + IntPtr.Size + EventType.Length * 2
-				       + IntPtr.Size + Data.Length
+				       + IntPtr.Size + _dataOnDisk.Length
 				       + IntPtr.Size + Metadata.Length;
 			}
 		}
@@ -102,7 +103,7 @@ namespace EventStore.Core.TransactionLog.LogRecords {
 					+ sizeof(long) /* TimeStamp */
 					+ StringSizeWithLengthPrefix(_eventTypeSize.Value) /* EventType */
 					+ sizeof(int) /* Data length */
-					+ Data.Length /* Data */
+					+ _dataOnDisk.Length /* Data */
 					+ sizeof(int) /* Metadata length */
 					+ Metadata.Length; /* Metadata */
 
@@ -158,7 +159,8 @@ namespace EventStore.Core.TransactionLog.LogRecords {
 			TimeStamp = timeStamp;
 			EventType = eventType ?? string.Empty;
 			_eventTypeSize = eventTypeSize;
-			Data = Flags.HasFlag(PrepareFlags.IsRedacted) ? NoData : data;
+			_dataOnDisk = data;
+			Data = Flags.HasFlag(PrepareFlags.IsRedacted) ? NoData : _dataOnDisk;
 			Metadata = metadata;
 			if (InMemorySize > TFConsts.MaxLogRecordSize) throw new Exception("Record too large.");
 		}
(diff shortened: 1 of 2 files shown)


src/EventStore.Core/TransactionLog/LogRecords/PrepareLogRecord.cs
```diff
@@ -42,6 +42,7 @@ public static class PrepareFlagsExtensions {
 
 	public class PrepareLogRecord : LogRecord, IEquatable<PrepareLogRecord>, IPrepareLogRecord<string> {
 		public const byte PrepareRecordVersion = 1;
+		private readonly ReadOnlyMemory<byte> _data;
```
shaan1337, line 45: suggestion: move this line below `public ReadOnlyMemory<byte> Data { get; }`

shaan1337:
```suggestion
		public ReadOnlyMemory<byte> Data =>
			Flags.HasFlag(PrepareFlags.IsRedacted) ? NoData : _data;
			
		private readonly ReadOnlyMemory<byte> _data;
```

we can also make the `Data` property return `NoData` if it the event is redacted, otherwise return `_data`

event writes would still use `_data` irrespective of whether the event is redacted or not

lakshdeepsingheventstore: >we can also make the Data property return NoData if it the event is redacted, otherwise return _data
Implemented this in the constructors

shaan1337: thanks @lakshdeepsingheventstore

src/EventStore.Core/TransactionLog/LogRecords/PrepareLogRecord.cs
```diff
@@ -55,6 +55,7 @@ public class PrepareLogRecord : LogRecord, IEquatable<PrepareLogRecord>, IPrepar
 		public string EventType { get; }
 		private int? _eventTypeSize;
 		public ReadOnlyMemory<byte> Data { get; }
+		private readonly ReadOnlyMemory<byte> _data;
```
shaan1337, line 58:
```suggestion
		private readonly ReadOnlyMemory<byte> _dataOnDisk;
```

suggestion: rename `_data` to `_dataOnDisk` to clearly distinguish between the two of them

src/EventStore.Core/TransactionLog/LogRecords/PrepareLogRecord.cs
```diff
@@ -158,7 +159,8 @@ public class PrepareLogRecord : LogRecord, IEquatable<PrepareLogRecord>, IPrepar
 			TimeStamp = timeStamp;
 			EventType = eventType ?? string.Empty;
 			_eventTypeSize = eventTypeSize;
-			Data = Flags.HasFlag(PrepareFlags.IsRedacted) ? NoData : data;
+			_data = data;
+			Data = Flags.HasFlag(PrepareFlags.IsRedacted) ? NoData : _data;
 			Metadata = metadata;
 			if (InMemorySize > TFConsts.MaxLogRecordSize) throw new Exception("Record too large.");
```
shaan1337, line 165: the following internal methods also need to be adjusted to use `_dataOnDisk` instead of `Data`
- InMemorySize - this one may seem counterintuitive but its actually used to check if the `PrepareLogRecord` doesn't exceed the max log record size, so we do need `_dataOnDisk` here
- SizeOnDisk
- Equals
- GetHashCode

lakshdeepsingheventstore: I guess we can use `_dataOnDisk` instead of `Data` for the methods `InMemorySize`, `SizeOnDisk`, and `GetHashCode`. For the `Equals` method, it is returning true or false based on whether 2 `PrepareLogRecord` objects are same or not (whether they have same object parameters or not). So maybe in the `Equals` method, instead of replacing `Data` with `_dataOnDisk`, we can keep `Data` and also add the equality check for `_dataOnDisk`.

shaan1337: sounds fair enough, `Equals` is used only in tests anyway so it shouldn't have any performance implications

shaan1337 (approved):
tested by replicating redacted data, looks good :+1:

PR #4305, opened 2024-06-21
[ESDB-139-3] Support multiple authorization policies

hayley-jean:
Added: Allow loading multiple policies for authorization

This PR introduces the ability to authorize operations against multiple policies. The policies are evaluated in order, with policies loaded through a plugin being evaluated first and the legacy policy last.

Policy plugins are currently loaded from the `EventStore:Plugins:Authorization:PolicyType` plugin config. There is work still to be done in future PRs to make use of the new plugin model for policy plugins and/or auth plugins.


src/EventStore.Core/Authorization/LegacyAuthorizationProviderFactory.cs
```diff
@@ -1,149 +1,16 @@
...
-			policy.RequireAuthenticated(Operations.Projections.Update);
-			policy.AddMatchAnyAssertion(Operations.Projections.UpdateConfiguration, Grant.Allow, OperationsOrAdmins);
-			policy.RequireAuthenticated(Operations.Projections.State);
-			policy.RequireAuthenticated(Operations.Projections.Status);
-			policy.RequireAuthenticated(Operations.Projections.Statistics);
-			policy.AddMatchAnyAssertion(Operations.Projections.Restart, Grant.Allow, OperationsOrAdmins);
-
-			return new PolicyAuthorizationProvider(new PolicyEvaluator(policy.AsReadOnly()));
+			return new PolicyAuthorizationProvider(
+			new MultiPolicyEvaluator(_policySelectors), true, false);
```
shaan1337, line 13:
```suggestion
			new MultiPolicyEvaluator(_policySelectors), logAuthorization: true, logSuccesses: false);
```

src/EventStore.Core/Authorization/LegacyAuthorizationProviderFactory.cs
```diff
@@ -1,149 +1,16 @@
-﻿using System;
```
shaan1337, line 1: we need to rename this file to `InternalAuthorizationProviderFactory.cs`

src/EventStore.Core/Authorization/PolicyAuthorizationProvider.cs
```diff
@@ -10,7 +10,7 @@
 namespace EventStore.Core.Authorization;
 
 public class PolicyAuthorizationProvider(IPolicyEvaluator policyEvaluator, bool logAuthorization = true, bool logSuccesses = false) : AuthorizationProviderBase {
-	static readonly ILogger Logger = Log.ForContext<PolicyEvaluator>();
+	static readonly ILogger Logger = Log.ForContext<IPolicyEvaluator>();
```
shaan1337, line 13:
```suggestion
	static readonly ILogger Logger = Log.ForContext<PolicyAuthorizationProvider>();
```

src/EventStore.Core/Authorization/PolicyEvaluator.cs
```diff
@@ -1,36 +1,40 @@
 ﻿using System;
+using System.Linq;
```
shaan1337, line 2: suggestion: rename this file to `MultiPolicyEvaluator`

src/EventStore.Core/Authorization/PolicySelectorsFactory.cs
```diff
@@ -0,0 +1,30 @@
...
+// TODO: Should this be moved to EventStore.Plugins?
+public interface IPolicySelectorFactory {
+	public string CommandLineName { get; }
+	public string Name { get; }
+	public string Version { get; }
+	public IPolicySelector Create(IPublisher publisher, ISubscriber subscriber);
+}
+
+public interface IStreamPermissionAssertion : IAssertion {
+}
```
shaan1337, line 15: suggestion: move to its own file

src/EventStore.Core/Authorization/PolicyEvaluator.cs
```diff
@@ -1,36 +1,40 @@
...
+				var policyInfo = policy.Information;
+				if (policy.TryGetAssertions(operation, out var assertions))
+					while (!assertions.IsEmpty && evaluation.Grant != Grant.Deny) {
+						if (ct.IsCancellationRequested) break;
+						var assertion = assertions.Span[0];
+						assertions = assertions.Slice(1);
+						var evaluate = assertion.Evaluate(cp, operation, policyInfo, evaluation);
+						if (!evaluate.IsCompleted)
+							return EvaluateAsync(evaluate, cp, operation, assertions, evaluation, ct);
+					}
```
shaan1337, line 36: it appears that there could be two bugs here:
1. when `assertion.Evaluate()` is processed synchronously: if `evaluation.Grant == Grant.Deny` , we don't break out of the `foreach` loop and return the result, but instead, we continue to process the next policy selector

2. when `assertion.Evaluate()` is processed asynchronously: we don't process the next policy selectors at all (e.g if the result is `Grant.Unknown` in `EvaluateAsync`, it'll be translated into a `DeniedByDefault` error but we should instead be processing the next policy selectors)

shaan1337: it would be much simpler to use `async`/`await` here, although i'm not sure if the performance difference would be significant (given that it's on the hot path). we could measure the difference to be sure it doesn't have any negative impact

hayley-jean: Thanks, I've updated `MultiPolicyEvaluator` to use `async`/`await`, and added tests for it

I did some benchmarks with the MicroBenchmarks project, and it looks like there's no significant performance difference between using `async`/`await` and not. So it makes sense to keep the code simple

When the call is made synchronously:

| Method             | Mean     | Error   | StdDev  | Ratio | RatioSD | Gen0   | Allocated | Alloc Ratio |
|------------------- |---------:|--------:|--------:|------:|--------:|-------:|----------:|------------:|
| Optimised_NoAwait  | 138.7 ns | 2.69 ns | 2.64 ns |  1.00 |    0.00 | 0.0739 |     464 B |        1.00 |
| NonOptimised_Await | 106.1 ns | 2.08 ns | 2.92 ns |  0.76 |    0.03 | 0.0612 |     384 B |        0.83 |

When the caller is using `async`/`await` themselves

| Method             | Mean     | Error   | StdDev  | Ratio | RatioSD | Gen0   | Allocated | Alloc Ratio |
|------------------- |---------:|--------:|--------:|------:|--------:|-------:|----------:|------------:|
| Optimised_NoAwait  | 157.5 ns | 3.15 ns | 6.86 ns |  1.00 |    0.00 | 0.0739 |     464 B |        1.00 |
| NonOptimised_Await | 135.7 ns | 2.38 ns | 2.11 ns |  0.86 |    0.03 | 0.0610 |     384 B |        0.83 |

src/EventStore.Core/Authorization/MultiPolicyEvaluator.cs
```diff
@@ -0,0 +1,42 @@
...
+				var policyInfo = policy.Information;
+				if (policy.TryGetAssertions(operation, out var assertions)) {
+					while (!assertions.IsEmpty && evaluationContext.Grant != Grant.Deny) {
+						if (ct.IsCancellationRequested) break;
+						var assertion = assertions.Span[0];
+						assertions = assertions.Slice(1);
+						await assertion.Evaluate(cp, operation, policyInfo, evaluationContext);
+					}
+				}
+			}
```
shaan1337, line 33: if all the assertions in a policy have passed, i think that we need to do an early return with the result here, otherwise it'll also evaluate the assertions in the next policies.

```suggestion
				}
                               if (evaluationContext.Grant == Grant.Allow)
                                   return evaluationContext.ToResult()
			}
```

src/EventStore.Core/Authorization/MultiPolicyEvaluator.cs
```diff
@@ -0,0 +1,42 @@
...
+
+			if (evaluationContext.Grant == Grant.Unknown) {
+				evaluationContext.Add(new AssertionMatch(_policyInfo, DeniedByDefault));
+			}
```
shaan1337, line 37:
```suggestion
			evaluationContext.Add(new AssertionMatch(_policyInfo, DeniedByDefault));
```

we can then remove this `if (evaluationContext.Grant == Grant.Unknown)` here as it'll always be true.

src/EventStore.Core/Authorization/PolicyEvaluator.cs
shaan1337: great, thanks!

shaan1337 (approved):

PR #4319, opened 2024-06-28
[ESDB-116-2] Simplify Persistent Subscription metrics and Projection metrics

timothycoleman:
Changed: internal changes

We store the current stats in the tracker and have observe methods that can read them and turn them into measurements. The metrics are then created using those observe methods


dhingrak (approved):
PR looks good to me. I have a question about metric names. 

We have different metric names from EventStore_Exporter metric names. Did we add this to differentiate between the subscriptions and persistent subscriptions metrics? The Marcin's metrics only mentioned subscriptions.

timothycoleman:
Thanks, yep exactly

PR #4366, opened 2024-08-02
Take2troubleshoot

coloradomorrissey:
one change to verbiage in admin ui to test process


realtonyyoung (approved):
Looks good to me

coloradomorrissey:
this seems like all the previous commits were pulled into this pull request... Is that correct? If so, that doesn't seem great.

alexeyzimarev:
@coloradomorrissey No, it's not right. Because the target branch is `master` but you created your branch from another branch, not `master`.

PR #4394, opened 2024-08-16
[ESDB-174-4] Remove HdrHistograms

timothycoleman:
Removed: /histogram/{name} endpoint

This isn't an obvious decision, but a number of factors are nudging towards removing them for now
- System.Diagnostics.Metrics is likely the way forward on balance. They aren't HDR but are easier to consume by other tools
- It's adding ~30 dependencies on MS libraries that the Qodana license audit cannot process automatically
- As far as we know they aren't being used, but its not hard to bring them back if necessary
- Saves a lookup in an (empty) concurrent dictionary for each operation (even when Histograms are disabled)


hayley-jean (approved):

src/EventStore.Core/Services/TcpSendService.cs
```diff
@@ -1,18 +1,14 @@
 ﻿using System.Diagnostics;
 using EventStore.Core.Bus;
 using EventStore.Core.Messages;
-using EventStore.Core.Services.Histograms;
 
 namespace EventStore.Core.Services {
 	public class TcpSendService : IHandle<TcpMessage.TcpSend> {
 		private readonly Stopwatch _watch = Stopwatch.StartNew();
```
hayley-jean, line 7: `_watch` is no longer used
```suggestion
```

hayley-jean (changes requested):

src/EventStore.Core/Services/TcpSendService.cs
timothycoleman: ah yeah thanks

hayley-jean (approved):

PR #4403, opened 2024-08-21
Make sure all systems that require a shutdown process complete before exiting.

YoEight:
Added: Make sure all systems that require a shutdown process complete before exiting.


src/EventStore.Core/Services/VNode/ClusterVNodeController.cs
```diff
@@ -130,6 +131,7 @@ private VNodeFSM CreateFSM() {
 				.When<SystemMessage.ServiceInitialized>().Do(Handle)
 				.When<SystemMessage.BecomeReadOnlyLeaderless>().Do(Handle)
 				.When<SystemMessage.BecomeDiscoverLeader>().Do(Handle)
+				// .When<SystemMessage.BecomeShuttingDown>().Do(Handle)
```
shaan1337, line 134:
```suggestion
```

shaan1337 (approved):
There are a few merge conflicts because of the recent async changes. Otherwise, the code looks good to me (just a few small suggestions)

shaan1337:
suggestion: we could also consider adding a timeout for subsystems to shut down (to prevent a broken plugin from making the server hang at shut down)

shaan1337 (approved):

PR #4408, opened 2024-08-23
[ESDB-169-3]: Migration of internal message dispatcher/scheduler to async

sakno:
Changed: `InMemoryBus` becomes asynchronous dispatcher. Async handlers now can use `IAsyncHandle<T>` interface to enable async execution.
Removed: `IBus` interface is removed with no replacement to clearly distinguish roles of message dispatcher and message scheduler. Removed unused schedulers and related tests.


sakno:
Depends on #4413 and #4412

timothycoleman:
# Buses and Messages

The ESDB server architecture is a message processing pipeline, with stages passing messages to each other via queues.

There are three main interfaces for message processing:

- `IPublisher`
- `I(Async)Handle` (previously: `IHandle`)
- `IEnvelope`

They are all similar in that they have a method that takes a message as a parameter. The primary difference is their _intent_.

- `IPublisher` is for sending messages.
	- If you want to send messages, get hold of an `IPublisher` implementation to send messages with.
- `I(Async)Handle` is for receiving messages.
	- If you want to receive messages, implement `I(Async)Handle<T>`
- `IEnvelope` is for sending messages in reply.
	- If you want a reply for the message you are sending, include an envelope with it for the receiver to call.

## Interface non-guarantees

None of the interfaces (`IPublisher`, `I(Async)Handle`, `IEnvelope`) guarantee anything about the processing of the message. E.g. the interfaces do not guarantee:
- What processing is done with the message before control returns to the caller.
- What recipients the message will be passed to or how many (none, one, many).
- What order messages are processed in.
- Whether the messages are processed in sequence or concurrently.

Such guarantees can be obtained, but they depend on the concrete implementations, and often depend on the way in which the concrete implementations are arranged in the pipeline.

## Best practices

- Ideally the sender of a message should care as little as possible about the way the message is handled by the pipeline. It just sends and that is it.
- Typically the sender of a message wanting a reply should include an envelope that does not burden the receiver. e.g. it should queue the reply rather than making the receiver do expensive work.

## Introducing Async Handling

Everything above was true before async handling was introduced and remains true after it. We are not changing the message processing principles, or the intentions behind the interfaces.

We want to be able to handle messages asynchronously (e.g. so that reads can reach out to async APIs such as S3). So naturally we introduce `IAsyncHandle`, but we want to minimise its impact on the codebase.

On the calling side, places that used to call `Handle` will now need to call `HandleAsync` where there is the possibility of asynchronous handling. Luckily there are not many of these (`InMemoryBus`)

On the implementation side, end handlers implementing `IHandle<T>` remain as-is. Synchronous handling is still supported via _async over sync_, which minimises code impact and is fine. 

When called non-concurrently (which is typical), the `InMemoryBus` provides strong guarantees (messages are processed one at a time, in order), so it must `await` the `HandleAsync` call to provide this guarantee. It can therefore no longer provide this guarantee through the `IPublisher` interface (because that returns `void` not `Task`). So `InMemoryBus` no longer implements `IPublisher`, which has the advantage of flagging, as compiler errors, every place that was relying on this guarantee when publishing (because `InMemoryBus` was the only `IPublisher` with such a guarantee). Those places can then be fixed by either:
	1. (TESTS ONLY): Derive a special class `SynchronousScheduler` from `InMemoryBus` that implements `IPublish` and hackily gives the immediate guarantee by doing _sync over async_. The tests can later be gradually changed according to approaches 2 or 3
	2. Convert the caller to async, have them `await bus.Handle` (e.g. we do this for `MonitoringService` when it collects stats)
	3. Have the caller publish on to a queue if they didn't actually need the immediate processing of the bus.

There is now no implementation of `IPublisher` (outside of the tests) that guarantees messages are processed one at a time and in order. Later it would be possible to add an async version of `IPublisher` and create such an implementation, but this would be high impact in the code and ought to be rarely used (see best practices) (unless we enqueue messages asynchronously - as in Channels).

timothycoleman (approved):
Nice :+1:

PR #4412, opened 2024-08-30
[ESDB-169-5] Group projection processing classes into namespaces

hayley-jean:
Changed: Group projection processing classes into namespaces

This is to make future work in the projections a bit easier to do and review.
- Removed `CheckpointTagJsonVersion` as it was unused.
- Some TODOs and comments that are no longer needed.

Open Questions:
1. Should we distinguish the parts that process projections specifically (`MultiStream`, `TransactionFile`, `ByEventType`, `SingleStream`) from the more general processing namespaces such as `Checkpointing` and `Emitting`?


timothycoleman (approved):
This PR doesn't change behaviour just moves code

PR #4414, opened 2024-09-04
Update streams.md

tomthetrainer:
Removed the misleading statement, 

Regarding deleted streams retaining one event that is readable, 

As per slack chat with Alexey that seems to not be an accurate representation. 

It seems there is enough additional information regarding the behavior of soft delete where appended events to a soft deleted streams start at previous offset plus one, that the best option is to simply delete this paragraph. 

At which version this paragraph was true? I don't have enough information to make that call, but certainly it is NOT true for current version.


timothycoleman:
hey @tomthetrainer, i think the paragraph is still true in current versions, but its pretty clumsy and we should definitely refine what it's saying.

i think what it is trying to say is that if you soft delete a stream, and then scavenge, then you will still find the last event in that stream when you read $all. is that consistent with what you're seeing?

PR #4442, opened 2024-09-23
Minor doc bug in v23.10 upgrade doc

realtonyyoung:
Zip pointed out what they believe is a doc bug on this page - https://developers.eventstore.com/server/v23.10/upgrade-guide.html#breaking-changes-from-21-10-and-earlier : ReplicationTcpPort should be ReplicationPort ? Can anyone confirm or deny?


alexeyzimarev:
The change seems to be too large, it looks like the source branch mismatch.

timothycoleman (approved):
yeah they're right :+1:

PR #4447, opened 2024-09-25
[ESDB-169-8] Remove functionality for upgrading 32bit hash indexes to 64bit

timothycoleman:
Removed: Support for extremely old (V1) PTables

- the logic that deals with the upgrade is complicating forward progress
- 64bit ptables have been created since v3.9.0, there probably aren't any 32bit ptables left anywhere
- if 32bit tables are present we detect them on startup and exit issuing advice (use a version between >= v3.9.0 and < 24.10.0 to upgrade the ptables, or wait for the index to be rebuilt
- the hash collision tests used to rely on V1 tables, instead we use V4 tables now but override the hash functions


src/EventStore.Core/Bus/InMemoryBus.cs
```diff
@@ -80,11 +80,11 @@ private async ValueTask DispatchAndWatchSlowMsg(MessageTypeHandler handlers, Mes
 		var elapsedMs = ts.ElapsedMilliseconds;
 		if (elapsedMs > _slowMsgThresholdMs) {
 			Log.Debug("SLOW BUS MSG [{bus}]: {message} - {elapsed}ms.",
-				Name, message.GetType().Name, elapsedMs);
+				Name, message.GetType().Name, (int)elapsedMs);
 			if (elapsedMs > QueuedHandlerThreadPool.VerySlowMsgThreshold.TotalMilliseconds &&
 			    message is not SystemMessage.SystemInit)
 				Log.Error("---!!! VERY SLOW BUS MSG [{bus}]: {message} - {elapsed}ms.",
-					Name, message.GetType().Name, elapsedMs);
+					Name, message.GetType().Name, (int)elapsedMs);
```
shaan1337, line 87: were these changes meant to be in this PR?

timothycoleman: they aren't strictly related, but it was a recent small regression that this fixes

shaan1337 (approved):

PR #4479, opened 2024-10-03
Add how to resign or change leader in eventstore in the api document.

CindyLoMinChung:
Add how to resign or change a leader under API documentation.

### DIFF
--- a/docs/http-api/api.md
+++ b/docs/http-api/api.md
@@ -1851,6 +1851,22 @@ To perform this operation, you must be authenticated by means of one of the foll
 basicAuth
 </aside>
 
+### Resigning a leader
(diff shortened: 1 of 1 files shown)


coloradomorrissey (approved):

timothycoleman (changes requested):
actually just hold on this one for a mo, i think we made it so you no longer have to reduce the priority - @hayley-jean ?

timothycoleman:
Ah yes it was here https://github.com/EventStore/EventStore/pull/4371 which affects only 24.10, so we can adjust that in the cherry pick to master

timothycoleman (approved):

PR #4518, opened 2024-10-11
[ESDB-139-8] Allow configuring auth policies through a stream

hayley-jean:
Added: Allow configuring the authorization policy type through a settings stream.

Adds the `$authorization-policy-settings` stream, which can be used to configure which auth policy is in use.
To change the auth policy, write an event to the stream with `streamAccessPolicyType` set to the policy type you want to use (`acl`, `streampolicy`, `fallback`):

```
{
    "streamAccessPolicyType": "streampolicy"
}
```

`IPolicySelectorFactory` can now be enabled or disabled depending on configuration. It should be possible to call Enable and Disable repeatedly.

`EventStore::Authorization:PolicyType` is now used to set the default policy type, and will be overridden by any policy configured in the settings stream.


shaan1337 (approved):
i didn't have enough time to think carefully about/test all cases but approving as we're short of time for the release

PR #4528, opened 2024-10-15
[ESDB-182-1] [ESDB-182-2] Archive completed, old & scavenged chunks

shaan1337:


src/EventStore.Core/Services/Archiver/Storage/FileSystemArchiveStorage.cs
```diff
@@ -16,11 +21,74 @@ public FileSystemArchiveStorage(FileSystemOptions options) {
...
+				path: tempPath,
+				options: new FileStreamOptions {
+					Mode = FileMode.CreateNew,
+					Access = FileAccess.ReadWrite,
+					Share = FileShare.None,
+					Options = FileOptions.Asynchronous,
+					PreallocationSize = new FileInfo(chunkPath).Length
+				});
+
+			await source.CopyToAsync(destination, ct);
```
sakno, line 54: Can be replaced with `File.Copy` to utilize the best effort from OS.

src/EventStore.Core/Services/Archiver/Storage/FileSystemArchiveStorage.cs
```diff
@@ -16,11 +21,74 @@ public FileSystemArchiveStorage(FileSystemOptions options) {
...
+			var directoryInfo = new DirectoryInfo(_archivePath);
+			for (var chunkNumber = chunkStartNumber; chunkNumber <= chunkEndNumber; chunkNumber++) {
+				foreach (var file in directoryInfo.EnumerateFiles($"chunk-{chunkNumber:000000}.*")) {
+					if (file.Name == exceptChunk)
+						continue;
+
+					File.Delete(file.FullName);
+				}
+			}
+		} catch (Exception ex) when (ex is not OperationCanceledException) {
```
sakno, line 79: If `ex` is not `OperationCanceledException` then the method throws instead of returning the task with wrapped exception.

src/EventStore.Core/Services/Archiver/Storage/IArchiveStorage.cs
```diff
@@ -2,17 +2,39 @@
...
+	/// <summary>Stores a chunk in the archive</summary>
+	/// <param name="chunkPath">The path of the chunk to archive</param>
+	/// <param name="ct">The token to monitor for cancellation requests</param>
+	/// <exception cref="ChunkDeletedException">Thrown if the chunk file is deleted while being archived</exception>
+	/// <exception cref="OperationCanceledException">Thrown if cancellation was requested</exception>
+	/// <returns>
+	/// <see langword="true"/> if the chunk was successfully archived<br/>
+	/// <see langword="false"/> if there was a recoverable error
+	/// </returns>
+	public ValueTask<bool> StoreChunk(string chunkPath, CancellationToken ct);
```
sakno, line 22: It's not clear what the caller needs to do with **false** when returned? What is a recoverable error?

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
```diff
@@ -1414,4 +1415,15 @@ internal readonly ref struct Slot(ReaderWorkItem[] array, int index) {
 			public ref ReaderWorkItem ValueRef => ref UnsafeGetElement(array, index);
 		}
 	}
+
+	public ChunkInfo GetChunkInfo() {
```
sakno, line 1419: Why not just a property? Especially if `ChunkInfo` is **struct**.

src/EventStore.Core/TransactionLog/Chunks/TFChunkDb.cs
```diff
@@ -22,13 +22,20 @@ public class TFChunkDb : IAsyncDisposable {
...
 
-	public TFChunkDb(TFChunkDbConfig config, ITransactionFileTracker tracker = null, ILogger log = null, DbTransformManager transformManager = null) {
+	public TFChunkDb(
+		TFChunkDbConfig config,
+		ITransactionFileTracker tracker = null,
+		ILogger log = null,
+		DbTransformManager transformManager = null,
+		Action<EventStore.Core.Data.ChunkInfo> onChunkLoaded = null,
+		Action<EventStore.Core.Data.ChunkInfo> onChunkCompleted = null,
+		Action<EventStore.Core.Data.ChunkInfo> onChunkSwitched = null) {
```
sakno, line 32: It's better for readability to move these optional callbacks to init-only property. `NoOp` delegate is not necessary because null-safe invocation can be done with `callback?.Invoke` expression.

src/EventStore.Core/TransactionLog/Chunks/TFChunkManager.cs
```diff
@@ -38,12 +39,26 @@ public int ChunksCount {
...
+	private readonly Action<ChunkInfo> _onChunkCompleted;
+	private readonly Action<ChunkInfo> _onChunkSwitched;
+
+	public TFChunkManager(
+		TFChunkDbConfig config,
+		ITransactionFileTracker tracker,
+		DbTransformManager transformManager,
+		Action<ChunkInfo> onChunkLoaded,
+		Action<ChunkInfo> onChunkCompleted,
+		Action<ChunkInfo> onChunkSwitched) {
```
sakno, line 52: Same as above, replace parameters with init-only or required properties

src/EventStore.Core/TransactionLog/Chunks/TFChunkManager.cs
```diff
@@ -289,6 +311,8 @@ public async ValueTask AddChunk(TFChunk.TFChunk chunk, CancellationToken token)
 			_chunksLocker.Release();
 		}
 
+		_onChunkSwitched(newChunk.GetChunkInfo());
```
sakno, line 314: This is one of the reasons why `callback?.Invoke()` is better. If the callback is **null**, no need to compute a value of `GetChunkInfo` method. It makes no sense to compute the value and pass it to no-op callback.

src/EventStore.Core/Services/Archiver/Storage/IArchiveStorage.cs
shaan1337: the caller needs to retry after some time when `false` is returned. By recoverable error, I mean that there was a temporary issue storing the chunk (e.g if the storage server is temporarily down or if there is an issue with the network connection)

shaan1337: i'll amend the code docs to clarify that, thanks

src/EventStore.Core/TransactionLog/Chunks/TFChunkDb.cs
shaan1337: if we move them to init-only properties in TFChunkDb, we'll also need to change `TFChunkManager Manager` to an init-only property as we build the TFChunkManager in TFChunkDb's constructor (this will affect many tests, the TFChunkDb class is used in ~99 places)

but for the TFChunkManager class, moving them to init-only properties should be fine

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
shaan1337: Fixed (i considered that initially and used a method instead so that it doesn't look like the "ChunkInfo" information is static - it can change when the chunk is completed. but i think it's fine to use a property as well)

src/EventStore.Core/Services/Archiver/Storage/FileSystemArchiveStorage.cs
shaan1337: isn't it the other way round? (if `ex` is `OperationCanceledException`, then it'll throw it, otherwise it'll log an error and return false).

maybe it would be more easily readable if we add an explicit catch for OperationCanceledException:
```
catch (OperationCanceledException) {
   throw;
} catch (Exception ex) {
...
}
```

shaan1337: fixed

src/EventStore.Core/Services/Archiver/Storage/FileSystemArchiveStorage.cs
shaan1337: i initially thought about using `File.Copy`, but then i've switched to `FileStream` for two main reasons:
- streams have async support
- to have better control over the file opening/creation options

i'm open to using `File.Copy()` if the above points don't matter much in this context

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
timothycoleman: ChunkInfo is a class at the moment though. since it can change maybe the property should be called `CurrentInfo`

src/EventStore.Core/Services/Archiver/ChunkPrioritizer.cs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Event Store Ltd and/or licensed to Event Store Ltd under one or more agreements.
+// Event Store Ltd licenses this file to you under the Event Store License v2 (see LICENSE.md).
+
+using System.Collections.Generic;
+
+namespace EventStore.Core.Services.Archiver;
+
+public class ChunkPrioritizer : IComparer<Commands.ArchiveChunk> {
+	public int Compare(Commands.ArchiveChunk x, Commands.ArchiveChunk y) {
+		int cmp = x!.ChunkStartNumber.CompareTo(y!.ChunkStartNumber);
```
timothycoleman, line 10: are the `!` needed here

src/EventStore.Core/Services/Archiver/Storage/IArchiveStorage.cs
```diff
@@ -2,17 +2,39 @@
...
 
 public interface IArchiveStorage {
-	public static IArchiveStorage None = new NoArchiveStorage();
-	public ValueTask StoreChunk(string path);
-}
+	/// <summary>Stores a chunk in the archive</summary>
+	/// <param name="chunkPath">The path of the chunk to archive</param>
+	/// <param name="ct">The token to monitor for cancellation requests</param>
+	/// <exception cref="ChunkDeletedException">Thrown if the chunk file is deleted while being archived</exception>
+	/// <exception cref="OperationCanceledException">Thrown if cancellation was requested</exception>
```
timothycoleman, line 17: open to opinions on this, but personally i dont think its worth it to document the meaning of the cancellation token and the possibility of operationcanceledexception when they're the normal pattern

src/EventStore.Core/Services/Archiver/Storage/IArchiveStorage.cs
timothycoleman: i could be persuaded but im not sure about the return value to indicate retryable errors vs exceptions to indicate nonretryable, unless it's a common pattern i'd probably just handle it all through returnvalue or exceptions (probably exceptions in this case)

src/EventStore.Core/TransactionLog/Chunks/TFChunkManager.cs
```diff
@@ -289,6 +303,8 @@ public async ValueTask AddChunk(TFChunk.TFChunk chunk, CancellationToken token)
 			_chunksLocker.Release();
 		}
 
+		OnChunkSwitched?.Invoke(newChunk.ChunkInfo);
```
timothycoleman, line 306: do we want to call this even if the chunk switching failed?

src/EventStore.Core/Services/Archiver/Storage/FileSystemArchiveStorage.cs
```diff
@@ -16,11 +21,78 @@ public FileSystemArchiveStorage(FileSystemOptions options) {
...
+				chunkStartNumber, chunkEndNumber, exceptChunk);
+			return ValueTask.FromResult(false);
+		}
+
+		return ValueTask.FromResult(true);
+	}
+
+	public IAsyncEnumerable<string> ListChunks(CancellationToken ct) {
+		return new DirectoryInfo(_archivePath)
+			.EnumerateFiles("chunk-*")
```
timothycoleman, line 94: not necessarily now but perhaps we should at least add a note that a `IVersionedFileNamingStrategy` should probably drive the filtering here (and elsewhere in this file where the "chunk" prefix/pattern is hard coded

src/EventStore.Core/Services/Archiver/ArchiverService.cs
```diff
@@ -2,54 +2,192 @@
...
+		_mainBus.Subscribe<ReplicationTrackingMessage.ReplicatedTo>(this);
+		_mainBus.Subscribe<SystemMessage.SystemStart>(this);
+		_mainBus.Subscribe<SystemMessage.BecomeShuttingDown>(this);
+	}
+
+	private void Start() {
+		Task.Factory.StartNew(
+			() => ArchiveChunks(_cts.Token),
+			_cts.Token,
+			TaskCreationOptions.LongRunning,
```
timothycoleman, line 71: LongRunning is suitable when the task is going to use the thread that its scheduled on for a long time (it's a hint to the scheduler that another thread might be handy)   but that isn't the case here. we can just use task.run

src/EventStore.Core/Services/Archiver/ArchiverService.cs
```diff
@@ -2,54 +2,192 @@
...
-			StorageType.S3 => new S3ArchiveStorage(options.S3),
-			_ => throw new InvalidOperationException(),
-		};
+	public void Handle(ReplicationTrackingMessage.ReplicatedTo message) {
+		_replicationPosition = Math.Max(_replicationPosition, message.LogPosition);
+		ProcessUncommittedChunks();
+	}
+
+	public void Handle(SystemMessage.SystemStart message) {
+		Task.Factory.StartNew(() => ScheduleExistingChunksForArchiving(_cts.Token), _cts.Token);
```
timothycoleman, line 102: task.run will do here for simplicity (https://blog.stephencleary.com/2013/08/startnew-is-dangerous.html)

src/EventStore.Core/Services/Archiver/ArchiverService.cs
```diff
@@ -2,54 +2,192 @@
...
-	public string Version => "0.0.1";
+		_uncommittedChunks = new();
+		_existingChunks = new();
+		_cts = new();
+		_archiveChunkCommands = Channel.CreateUnboundedPrioritized(
+			new UnboundedPrioritizedChannelOptions<Commands.ArchiveChunk> {
+			SingleWriter = true,
+			SingleReader = true,
+			Comparer = new ChunkPrioritizer()
+		});
```
timothycoleman, line 52:
```suggestion
				SingleWriter = true,
				SingleReader = true,
				Comparer = new ChunkPrioritizer()
			});
```

src/EventStore.Core/Services/Archiver/ChunkPrioritizer.cs
shaan1337: yes, because that's how IComparer is defined:
```
  public interface IComparer<in T>
  {
    int Compare(T? x, T? y);
  }
```

[strictly speaking](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.icomparer-1.compare?view=net-8.0#remarks) we should implement the case where x and y can be null but in this case it should never happen

src/EventStore.Core/Services/Archiver/Storage/IArchiveStorage.cs
shaan1337: yup, fixed :+1:

src/EventStore.Core/TransactionLog/Chunks/TFChunkManager.cs
shaan1337: good catch! fixed

src/EventStore.Core/Services/Archiver/ArchiverService.cs
shaan1337: fixed, thanks for the clarification

src/EventStore.Core/Services/Archiver/ArchiverService.cs
shaan1337: fixed

src/EventStore.Core/Services/Archiver/ArchiverService.cs
shaan1337: fixed :+1:

src/EventStore.Core/Services/Archiver/Storage/FileSystemArchiveStorage.cs
sakno: Anyway, it's mixed with `File.Move` which is not async. `File.Copy` is much more performant because it's a system call.

src/EventStore.Core/TransactionLog/Chunks/TFChunkDb.cs
sakno: Valid concern

src/EventStore.Core/Services/Archiver/ChunkPrioritizer.cs
timothycoleman: i think thats with nullable enabled though, x and y are not nullable in ChunkPrioritizer.cs

> public int Compare(Commands.ArchiveChunk x, Commands.ArchiveChunk y)

probably the best thing would be to `#nullable enable` the file (this would make sense as routine for new files)

src/EventStore.Core/Services/Archiver/ArchiverService.cs
```diff
@@ -2,54 +2,188 @@
...
+		_mainBus = mainBus;
+		_archiveStorage = archiveStorage;
 
-	public string Version => "0.0.1";
+		_uncommittedChunks = new();
+		_existingChunks = new();
+		_cts = new();
+		_archiveChunkCommands = Channel.CreateUnboundedPrioritized(
+			new UnboundedPrioritizedChannelOptions<Commands.ArchiveChunk> {
+				SingleWriter = true,
```
timothycoleman, line 49:
```suggestion
				SingleWriter = false,
```
i think writes can come from the main queue but also from `ScheduleExistingChunksForArchiving`

src/EventStore.Core/Services/Archiver/Storage/FileSystemArchiveStorage.cs
shaan1337: good idea, fixed (but not tested yet)

src/EventStore.Core/Services/Archiver/ChunkPrioritizer.cs
shaan1337: fixed :+1:

src/EventStore.Core/Services/Archiver/Storage/FileSystemArchiveStorage.cs
shaan1337: i agree that `File.Move()` is also not async but renames are a much shorter operation than file copies.

chunk files are quite large and could take a few seconds to copy. for e.g. consider the case where the directory is an NFS share, it could take several seconds to copy the chunk and we would be blocking that thread during the entire duration (the time taken to copy the chunk to the archive isn't too critical here as it's done in the background anyway)

src/EventStore.Core/Services/Archiver/ArchiverService.cs
```diff
@@ -2,54 +2,188 @@
...
+			if (chunkInfo.ChunkEndPosition > _replicationPosition)
+				break;
+
+			_uncommittedChunks.Dequeue();
+			ScheduleChunkForArchiving(chunkInfo, "new");
+		}
+	}
+
+	private void ScheduleChunkForArchiving(ChunkInfo chunkInfo, string chunkType) {
+		_archiveChunkCommands.Writer.TryWrite(new Commands.ArchiveChunk {
```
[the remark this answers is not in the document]
timothycoleman: im not sure, is it a documented behaviour of unbounded priority channels that trywrite will never return false? or just an implementation detail

PR #4533, opened 2024-10-16
[DBX-77] Remove deprecated options for 25.2

hayley-jean:
Removed: Deprecated options which had no effect - `AlwaysKeepScavenged`, `GossipOnSingleNode`, `DisableInternalTcpTls`, `OptimizeIndexMerge`.
Removed: Deprecated option `IntIp`, use `ReplicationIp` instead.
Removed: Deprecated option `ExtIp`, use `NodeIp` instead.
Removed: Deprecated option `HttpPort`, use `NodePort` instead.
Removed: Deprecated option `IntTcpPort`, use `ReplicationPort` instead.
Removed: Deprecated option `IntHostAdvertiseAs`, use `ReplicationHostAdvertiseAs` instead.
Removed: Deprecated option `AdvertiseHttpPortToClientAs`, use `AdvertiseNodePortToClientAs` instead.
Removed: Deprecated option `HttpPortAdvertiseAs`, use `NodePortAdvertiseAs` instead.
Removed: Deprecated option `IntTcpPortAdvertiseAs`, use `ReplicationTcpPortAdvertiseAs` instead.
Removed: Deprecated option `IntTcpHeartbeatTimeout`, use `ReplicationHeartbeatTimeout` instead.
Removed: Deprecated option `IntTcpHeartbeatInterval`, use `ReplicationHeartbeatInterval` instead.

Old scavenge is no longer set up in `ClusterVNode`.


timothycoleman (approved):

alexeyzimarev:
We get complains from users that they can hardly find what options are deprecated when doing a jump upgrade over several versions. They need to check each release notes document to compose the list of deprecated settings.

I think it can be solved by having a list of deprecated options in each release docs on a separate page, which would have a table of all options with server versions when the option was added or removed, and what it was replaced with (if that's the case). @coloradomorrissey

coloradomorrissey:
Thanks @alexeyzimarev.  I want to build a "template" with Engineering to capture expectations for release documentation.  This seems applicable.

PR #4558, opened 2024-10-22
Fix config reloading

timothycoleman:
Fixed: Config hot reload (bug introduced since the previous release)

SectionProvider wasn't monitoring the wrapped provider for changes.


shaan1337 (approved):
tested, looks good. the failing test just needs to be fixed.

shaan1337 (approved):

PR #4578, opened 2024-10-28
Treat MACHINE_LIMIT_EXCEEDED as a conclusive license validation error

timothycoleman:
Changed: Treat MACHINE_LIMIT_EXCEEDED as a conclusive license validation error

### DIFF
--- a/src/EventStore.Licensing/Keygen/KeygenLifecycleService.cs
+++ b/src/EventStore.Licensing/Keygen/KeygenLifecycleService.cs
@@ -85,12 +85,7 @@ public sealed class KeygenLifecycleService : IHostedService, IDisposable {
 	async Task<LicenseInfo> Validate(CancellationToken cancellationToken) {
 		var restResponse = await _client.ValidateLicense(_fingerprint, cancellationToken);
 		if (!TryGetParsedResponse(restResponse, "License validation", out var response, out var error)) {
-			// todo: other conclusive errors?
-			return error.Code switch {
-				"LICENSE_INVALID" or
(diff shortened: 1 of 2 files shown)


src/EventStore.Licensing/Keygen/KeygenLifecycleService.cs
```diff
@@ -259,7 +249,16 @@ static bool TryGetParsedResponse<T>(
...
 static class ErrorExtensions {
-	public static LicenseInfo.Conclusive ToConclusiveInfo(this KeygenError error) {
-		return LicenseInfo.Conclusive.FromError($"{error.Title}. {error.Detail}");
-	}
+	static readonly string[] _conclusiveErrors = [
+		"LICENSE_INVALID",
+		"LICENSE_SUSPENDED",
+		"MACHINE_CORE_LIMIT_EXCEEDED",
+		"MACHINE_LIMIT_EXCEEDED",
+		"MACHINE_PROCESS_LIMIT_EXCEEDED",
```
hayley-jean, line 257: Should there be a test for the `MACHINE_PROCESS_LIMIT_EXCEEDED` case?

hayley-jean (approved):

src/EventStore.Licensing/Keygen/KeygenLifecycleService.cs
timothycoleman: I decided to include it - so far we aren't using this but it gives us the option in the keygen policy if we want to and it's clear what it means

PR #4634, opened 2024-11-15
Changes to HTTP API Documentation for operations requiring authentication

lakshdeepsingheventstore:
Fixed : The documentation for certain operations via the HTTP API mentions that authentication is not required, however these operations do require authentication. For example resetting a user password, creating a new user via HTTP API. This PR addresses these documentation mistakes.


coloradomorrissey (approved):

timothycoleman:
Does it have to be basic auth in particular though?

coloradomorrissey:
@lakshdeepsingheventstore @timothycoleman, is there still an open question about this PR?

lakshdeepsingheventstore:
> Does it have to be basic auth in particular though?

I have tried to remain consistent with the already existing docs. The already existing docs state basic auth for operations such as scavenging, node shutdown etc. Also, for the operations that have been changed in this PR, I have tested all the operations with basic auth and it does work for all the operations changed.

timothycoleman (approved):
interesting, i suspect other mechanisms might work besides basic auth, but if this is consistent with the rest of the docs then that's fine by me and we can revisit it more generally later

PR #4641, opened 2024-11-20
[KDB-454] Add support for Partial Reads

timothycoleman:
Added: Archive: Support for partial reads


src/EventStore.Core.XUnit.Tests/Services/Archive/Storage/ArchiveStorageReaderTests.cs
```diff
@@ -1,16 +1,86 @@
...
+
+		// read the local chunk
+		var localContent = await File.ReadAllBytesAsync(chunkPath);
+
+		// read the uploaded chunk
+		using var chunkStream = await sut.GetChunk(chunkPath, CancellationToken.None);
+		var chunkStreamContent = chunkStream.ToByteArray();
+
+		// then
+		Assert.Equal(localContent, chunkStreamContent);
```
shaan1337, line 36: in the past, i've measured the time to compare byte arrays using `Assert.Equal` and it was much faster to use Span.SequenceEqual: `Assert.True(localContent.Span.SequenceEqual(chunkStreamContent))`

i'm not sure if that's still the case

shaan1337 (approved):

PR #4650, opened 2024-11-25
[KDB-402]: Move log record read operation to async

sakno:
Changed: `TFChunk` read side internal API has moved to async I/O


timothycoleman (approved):
Nice

Approved but it might be sensible to made an adjustment to satisfy the qodana false positive 'failure'

timothycoleman (approved):

PR #4673, opened 2024-12-11
[KDB-566]: Async transformation infrastructure

sakno:
Changed: migration to span-friendly `IncrementalHash` class + transformation infrastructure compatible with async code.

### DIFF
--- a/src/EventStore.Core.Tests/Transforms/BitFlip/BitFlipChunkWriteTransform.cs
+++ b/src/EventStore.Core.Tests/Transforms/BitFlip/BitFlipChunkWriteTransform.cs
@@ -2,6 +2,8 @@
 // Event Store Ltd licenses this file to you under the Event Store License v2 (see LICENSE.md).
 
 using System;
+using System.Threading;
+using System.Threading.Tasks;
 using EventStore.Plugins.Transforms;
 
 namespace EventStore.Core.Tests.Transforms.BitFlip;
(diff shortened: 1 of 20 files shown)


src/EventStore.Core.Tests/Transforms/BitFlip/BitFlipChunkWriteTransform.cs
```diff
@@ -14,20 +16,19 @@ public ChunkDataWriteStream TransformData(ChunkDataWriteStream dataStream) {
...
 		var alignedSize = GetAlignedSize(chunkHeaderAndDataSize + footerSize, alignmentSize);
 		var paddingSize = alignedSize - chunkHeaderAndDataSize - footerSize;
-		if (paddingSize > 0) {
-			var padding = new byte[paddingSize];
-			_transformedStream.ChunkFileStream.Write(padding);
-			_transformedStream.ChecksumAlgorithm.TransformBlock(padding, 0, padding.Length, null, 0);
-		}
+
+		return paddingSize > 0
+			? _transformedStream.WriteAsync(new byte[paddingSize], token)
```
shaan1337, line 25: the bits in the padding will flipped as it's going through the transformed stream, but i think that it doesn't really matter here as we're not testing if the padding has only zeroes

sakno: Correct. That's why I left this code as-is. Padding for production code is zeroed.

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
```diff
@@ -605,17 +602,20 @@ public async ValueTask VerifyFileHash(CancellationToken token) {
...
+		VerifyHash(footer.MD5Hash, md5);
 
-		// Perf: use hardware accelerated byte array comparison
-		if (!footer.MD5Hash.SequenceEqual(hash))
-			throw new HashValidationException();
+		static void VerifyHash(ReadOnlySpan<byte> expected, IncrementalHash actual) {
+			Span<byte> buffer = stackalloc byte[ChunkFooter.ChecksumSize];
+
+			// Perf: use hardware accelerated byte array comparison
+			if (!actual.TryGetCurrentHash(buffer, out var bytesWritten)
```
shaan1337, line 615: is it better to use `actual.GetCurrentHash(...)` instead of `actual.TryGetCurrentHash(...)` here ? as we guarantee that the destination buffer is large enough to hold the checksum.

src/EventStore.Core/Util/MD5Hash.cs
```diff
@@ -59,14 +59,12 @@ public static async ValueTask ContinuousHashFor(HashAlgorithm md5, Stream s, int
 
 		var buffer = ArrayPool<byte>.Shared.Rent(4096);
 		try {
-			long toRead = count;
-			while (toRead > 0) {
-				int read = await s.ReadAsync(buffer.AsMemory(0, (int)Math.Min(toRead, buffer.Length)), token);
-				if (read is 0)
+			for (int bytesRead; count > 0L; count -= bytesRead) {
+				bytesRead = await s.ReadAsync(buffer.AsMemory(0, (int)Math.Min(count, buffer.Length)), token);
```
shaan1337, line 63:
```suggestion
				bytesRead = await s.ReadAsync(buffer.AsMemory(0, (int)Math.Min(count, 4096)), token);
```

per our earlier discussion, i agree that `ArrayPool` guarantees that an array of length 4096 is returned here as it's a power of 2, but it's an implementation detail, isn't it ? (i don't see it mentioned in the API docs)

shaan1337 (approved):
Approved, with 2 small improvement suggestions

src/EventStore.Core/TransactionLog/Chunks/TFChunk/TFChunk.cs
sakno: From perf point of view, they are the same. `GetCurrentHash` throws an exception if the size is not enough, we can avoid that branch.

sakno: Fixed

src/EventStore.Core/Util/MD5Hash.cs
sakno: Also, if there was a buffer larger than 4k rented previously, the pool returns that buffer. It can be 8k, 16k, or any other size. Larger buffer could improve I/O performance, because we can read more bytes at the cost of single system call, which is more beneficial.

timothycoleman (approved):
Shaan's review has been dismissed by the new changes, reapproving as a proxy :)

PR #4676, opened 2024-12-13
[KDB-447] Skip chunk merging on all nodes when Archive is enabled

timothycoleman:
Changed: Disable chunk merging on all nodes when Archive is enabled

### DIFF
--- a/src/EventStore.Core/Services/Archive/ArchiveOptions.cs
+++ b/src/EventStore.Core/Services/Archive/ArchiveOptions.cs
@@ -4,7 +4,6 @@
 namespace EventStore.Core.Services.Archive;
 
 public class ArchiveOptions {
-	public bool Enabled { get; init; } = true;
 	public StorageType StorageType { get; init; } = StorageType.None;
 	public FileSystemOptions FileSystem { get; init; } = new();
 	public S3Options S3 { get; init; } = new();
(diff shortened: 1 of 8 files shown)


src/EventStore.Core/Services/Archive/ArchiveOptions.cs
```diff
@@ -4,7 +4,6 @@
 namespace EventStore.Core.Services.Archive;
 
 public class ArchiveOptions {
-	public bool Enabled { get; init; } = true;
```
shaan1337, line 7: i think that it's still useful to have an explicit `Enabled` property for a few reasons:
- it's consistent with the configuration options of other plugins
- a user may want to temporarily disable archiving. to do so, the user can just change `Enabled` to `false` without needing to comment out the whole archive config section (but `StorageType: None` is less easy to remember)

timothycoleman: yeah true, although the part i was uncomfortable with was having the Enabled bool default to true while having the whole archiving mechanism disabled 'by default' if there is no config. It does make sense but it's a little bit complicated. Would it be alright with you if we set the Enabled bool to false by default? The user would have to explicitly enable it as well as setting the storage type, but it's workable for me

shaan1337 (approved):

PR #4679, opened 2024-12-13
[KDB-567] Deal with requested scavenge threads being out of range

timothycoleman:
Fixed: Scavenge no longer fails when requested threads is out of the range 1-4

Previously if you specified more than 4 threads in the scavenge request, the scavenge will proceed until the chunk execution phase and then stop with an error about the pool being empty. The solution is to run the scavenge again with fewer threads (1-4) and it will continue from where it left off. This fix clamps the threads to the valid range, the same as old scavenge does.

### DIFF
--- a/src/EventStore.Core/TransactionLog/Scavenging/Stages/ChunkExecutor.cs
+++ b/src/EventStore.Core/TransactionLog/Scavenging/Stages/ChunkExecutor.cs
@@ -40,7 +40,13 @@ public class ChunkExecutor<TStreamId, TRecord> : IChunkExecutor<TStreamId> {
 		_chunkSize = chunkSize;
 		_unsafeIgnoreHardDeletes = unsafeIgnoreHardDeletes;
 		_cancellationCheckPeriod = cancellationCheckPeriod;
-		_threads = threads;
+
+		if (threads < 1 || threads > TFChunkScavenger.MaxThreadCount) {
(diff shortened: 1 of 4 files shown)


src/EventStore.Core/TransactionLog/Scavenging/Stages/ChunkExecutor.cs
```diff
@@ -40,7 +40,13 @@ public ChunkExecutor(
 		_chunkSize = chunkSize;
 		_unsafeIgnoreHardDeletes = unsafeIgnoreHardDeletes;
 		_cancellationCheckPeriod = cancellationCheckPeriod;
-		_threads = threads;
+
+		if (threads < 1 || threads > TFChunkScavenger.MaxThreadCount) {
+			_logger.Warning("SCAVENGING: Number of threads specified ({SpecifiedThreads}) is too high, clamping to {AdjustedThreads}",
```
hayley-jean, line 45: This is pretty pedantic, but if the provided thread count is 0 or -1, the value would be clamped to 1 instead of `MaxThreadCount`

Perhaps it would be better to move this log line to after `_threads` are set, and simply log a warning that the thread count was clamped if `threads != _threads`?

timothycoleman: yeah that's much better, thanks!

hayley-jean (approved):

PR #4706, opened 2024-12-23
Added impact of resetting projection

CindyLoMinChung:
Added impact of resetting projections base on gothub issue https://github.com/EventStore/documentation/issues/687

### DIFF
--- a/docs/server/features/projections/README.md
+++ b/docs/server/features/projections/README.md
@@ -87,3 +87,7 @@ etc. If anyone can append to the emitted streams, then the projection would have
 in terms of processing. Therefore, it can no longer trust that the projection itself emitted that event or if
 something else did.
 
+### Impact of resetting projections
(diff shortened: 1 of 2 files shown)


hayley-jean:
The text looks good, but perhaps it would be better to add it to the `Introduction to projections` doc under [Limitations](https://github.com/EventStore/EventStore/blob/d1bd648afc4432f2b4ce801350566e65ae6281be/docs/server/features/projections/README.md#limitations) rather than in the custom user projections docs?

I say this because resetting a projection isn't limited to custom projections. While it's somewhat rarer that you would do it for a system projection, it is still possible, and sometimes required for recovery scenarios

coloradomorrissey:
@CindyLoMinChung, on line 91, please update to read "EventStoreDB" (as opposed to EventStore).  Sorry to be a stickler, but I am trying to keep naming conventions consistent.  Once resubmitted, I'll approve and merge.

coloradomorrissey (approved):

PR #4707, opened 2024-12-24
[KDB-589] Rename configuration prefixes and sections

hayley-jean:
- Rename `EventStore` prefixes and config sections to `Kurrent`.
- Add fallback environment variable and command line sources which load options provided with the `EventStore` prefix and changes the prefix to `Kurrent`.
- Rename default locations from `eventstore` to `kurrentdb`.
- Rename default config file from `eventstore.conf` to `kurrentdb.conf`.

Still to do:
- Support `EventStore` section in json config files.
- Add a warning when options are provided with the `EventStore` prefix or section.


src/EventStore.Core/Util/DefaultFiles.cs
```diff
@@ -5,9 +5,8 @@
 
 namespace EventStore.Core.Util;
 
-
 public static class DefaultFiles {
     public static readonly string DefaultConfigFile = RuntimeInformation.IsWindows
         ? string.Empty
-        : "eventstore.conf";
+        : "kurrentdb.conf";
```
YoEight, line 11: Shouldn't we have some grace period with that file? It breaks basically our `.deb` file and docker configuration in the wild.

What about issuing a deprecation notice when `eventstore.conf` is used instead? We could get rid completely of the `eventstore.conf` in the next release?

hayley-jean: @YoEight Yeah, I'll add support for still using the `eventstore.conf` file if `kurrentdb.conf` doesn't exist

Do you think we should provide the same grace period for the default locations such as `/etc/eventstore` and `/var/lib/eventstore` as well? That's trickier, but should still be doable

timothycoleman (approved):

timothycoleman (approved):

PR #4713, opened 2025-01-03
[KDB-439]: Convert `TFChunk.MarkForDeletion` to async

sakno:
Changed: Convert `TFChunk.MarkForDeletion` to async to avoid sync-over-async when flushing the stream


sakno:
Closed as not needed

PR #4715, opened 2025-01-03
[KDB-437]  Minor bug fix, unwrap AggregateException correctly

timothycoleman:
Fixed: Improved user-friendliness of message if exception is thrown while opening the database

now that we are calling Parallel.ForEachAsync instead of Parallel.ForEach

### DIFF
--- a/src/EventStore.Core/ClusterVNode.cs
+++ b/src/EventStore.Core/ClusterVNode.cs
@@ -1586,9 +1586,9 @@ public class ClusterVNode<TStreamId> :
 		void StartNodeUnwrapException(IApplicationBuilder app) {
 			try {
 				StartNode(app);
-			} catch (AggregateException aggEx) {
+			} catch (AggregateException aggEx) when (aggEx.InnerException is { } innerEx) {
 				// We only really care that *something* is wrong - throw the first inner exception.
-				throw aggEx.InnerException;
+				throw innerEx;
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


src/EventStore.Core/ClusterVNode.cs
```diff
@@ -1583,6 +1583,15 @@ void ConfigureNode(IApplicationBuilder app) {
 					$"Unknown {nameof(options.Database.Transform)} specified: {options.Database.Transform}");
 		}
 
+		void StartNodeUnwrapException(IApplicationBuilder app) {
+			try {
+				StartNode(app);
+			} catch (AggregateException aggEx) {
+				// We only really care that *something* is wrong - throw the first inner exception.
+				throw aggEx.InnerException;
```
sakno, line 1591: There is Qodana warning about possible NRE. It's better to write `throw aggEx?.InnerException ?? aggEx` or add **when** filter to **catch** statement like `when (aggEx.InnerException is { } innerEx)`

sakno: Another approach is to use list pattern matching:  `when (aggEx.InnerExceptions is [var innerEx, ..])`

sakno (approved):

src/EventStore.Core/ClusterVNode.cs
timothycoleman: ah yeah, thanks :+1:

PR #4717, opened 2025-01-07
[KDB-437]: Abstract away the file system for existing chunks

sakno:
Changed: Abstract away the file system for existing chunks


timothycoleman:
For future reference this is not an abstraction of local/remote, it is a gathering together of the code that accesses the file system (Directory, Path classes etc), although it does open the opportunity for drop in replacements if we ever want later.

timothycoleman (approved):

PR #4739, opened 2025-01-10
[KDB-587] Don't rearchive logical chunks that are already present in the archive, unless if it's a switched chunk

shaan1337:
Fixed: Don't rearchive logical chunks that are already present in the archive, unless if it's a switched chunk


shaan1337:
After discussion with @timothycoleman , given that it's a quite rare occurrence (it can happen mainly when upload of a scavenged/merged chunk to the archive is interrupted), the optimization is not worth it due to the additional complexity

PR #4748, opened 2025-01-17
[KDB-430]: Implement monotonous archiving process

sakno:
Changed: Implement monotonous archiving process


src/EventStore.Core/Services/Archive/Archiver/ArchiverService.cs
```diff
@@ -3,240 +3,116 @@
...
+	private long _replicationPosition; // volatile
+	private volatile CancellationTokenSource _cts;
 
+	// systemStart
 	public ArchiverService(
 		ISubscriber mainBus,
 		IArchiveStorage archiveStorage,
-		IChunkUnmerger chunkUnmerger) {
-		_mainBus = mainBus;
+		IChunkRegistry<IChunkBlob> chunkChunkManager) {
```
shaan1337, line 41:
```suggestion
		IChunkRegistry<IChunkBlob> chunkManager) {
```

src/EventStore.Core/Services/Archive/Archiver/ArchiverService.cs
```diff
@@ -3,240 +3,116 @@
...
+		mainBus.Subscribe<SystemMessage.BecomeShuttingDown>(this);
 	}
 
-	public void Handle(SystemMessage.ChunkLoaded message) {
-		if (!message.ChunkInfo.IsCompleted)
-			return;
-
-		// queue chunk for archiving
-		_existingChunks[Path.GetFileName(message.ChunkInfo.ChunkLocator)!] = message.ChunkInfo;
+	public void Handle(SystemMessage.SystemStart message) {
```
shaan1337, line 55: the archiver node may need to truncate when it joins the leader, so it might not be safe to start archiving at this point

src/EventStore.Core/Services/Archive/Archiver/ArchiverService.cs
```diff
@@ -3,240 +3,116 @@
...
 	private readonly IArchiveStorage _archive;
-	private readonly Queue<ChunkInfo> _uncommittedChunks;
-	private readonly ConcurrentDictionary<string, ChunkInfo> _existingChunks;
-	private readonly CancellationTokenSource _cts;
-	private readonly Channel<Commands.ArchiveChunk> _archiveChunkCommands;
-	private readonly IChunkUnmerger _chunkUnmerger;
+	private readonly CancellationToken _lifetimeToken;
+	private readonly AsyncAutoResetEvent _archivingSignal;
+	private readonly IChunkRegistry<IChunkBlob> _chunkManager;
+	private readonly ConcurrentBag<ChunkInfo> _switchedChunks;
```
shaan1337, line 31: would it be better if we use a set data structure, using the `ChunkEndNumber` as key?

1. it would ensure we can iterate on the switched chunks in order when archiving them
2. consider the situation where the archive storage is down for a long time (e.g a few days). in that case `_archive.StoreChunk` will take a long time to complete. in the mean time, it may have been possible that the archive was scavenged several times. so we would be uploading the scavenged chunks multiple times when the archive storage comes back up. with the set data structure, there wouldn't be duplicates.

sakno: > it would ensure we can iterate on the switched chunks in order when archiving them

The order still not guaranteed. The background job can take switched chunk `1` to upload it, in the same time `ChunkSwitched` can happen concurrently and add chunk `0`.

src/EventStore.Core/Services/Archive/Archiver/ArchiverService.cs
timothycoleman: it's ok because its guarded by the replication checkpoint though?

src/EventStore.Core/Services/Archive/Archiver/ArchiverService.cs
timothycoleman: it's a good point about not uploading multiple times though

src/EventStore.Core/Services/Archive/Archiver/ArchiverService.cs
sakno: I'm not sure, but the truncator runs at very early stage of the node, in `ClusterVNode.StartNode` method. According to the source code, `TruncateDb()` happens before `_mainQueue.Start();`. I can assume that `SystemStart` happens after `TruncateDb()`.

timothycoleman: it's a different part of the truncation that shaan is concerned about, when it makes the decision to truncate rather the actual truncation.

still, i think its ok because we're checking the replication checkpoint, unless @shaan1337 sees it differently

src/EventStore.Core/Services/Archive/Archiver/ArchiverService.cs
timothycoleman: actually this becomes a bit complicated, lets leave the deduplication out for now and reconsider it when we figure out the details of scavenging the archive

src/EventStore.Core/Services/Archive/Archiver/ArchiverService.cs
shaan1337: i think that it's fine right now, as the replication checkpoint is initialized to -1. but say in the future, somebody comes along and changes the replication checkpoint to be initialized to another value (e.g the index checkpoint, which is also a valid initialization value), then the behaviour could be wrong:

e.g consider the case where the (leader's and follower's) data is restored from a backup but the archiver is still ahead of them: the archiver would already start uploading chunks to the archive before it joins the cluster, then it would truncate.

i'm not sure if we need to worry about that case for now though.

timothycoleman: yeah i dont think we need to worry about that, it makes sense to me for the archiver to be allowed to archive anything that it knows is replicated. if we want to stop it from doing that because we've restored a backup to multiple nodes, then we ought to restore that backup to the archiver node as well

shaan1337 (dismissed):
nice work :+1:

timothycoleman (approved):

timothycoleman (approved):

PR #4761, opened 2025-01-23
Improved log record write performance in TFChunk

sakno:
Changed: Improved log record write performance in TFChunk. When the buffered stream has enough space in its internal buffer, the write can be done synchronously.


timothycoleman:
I get about 170k events written per second with or without this change

non real world scenario:

```
>>> wrfl 2 100000 1 4 100 t
```
single node, insecure

timothycoleman (approved):

PR #4799, opened 2025-01-31
[KDB-621] Validate against attempts to set metadata for the "" stream

timothycoleman:
Fixed: Setting stream metadata for "" stream

"" has never been a valid stream name. Attempting to set the metadata for it results in an attempt to write to the stream "$$" which, until now, has been a valid stream name.

However, writing to $$ involves checking on the "" stream, to see if it is soft deleted. This results in the storage writer exiting which shuts down the server to avoid a 'sick but not dead' scenario

This PR makes "$$" an invalid stream name, and so the attempt to write to it is rejected at an early stage.

Depends on: https://github.com/EventStore/EventStore/pull/4800
Fixes: https://github.com/EventStore/EventStore/issues/4767


shaan1337 (approved):
tested, crash no longer happens with the fix

PR #4854, opened 2025-02-24
Update mongo.md - add tutorial link

mzhai20:
Added tutorial link


alexeyzimarev (changes requested):
Please avoid using absolute URLs when linking to the docs. First, the URL is wrong, it should be `docs.kurrent.io`. Second, it should be `/tutorials/...` so preview build will work.

alexeyzimarev:
Repositories don't matter as it all builds to one website.
About preview builds, if the link is absolute, the link won't work in Cloudflare preview websites as it points to production. It should work for previews too so we can check if everything is ok before publishing it to production.

coloradomorrissey (approved):

coloradomorrissey (approved):

PR #4857, opened 2025-02-25
Prevent swallowing of gRPC exceptions that lead to hanging writes on client

timothycoleman:
Reverting f9791cf - Reduce gRPC log level on the server

If we swallow the exception the call terminates successfully, which the client isn't expecting in case of failure. e.g. in the dotnet client the user's Append call will not be terminated.

There is room to discuss later if we should be converting some of these to RpcExceptions, to avoid logging Errors for them.

### DIFF
--- a/src/EventStore.Core/Services/Transport/Grpc/Streams.BatchAppend.cs
+++ b/src/EventStore.Core/Services/Transport/Grpc/Streams.BatchAppend.cs
@@ -40,17 +40,8 @@ partial class Streams<TStreamId> {
 			requestStream, responseStream,
 			context.GetHttpContext().User, _maxAppendSize, _writeTimeout,
 			GetRequiresLeader(context.RequestHeaders));
-		try {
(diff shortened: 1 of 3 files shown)


alexeyzimarev:
I'd say we catch `OperationCanceledException` and throw `RpcException.Timeout` instead

```
public static class RpcExceptions {
	public static Exception Timeout(string message) => new RpcException(new Status(StatusCode.Aborted, $"Operation timed out: {message}"));
```

timothycoleman:
> I'd say we catch `OperationCanceledException` and throw `RpcException.Timeout` instead

Maybe but I think it's part of a wider discussion. There are multiple endpoints we might want to do that for, it might be better in an interceptor. For the minute I think just 'putting it back' is the right incremental step

alexeyzimarev (approved):

hayley-jean (approved):

PR #4868, opened 2025-02-27
[DEVEX-240] Add connector factory metrics and update metrics configuration

w1am:
1. Add sink and connector metrics to connectors factory
2. Update metrics config to include these meters


w1am:
Addressed in https://github.com/kurrent-io/EventStore/pull/4960

PR #4875, opened 2025-02-28
Don't return 401 for the /oauth path

hayley-jean:
This only fixes the oauth login for the legacy UI, it doesn't add support for OAuth to the new embedded UI


hayley-jean:
Closing in favour of https://github.com/EventStore/EventStore/pull/4876

PR #4883, opened 2025-03-04
[KDB-697]: Removed buffer copy

sakno:
Fixed: default implementation of `ChunkDataReadStream.Read(Span<byte>)` doesn't copy the input span to the rented buffer anymore


timothycoleman (approved):

timothycoleman:
(note no need for cherry pick to master because master already has this change and more)

PR #4892, opened 2025-03-09
[KDB-710] Make sure not to send invalid checkpoint when filtered $all subscription transitions to live

timothycoleman:
Before this change it was possible to incorrectly send a checkpoint of (ulong.max, ulong.max) to a filtered $all subscription if
- the filtered subscription started from the beginning of the database and
- none of the events in the database matched the filter and
- checkpoint interval is greater than the number of events in the database.

In these circumstances the first checkpoint would be when the subscription transitions to live, but the checkpoint would be (-1, -1)
i.e. outside of the log. The checkpoint datastructure we send to the client is only able to express positions within the log, not
a position before the beginning, and it gets translated to (ulong.max, ulong.max).

This fix updates the checkpoint to a valid value before sending it when transitioning to live, since we do have one.

We also add a failsafe to make sure that invalid checkpoints are not sent under any circumstances.

### DIFF
--- a/src/EventStore.Core.Tests/AssertEx.cs
+++ b/src/EventStore.Core.Tests/AssertEx.cs
@@ -10,6 +10,11 @@ using NUnit.Framework;
 namespace EventStore.Core.Tests;
 
 public static class AssertEx {
+	public static T IsType<T>(object o) {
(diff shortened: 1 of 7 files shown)


shaan1337 (approved):
nice catch!

PR #4908, opened 2025-03-13
Update persistent-subscriptions.md

coloradomorrissey:
updated stream name for parked messages based on this issue:  https://github.com/EventStore/EventStore/issues/4139#issuecomment-2722254741

@hayley-jean or @timothycoleman could you confirm the feedback in the issue is correct?


coloradomorrissey:
There is some updated language, but the primary purpose of this PR is the following (which is referenced in the issue above):  
image

hayley-jean (approved):

PR #4914, opened 2025-03-17
Flush writer before chaser during Archive catchup

timothycoleman:
To avoid the (unlikely) scenario of the chaser moving ahead of the writer


timothycoleman:
Closing this for now
- if there is a power failure between flushing the chaser and flushing the writer, then the archive catchup process will resume next start and it copes with the chaser being ahead of the writer.
- the chaser being too far behind the writer causes performance problems on startup

PR #4915, opened 2025-03-17
Nest messages source files for easier navigation

alexeyzimarev:


timothycoleman:
I'm not sure I understand why this makes navigation easier. At least on VS it just nests the files under the parent file, but there is no indication that expanding that file will reveal more files

PR #4921, opened 2025-03-17
Enable claim mapping

alexeyzimarev:
When mapping claims is disabled, the policy that matches roles (and uses full claim namespace) doesn't work.


alexeyzimarev:
Branch name is wrong btw

src/EventStore.Core/ClusterVNodeStartup.cs
```diff
@@ -213,9 +213,8 @@ public void ConfigureServices(IServiceCollection services) {
 					}
 					o.SaveTokens = true;
 					o.GetClaimsFromUserInfoEndpoint = true;
-					o.MapInboundClaims = false;
+					// o.MapInboundClaims = false;
```
hayley-jean, line 216:
```suggestion
```

hayley-jean (dismissed):

hayley-jean (approved):

PR #4951, opened 2025-03-27
[KDB-678] Rename unix socket from `eventstore.sock` to `kurrent.sock`

hayley-jean:
Check for and clean up `eventstore.sock` in addition to `kurrent.sock` at startup.
Update the namespaces for the redaction proto.

I've labeled this as a breaking change because it won't work with older versions of the redactor tool, but there is no breaking change between server versions.

This will only be compatible with the `KurrentDB.Redactor` tool

### DIFF
--- a/src/KurrentDB/Program.cs
+++ b/src/KurrentDB/Program.cs
@@ -308,17 +308,11 @@ static void TryListenOnUnixSocket(ClusterVNodeHostedService hostedService, Kestr
 	}
 
 	try {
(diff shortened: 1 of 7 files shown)


timothycoleman (approved):

src/KurrentDB/Program.cs
bretthowell714-source, line 1: Who ever wrote this did a hell of a job

bretthowell714-source: Thanks

PR #4975, opened 2025-04-03
[KDB-744] Fix: Filtered $all subscriptions checkpoint behaviour change

shaan1337:
Fixes: https://github.com/kurrent-io/EventStore/issues/4970

If a user only stored the commit position of a checkpoint yielded by a filtered $all subscription and used it as both the commit position and prepare position when resuming the subscription, then it is possible under certain circumstances (https://github.com/kurrent-io/EventStore/issues/4970#issuecomment-2779277520) for the subscription to miss an event.

In previous versions (24.10 and earlier) this technique of storing only the commit position was viable as long as explicit transactions had not been written to the database. This PR reinstates that behaviour.

Thanks to @johnbywater for detecting this bug!

A regression was introduced in https://github.com/kurrent-io/EventStore/pull/4892 :

- We are using `completed.CurrentPos` as checkpoint but this position can be the position of an event that hasn't been processed yet - more precisely when `completed.CurrentPos == completed.NextPos`

- `completed.CurrentPos` can be equal to `completed.NextPos` when an empty page was read

- An empty page can be read when the previous page that was read was full and we've reached the end of the log


timothycoleman (approved):

timothycoleman:
The circumstances when this might affect a user are quite specific (https://github.com/kurrent-io/EventStore/issues/4970#issuecomment-2779277520 detailed here).

Most notably, to be affected, the user needs to be relying on only the commit position of the checkpoint and not the full checkpoint (commit/prepare pair). Doing that is reasonable enough though, and it used to work in 24.10, so definitely a good thing to be reinstating here.

hayley-jean (approved):

PR #4981, opened 2025-04-04
Release/v25.0

coloradomorrissey:


timothycoleman:
It's probably best not to pull between general branches like this, because someone might merge a change into v25 that we don't want to include in master.

It's possible to create a branch where v25 currently is though and open a pr for that branch into master

PR #4992, opened 2025-04-09
[KDB-576] Rename remaining assemblies

hayley-jean:
I tried to get away with as few functional changes as possible. Commits titled `Rename ...` are as close to Rider's automated refactoring and a find/replace as I could get them. The `Clean namespaces` commit is an automated code reformat and cleanup to reorder and remove namespaces that may have got out of order.

The following have not been renamed:
- gRPC and TCP proto namespaces
- LogV3 projects and namespaces
- EventStore.Core.Tests

If we want to rename `EventStore.Core.Tests`, then it should be done in a separate PR as it breaks some of the ClientAPI tests.

This PR skips the OAuth tests, as these need to be reworked to support both the name changes and the restructuring of the plugins in KurrentDB.

Of note:
- System projections use [the full type name](https://github.com/kurrent-io/EventStore/blob/0dbf9121bf6f95e87e868d7d20a2592e55497043/src/EventStore.Projections.Core/Services/IProjectionStateHandler.cs#L91) of the standard projection class, so they had to be changed to support both the old and the new namespaces. These namespaces aren't changed/upgraded when they are loaded or saved. This will affect creating system projections in [the legacy UI](https://github.com/kurrent-io/EventStore.UI.Commercial/blob/30b8a65602355e4d0ee90d5cad367c5b12b4fc3b/src/js/modules/projections/controllers/ProjectionsStandardCtrl.js#L10) as well.
- The [AfterInitializationIsFilledWithZeroes](https://github.com/kurrent-io/EventStore/blob/0dbf9121bf6f95e87e868d7d20a2592e55497043/src/EventStore.Core.XUnit.Tests/DataStructures/ProbabilisticFilter/PersistenceStrategyTests.cs#L29) test case in `PersistenceStrategyTests` started failing after the rename, but only when run alongside the other persistent strategy tests. This was fixed by giving this test case a unique file name, but I'm unsure what about the rename caused it to break in the first place.
- The [legacy meter names](https://github.com/kurrent-io/EventStore/blob/0dbf9121bf6f95e87e868d7d20a2592e55497043/src/EventStore.Common/Configuration/MetricsConfiguration.cs#L128) should be unchanged and still supported.
- I don't anticipate any issues with node-node communication or replication since the protos are unchanged, but we should thoroughly test forwarding of HTTP requests, and gossip and replication between nodes of different versions.

### DIFF
--- a/src/KurrentDB.Core.XUnit.Tests/DataStructures/ProbabilisticFilter/PersistenceStrategyTests.cs
+++ b/src/KurrentDB.Core.XUnit.Tests/DataStructures/ProbabilisticFilter/PersistenceStrategyTests.cs
@@ -6,7 +6,7 @@ using Xunit;
 
 namespace KurrentDB.Core.XUnit.Tests.DataStructures.ProbabilisticFilter;
 
-public abstract class PersistenceStrategyTests {
+public abstract class PersistenceStrategyTests : DirectoryPerTest<PersistenceStrategyTests> {
 	protected abstract IPersistenceStrategy GenSut(
 		long size, bool create, string fileName);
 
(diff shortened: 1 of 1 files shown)


timothycoleman (approved):
Replication is working fine cross version (new -> old and vice versa). As is http forwarding and grpc forwarding

timothycoleman:
(i imagine the qodana check will start working once this has been merged, since the qodana config in master is looking for the old solution file that doesn't exist on this branch)

PR #5000, opened 2025-04-09
Compare v25 with v24

Deflaimun:


Deflaimun:
This was done just to compare the two versions. Looks like we have 31 conflicts in docs between versions. Mostly around EventStoreDB vs. KurrentDB nomeclature between versions. Kinda of expected. 

Let this be reference for future merge conflicts.

PR #5025, opened 2025-04-25
[KDB-769] Support more flexible virtual stream readers

oskardudycz:
Today, we support $mem streams by registering specific readers for particular streams prefixed with `$mem-`. The issue is that for indexes being available for reads like streams, we need to support the situation when one reader (category index reader, for example) would be able to handle multiple streams with a specific prefix, where the stream names it is able to handle aren’t known upfront and we can only rely on a certain prefix.

Resolving this issue means: 

- supporting readers by prefix rather than full stream name
- supporting multiple prefixes, not just one, because for index “streams” we’d not use `$mem` but something like `$idx`


timothycoleman (approved):
:+1:

PR #5027, opened 2025-04-28
[KDB-777] Add Properties to PrepareLogRecord

hayley-jean:
Based on PR https://github.com/kurrent-io/KurrentDB/pull/5001

- Add support for writing Properties (a map of string and dynamic values) to PrepareLogRecords.
    - This increases the PrepareRecordVersion to v2.
    - If a PrepareLogRecord does not contain Properties then it is written as a v1 record to allow for a rolling upgrade from previous versions.
- Metadata received on events appended over gRPC is copied into Properties when creating the write request, and converted back into Metadata when reading the records over gRPC.

**Compatibility**:
A rolling upgrade can be done from older versions provided records are not written with properties until all nodes in the cluster are on the same version. It is safest to upgrade the leader of the cluster last to ensure that no v2 records are written during the upgrade.

It will not be possible to downgrade once records containing properties have been written to the database.


timothycoleman (approved):
:+1:

PR #5035, opened 2025-04-30
Added abstraction for Secondary Indexes building

oskardudycz:
Added needed abstractions for handling secondary indexes and plugged them to plugin. This is the first step before plugging the real DuckDB implementation. 

1. `SecondaryIndexSubscription` - responsible for listening to log and indexing events that were stored there.
2. Indexing and committing happens in the implementation of `ISecondaryIndexProcessor`.
3. `SecondaryIndexCheckpointTracker` is responsible for triggering index commits. Commits happen when the maximum size of pending batches is reached or when the maximum time has elapsed since the previous commit.
4. Commit settings can be configured through options: CheckpointCommitBatchSize and CheckpointCommitDelayMs exposed through database settings.
5. All of that is handled by the `SecondaryIndexBuilder` hosted service, responsible for running the subscription.

You can see how it will work for DuckDB in simplified fake in-memory implementations.

Besides that plugged integration tests based on the Connectors plugins. Expanded them to make it possible to run them in parallel.

**_Note:_ I also made `_clusterNodeMutex` NOT be acquired for the in-memory database in `ClusterVNodeHostedService`  (so just like the db block). I also released it on dispose it accordingly.** This enables having multiple instances of `ClusterVNodeHostedService` running in integration tests. Previous tests in connectors assumed that only one `ClusterVNodeHostedService` would be running in the assembly fixture, but that doesn't allow for testing different configuration settings (e.g. if the plugin was enabled or disabled).

TODO:
- [ ] - add more tests,
- [ ] - polish implementation,
- [ ] - expand description, docs if needed


src/Connectors/KurrentDB.Connectors.Tests/ClusterVNodeApp.cs
```diff
@@ -54,6 +55,9 @@ public class ClusterVNodeApp : IAsyncDisposable {
             .With(x => x.Services.AddSingleton<IHostedService>(esdb))
             .With(x => configureServices?.Invoke(x.Services));
 
+        if(useRandomPort)
+			builder.WebHost.ConfigureKestrel(serverOptions => serverOptions.ListenAnyIP(Random.Shared.Next(5001, 6000)));
```
oskardudycz, line 59: Needed not to have conflicts on an open port for the cluster vNode. I didn't pass a specific port, as for now, it's not needed.

src/KurrentDB/ClusterVNodeHostedService.cs
```diff
@@ -355,6 +360,9 @@ public Task StopAsync(CancellationToken cancellationToken) =>
 		Node.StopAsync(cancellationToken: cancellationToken);
 
 	public void Dispose() {
+		if (_clusterNodeMutex is { IsAcquired: true })
```
oskardudycz, line 363: @timothycoleman added this release on dispose when mutex was acquired to allow cleaning it up for parallel tests.

src/KurrentDB.Core.Tests/Helpers/MiniClusterNode.cs
```diff
@@ -194,7 +194,7 @@ public MiniClusterNode(string pathname, int debugIndex, IPEndPoint internalTcp,
 						options.Application.AllowAnonymousEndpointAccess,
 						options.Application.AllowAnonymousStreamAccess,
 						options.Application.OverrideAnonymousEndpointAccessForGossip).Create(components.MainQueue)]))),
-			virtualStreamReaders ?? [],
+			null,
```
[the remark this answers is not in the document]
oskardudycz: Fine by me, I'll do it :+1:

src/Connectors/KurrentDB.Connectors.Tests/ClusterVNodeApp.cs
timothycoleman: there is a way to make bind to some available port i think, possibly specifying port 0?

src/KurrentDB.Core.Tests/Helpers/MiniClusterNode.cs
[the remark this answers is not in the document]
oskardudycz: Updated in https://github.com/kurrent-io/KurrentDB/pull/5035/commits/df4531fcf7a346964f706a7ed05cf92c201b2d39.

timothycoleman (dismissed):

timothycoleman (dismissed):

timothycoleman (approved):

PR #5037, opened 2025-05-02
Naive POC: recycle buffers used for chunk cache

timothycoleman:
Added: experiment


timothycoleman:
No need for this, the problem we encountered wasn't related to chunk allocations / fragmentation but windows Commit limit being exceeded

PR #5045, opened 2025-05-09
Increase log record max size to 32mb

timothycoleman:
Added: Larger log record limit


hayley-jean:
Closing this as we're not going to increase the log record max size. This was a temporary measure to fix a projection that was trying to write a too-large state event.

We may consider decreasing the default `MaxProjectionStateSize` so that users have some warning and some time to fix issues with projections before the max log record size is reached.

PR #5049, opened 2025-05-12
Shaan1337/add properties to msa

timothycoleman:


timothycoleman:
-> https://github.com/kurrent-io/KurrentDB/pull/5050

PR #5055, opened 2025-05-13
Fix spelling of Committed/er/ing

timothycoleman:


timothycoleman:
(draft for now in case of merge conflicts with index work)

PR #5089, opened 2025-05-20
Add release notes for 25.0.1

hayley-jean:


Deflaimun (approved):
Thanks !

docs/server/release-schedule/release-notes.md
```diff
@@ -6,6 +6,75 @@ order: 1
 
 This page contains the release notes for KurrentDB 25.0
 
+## [25.0.1](https://github.com/EventStore/EventStore/releases/tag/v25.0.1)
+
+20 May 2025
+
+### Fix filtered $all subscriptions checkpoint behavior (PR [#4991](https://github.com/kurrent-io/KurrentDB/pull/4991))
+
+Fixed an issue where an empty page read would cause filtered $all subscriptions to send an incorrect checkpoint with commit position equal to the end position of the log and prepare position 0. This can cause an event to the missed if the client stores only the commit position instead of both.
```
Deflaimun, line 15:
```suggestion
Fixed an issue where an empty page read would cause filtered `$all` subscriptions to send an incorrect checkpoint with commit position equal to the end position of the log and prepare position 0. This can cause an event to the missed if the client stores only the commit position instead of both.
```

dharmabumstead: I suspect you mean to say 'This can cause an event to be missed' instead of 'This can cause an event to the missed'.

timothycoleman (approved):

PR #5093, opened 2025-05-22
Added the Default Secondary Index implementation with DuckDB

oskardudycz:
:construction: :construction_worker:  **WIP**

That includes DuckDB setup, default index, processor, reader, etc.

Also extended the database reader's code to support reading real events based on the indexed position.


alexeyzimarev:
Part of #5115

PR #5094, opened 2025-05-22
Fixed race condition in Secondary Index Checkpoint Tracker

oskardudycz:
In the CI where the thread number is limited, 500ms may be too short for high threading tests, increased the max waiting time in `SecondaryIndexCheckpointTrackerTests`.

### DIFF
--- a/src/KurrentDB.SecondaryIndexing.Tests/Subscriptions/SecondaryIndexCheckpointTrackerTests.cs
+++ b/src/KurrentDB.SecondaryIndexing.Tests/Subscriptions/SecondaryIndexCheckpointTrackerTests.cs
@@ -25,7 +25,7 @@ public class SecondaryIndexCheckpointTrackerTests {
 			tracker.Increment();
 		}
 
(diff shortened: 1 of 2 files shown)


src/KurrentDB.SecondaryIndexing.Tests/Subscriptions/SecondaryIndexCheckpointTrackerTests.cs
```diff
@@ -290,5 +290,5 @@ public async Task Disposal_During_Active_Commit_Allows_Commit_To_Complete() {
 		Assert.True(commitCompleted);
 	}
 
-	private static readonly TimeSpan HalfOfASecond = TimeSpan.FromMilliseconds(500);
+	private static readonly TimeSpan FiveSeconds = TimeSpan.FromSeconds(5);
```
sakno, line 293: Nit: It's better not to reflect the value as well as the type of the variable in its name. Just choose to `CommitSignalTimeout`

timothycoleman (approved):

PR #5115, opened 2025-05-26
Secondary indexes (streams, category, event type)

alexeyzimarev:
* Add DuckDB init code
* Add subscription for building the default index tables
* Add index builders (stream, event type, category)
* Add index readers


src/KurrentDB.Core/Services/Transport/Enumerators/Enumerator.StreamSubscription.cs
```diff
@@ -35,6 +35,7 @@ public sealed class StreamSubscription<TStreamId> : StreamSubscription {
 		private readonly bool _resolveLinks;
 		private readonly ClaimsPrincipal _user;
 		private readonly bool _requiresLeader;
+		readonly int _readBatchSize;
```
shaan1337:
```suggestion
		private readonly int _readBatchSize;
```

src/KurrentDB.SecondaryIndexing.LoadTesting/Observability/MessagesBatchObserver.cs
```diff
@@ -0,0 +1,36 @@
...
+			EventTypes.AddOrUpdate(messagesByType.Key, eventTypeCount, (_, current) => current + eventTypeCount);
+		}
+
+		Interlocked.Add(ref _totalCount, messagesCount);
+	}
+
+	public ConcurrentDictionary<string, long> Categories { get; } = new();
+	public ConcurrentDictionary<string, long> EventTypes { get; } = new();
+
+	public long TotalCount => _totalCount;
```
shaan1337:
```suggestion
	public long TotalCount => Interlocked.Read(ref _totalCount);
```

src/KurrentDB.SecondaryIndexing.LoadTesting/Environments/DuckDB/RawQuackMessageBatchAppender.cs
```diff
@@ -0,0 +1,82 @@
...
+				throw;
+			}
+		}
+
+		return ValueTask.CompletedTask;
+	}
+
+	public ValueTask DisposeAsync() {
+		 _dbDataSource.Dispose();
+		 _defaultIndexAppender.Dispose();
```
shaan1337, line 78: should the appender be disposed before the db data source? or is it fine?

src/KurrentDB.SecondaryIndexing/Indexes/Category/CategoryIndexProcessor.cs
```diff
@@ -0,0 +1,66 @@
...
+
+namespace KurrentDB.SecondaryIndexing.Indexes.Category;
+
+internal class CategoryIndexProcessor {
+	readonly Dictionary<string, int> _categories;
+	readonly Dictionary<int, long> _categorySizes = new();
+	readonly DuckDbDataSource _db;
+
+	int _seq;
+	public long LastIndexesPosition { get; private set; }
```
shaan1337:
```suggestion
	public long LastIndexedPosition { get; private set; }
```

src/KurrentDB.SecondaryIndexing/Indexes/Category/CategoryIndexProcessor.cs
```diff
@@ -0,0 +1,66 @@
...
+
+		_db.Pool.ExecuteNonQuery<AddCategoryStatementArgs, AddCategoryStatement>(new(id, categoryName));
+		LastIndexesPosition = resolvedEvent.Event.LogPosition;
+		return new(id, 0);
+	}
+
+	public long GetLastEventNumber(int categoryId) =>
+		_categorySizes.TryGetValue(categoryId, out var size) ? size : ExpectedVersion.NoStream;
+
+	public int GetCategoryId(string categoryName) =>
```
shaan1337, line 60: suggestion: rename to `TryGetCategoryId(string categoryName, out int categoryId)` for better semantics

src/KurrentDB.SecondaryIndexing/Indexes/Category/CategoryIndexProcessor.cs
```diff
@@ -0,0 +1,66 @@
...
+		_categories.TryGetValue(categoryName, out var categoryId) ? categoryId : -1;
+
+	private static string GetStreamCategory(string streamName) {
```
shaan1337, line 63: suggestion: rename to `TryGetStreamCategory(string streamName, out string category)` for better semantics

src/KurrentDB.SecondaryIndexing/Indexes/Category/CategoryIndexProcessor.cs
```diff
@@ -0,0 +1,66 @@
...
+		var sequences = connection.Query<(int Id, long Sequence), GetCategoriesMaxSequencesQuery>();
+		foreach (var sequence in sequences) {
+			_categorySizes[sequence.Id] = sequence.Sequence;
+		}
+
+		_seq = _categories.Count > 0 ? _categories.Values.Max() : 0;
+	}
+
+	public SequenceRecord Index(ResolvedEvent resolvedEvent) {
+		var categoryName = GetStreamCategory(resolvedEvent.OriginalStreamId);
```
shaan1337, line 37: currently stream names that don't contain a `-` use the stream name as the category. is this intentional? as there can be millions of stream names and it'll use a lot of memory in the _categories/_categorySizes dictionaries. (or are the streams already filtered at this point?)

src/KurrentDB.SecondaryIndexing/Indexes/Category/CategoryIndexReader.cs
```diff
@@ -0,0 +1,47 @@
...
+	CategoryIndexProcessor processor,
+	IReadIndex<string> index,
+	QueryInFlightRecords<CategoryRecord> queryInFlightRecords
+) : SecondaryIndexReaderBase(index) {
+	protected override long GetId(string streamName) {
+		if (!streamName.StartsWith(CategoryIndex.IndexPrefix)) {
+			return ExpectedVersion.Invalid;
+		}
+
+		var categoryName = streamName[8..];
```
shaan1337:
```suggestion
		var categoryName = streamName[CategoryIndex.IndexPrefix.Length..];
```

src/KurrentDB.SecondaryIndexing.Tests/IntegrationTests/IndexingDisabledTests.cs
```diff
@@ -0,0 +1,27 @@
...
+[Trait("Category", "Integration")]
+[Collection("SecondaryIndexingPluginDisabled")]
+public class IndexingDisabledTests(
+	SecondaryIndexingDisabledFixture fixture,
+	ITestOutputHelper output
+) : SecondaryIndexingTestBase(fixture, output) {
+	private readonly string[] _expectedEventData = ["""{"test":"123"}""", """{"test":"321"}"""];
+
+	[Fact]
+	public async Task Index_streams_should_not_be_found() {
```
shaan1337, line 18:
```suggestion
	public async Task index_streams_should_not_be_found() {
```

use snake_case or PascalCase, for consistency with other tests

src/KurrentDB.Core/Services/SubscriptionsService.cs
```diff
@@ -307,6 +309,30 @@ private static Message CloneReadRequestWithNoPollFlag(Message originalRequest) {
 		ReissueReadsFor(message.Event.EventStreamId, message.CommitPosition, message.Event.EventNumber);
 	}
 
+	async ValueTask IAsyncHandle<StorageMessage.SecondaryIndexRecordCommitted>.HandleAsync(StorageMessage.SecondaryIndexRecordCommitted message, CancellationToken token) {
+		_lastSeenSecondaryIndexLogPosition = message.LogPosition;
+
+		var streamCategory = SystemStreams.GetStreamCategory(message.Event.EventStreamId ?? message.Event.EventStreamId);
```
shaan1337:
```suggestion
		var streamCategory = SystemStreams.GetStreamCategory(message.Event.EventStreamId ?? message.Event.EventStreamId);
```

`message.Event.EventStreamId` is specified twice

src/KurrentDB.SecondaryIndexing/Indexes/EventType/EventTypeIndexProcessor.cs
```diff
@@ -0,0 +1,61 @@
...
+
+		_db.Pool.ExecuteNonQuery<AddEventTypeStatementArgs, AddEventTypeStatement>(new(id, eventTypeName));
+		LastIndexedPosition = resolvedEvent.Event.LogPosition;
+		return new(id, 0);
+	}
+
+	public long GetLastEventNumber(int eventTypeId) =>
+		_eventTypeSizes.TryGetValue(eventTypeId, out var size) ? size : ExpectedVersion.NoStream;
+
+	public int GetEventTypeId(string eventTypeName) =>
```
shaan1337, line 61:
```suggestion
	public bool TryGetEventTypeId(string eventTypeName, out int eventTypeId) =>
```
(for better semantics)

src/KurrentDB.SecondaryIndexing/Indexes/EventType/EventTypeIndexReader.cs
```diff
@@ -0,0 +1,47 @@
...
+	EventTypeIndexProcessor processor,
+	IReadIndex<string> index,
+	QueryInFlightRecords<EventTypeRecord> queryInFlightRecords
+) : SecondaryIndexReaderBase(index) {
+	protected override long GetId(string streamName) {
+		if (!streamName.StartsWith(EventTypeIndex.IndexPrefix)) {
+			return ExpectedVersion.Invalid;
+		}
+
+		var eventTypeName = streamName[8..];
```
shaan1337:
```suggestion
		var eventTypeName = streamName[EventTypeIndex.IndexPrefix.Length..];
```

src/KurrentDB.SecondaryIndexing.LoadTesting/Assertions/DuckDb/DuckDbIndexingSummaryAssertion.cs
```diff
@@ -0,0 +1,34 @@
...
+public class DuckDbIndexingSummaryAssertion(DuckDbDataSource db): IIndexingSummaryAssertion {
+	public async ValueTask IndexesMatch(IndexingSummary summary) {
+		await AssertCategoriesAreIndexed(summary);
+		await AssertEventTypesAreIndexed(summary);
+	}
+
+	private ValueTask AssertCategoriesAreIndexed(IndexingSummary summary) {
+		var categories = db.Pool.Query<CategorySummary, GetCategoriesSummaryQuery>();
+
+		if (categories.All(c => summary.Categories.ContainsKey(c.Name)))
```
shaan1337, line 20: should a check also be done to see if summary.Categories contains extra categories that were not in the DB? (same applies for event types below)

src/KurrentDB.SecondaryIndexing/Indexes/Default/DefaultIndexProcessor.cs
```diff
@@ -0,0 +1,124 @@
...
+	public IEnumerable<T> QueryInFlightRecords<T>(Func<InFlightRecord, bool> query, Func<InFlightRecord, T> map) {
+		for (var i = 0; i < _inFlightRecordsCount; i++) {
+			var current = _inFlightRecords[i];
+			if (query(current)) yield return map(current);
+		}
+	}
+
+	protected override void Dispose(bool disposing) {
+		if (disposing) {
+			Commit();
```
shaan1337, line 147: i'm not sure if it's a good idea to commit when disposing - the disposal could be happening due to a fatal error in the server

src/KurrentDB.SecondaryIndexing/Indexes/Diagnostics/SecondaryIndexProgressTracker.cs
```diff
@@ -0,0 +1,113 @@
...
+			Interlocked.Increment(ref _pendingEvents);
+		} catch (Exception exc) {
+			Log.Error(exc, "Error recording metrics of event indexed in secondary index.");
+		}
+	}
+
+	public void RecordAppended(EventRecord eventRecord) {
+		try {
+			Interlocked.Exchange(ref _lastLogPosition, eventRecord.LogPosition);
+			Interlocked.Exchange(ref _lastAppendedAt, eventRecord.TimeStamp.Ticks);
```
shaan1337: there's no guarantee that the timestamp increases monotonically - for example if there's a leader change, some events in a chunk will have used the previous leader's clock while the new events will use the new leader's clock (both clocks should be quite close to each other though)

so, it might be worth adding a check to take the max of the current and new timestamp

src/KurrentDB.SecondaryIndexing/Indexes/Diagnostics/SecondaryIndexProgressTracker.cs
```diff
@@ -0,0 +1,113 @@
...
+			var streamPos = Interlocked.Read(ref _lastLogPosition);
+			var indexedPos = Interlocked.Read(ref _lastIndexedPosition);
+
+			return new Measurement<long>(streamPos - indexedPos);
+		}
+
+		private Measurement<long> ObserveLag() {
+			var lastAppendedAt = Interlocked.Read(ref _lastAppendedAt);
+			var lastIndexedAt = Interlocked.Read(ref _lastIndexedAt);
+			var lag = lastAppendedAt - lastIndexedAt;
```
shaan1337:
```suggestion
			var lag = Math.Max(0L, lastAppendedAt - lastIndexedAt);
```

(related to the above comment - the value could be negative)

src/KurrentDB.Core/Services/Transport/Grpc/Streams.Read.cs
```diff
@@ -208,7 +215,10 @@ private IAsyncEnumerator<ReadResponse> CreateEnumerator(
 					request.Options.ResolveLinks,
 					user,
 					requiresLeader,
-					cancellationToken),
+					// TODO: temp hack, it should be part of the client message in the future
```
shaan1337, line 218: move comment above, near:
```
SystemStreams.IsIndexStream(request.Options.Stream.StreamIdentifier) ? 2000 : Enumerator.DefaultReadBatchSize
```

src/KurrentDB.SecondaryIndexing.Tests/Observability/MessagesBatchObserver.cs
```diff
@@ -0,0 +1,77 @@
...
+	public IndexingSummary Summary { get; }
+}
+
+public record IndexingSummary(
+	IDictionary<string, long> Categories,
+	IDictionary<string, long> EventTypes,
+	long TotalCount
+);
+
+public class StreamsMessagesBatchObserver : IMessagesBatchObserver {
```
shaan1337, line 25: should this unused class be removed? it looks identical to `SimpleMessagesBatchObserver`

src/KurrentDB.SecondaryIndexing.Tests/IntegrationTests/IndexingEnabledTests.cs
```diff
@@ -0,0 +1,166 @@
...
+			await ValidateSubscription(CategoryIndex.Name(category), expectedEvents);
+		}
+	}
+
+	[Fact]
+	public async Task SubscriptionReturnsAllEventsFromEventTypeIndex() {
+		foreach (var eventType in EventTypes) {
+			var expectedEvents = AppendedBatches.ToEventTypeIndexResolvedEvents(eventType);
+			await ValidateSubscription(EventTypeIndex.Name(eventType), expectedEvents);
+		}
```
shaan1337: it might be worth adding a test when the subscription reaches the end of the stream and goes live too: i.e. append a few new events and see if they are received by the subscription

src/KurrentDB.SecondaryIndexing/Subscriptions/SecondaryIndexSubscription.cs
```diff
@@ -1,70 +1,84 @@
...
-			_checkpointCommitBatchSize,
-			_checkpointCommitDelayMs,
-			ct => index.Processor.Commit(ct),
-			linkedCts.Token
-		);
 
-		_subscription = new Enumerator.AllSubscription(
+	public void Subscribe() {
+		var position = indexProcessor.GetLastPosition();
+		var startFrom = position is null ? Position.Start : Position.FromInt64((long)position, (long)position);
```
shaan1337, line 33: Using only the `prepare position` as checkpoint will not work properly if there are explicit transactions in the log (it might actually even move the checkpoint backwards)

It is better to use the full TFPos as checkpoint (`resolvedEvent.EventPosition!.Value` instead of `resolvedEvent.Event.LogPosition`) in DefaultIndexProcessor, CategoryIndexProcessor, etc.

shaan1337: Alternatively if we use only the `commit position` as checkpoint, then if we checkpoint in the middle of an explicit transaction,  resuming it from that checkpoint (`commit position`, `commit position`) will skip some events in the explicit transaction (which is bad) as the checkpoint is exclusive. The checkpointing mechanism would need to make the checkpoint inclusive when resuming (which adds complexity).

It is thus simpler to just use the full TFPos as checkpoint.

shaan1337: (a test probably needs to be added for that case too - i.e with explicit transactions)

src/KurrentDB.SecondaryIndexing.Tests/PluginTests.cs
```diff
@@ -4,35 +4,33 @@
...
 		// when
-		TestPluginStartup.Configure(sut);
+		using var app = TestPluginStartup.Configure(sut);
 
 		// then
 		Assert.False(sut.Enabled);
 	}
 
-	[Theory]
+	[Theory(Skip = "TODO: make a proper configuration")]
```
shaan1337, line 25: fix TODO

src/KurrentDB.SecondaryIndexing.Tests/Subscriptions/CommitterTests.cs
```diff
@@ -0,0 +1,280 @@
+// // Copyright (c) Kurrent, Inc and/or licensed to Kurrent, Inc under one or more agreements.
```
shaan1337, line 1: Will this file be deleted? or must the tests be adjusted to pass?

src/KurrentDB.SecondaryIndexing.Tests/Indexes/DefaultIndexProcessorTests.cs
```diff
@@ -0,0 +1,462 @@
...
+	}
+
+	public override Task DisposeAsync() {
+		_processor.Dispose();
+		return base.DisposeAsync();
+	}
+}
+
+public class CleanUpTests {
+	[Fact(Skip = "TODO: Check why is it failing")]
```
shaan1337, line 448: fix TODO / test

src/KurrentDB.Core/Services/Storage/ReaderIndex/IndexBackend.cs
```diff
@@ -96,9 +115,10 @@ public SystemSettings GetSystemSettings() {
 		return _systemSettings;
 	}
 
-	public struct EventNumberCached(int version, long? lastEventNumber) {
+	public struct EventNumberCached(int version, long? lastEventNumber, long? secondaryIndexId) {
 		public readonly int Version = version;
 		public readonly long? LastEventNumber = lastEventNumber;
+		public readonly long? SecondaryIndexId = secondaryIndexId;
```
shaan1337, line 121: would it be better if the secondary index plugin has its own, separate cache?

Right now, even if the plugin is disabled, each item in the cache will consume an additional 8 bytes which is a significant proportion of the cache item's size in memory (a 50% increase or more depending on padding, etc.) 

Having its own cache would also remove the need to inject `IndexBackend` in DI.

src/KurrentDB.SecondaryIndexing/Indexes/EventType/EventTypeIndexProcessor.cs
```diff
@@ -0,0 +1,63 @@
...
+using KurrentDB.Core.Data;
+using KurrentDB.Core.Messages;
+using KurrentDB.SecondaryIndexing.Readers;
+using KurrentDB.SecondaryIndexing.Storage;
+using static KurrentDB.SecondaryIndexing.Indexes.EventType.EventTypeSql;
+
+namespace KurrentDB.SecondaryIndexing.Indexes.EventType;
+
+internal class EventTypeIndexProcessor {
+	private readonly Dictionary<string, int> _eventTypes;
```
shaan1337, line 14: usually the number of event types should be relatively low (hundreds or thousands max), but there may be some users  who are using the system in an unconventional way. we could consider using an LRU cache here (or adding a metric to monitor the size of the dictionary)

src/KurrentDB.SecondaryIndexing/Indexes/Category/CategoryIndexProcessor.cs
```diff
@@ -0,0 +1,67 @@
...
+using KurrentDB.Core.Data;
+using KurrentDB.Core.Messages;
+using KurrentDB.SecondaryIndexing.Readers;
+using KurrentDB.SecondaryIndexing.Storage;
+using static KurrentDB.SecondaryIndexing.Indexes.Category.CategorySql;
+
+namespace KurrentDB.SecondaryIndexing.Indexes.Category;
+
+internal class CategoryIndexProcessor {
+	private readonly Dictionary<string, int> _categories;
```
shaan1337, line 14: same here as for event types - consider using an LRU cache or adding a metric

shaan1337 (commented):
some things that need particular attention, the other suggestions are mostly minor/quick fixes:
- SecondaryIndexSubscription: Use the full (commit, prepare) position as checkpoint for the index subscription: Using only the prepare position could result in incorrect behaviour when there are explicit transactions in the log. And using the commit position could result in missing events when resuming.
- New Metrics: there's no guarantee that events have a monotonically increasing timestamp (non increasing timestamps must be handled)
- IndexBackend: Use a separate cache for the secondary index plugin otherwise this will result in a performance regression when the plugin is disabled. (the `LastEventNumber` cache will be able to fit significantly less items than before). This will also avoid the need to inject `IndexBackend` into DI.
- CategoryIndexProcessor/EventTypeProcessor: consider using an LRU cache to store the event types and categories (or adding a metric)
- CategoryIndexProcessor: verify the behaviour of `GetStreamCategory(string streamName)` when the stream doesn't contain a `-`
- DefaultIndexProcessor: verify if it's safe to Commit() when disposing (as disposal may be triggered due to failure)
- DefaultIndexReader/CategoryIndexReader/etc: what guarantees do we have that the in flight records and the db records are mutually exclusive? do the records need to be de-duplicated? (consider the case when the index is being read while the appender is being flushed)

other considerations:
- How will backups of the duckdb database be done while the system is live? (apart from volume snapshots)

src/KurrentDB.SecondaryIndexing/SecondaryIndexingPlugin.cs
```diff
@@ -18,39 +26,70 @@ namespace KurrentDB.SecondaryIndexing;
...
-				sp.GetRequiredService<ISubscriber>(),
-				options
+		var options = configuration
+			.GetSection($"{KurrentConfigurationKeys.Prefix}:SecondaryIndexing:Options")
+			.Get<SecondaryIndexingPluginOptions>() ?? new();
+		services.AddSingleton(options);
+
+		services.AddSingleton<DuckDbDataSourceOptions>(sp => {
+			var dbConfig = sp.GetRequiredService<TFChunkDbConfig>();
+			var connectionString = $"Data Source={Path.Combine(dbConfig.Path, "index.db")};";
```
shaan1337: consider storing the database under a new folder: `index/secondary-indexes` instead of the main db directory

src/KurrentDB.SecondaryIndexing/Indexes/Default/DefaultIndexReader.cs
```diff
@@ -0,0 +1,33 @@
...
+) : SecondaryIndexReaderBase(index) {
+	protected override long GetId(string streamName) => 0;
+
+	protected override long GetLastIndexedSequence(long id) => processor.LastSequence;
+
+	protected override IEnumerable<IndexedPrepare> GetIndexRecords(long _, long fromEventNumber, long toEventNumber) {
+		var range = db.Pool.Query<(long, long), AllRecord, DefaultSql.DefaultIndexQuery>((fromEventNumber, toEventNumber));
+		if (range.Count < toEventNumber - fromEventNumber + 1) {
+			var inFlight = inFlightRecords.TryGetInFlightRecords(fromEventNumber, toEventNumber);
+			range.AddRange(inFlight);
```
shaan1337, line 23: what guarantees do we have that the in flight records and the db records are mutually exclusive? do the records need to be de-duplicated? (consider the case where the index is being read while the appender is being flushed)

src/KurrentDB.SecondaryIndexing/Subscriptions/SecondaryIndexSubscription.cs
alexeyzimarev: I don't like that. We can read prepare records using one position. If we have a deficient enumerator that is incapable of figuring out where to start reading - that must be fixed. TFPos spreads out everywhere, including client APIs, and it just confuses people.

src/KurrentDB.Core/Services/Storage/ReaderIndex/IndexBackend.cs
alexeyzimarev: We do not plan to disable the plugin. The feature flag is there in case we can't make it work properly. Practically, eviction of last event number should be on par with eviction of stream id. Ideally, this cache needs to be revisited anyway but for a different reason. LRU cache doesn't serve well in this case, it should be LFU-based eviction. We all know that, so we should at least look at it. If that happens, I am fine with moving it out.

src/KurrentDB.SecondaryIndexing/Indexes/EventType/EventTypeIndexProcessor.cs
alexeyzimarev: We don't really need a metric for that. We will have this information available for the user.

src/KurrentDB.SecondaryIndexing/Indexes/Category/CategoryIndexProcessor.cs
alexeyzimarev: Same as I replied above.

src/KurrentDB.SecondaryIndexing/SecondaryIndexingPlugin.cs
alexeyzimarev: Right now it's not desirable. We'll consider it when e make sure the index can be reliably rebuild from the log, and even if it makes sense.

src/KurrentDB.SecondaryIndexing/Indexes/Default/DefaultIndexReader.cs
alexeyzimarev: Because in-flight records get cleared on appender flush.

src/KurrentDB.SecondaryIndexing/Subscriptions/SecondaryIndexSubscription.cs
shaan1337: As it is, the checkpoint is derived from duckdb by taking the `max(log_position)`, so it's fine. But the in-memory checkpoint (`LastIndexedPosition`) should be updated to be `max(LastIndexedPosition, newPosition)` as `newPosition` can move backwards when there are explicit transactions.

src/KurrentDB.SecondaryIndexing/Indexes/Default/DefaultIndexReader.cs
shaan1337: > Because in-flight records get cleared on appender flush.

consider these cases where the calls are interleaved:

case 1:

1. DefaultIndexReader: GetIndexRecords()::db.Pool.Query
2. DefaultIndexProcessor: Appender is flushed
3. DefaultIndexProcessor: InFlight records are cleared
4. DefaultIndexReader: GetIndexRecords()::inFlightRecords.TryGetInFlightRecords()

the result will be missing some records as the in-flight records were already cleared

case 2:

as far as i understand, duckdb's appender flush is not atomic.

say there are 10 events: 1-5 are in duckdb, 6-10 are in-flight
there is an incoming index read to read events 1-10

1. DefaultIndexProcessor: Appender flushes events 6,7
2. DefaultIndexReader: GetIndexRecords()::db.Pool.Query // receives 1-7
3. DefaultIndexReader: GetIndexRecords()::inFlightRecords.TryGetInFlightRecords() // receives 6-10
4. DefaultIndexProcessor: Appender flushes events 8,9,10
5. DefaultIndexProcessor: InFlight records are cleared

the records 6 & 7 will be present in the result twice

src/KurrentDB.SecondaryIndexing/SecondaryIndexingPlugin.cs
shaan1337: note that the `index` folder can be configured to be on a separate disk (and it's a recommended practice).
right now the IOPS/disk space will be shared between KurrentDB & DuckDb index operations.

src/KurrentDB.Core/Services/Storage/ReaderIndex/IndexBackend.cs
shaan1337: another aspect is that it could be a breaking change:
In some production environments, `StreamInfoCacheCapacity` may have already been tuned according to the workload. After upgrading, the caches will suddenly be taking 50%-70% more memory than before which could lead to OOMs or performance degradation if there isn't a lot of free memory left.

PR #5135, opened 2025-06-06
Collect together release notes for 22.10 and 22.6

hayley-jean:


docs/server/release-notes.md
```diff
@@ -0,0 +1,271 @@
...
+
+We have reduced the computational complexity of the methods run at startup to lower the impact of the regression and improve startup times.
+
+This fix was introduced in EventStoreDB 21.10.2.
+
+### Breaking changes
+
+There are no breaking changes in terms of the database format, configuration, or rolling upgrades.
+
+However, there are some breaking changes since 22.10.0 that will affect external clients trying to connect to EventStoreDB 22.10.0 nodes:
```
timothycoleman: this may need updating having been copy-pasted from somewhere else?

docs/server/release-notes.md
```diff
@@ -0,0 +1,271 @@
...
+CertificateSubjectName: eventstoredb-node
+```
+
+#### Packages for Ubuntu 22.04
+
+This release includes packages for Ubuntu 22.04. While we have done preliminary testing on these packages, we have not yet run a full QA on them.
+
+As such we consider these packages as a preview, where you can use them at your own risk. If we do not find any issues with these packages, they will be officially supported with the release of 22.10.0 later this year.
+
+### Breaking changes for community gRPC clients
```
timothycoleman, line 246: Just checking we do want this here and also in the 22.10 section?

hayley-jean: These were both from the original blog article, but I don't think we really need them here. I'll just remove them

PR #5144, opened 2025-06-11
Add Server Configuration Option for TCP read expiry: TcpReadTimeoutMs

timothycoleman:
In gRPC this is already driven by the request itself. In TCP we have no such mechanism. We opt for this solution instead of adding it to the TCP protocol to make the change smaller, safer, and easier to roll out while still being useful.

### DIFF
--- a/src/EventStore.Core.Tests/Services/Transport/Tcp/TcpClientDispatcherTests.cs
+++ b/src/EventStore.Core.Tests/Services/Transport/Tcp/TcpClientDispatcherTests.cs
@@ -347,7 +347,7 @@ namespace EventStore.Core.Tests.Services.Transport.Tcp {
 				resolveLinkTos: true,
 				requireLeader: true);
 
-			var package = new TcpPackage(TcpCommand.ReadStreamEventsForward, Guid.NewGuid(), dto.Serialize());
+			var package = new TcpPackage(TcpCommand.ReadStreamEventsBackward, Guid.NewGuid(), dto.Serialize());
 
 			var message = _dispatcher.UnwrapPackage(
 				package: package,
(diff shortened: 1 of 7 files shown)


src/EventStore.Core.Tests/Services/Transport/Tcp/TcpClientDispatcherTests.cs
```diff
@@ -295,5 +281,212 @@ private EventRecord CreateLinkEventRecord() {
...
+			var readEvent = message as ClientMessage.ReadAllEventsBackward;
+			Assert.AreEqual(2000, readEvent.CommitPosition);
+			Assert.AreEqual(1000, readEvent.PreparePosition);
+			Assert.AreEqual(6, readEvent.MaxCount);
+			Assert.AreEqual(true, readEvent.ResolveLinkTos);
+			Assert.AreEqual(true, readEvent.RequireLeader);
+			Assert.That(readEvent.Expires, Is.EqualTo(DateTime.UtcNow + _readTimeout).Within(_tolerance));
+		}
+
+				[Test]
```
shaan1337: fix indentation

src/EventStore.Core.Tests/Services/Transport/Tcp/TcpClientDispatcherTests.cs
```diff
@@ -295,5 +281,212 @@ private EventRecord CreateLinkEventRecord() {
...
+		[Test]
+		public void unwraps_ReadStreamEventsBackward() {
+			var dto = new ReadStreamEvents(
+				eventStreamId: "my-stream",
+				fromEventNumber: 5,
+				maxCount: 6,
+				resolveLinkTos: true,
+				requireLeader: true);
+
+			var package = new TcpPackage(TcpCommand.ReadStreamEventsForward, Guid.NewGuid(), dto.Serialize());
```
shaan1337: should TcpCommand.ReadStreamEventsForward be TcpCommand.ReadStreamEventsBackward?

src/EventStore.Core.Tests/Services/Transport/Tcp/TcpClientDispatcherTests.cs
timothycoleman:
```suggestion
		[Test]
```

src/EventStore.Core.Tests/Services/Transport/Tcp/TcpClientDispatcherTests.cs
timothycoleman: yes thanks, i'll fix

shaan1337 (approved):

PR #5151, opened 2025-06-16
Allow users to see loaded plugins

alexeyzimarev:
* Options converted to a class with mutable properties, so the options object registered in the DI container stays actual.
* New UI page to show loaded plugins


alexeyzimarev:
It's already merged to the branch

PR #5157, opened 2025-06-19
[KDB-912]: Introduced a new scheduler for the messages

sakno:
Changed: introduced a new scheduler for the messages that is relying on `ThreadPool` and can be scaled automatically based on the resource limits

### DIFF
--- a/src/KurrentDB.Core.Tests/Bus/queued_handler_should.cs
+++ b/src/KurrentDB.Core.Tests/Bus/queued_handler_should.cs
@@ -31,7 +31,7 @@ public abstract class queued_handler_should : QueuedHandlerTestWithNoopConsumer 
 [TestFixture]
 public class queued_handler_threadpool_should : queued_handler_should {
 	public queued_handler_threadpool_should()
-		: base((consumer, name, timeout) =>
+		: base(static (consumer, name, timeout) =>
 			new QueuedHandlerThreadPool(consumer, name, new QueueStatsManager(), new(), false, null, timeout)) {
 	}
 }
--- a/src/KurrentDB.Core.Tests/Bus/when_publishing_to_queued_handler.cs
+++ b/src/KurrentDB.Core.Tests/Bus/when_publishing_to_queued_handler.cs
@@ -85,7 +85,16 @@ public abstract class when_publishing_to_queued_handler : QueuedHandlerTestWithW
 [TestFixture, Category("LongRunning")]
 public class when_publishing_to_queued_handler_threadpool : when_publishing_to_queued_handler {
 	public when_publishing_to_queued_handler_threadpool()
-		: base((consumer, name, timeout) =>
+		: base(static (consumer, name, timeout) =>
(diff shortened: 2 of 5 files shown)


timothycoleman:
Changes Notes:

Transaction File Readers
- There is no longer a pool of `ITransactionFileReaders`, instead there is just one stateless one. The callers pass in Cursor that keeps track of the position.
- `TFReaderLease` is therefore no longer needed and is removed.

Worker Threads
- We are getting rid of the worker thread count. In this PR it is deprecated but remains present.
- There is only one worker bus and it performs work on the thread pool without a concurrency limit because it is on the output side of the server.
- Order is still preserved per connection using `Affinity`.

Reader Threads
- We are keeping `ReaderThreadCount` because it will have an equivalent in rate limiting.
- There is now only one `StorageReaderWorker`
- After this PR a Rate Limiter will be added to limit the number of inprocess reads.

MultiQueuedHandler
- `MultiQueuedHandler` (which distributes messages across a pool of `IQueuedHandlers`) is no longer used.
- Storage Readers and Workers used to use it.
- It is replaced with `ThreadPoolMessageScheduler` which differs in that it does not have a fixed number of queues so
    unrelated messages needn't get stuck queueing behind each other. messages for the same affinity maintain their respective order.
			
QueuedHandlerThreadPool
- `QueuedHandlerThreadPool` is no longer used by the core server (only by projections)
- It is replaced with `ThreadPoolMessageScheduler` with `SynchronizeMessagesWithUnknownAffinity` set to true so that it has the ordering semantics of a queue.
- It is guaranteed that messages published to the `ThreadPoolMessageScheduler` are handled on the thread pool.
- The queues changed to the new implementation are:
	- Workers handler
	- Monitoring queue
	- Subscription queue
	- PersistentSubscription queue
	- Redaction queue
	- StorageReaderService queue
	- StorageWriterQueue
	- MainQueue
	
Cursors
- `ITransactionFileReader` used to maintain it's own position, so we had a pool of them to check out of for use.
- Now the caller maintains the position and passes it into the `TryReadNext/Prev` call as a cursor that the call can mutate.
- We can't pass the cursor in as a `ref long` because the `TryReadNext/Prev` implementations are `async`.
- So we pass in a `ValueReference<long>`, which is a reference to a long stored somewhere with appropriately long lifetime.
- We wrap the `ValueReference<long>` in a `IReadCursor` implementation so that `TryReadNext/Prev` don't _always_ have to use `ValueReference<long>`
- We store the actual `long` in the unmanaged heap to reduce GC pressure.
- We use a `Scope` to release the long when we're done.
- The type parameter `TCursor` in  `TryReadPrev<TCursor>(TCursor)` avoids boxing the cursor struct as would be the case if passed as `IReadCursor`


For follow up in future PR
- Rate limiter for read requests
- Metrics and slow message handling for ThreadPoolMessageScheduler
- Considering replacing cursor mechanism with one where the position is passed into the method as a long and returned out in the result. Such a mechansim is simpler to understand conceptually, but may be more fiddly for the caller to use correctly.

timothycoleman:
(i approve this pr)

timothycoleman:
deprecates the WorkerThreads setting

PR #5185, opened 2025-07-09
[KDB-920]: Initial version of release test automation workflow

sakno:
Changed: Initial version of release test automation workflow


sakno:
Moving to TrainStation repo

PR #5190, opened 2025-07-15
[KDB-803] Only store non-redundant log record properties

timothycoleman:
Log record properties that are already captured elsewhere in the log record need not be stored as properties. They are reconstituted on read.

i.e.
- EventType
- ContentType if it is fully described by the IsJson flag (`application/json` or `application/octet-stream`)


src/KurrentDB.Core/Services/Transport/Grpc/Constants.cs
```diff
@@ -72,6 +72,9 @@ public static class Properties {
 		public static class DataFormats {
 			// can be others (protobuf, avro, ...) but json is the only one the server cares about
 			public const string Json = "json";
+			public const string Bytes = "bytes";
```
RagingKore, line 75: On a future PR we should ensure we use the typed proto enum for the Data Formats.

PR #5191, opened 2025-07-17
Do not populate log record properties on streams v1 API

timothycoleman:
Settled on design where log record properties can only be set on the v2 streams api. Design documented in developers guide (which should be moved to this repo quite soon)

Reject writes that try to specify both properties and metadata


RagingKore (approved):
LGTM

PR #5192, opened 2025-07-17
[KDB-803] Improve serialization of properties to synthetic metadata

timothycoleman:
- Preserve numbers and bools rather than converting to string
- Preserve key names
- Avoid unicode \u002B in timestamp. Format in ISO.
- Avoiding boxing present in DynamicValueMapper TryMapValue<T>
- Only implement the paths we need, make specific to SyntheticMetadata


timothycoleman:
Closing, this is now covered by https://github.com/kurrent-io/KurrentDB/pull/5208

PR #5131, opened 2025-06-04
[DEVEX-278] Schema Registry

w1am:
This PR moves the schema registry logic from the Surge repo into this repo and introduces several improvements and fixes across testing, compatibility checks, and CI configuration.

**Changes:**

* Migrated schema registry codebase from Surge to this repository.
* Added full integration test coverage for all schema registry operations.
* Add support for `BACKWARD_ALL`, `FORWARD_ALL` and `FULL_ALL` compatibility mode.
* Fixed compatibility checker for JSON schemas with nested `$ref` references.
* Increase test coverage for JSON compatibility checker.
* Removed the WaitUntilCaughtUp test utility, which caused DuckDB transaction issues and led to test hangs.
* Updated CI pipeline to support TUnit and MTP.

PR #5193, opened 2025-07-18
Remove AsyncCallbackEnvelope, also Reduce allocations

timothycoleman:


timothycoleman:
some of these were encorporated into the shema reg pr https://github.com/kurrent-io/KurrentDB/pull/5131
the other significant suggestion is now its own pr https://github.com/kurrent-io/KurrentDB/pull/5230

PR #5197, opened 2025-07-22
Lower Scavenge API GET calls to Verbose

timothycoleman:
The auto-scavenge checks on the status of scavenges frequently, which before this PR is producting a lot of logs


rmfaheem:
LGTM :+1:

shaan1337 (approved):

PR #5209, opened 2025-07-24
Store Properties in the Metadata slot

timothycoleman:
- Earlier this release we added properties alongside metadata in log records (unreleased)
- The updated design has properties as a replacement for metadata, essentially a replacement _format_ for metadata
- This PR changes the properties to be stored in the metadata slot for the log record
- Which format is used is determined by a prepareflag
- No bump in prepare log record version is required any more


RagingKore (dismissed):
LGTM

PR #5250, opened 2025-08-29
[DEV-803] Secondary indexes without references

alexeyzimarev:
Second iteration:

* Do not use references (category, event type) and use strings instead, embracing dictionary vectors
* Move from stream reads to read all API with stream prefix as index name, using ReadIndexEvents internal messages
* Add DB Stats and Query pages to the embedded UI
* Thread-safe inflight records
* Inline function to query inflight records
* Removed macros
* Align indexes and schema registry to use the same DuckDB connection pool and database file
* Change ad-hoc queries to use CTEs and query parsing


shaan1337 (approved):
great work @alexeyzimarev @oskardudycz @sakno - this was a true marathon effort!

PR #5251, opened 2025-08-30
[DB-1607] Make InMemoryBus.DefaultSlowMessageThreshold configurable

lakshdeepsingheventstore:
The default value was configured at 48 ms. Any message taking longer than this to be processed was logged as a SLOW BUS/QUEUE MSG in the KurrentDB logs. This PR makes this parameter configurable. If it is not set/configured, it will stick to the previous default of 48 ms.


timothycoleman:
closing in favour of #5440 which is similar but more general

PR #5263, opened 2025-09-09
[DEV-906] OTLP logs export support

alexeyzimarev:
Add support for exporting logs via OTLP protocol using the same configuration as the metrics export


alexeyzimarev:
image

PR #5273, opened 2025-09-18
[DB-1674] Use nuget-license to generate license audit/notice

timothycoleman:
The NOTICE.html (from Qodana) is replaced with NOTICE.md


timothycoleman:
just manually for now since it changes rarely, but some kind of CI integration might be handy later

PR #5282, opened 2025-09-23
Projections code cleanup (no behaviour change)

alexeyzimarev:


timothycoleman:
(put back in draft for failing tests, Alexey is considering whether its worth tracking it down)

PR #5283, opened 2025-09-24
Added rate limiter with two queues to control in-flight requests

sakno:
Changed: Added proper way to handle multiple cancellation tokens without the allocation of CTS for every linkage. Added rate limiter.


sakno:
Closing this PR in favor of #5347.

PR #5284, opened 2025-09-24
[DB-1648]: Handle multiple CTs without the allocation

sakno:
Changed: Added proper way to handle multiple cancellation tokens without the allocation of CTS for every linkage


timothycoleman:
looks good, just needs bumped dotnext version to address the pooling bug then the tests should pass and i can approve

timothycoleman (approved):

PR #5287, opened 2025-09-26
[DB-1640] Support Google cloud storage as archiving backend

shaan1337:
Added: Support for Google cloud storage as archiving backend

### DIFF
--- a/src/KurrentDB.Core/Services/Archive/Storage/Gcp/GcpBlobStorage.cs
+++ b/src/KurrentDB.Core/Services/Archive/Storage/Gcp/GcpBlobStorage.cs
@@ -36,6 +36,9 @@ public class GcpBlobStorage : IBlobStorage {
 	public async ValueTask<int> ReadAsync(string name, Memory<byte> buffer, long offset, CancellationToken ct) {
 		ArgumentOutOfRangeException.ThrowIfNegative(offset);
 
+		if (buffer.IsEmpty)
+			return 0;
+
 		var destination = StreamSource.AsSynchronousStream(new MemoryWriter(buffer));
 		try {
 			await _storageClient.DownloadObjectAsync(


src/KurrentDB.Core/Services/Archive/Storage/Gcp/GcpBlobStorage.cs
```diff
@@ -0,0 +1,87 @@
...
+
+		_storageClient = StorageClient.Create();
+	}
+
+	public async ValueTask<int> ReadAsync(string name, Memory<byte> buffer, long offset, CancellationToken ct) {
+		ArgumentOutOfRangeException.ThrowIfNegative(offset);
+
+		try {
+			await using (var stream = _memoryStreamManager.GetStream(nameof(GcpBlobStorage), requiredSize: buffer.Length)) {
+				await _storageClient.DownloadObjectAsync(
```
sakno: This is suboptiomal, because we need to copy the buffer to the rented temporary stream, and then copy the same buffer back to the passed buffer. Is there any way to remove this temporary destination for the buffer? If not, we can consider a custom implementation of the stream that just wraps `buffer`.

sakno: Let's add TODO or task to finish it in future.

shaan1337: Thanks, I've created a task for it here: https://linear.app/kurrent/issue/DB-1699/avoid-extra-copy-in-gcpblobstorage-readasync

sakno: The fix is submitted to this PR.

PR #5294, opened 2025-10-02
Fixing system store

alexeyzimarev:
* Use one store implementation
* Fix reads
* Add read forwards test
* Fix stream not found when reading backwards


RagingKore (approved):
LGTM.

Not sure if we needed to change the StreamExists operation, but it is not a blocker.

PR #5296, opened 2025-10-07
[DB-1639]: Azure Blob support for KDB Archiving Feature

sakno:
Changed: Added Azure provider to store database archived transaction log files

### DIFF
--- a/src/KurrentDB.Core/Services/Archive/ArchiveOptions.cs
+++ b/src/KurrentDB.Core/Services/Archive/ArchiveOptions.cs
@@ -81,7 +81,7 @@ public class AzureOptions {
 	public string Container { get; init; } = "";
 
 	/// <summary>
-	/// Gets or sets service URL or connection string depending on <see cref="Authentication"/> type
+	/// Gets or sets service URL or connection string depending on <see cref="Authentication"/> type.
 	/// </summary>
 	public string ConnectionStringOrServiceUrl { get; init; } = "";
 


src/Directory.Packages.props
```diff
@@ -23,9 +23,11 @@
     <PackageVersion Include="Eventuous.Testing" Version="0.15.0" />
     <PackageVersion Include="Extensions.MudBlazor.StaticInput" Version="3.2.1" />
     <PackageVersion Include="FluentAssertions" Version="6.12.2" />
-    <PackageVersion Include="FluentAssertions.Analyzers" Version="0.34.1" />
+    <PackageVersion Include="FluentAssertions.Analyzers" Version="0.33.0" />
```
shaan1337, line 28: is the version downgrade intentional? (the new version was probably added to `master` recently)

.github/workflows/build-reusable.yml
```diff
@@ -9,6 +9,13 @@ on:
       arch:
         required: true
         type: string
+      azurite:
```
shaan1337: is this variable (`azurite`) used, or can it be removed?

src/Directory.Packages.props
sakno: Yes, it's intentional. I have incompatibility problems with my local .NET SDK 8.

.github/workflows/build-reusable.yml
sakno: Fixed

src/KurrentDB.Core/Services/Archive/ArchiveOptions.cs
```diff
@@ -72,6 +77,78 @@ public void Validate() {
...
+	public required AuthenticationType Authentication { get; init; }
+
+	public void Validate() {
+		string message;
+		if (string.IsNullOrWhiteSpace(Container)) {
+			message = "Please provide a Container for the Azure archive";
+		} else {
+			switch (Authentication) {
+				case AuthenticationType.ConnectionString when string.IsNullOrWhiteSpace(ConnectionString):
+					message = "Please provide connection string for Azure BLOB";
```
shaan1337:
```suggestion
					message = "Please provide a connection string (using ConnectionStringOrServiceUrl) for the Azure archive's storage account";
```

src/KurrentDB.Core/Services/Archive/ArchiveOptions.cs
```diff
@@ -72,6 +77,78 @@ public void Validate() {
...
+					break;
+				case AuthenticationType.UserAssignedClientId when string.IsNullOrWhiteSpace(UserIdentity):
+					message = "Please provide user identity";
```
shaan1337:
```suggestion
					message = "Please provide a UserIdentity for the Azure archive";
```

src/KurrentDB.Core/Services/Archive/ArchiveOptions.cs
```diff
@@ -72,6 +77,78 @@ public void Validate() {
...
+					break;
+				default:
+					return;
+			}
+		}
+
+		throw new InvalidConfigurationException(message);
+	}
+
+	public enum AuthenticationType {
```
shaan1337, line 132: right now, when providing an invalid `Authentication` type, an unintuitive exception is thrown. it would be great if an error message like this could be displayed:
```
Please provide a valid Authentication type (e.g Default, ConnectionString, SystemAssignedIdentity, UserAssignedClientId) for the Azure archive
```

src/KurrentDB.Core/Services/Archive/ArchiveOptions.cs
```diff
@@ -105,7 +105,7 @@ public void Validate() {
 			message = "Please provide a Container for the Azure archive";
 		} else {
 			switch (Authentication) {
```
shaan1337, line 106: it looks like `Default` is missing in the switch statement:

```
				case AuthenticationType.Default when string.IsNullOrWhiteSpace(ConnectionStringOrServiceUrl):
                        message = "Please provide a Service URL (using ConnectionStringOrServiceUrl) for the Azure archive's storage account";
                        break;
```

right now, it throws an unfriendly exception:
```
System.ArgumentNullException: Value cannot be null. (Parameter 'uriString')
   at System.ArgumentNullException.Throw(String paramName)
   at System.ArgumentNullException.ThrowIfNull(Object argument, String paramName)
   at System.Uri..ctor(String uriString, UriKind uriKind)
...
```

src/KurrentDB.Core/Services/Archive/ArchiveOptions.cs
```diff
@@ -72,6 +77,90 @@ public void Validate() {
 	}
 }
 
+public class AzureOptions {
+	public string Container { get; init; } = "";
+
+	/// <summary>
+	/// Gets or sets service URL or connection string depending on <see cref="Authentication"/> type
```
shaan1337:
```suggestion
	/// Gets or sets service URL or connection string depending on <see cref="Authentication"/> type.
```

PR #5310, opened 2025-10-10
[DB-1650]: Use single file handle in PTable

sakno:
Changed: Index shares the single file handle across all readers


src/KurrentDB.Core/Configuration/ClusterVNodeOptions.cs
```diff
@@ -427,6 +427,7 @@ public record DatabaseOptions {
 		public int InitializationThreads { get; init; } = 1;
 
 		[Description("The number of reader threads to use for processing reads. Set to '0' to scale automatically (Default)")]
+		[Deprecated("This setting no longer has an effect. The workers automatically scale as necessary")]
```
shaan1337:
```suggestion
		[Deprecated("The ReaderThreadsCount parameter has been deprecated as of version 26.0.0 and currently has no effect. The workers automatically scale as necessary."")]
```

PR #5317, opened 2025-10-14
[DB-1713] Query editor: check if data is valid JSON

shaan1337:
Fixed: Query editor: check if data is valid JSON

Non-JSON data is now displayed as `<non-json data>` (instead of resulting in a query error previously)

Screenshot from 2025-10-15 16-55-28

### DIFF
--- a/src/KurrentDB.Core/DuckDB/InlineFunctions.cs
+++ b/src/KurrentDB.Core/DuckDB/InlineFunctions.cs
@@ -41,7 +41,7 @@ public class KdbGetEventSetup(IPublisher publisher) : IDuckDBSetup {
 		DateTime created,
 		ReadOnlyMemory<byte> data,
 		ReadOnlyMemory<byte> meta) {
-		var dataString = data.IsValidUtf8Json() ? Helper.UTF8NoBom.GetString(data.Span) : "\"<raw data>\"";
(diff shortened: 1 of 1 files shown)


timothycoleman:
would `KurrentDB.Common.Utils.IsValidUtf8Json` be a sensible lighter weight choice here

shaan1337:
> would `KurrentDB.Common.Utils.IsValidUtf8Json` be a sensible lighter weight choice here

yup, that's better, thanks. fixed

src/KurrentDB.Core/DuckDB/InlineFunctions.cs
```diff
@@ -41,7 +41,7 @@ private static string AsDuckEvent(string stream,
 		DateTime created,
 		ReadOnlyMemory<byte> data,
 		ReadOnlyMemory<byte> meta) {
-		var dataString = Helper.UTF8NoBom.GetString(data.Span);
+		var dataString = data.IsValidUtf8Json() ? Helper.UTF8NoBom.GetString(data.Span) : "\"<raw data>\"";
```
shaan1337:
```suggestion
		var dataString = data.IsValidUtf8Json() ? Helper.UTF8NoBom.GetString(data.Span) : "\"<non-json data>\"";
```

PR #5347, opened 2025-11-05
[DB-1715]: Simple rate limiter for read requests

sakno:
Changed: Introduced simple rate limiter for read requests, which is off by default.


timothycoleman:
(sorry clicked the wrong button and closed it accidentally)

PR #5366, opened 2025-11-18
[DEV-1010] Add POST method support to list connectors

w1am:
Support querying connectors using `POST /connectors` while still supporting `GET /connectors` with query params.


w1am:
Cancelling this because querying with search parameters is preferable

PR #5497, opened 2026-02-09
[DB-1912] Persistent subscriptions: Fix wrong checkpoint when using pinned strategy

shaan1337:
Persistent subscription consumer strategies that can skip events can produce wrong checkpoints when events are skipped. The pinned consumer strategy is currently the only strategy that can skip events.

This PR fixes the following issues:
- The logic for assigning sequence numbers / previous event positions to `OutstandingMessage` did not take skipped events into consideration - this could result into wrongly assigned sequence numbers / previous event positions and subsequently to wrong checkpoints too.
- The checkpointing mechanism only considers messages in the retry list but skipped events were left in the buffer - this could also result into wrong checkpoints.

### DIFF
--- a/src/KurrentDB.Core.Tests/Services/PersistentSubscription/PersistentSubscriptionTests.cs
+++ b/src/KurrentDB.Core.Tests/Services/PersistentSubscription/PersistentSubscriptionTests.cs
@@ -2543,6 +2543,61 @@ public class CheckpointingWithSkippedEvents {
 	}
 }
 
+public class CheckpointingWithNoMoreCapacity {
+	[Test]
+	public void checkpoint_is_correct_when_message_hits_no_more_capacity_then_succeeds() {
+		var version = 0;
+
+		IPersistentSubscriptionStreamPosition checkpoint = null;
+		var clientEnvelope = new FakeEnvelope();
+		var reader = new FakeCheckpointReader();
+		const string subscriptionStream = "streamName";
+		var sub = new KurrentDB.Core.Services.PersistentSubscription.PersistentSubscription(
+			PersistentSubscriptionToStreamParamsBuilder.CreateFor(subscriptionStream, "groupName")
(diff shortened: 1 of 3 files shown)


timothycoleman:
### Originally (before pinned strategies)

- Events were leaving the _buffer in order, and assigned sequence numbers as they did so.
- Checkpointing worked by looking at the lowest sequence number that is in flight or waiting to be retried. i.e. the lowest one that we've started but not finished.

### With the addition of pinned strategies

- Introduced 'Skipped' consumer push result, where we leave an event in the _buffer (with no sequence number) if its consumer is too busy.
- Events can therefore leave the buffer out of order and receive sequence numbers out of order.
- Checkpointing unchanged, does not consider events in _buffer that were skipped, only in-flight and retries.

Problems (these only affect pinned strategies):
1. skipped events, when they are pushed to client later, will be assigned a sequence number out of order, which is not suitable for checkpointing.
1. skipped events are not visible to the checkpointing mechanism

### This PR

To solve this, we
1. ensure that the sequence numbers are assigned in order even when an event is skipped
1. ensure that skipped events are visible to the checkpoint mechanism

The easiest way to achieve both these things is, when we skip an event we give it a sequence number and move it to the _retry list immediately. This is natural because conceptually the `StreamBuffer` of events ready to push is `_retry ++ _buffer`.

Alternatively, when skipping an event we could give it a sequence number and leave it in the `_buffer`, and give the checkpointer a way to check the buffer (it would only check the initial section that have sequence numbers, not the whole buffer), but this seems unnecessarily complicated.

Note that:

1. Backpressure is maintained because StreamBuffer.CanAccept considers the count of _buffer and _retry
1. Stats will show a higher number of events in retry. this is reasonable enough because although we didn't send them to the client, but we did try to.

timothycoleman (approved):

PR #5499, opened 2026-02-10
[DB-1915]: Use fixed behavior in PoolingBufferedStream class

sakno:
Changed: Use fixed behavior in PoolingBufferedStream class.

Related PR #5490.


timothycoleman (approved):
:+1:

timothycoleman:
this PR removes a workaround now that the library class has the desired behaviour. it is an internal refactor not affecting behaviour

PR #5544, opened 2026-03-07
feat(projections-v2): V2 projection engine with biState support

alexeyzimarev:
## Summary

- Restructure projections into modular assemblies (JavaScript, Management, Shared, V1, V2)
- Implement V2 projection engine with channel-based partition processing, biState support, deferred partition key computation for thread-safe Jint usage, and checkpoint coordination with atomic multi-stream writes
- Fix null-state partition caching bug that caused repeated `Initialize()`/`ProcessPartitionCreated()` calls corrupting shared state in biState projections
- Add account balancer biState integration test validating credit/debit processing, shared state accumulation, and custom partition key filtering

## Key Changes

**V2 Engine Architecture:**
- `PartitionDispatcher` — routes events to partition channels, supports deferred partition key mode for non-thread-safe Jint engine
- `PartitionProcessor` — processes events per partition with biState (shared + per-partition) state management
- `CheckpointCoordinator` — coordinates checkpoints across partitions with atomic multi-stream writes
- `ReadStrategyFactory` — creates read strategies (FilteredAll, StreamSet) from source definitions

**Bug Fixes:**
- `PartitionProcessor`: Track null-state partitions in cache (`Dictionary<string, string?>`) to prevent re-initialization
- `AccountBalancerSpecTests`: Use `ConcurrentQueue` instead of `ConcurrentBag` to preserve write ordering

## Test plan

- [x] Account balancer biState integration test passes (verifies ESDBB-01 balance=550, ESDBB-S01 balance=300, shared state totalBalance=850)
- [x] Existing projections tests still pass
- [x] Unit tests for PartitionDispatcher, OutputBuffer, and engine lifecycle


alexeyzimarev:
## Assessment of Qodo Review Items

### Items that are NOT bugs (dismiss):

**1. "Create missing ApiCallback"** — `ProjectionManagement` is not an API v2 service. It uses `CallbackEnvelope`, the standard pattern for the management layer. Mixing in `ApiCallback` would cross abstraction boundaries.

**2. "Create throws generic exception"** — Same reasoning. This isn't an API v2 endpoint, so `RequestValidator` doesn't apply here.

**3. "Engine v2 always faults — CreateProcessingPhases throws NotSupportedException"** — **Intentional by design.** The message says `"V2 engine does not use v1 processing phases. Use ProjectionEngineV2 directly."` The strategy class exists as a marker for the management layer. The V2 engine is invoked through `ProjectionEngineV2`, not V1's phase pipeline.

**7. "Checkpoint sequence reset drops buffers"** — Correct behavior. If a newer checkpoint marker arrives before all partitions reported for the old one, abandoning the stale checkpoint in favor of the newer one is the right thing to do. Writing an old checkpoint when a newer one is pending would be wrong.

**9. "Null passed to Load"** — `JintProjectionStateHandler.Load(null)` explicitly handles null by calling `LoadCurrentState(JsValue.Null)`. This is correct for partitions where `$created` sets `s[0] = null` (like the "description" partition in the account-balancer spec).

### Items fixed in 4f71abf:

**4. "StateHandler shared concurrently"** — Fixed. Config now uses `StateHandlerFactory` (`Func<IProjectionStateHandler>`) instead of a single instance. Each partition processor creates its own handler via the factory.

**5. "BiState shared state races"** — Fixed. BiState projections are now forced to `PartitionCount=1` at engine startup (shared state requires sequential processing). A warning is logged if the configured count was > 1.

**6. "Custom partition routing mismatch"** — Fixed. Removed the deferred partition key mode. Partition keys are now always computed on the read loop thread using a dedicated handler instance, and the dispatcher routes by actual partition key hash (not stream ID hash). This guarantees all events for the same partition land on the same processor.

**10. "EmitEnabled flag unused"** — Fixed. The partition processor now gates emitted events on the `EmitEnabled` config flag.

### Low-priority improvement (noted, not blocking):

**8. "Checkpoint write can hang"** — `await tcs.Task` has no timeout. In practice the bus is in-process so message loss is extremely unlikely, but adding a timeout would be more defensive. Low priority.

---

*🤖 Generated with [Claude Code](https://claude.com/claude-code)*

src/KurrentDB.Projections.JavaScript/Services/Interpreted/JintProjectionStateHandler.cs
George-Payne, line 1: ## 1. LinkStreamTo indexes out of bounds

Lines ~337-343. When 3 parameters are passed, it accesses `parameters.At(4)` instead of `parameters.At(2)`. Also has an empty `if` block that looks like an incomplete refactor:

```csharp
JsValue LinkStreamTo(JsValue thisValue, JsValue[] parameters) {
    var stream = EnsureNonNullStringValue(parameters.At(0), "streamId");
    var linkedStreamId = EnsureNonNullStringValue(parameters.At(1), "linkedStreamId");
    if (parameters.Length == 3) {
        // empty block - looks unfinished
    }

    ExtraMetaData? metadata = null;
    if (parameters.Length == 3) {
        var md = parameters.At(4).AsObject(); // should be At(2)
```

Would throw at runtime whenever metadata is passed to `linkStreamTo`. Likely untested code path.

## 2. Multi-parameter log doesn't assemble output correctly

Lines ~376-389. Two issues:
- Separator starts at `i > 1` instead of `i > 0`, so first two items have no separator
- Primitives are logged individually via `Log(p.ToString())` instead of appended to the StringBuilder, so the final `Log(sb.ToString())` only contains objects

```csharp
for (int i = 0; i < parameters.Length; i++) {
    if (i > 1)          // should be i > 0
        sb.Append(" ,");
    var p = parameters.At(i);
    if (p != null && p.IsPrimitive())
        Log(p.ToString());    // should be sb.Append(p.ToString())
    if (p is ObjectInstance oi)
        sb.Append(Serialize(oi));
}
Log(sb.ToString());
```

`log("a", "b", "c")` would emit three separate log lines for each primitive, then an empty assembled line.

## 3. EnsureBody casts JsNull to ObjectInstance

Line ~884. If the parsed body is `JsNull` (raw data is `"null"`), it casts to `ObjectInstance` which would throw:

```csharp
var body = raw.IsNull() ? raw : _parser.Parse(raw.AsString());
// ...
objectInstance = (ObjectInstance)body; // throws if body is JsNull
```

## 5. BiState PrepareOutput checks wrong value for IsString

Line ~214 in `PrepareOutput`. Checks `_state.IsString()` (the biState array) instead of `state.IsString()` (the extracted element from index 0):

```csharp
var arr = _state.AsArray();
newState = arr.TryGetValue(0, out var state)
    ? (_state.IsString() ? _state.AsString() : ConvertToStringHandlingNulls(state))
    //  ^^^^^^ should be `state.IsString()` and `state.AsString()`
    : "";
```

Would cause incorrect serialization if a biState projection stored a raw string as partition state (s[0] = "someString"). The string would go through JSON serialization instead of being returned directly.

## 6. NaN/Infinity crashes the serializer

`SerializePrimitive` calls `writer.WriteNumberValue(value.AsNumber())` without checking for NaN/Infinity. `Utf8JsonWriter` throws `ArgumentException` for non-finite doubles. A projection handler that produces NaN or Infinity state values would fault the projection.

timothycoleman: Alexey's filed this as DB-1969

timothycoleman (approved):
Further work to consider, filed in ticket DB-2020:
```
Limitations in this preview / Potential further work
- outputstate() not yet implemented
- consider storing the checkpoint position in the state checkpoints?
- reset not yet implemented
- projection progress % isn't reported, possibly other stats too. processing rate is reported.
- Error when editing v2 projection in legacy UI: 'Failed to fetch projection result'
- Projections.V2.Tests take a curiously long time to run. engine_with_multiple_partitions_processes_events, v2_tombstone_while_running_with_more_events  seems to be flaky
- v2 always writes the state even if it hasn't changed (not sure if important)
- Various todos in the Projections.V2 project. Highlights:
	- // todo: is this less rich than CheckpointTag? should it be?
	- two unbounded partition dictionaries may need evication strategy
	- consider checkpointing when the source subscription has filtered out a lot of events
	- maxProjectionStateSize warnings
	- projection version/epoch not plumbed
```

PR #5562, opened 2026-03-23
Projections v2 engine suggestions

timothycoleman:
Suggestions/fixesefor #5544


AlexanderJohnston:
🫨Partitioned processing?! Let me stand up and cheer for you all because this will have a big impact on my work. I love this change!

I think there may be a correctness bug under overlapping checkpoints in the new v2 engine. The design doc calls out this invariant:
- At most 1 outstanding checkpoint
- 2 buffers per partition are only sufficient given that invariant

But I don't see that enforced in the implementation or the tests. For example, in `ProjectionEngineV2LifecycleTests`, the multi-partition test intentionally uses a high checkpoint threshold with the comment:

> `avoiding a race in CheckpointCoordinator when multiple markers are injected and partitions process at different speeds`

That suggests the race is already suspected, but I do not see a test that validates overlapping checkpoint markers are actually safe. Is that covered in a different PR?

## Relevant Code

`ProjectionEngineV2.RunReadLoop()` injects checkpoint markers whenever thresholds fire, but it does not wait for the previous checkpoint to complete:

```csharp
if (elapsedMs >= _config.CheckpointAfterMs &&
      (eventsProcessed >= _config.CheckpointHandledThreshold ||
       bytesProcessed >= _config.CheckpointUnhandledBytesThreshold)) {
      await dispatcher.InjectCheckpointMarker(lastLogPosition, ct);
      eventsProcessed = 0;
      bytesProcessed = 0;
      lastCheckpointTime = Stopwatch.GetTimestamp();
  }
```

`PartitionProcessor.HandleCheckpointMarker()` only has two buffers and rotates them like this:

```csharp
var bufferToFlush = _activeBuffer;
_activeBuffer = _frozenBuffer;
_activeBuffer.Clear();
_frozenBuffer = bufferToFlush;

await onCheckpointMarker(sequence, bufferToFlush);
```

This is only safe if the previous frozen buffer is guaranteed not to still belong to an in-flight checkpoint.

Only the last partition to report a marker seems to block on `WriteCheckpoint()`:

```csharp
lock (_lock) {
    if (_currentMarkerSequence != markerSequence) {
        _currentMarkerSequence = markerSequence;
        _collectedCount = 0;
        Array.Clear(_collectedBuffers);
    }

    _collectedBuffers[partitionIndex] = buffer;
    _collectedCount++;
    allCollected = _collectedCount == partitionCount;
}

if (allCollected) {
    await WriteCheckpoint(markerSequence);
}
```

Earlier partitions return immediately and continue processing, so they can reach the next marker before the current checkpoint write has finished.

## Failure Case

Let's assume there's two partitions, `P0` and `P1`, and checkpoint sequence is at `100`. 

1. `P0` reaches marker `100` first.
2. `P0` freezes buffer `A`, switches to buffer `B`, reports marker `100`, and returns immediately because not all partitions have reported yet.
3. `P0` continues processing new events into `B`.
4. `P1` later reaches marker `100` and is the last partition to report, so it starts `WriteCheckpoint(100)`.
5. While checkpoint `100` is still in flight, `P0` reaches marker `101`.
6. `P0` executes:

```csharp
  var bufferToFlush = _activeBuffer;    // B
   _activeBuffer = _frozenBuffer;      // A
   _activeBuffer.Clear();             // clears A
   _frozenBuffer = bufferToFlush;    // frozen = B
```

7. Buffer `A` was still part of checkpoint `100`, but marker `101` just cleared/reused it.

At that point checkpoint `100` can be written with missing or corrupted emitted events / state from `P0`.

## Notes 

I could try proposing a fix to this, but if I'm being honest... I've only been watching this PR for a week and it's full of nuanced choices. Can anyone confirm that my analysis makes sense before I put more time into identifying a targeted fix? It's possible I've misunderstood the intended flow here with concurrency.

**Initial guesses:**

- Enforce at most one outstanding checkpoint before injecting the next marker. This seems to align with the design doc and makes the buffer-rotation safe.
- Support overlapping checkpints explicitly, but that is more complex.

alexeyzimarev:
Hey @AlexanderJohnston thanks for looking into it. We'll check on Tuesday.

alexeyzimarev:
Great analysis, Alexander — the race condition you identified in the original code was real and would have corrupted checkpoint data. Tim Coleman's review caught this exact bug and fixed it in commit 39c1e658 ("Important: Sort out checkpointing").

**What changed**: `CheckpointCoordinator` now uses `AsyncExclusiveLock` (DotNext). The key invariant:

```
InjectCheckpointMarker() acquires the lock BEFORE sending markers to partitions
WriteCheckpoint() releases the lock in its finally block
```

So the read loop in `RunReadLoop` calls `await coordinator.InjectCheckpointMarker(...)` which blocks at `_checkpointLock.AcquireAsync()` if a previous checkpoint is still writing. No partition can receive marker N+1 until checkpoint N completes. The double-buffer rotation is safe because only one set of frozen buffers is ever "in flight."

Your failure case now plays out as:
1. P0 reaches marker 100, freezes buffer A, reports to coordinator
2. P1 reaches marker 100, triggers `WriteCheckpoint()` (async, releases lock when done)
3. Read loop hits checkpoint threshold again, calls `InjectCheckpointMarker` — **blocks on the lock** until checkpoint 100's write finishes
4. Only after the lock is released does marker 101 get injected
5. Buffer A is no longer being read by checkpoint 100 when P0 rotates it

The test comment you quoted ("avoiding a race...") was from the code *before* Tim's fix and is now moot.

Also worth noting Tim's commit message captured the rationale well:

> Previously, nothing prevented multiple checkpoints from being in progress at once. This would have led to various race conditions and corruption of the frozen buffers, with different parts of the system reading/writing them concurrently while believing they have exclusive access.

alexeyzimarev (approved):

PR #5580, opened 2026-04-14
[DB-2019] Add persistent subscriptions support for secondary indexes

alexeyzimarev:
## Summary

- Adds `PersistentSubscriptionIndexService` — a sibling service to `PersistentSubscriptionService` that handles persistent subscriptions targeting secondary indexes
- gRPC layer detects index-prefix filter on Create requests and publishes `…ToIndex` messages; Delete/Update/Connect use forwarding from the existing service
- Reuses existing `PersistentSubscription` state machine, checkpoint reader/writer, parking, ack/nack, and consumer strategies
- On `SecondaryIndexDeleted`, active connections are dropped and the subscription config is removed

### Key design decisions

- **Wire API unchanged** — clients subscribe by targeting `$all` with a `StreamIdentifier` filter whose single prefix is an index name (same convention as catch-up subscriptions)
- **Separate service (A.1)** — `PersistentSubscriptionIndexService` is a sibling on the same `perSubscrBus`, cleanly isolating index lifecycle from stream/all subscriptions
- **Forwarding for Delete/Update/Connect** — these proto messages don't carry a filter, so the existing service checks config entries and forwards to the index service when the group belongs to an index subscription
- **Drop + delete on index deletion** — no dormant configs; index deletion is a terminal event

### New files

- `PersistentSubscriptionIndexService.cs` — handles Create/Update/Delete/Connect, SecondaryIndexCommitted (live events), SecondaryIndexDeleted (cleanup), TimerTick, lifecycle
- `PersistentSubscriptionIndexEventSource.cs` — `IPersistentSubscriptionEventSource` for indexes (TFPos-based, no filter)
- `PersistentSubscriptionToIndexParamsBuilder.cs` — factory for index subscription params
- `FilterRouting.cs` — shared helper extracted from `Streams.Read.cs` for detecting index-prefix filters

## Test plan

- [x] Unit tests: `PersistentSubscriptionIndexEventSourceTests` (11 tests), `FilterRoutingTests` (8 tests)
- [x] Integration tests: `PersistentSubscriptionTests` — create/delete lifecycle, unknown index rejection, duplicate detection, non-existent delete (4 tests)
- [x] Existing persistent subscription tests: 480 passed, 0 failed, 1 skipped (unchanged)
- [x] Full xUnit suite: 1277 passed, 1 pre-existing failure (OOM finalizer, unrelated)
- [ ] Manual smoke test with dev server

### DIFF
--- a/src/KurrentDB.Core/ClusterVNode.cs
+++ b/src/KurrentDB.Core/ClusterVNode.cs
@@ -1262,6 +1262,9 @@ public class ClusterVNode<TStreamId> :
 		perSubscrBus.Subscribe<ClientMessage.UpdatePersistentSubscriptionToIndex>(persistentSubscriptionIndex);
 		perSubscrBus.Subscribe<ClientMessage.DeletePersistentSubscriptionToIndex>(persistentSubscriptionIndex);
 		perSubscrBus.Subscribe<ClientMessage.ConnectToPersistentSubscriptionToIndex>(persistentSubscriptionIndex);
+		perSubscrBus.Subscribe<ClientMessage.UnsubscribeFromStream>(persistentSubscriptionIndex);
+		perSubscrBus.Subscribe<ClientMessage.PersistentSubscriptionAckEvents>(persistentSubscriptionIndex);
+		perSubscrBus.Subscribe<ClientMessage.PersistentSubscriptionNackEvents>(persistentSubscriptionIndex);
 		perSubscrBus.Subscribe<StorageMessage.SecondaryIndexCommitted>(persistentSubscriptionIndex);
 		perSubscrBus.Subscribe<StorageMessage.SecondaryIndexDeleted>(persistentSubscriptionIndex);
 		perSubscrBus.Subscribe<SubscriptionMessage.PersistentSubscriptionTimerTick>(persistentSubscriptionIndex);
--- a/src/KurrentDB.Core/ClusterVNode.cs
+++ b/src/KurrentDB.Core/ClusterVNode.cs
@@ -1246,6 +1246,27 @@ public class ClusterVNode<TStreamId> :
 		perSubscrBus.Subscribe<SubscriptionMessage.PersistentSubscriptionTimerTick>(persistentSubscription);
 		perSubscrBus.Subscribe<SubscriptionMessage.PersistentSubscriptionsRestart>(persistentSubscription);
 
+		// PERSISTENT SUBSCRIPTIONS TO INDEX
+		_mainBus.Subscribe<StorageMessage.SecondaryIndexCommitted>(perSubscrQueue);
+		_mainBus.Subscribe<StorageMessage.SecondaryIndexDeleted>(perSubscrQueue);
+		_mainBus.Subscribe<ClientMessage.CreatePersistentSubscriptionToIndex>(perSubscrQueue);
+		_mainBus.Subscribe<ClientMessage.UpdatePersistentSubscriptionToIndex>(perSubscrQueue);
+		_mainBus.Subscribe<ClientMessage.DeletePersistentSubscriptionToIndex>(perSubscrQueue);
+		_mainBus.Subscribe<ClientMessage.ConnectToPersistentSubscriptionToIndex>(perSubscrQueue);
+
+		var persistentSubscriptionIndex = new PersistentSubscriptionIndexService(
+			perSubscrQueue, psubDispatcher, _mainQueue, consumerStrategyRegistry, secondaryIndexReaders);
+		perSubscrBus.Subscribe<SubscriptionMessage.PersistentSubscriptionIndexEntriesLoaded>(persistentSubscriptionIndex);
+		perSubscrBus.Subscribe<ClientMessage.CreatePersistentSubscriptionToIndex>(persistentSubscriptionIndex);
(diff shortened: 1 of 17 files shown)


alexeyzimarev:
# PR Review Summary

## Critical Issues (3 found)

| # | Issue | Location |
|---|-------|----------|
| 1 | **Config race condition**: Both services write `$persistentSubscriptionConfig` with `ExpectedVersion.Any`. Read-merge-write in the index service can clobber entries written by the main service between its read and write. | `PersistentSubscriptionIndexService.cs:565-612` |
| 2 | **Translating CallbackEnvelopes silently swallow non-matching messages**: The forwarding envelopes for Update/Delete have an `if` with no `else`. If a `NotHandled` or unexpected message arrives, the client hangs forever — no reply sent. | `PersistentSubscriptionService.cs:651,849` |
| 3 | **Hard cast in CallbackEnvelope**: `(ReadIndexEventsForwardCompleted)msg` with no type check. A `NotHandled` message would crash with `InvalidCastException` on the callback thread. | `PersistentSubscriptionStreamReader.cs:108` |

## Important Issues (7 found)

| # | Issue | Location |
|---|-------|----------|
| 4 | **Ambiguous group-name-only routing**: Forwarding uses `FirstOrDefault(e => e.IndexName != null && e.Group == message.GroupName)` — if two indexes share a group name, wrong subscription is targeted. | `PersistentSubscriptionService.cs:648,846,1043` |
| 5 | **No `_started` lifecycle guard**: Create/Connect/etc. can be processed before bootstrap or after shutdown. | `PersistentSubscriptionIndexService.cs` |
| 6 | **`SecondaryIndexDeleted` calls `Shutdown()` but not `Delete()`**: Checkpoint and parked-message streams are leaked. The explicit Delete handler correctly calls `Delete()`. | `PersistentSubscriptionIndexService.cs:400-406` |
| 7 | **Save-path errors silently abandon caller**: `HandleSaveReadCompleted` returns without calling `continueWith` on error — client hangs forever. | `PersistentSubscriptionIndexService.cs:508-525` |
| 8 | **No timeout/retry for index reads**: `FromStream`/`FromAll` branches have exponential-backoff retry on timeout. `FromIndex` has none — subscription silently hangs if read is lost. | `PersistentSubscriptionStreamReader.cs:99-118` |
| 9 | **Dead code**: `LoadConfiguration` / `HandleLoadCompleted` methods are never called. | `PersistentSubscriptionIndexService.cs:535` |
| 10 | **`Debug.Assert` stripped in Release**: Checkpoint parsing uses `Debug.Assert` for validation — malformed checkpoints produce cryptic `FormatException` in production. | `PersistentSubscriptionIndexEventSource.cs:39-42` |

## Suggestions (6 found)

| # | Issue | Location |
|---|-------|----------|
| 11 | **Boolean tristate → enum**: `FromStream`/`FromAll`/`FromIndex` should be an `EventSourceKind` enum for exhaustiveness checking. | `IPersistentSubscriptionEventSource.cs` |
| 12 | **Seal new classes**: `PersistentSubscriptionIndexEventSource`, `PersistentSubscriptionIndexService`, `PersistentSubscriptionToIndexParamsBuilder` should be sealed. | Multiple files |
| 13 | **Extract `$index-` as a constant**: Hardcoded in 4 locations across 3 files. | Multiple files |
| 14 | **Log level**: Timeout retry uses `Information` instead of `Warning`. | `PersistentSubscriptionIndexService.cs:622` |
| 15 | **Null `mainQueue` accepted at construction**: Old 2-param constructor allows `null` that only fails at read-time. | `PersistentSubscriptionStreamReader.cs:30-31` |
| 16 | **Significant test gaps**: Update, Connect, live event routing, SecondaryIndexDeleted cleanup, forwarding paths, bootstrap, and gRPC entry point are all untested. | Tests |

## Strengths

- **Clean architecture**: Separate sibling service with clear boundaries, sharing the state machine and checkpoint infrastructure.
- **FilterRouting helper**: Stateless, pure, excellent use of `[NotNullWhen(true)]`, clean guard clauses.
- **Consistent patterns**: New code follows established `PersistentSubscriptionService` patterns closely.
- **Config partitioning**: Existing service partitions entries by `IndexName != null` and forwards — elegant bootstrap design.
- **Integration tests**: Test fixture helpers use `TcsEnvelope` correctly per threading guidelines.

## Recommended Action

**Fix before merge (Critical + top Important):**
1. Add `else` branches to translating envelopes — reply with `Fail` on unexpected messages
2. Add type-check before the cast in `PersistentSubscriptionStreamReader` — route `NotHandled` to `_onError`
3. Call `sub.Delete()` in `SecondaryIndexDeleted` handler (not just `Shutdown()`)
4. Add `_started` lifecycle guard to index service
5. Fix `HandleSaveReadCompleted` — call `continueWith` or reply with error on failure paths
6. Remove dead `LoadConfiguration` / `HandleLoadCompleted`

**Address soon (remaining Important):**
7. Config race condition — use optimistic concurrency or single-writer
8. Ambiguous group-name routing — match on `IndexName` as well as `GroupName`
9. Add timeout/retry for index reads
10. Replace `Debug.Assert` with proper validation

**Consider later (Suggestions):**
11. Enum for event source kind, seal classes, extract constants, fix log level, expand test coverage

---

PR #5588, opened 2026-04-21
[DB-1970]: Arrow Flight SQL protocol support

sakno:
Changed: Added Arrow Flight SQL protocol support.

# Python
Example for Python:
```python
import adbc_driver_flightsql.dbapi as dbapi

with dbapi.connect("grpc://localhost:41119") as conn:
    with conn.cursor() as cur:
        cur.execute("SELECT * FROM kdb.records WHERE log_position>?", parameters=[42])
        print(cur.fetchall())          # list of tuples
```

Required dependencies:
```sh
pip install pyarrow
pip install adbc-driver-flightsql
```

# Java/JDBC
Example for Java:
```java
package org.example;

import java.sql.*;

public class Main {
    public static void main(String[] args) throws SQLException {
        String url = "jdbc:arrow-flight-sql://localhost:41119?useEncryption=false";

        try (Connection conn = DriverManager.getConnection(url);
             PreparedStatement stmt = conn.prepareStatement(
                     "SELECT * FROM kdb.records WHERE log_position > ?")) {

            var meta = stmt.getMetaData();
            System.out.println(meta.getColumnCount());

            stmt.setInt(1, 42);  // 1-based index

            try (ResultSet rs = stmt.executeQuery()) {
                while (rs.next()) {
                    // access columns by index or name
                    System.out.println(rs.getLong("log_position"));
                }
            }
        }
    }
}

```

Maven POM file
```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>

    <groupId>org.example</groupId>
    <artifactId>AdbcTest</artifactId>
    <version>1.0-SNAPSHOT</version>

    <properties>
        <maven.compiler.source>17</maven.compiler.source>
        <maven.compiler.target>17</maven.compiler.target>
        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
    </properties>

    <dependencies>
        <dependency>
            <groupId>org.apache.arrow</groupId>
            <artifactId>flight-sql-jdbc-driver</artifactId>
            <version>19.0.0</version>
        </dependency>
    </dependencies>

</project>
```

Tested on JDK 17, I wasn't able to fix the problem with Netty default memory allocator on the latest JDK 26.

# ADO.NET
ADO.NET example is a part of the integration test.


src/KurrentDB.SecondaryIndexing/Query/QueryPreparationOptions.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace KurrentDB.SecondaryIndexing.Query;
+
+/// <summary>
+/// Represents query preparation options.
+/// </summary>
+[StructLayout(LayoutKind.Auto)]
+public readonly struct QueryPreparationOptions {
+	/// <summary>
+	/// Indicates that the prepared query must be signed to protect it from the modification.
```
shaan1337:
```suggestion
	/// Indicates that the prepared query must be signed to protect it from modifications.
```

src/Directory.Packages.props
```diff
@@ -58,7 +58,8 @@
     <PackageVersion Include="Kurrent.Surge.Core" Version="1.1.1-alpha.1.65" />
     <PackageVersion Include="Kurrent.Surge.DataProtection" Version="1.1.1-alpha.1.57" />
     <PackageVersion Include="Kurrent.Surge.DuckDB" Version="1.1.1-alpha.1.57" />
-    <PackageVersion Include="Kurrent.Quack" Version="0.0.0-alpha.164" />
+		<PackageVersion Include="Kurrent.Quack" Version="0.0.0-alpha.176" />
```
shaan1337: fix indentation

src/Directory.Packages.props
```diff
@@ -167,5 +168,9 @@
     <PackageVersion Include="xunit.v3" Version="3.1.0" />
     <PackageVersion Include="xunit.v3.extensibility.core" Version="3.1.0" />
     <PackageVersion Include="YamlDotnet" Version="16.3.0" />
+		<PackageVersion Include="Apache.Arrow.Flight.AspNetCore" Version="22.1.0" />
+		<PackageVersion Include="Apache.Arrow.Flight.Sql" Version="22.1.0" />
+		<PackageVersion Include="Apache.Arrow.Adbc.Drivers.FlightSql" Version="0.21.0" />
+		<PackageVersion Include="Apache.Arrow.Adbc.Client" Version="0.23.0" />
```
shaan1337: fix indentation

src/KurrentDB.SecondaryIndexing/FlightSql/ConnectionState.PreparedStmt.cs
```diff
@@ -0,0 +1,161 @@
...
+	/// <param name="statement">The prepared statement object.</param>
+	/// <param name="parameters">Arrow schema of the prepared statement parameters.</param>
+	/// <returns>
+	/// <see langword="true"/> if the internal registry has enough space to place a new prepared statement;
+	/// <see langword="false"/> if there is no space in the internal registry.
+	/// </returns>
+	public bool CreatePreparedStatement(ReadOnlySpan<char> query,
+		out Guid handle,
+		[NotNullWhen(true)] out PreparedStatement? statement,
+		[NotNullWhen(true)]out Schema? parameters) {
```
shaan1337:
```suggestion
		[NotNullWhen(true)] out Schema? parameters) {
```

src/KurrentDB.SecondaryIndexing/FlightSql/ConnectionState.PreparedStmt.cs
```diff
@@ -0,0 +1,161 @@
...
+		// nothing to bind, the batch is empty
+		if (arguments.Length is 0) {
+			arguments.Dispose();
+			return true;
+		}
+
+		// save the binding context
+		var result = statement.TryBind(arguments, out var oldBatch);
+		oldBatch?.Dispose();
+		statement.DecrementRef();
```
shaan1337, line 88: should `statement.DecrementRef()` be called only if `result` is false ?

src/KurrentDB.SecondaryIndexing/FlightSql/FlightSqlServer.cs
```diff
@@ -0,0 +1,156 @@
...
+/// registry of the connections associated with the prepared statements. This is very expensive.
+/// Instead, the prepared statement handle is just a unique identifier that represents the transformed SQL query.
+/// The query is not prepared at DuckDB level, so on every execution the database needs to parse it, build the plan and execute.
+/// In other words, we mimic the prepared statement concept in FlightSQL, which leads to lower performance than true
+/// DuckDB prepared statement.
+/// </remarks>
+/// <param name="engine">The query engine.</param>
+internal sealed partial class FlightSqlServer(IQueryEngine engine) : FlightServer {
+	public override Task<FlightInfo> GetFlightInfo(FlightDescriptor request, ServerCallContext context) {
+		//A google.protobuf.Any message has a specific binary signature. It consists of two fields:
```
shaan1337:
```suggestion
		// A google.protobuf.Any message has a specific binary signature. It consists of two fields:
```

src/KurrentDB.SecondaryIndexing/FlightSql/FlightSqlServer.MemoryMgmt.cs
```diff
@@ -0,0 +1,77 @@
...
+		}
+
+		static string GetTypeUrl(MessageDescriptor descriptor)
+			=> string.Concat("type.googleapis.com/", descriptor.FullName);
+	}
+
+	private ByteString PackRaw(IMessage message) {
+		message.WriteTo(PrepareWriter());
+		var buffer = bufferWriter.DetachBuffer();
+		return UnsafeByteOperations.UnsafeWrap(buffer.Memory);
```
shaan1337: Should the buffer be registered for disposal here too ?
```suggestion
		return WrapAndRegisterOnDispose(buffer);
```

src/KurrentDB.SecondaryIndexing/FlightSql/ConnectionState.PreparedStmt.cs
sakno: No, we incremented the ref counter for the statement in the beginning of the method:
```cs
if (!_statements.TryGetValue(handle, out var statement) || !statement.TryIncrementRef())
			return false;
```

So we have to release the counter when leaving the method regardless the binding operation result.

src/KurrentDB.SecondaryIndexing/FlightSql/FlightSqlServer.MemoryMgmt.cs
sakno: Good catch

src/Directory.Packages.props
sakno: Fixed

shaan1337 (approved):

src/KurrentDB.SecondaryIndexing/FlightSql/ConnectionState.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace KurrentDB.SecondaryIndexing.FlightSql;
+
+/// <summary>
+/// Maintains the connection-level state required to implement stateful FlightSQL operations.
+/// </summary>
+/// <param name="engine"></param>
+internal sealed partial class ConnectionState(IQueryEngine engine) : Disposable {
+	protected override void Dispose(bool disposing) {
+		if (disposing) {
+			foreach (var statement in _statements.Values) {
```
shaan1337, line 16: is it possible for a prepared statement to be concurrently added to _statements while we are disposing the others? (and we miss its disposal)

sakno: In theory yes. `ConnectionState` is local to the client connection. Somehow we can start the request for statement preparation and close the connection concurrently. This is not fatal since the GC will call the finalizers for the necessary objects.

PR #5592, opened 2026-04-23
[DB-2040] Projections V2 — bound unbounded partition-state caches

alexeyzimarev:
## Summary

- Replaces the two unbounded partition-state caches in Projections V2 (`PartitionProcessor._stateCache`, `ProjectionEngineV2._partitionStates`) with bounded SIEVE caches via a thin wrapper over DotNext `RandomAccessCache`.
- Adds a system-wide `ClusterVNodeOptions.Projection.MaxPartitionStateCacheSize` (default `100_000`), threaded end-to-end to `ProjectionEngineV2Config`.
- Silent eviction preserves correctness — both caches already have `-state` stream fallbacks (`LoadPartitionState` and `CoreProjectionV2.Handle(GetState)`).
- Stats surface gains `PartitionStateCacheSize` and `PartitionStateCacheEvictions` (`long`), populated from the shared cache and forwarded through the HTTP projection statistics endpoint. Eviction log at `Verbose`.

Design on Linear: DB-2040. Plan: `docs/plans/2026-04-23-projections-v2-memory-control.md`.

## What changed

**New:**
- `src/KurrentDB.Projections.V2/Services/Processing/V2/PartitionStateCache.cs` — wrapper with `TryGet` / `Set` / `Count` / `Evictions`.
- `src/KurrentDB.Projections.V2.Tests/Unit/PartitionStateCacheTests.cs` — 6 unit tests.
- `src/KurrentDB.Projections.V2.Tests/Integration/PartitionStateCacheEvictionTests.cs` — 2 integration tests (40 partitions / capacity 4: ≥ 36 evictions; cold-engine recovery via `-state` stream fallback).

**Modified (wiring `MaxPartitionStateCacheSize`):** `ClusterVNodeOptions.cs`, `ProjectionsSubsystem.cs` (`ProjectionSubsystemOptions` record), `ProjectionsStandardComponents.cs`, `ProjectionCoreService.cs`, `ProcessingStrategySelector.cs`, `ProjectionProcessingStrategyV2.cs`, `CoreProjectionV2.cs`, `ClusterVNodeHostedService.cs`, `ProjectionEngineV2Config.cs` (new `required int`), `ProjectionEngineV2.cs` (cache + `GetCacheMetrics`), `PartitionProcessor.cs` (cache + dispose), `ProjectionStatistics.cs` + `ProjectionStatisticsHttpFormatted.cs`.

## Operator surface

- Setting: `KurrentDB.Projection.MaxPartitionStateCacheSize` (default `100_000`, mirrors `StreamInfoCacheCapacity`).
- Metrics on `GET /projection/:name/statistics`: `PartitionStateCacheSize`, `PartitionStateCacheEvictions` (both `long`).
- `Verbose` log on each eviction: `"Partition state cache {Cache} for projection {Projection} evicted partition {Partition}"`.

## Design rationale (deviations from the ticket worth flagging)

- **Size/evictions reported from the shared engine cache only**, not summed across per-slot caches. Every partition-state write flows through the shared cache, so it's the faithful summary of memory pressure; summing across pools would double-count partition keys.
- **Stats typed `long`** (not `int`) to avoid truncation on long-running projections.
- **DotNext `RandomAccessCache` uses SIEVE** (scan-resistant LRU variant). Described accurately in the option's `[Description]`.

## Known follow-ups (out of scope for DB-2040)

- gRPC `StatisticsResp.Types.Details` proto not extended for the new fields — HTTP path is wired; gRPC extension is a separate compat concern.
- Null-state-after-eviction: when a projection returns `null` for a partition that previously had non-null state, eviction causes the next load to re-read the last non-null state from the `-state` stream. This is the same behaviour you already get on engine restart (the `-state` stream doesn't carry null sentinels), so not a new failure mode — but bounded caches surface it on a shorter timeline. Mitigation would need null sentinels in the stream.

## Test plan

- [ ] `dotnet test --project src/KurrentDB.Projections.V2.Tests/KurrentDB.Projections.V2.Tests.csproj -c Release --framework net10.0` — expect 80 total / 68 pass / 12 skip / 0 fail. Of those, 2 new integration tests and 6 new unit tests for the cache wrapper.
- [ ] `dotnet test --project src/KurrentDB.Projections.Management.Tests/KurrentDB.Projections.Management.Tests.csproj -c Release --framework net10.0` — expect existing pass (fixtures updated for new ctor params).
- [ ] Smoke test: start server with a custom `MaxPartitionStateCacheSize` (e.g. `10`), run a high-cardinality projection, confirm `PartitionStateCacheEvictions > 0` in the stats endpoint and state queries still return the correct value for evicted partitions.
- [ ] Acceptance: 2× cap high-cardinality run — cache `Count` stays ≤ cap-equivalent (asserted via eviction-counter lower bound = `partitionCount - cacheSize`).


src/KurrentDB.Projections.V2/Services/Processing/V2/PartitionStateCache.cs
```diff
@@ -0,0 +1,70 @@
...
+using DotNext.Runtime.Caching;
+using Serilog;
+
+namespace KurrentDB.Projections.Core.Services.Processing.V2;
+
+public sealed class PartitionStateCache : IAsyncDisposable {
+	private static readonly ILogger Log = Serilog.Log.ForContext<PartitionStateCache>();
+
+	// RandomAccessCache requires TValue : notnull, so we wrap nullable strings in a record.
+	private readonly record struct CachedValue(string? State);
```
sakno, line 18: Rather than `struct` wrapper, can we use `string.Empty` to indicate empty value?

src/KurrentDB.Projections.V2/Services/Processing/V2/PartitionStateCache.cs
```diff
@@ -0,0 +1,70 @@
...
+			throw new ArgumentOutOfRangeException(nameof(capacity), capacity, "Capacity must be positive");
+		_name = name;
+		_projectionName = projectionName;
+		_cache = new RandomAccessCache<string, CachedValue>(capacity) {
+			Eviction = OnEviction,
+			KeyComparer = StringComparer.Ordinal,
+		};
+	}
+
+	public long Evictions => Interlocked.Read(ref _evictions);
```
sakno, line 37: `Volatile.Read` is enough. `Interlocked.Read` is a special method to read 64-bit values on 32-bit platforms atomically, which is not our case.

src/KurrentDB.Projections.V2/Services/Processing/V2/PartitionStateCache.cs
```diff
@@ -0,0 +1,70 @@
...
+	public long Count => Interlocked.Read(ref _count);
```
sakno, line 38: Same as above

src/KurrentDB.Projections.V2/Services/Processing/V2/PartitionStateCache.cs
```diff
@@ -0,0 +1,70 @@
...
+		}
+
+		value = null;
+		return false;
+	}
+
+	public async ValueTask Set(string key, string? value, CancellationToken ct) {
+		// Sync Dispose, not DisposeAsync: DotNext recommends sync disposal on non-rate-limited
+		// caches because DisposeAsync serializes writers behind a single promotion thread.
+		using var session = await _cache.ChangeAsync(key, ct).ConfigureAwait(false);
```
sakno: As far as I know, we're not using `ConfigureAwait` explicitly across the code.

src/KurrentDB.Projections.V2/Services/Processing/V2/PartitionStateCache.cs
```diff
@@ -0,0 +1,70 @@
...
+		var isNew = !session.TryGetValue(out _);
+		session.SetValue(new CachedValue(value));
```
sakno, line 57: Not a matter of code change, just for clarity: due to concurrency, the reader can see two different values for the same key, because `isNew` flag is ignored here. Typically, `ChangeAsync` is used to implement `GetOrAdd` semantics, rather than direct set:
```cs
if (session.TryGetValue(out var value))
  return value; // because the value is already exist, we don't need to rewrite it
```

This approach guarantees than when the cache entry once added, its value cannot be changed.

src/KurrentDB.Projections.V2/Services/Processing/V2/PartitionStateCache.cs
alexeyzimarev: When the state is `null` it won't parse to a proper JSON value in the JS engine, we need to pass null there.

src/KurrentDB.Projections.V2/Services/Processing/V2/PartitionStateCache.cs
timothycoleman: my suggestions removed this one

timothycoleman (approved):

PR #5603, opened 2026-04-28
[DEV-1625] Require Unicode NFC normalization for schema names

w1am:
Now that schema names can contain Unicode, inputs in NFD form would silently split into distinct schemas: core persists and hashes raw bytes without normalizing, so "café" (precomposed e-acute) and "cafe" + combining acute become different streams.


RagingKore (approved):
LGTM

PR #5610, opened 2026-04-30
[DB-2085] Fix three JintProjectionStateHandler correctness bugs

alexeyzimarev:
## Summary

- Projections reading `event.body` no longer fault on events with primitive JSON
  bodies (numbers, strings, booleans, `null`, as opposed to objects).
- Projections that return string state — and projections defined without explicit
  `when()` handlers — now persist and reload state correctly across restarts.
- Projection state containing `NaN` or `Infinity` is now persisted as `null`
  (matching `JSON.stringify`) instead of faulting the projection.

## Details

Fixes three correctness bugs in `JintProjectionStateHandler` (shared by V1 and V2 projection engines). Each bug faults the projection on a real-world input.

- **`EnsureBody` cast** (`JintProjectionStateHandler.cs:884`) — `(ObjectInstance)body` throws `InvalidCastException` whenever `bodyRaw` parses to a non-object: `null`, a number, a string, or a boolean. Cast removed; the `out` parameter was already typed `JsValue`. The sibling `EnsureMetadata` was already correct — this matches it.
- **`PrepareOutput` raw string output** (`JintProjectionStateHandler.cs:207-217`) — both bi-state and uni-state paths emitted JS string state raw (e.g. `hello`) instead of JSON-encoded, so `Load()`'s `JsonParser.Parse` faults on restart. Always JSON-encode via
  `ConvertToStringHandlingNulls`, restoring the V8 prelude's always-`JSON.stringify` contract.
- **`SerializePrimitive` NaN/Infinity** (`JintProjectionStateHandler.cs:1226`) — `Utf8JsonWriter.WriteNumberValue(double)` throws `ArgumentException` on non-finite doubles regardless of `SkipValidation`. Now guarded with `double.IsFinite`; non-finite values serialize as `null` to match `JSON.stringify` semantics.

Linear issue: [DB-2085](https://linear.app/kurrent/issue/DB-2085/jintprojectionstatehandler-three-correctness-bugs-body-cast-bistate)

## Test plan

TDD — each test was watched fail with the exact reported error before the fix landed.

- [x] `when_accessing_event_body_with_non_object_data` — raw `null`, raw `42`, raw `"hello"` event bodies are accessible via `event.body` without `InvalidCastException`.
- [x] `when_running_bi_state_js_projection_with_string_state` — biState returning `["foo", {…}]` produces `newState == "foo"` (raw, not JSON-quoted).
- [x] `when_serializing_state` — `JsNumber(NaN)`, `+Infinity`, `-Infinity`, and an object containing those properties all serialize as `null`.
- [x] Full Jint test suite: 8 new tests green; 5 pre-existing failures unrelated to this fix (DotNext message routing in projection-management subscription tests) confirmed present on `master` baseline before the change.

### DIFF
--- a/src/KurrentDB.Projections.JavaScript/Services/Interpreted/JintProjectionStateHandler.cs
+++ b/src/KurrentDB.Projections.JavaScript/Services/Interpreted/JintProjectionStateHandler.cs
@@ -209,25 +209,12 @@ public class JintProjectionStateHandler : IProjectionStateHandler {
 		_emitted.Clear();
 		if (_definitionBuilder.IsBiState && _state.IsArray()) {
 			var arr = _state.AsArray();
-			if (arr.TryGetValue(0, out var state)) {
-				if (state.IsString()) {
-					newState = state.AsString();
-				} else {
-					newState = ConvertToStringHandlingNulls(state);
-				}
-			} else {
-				newState = "";
-			}
-
-			if (arr.TryGetValue(1, out var sharedState)) {
-				newSharedState = ConvertToStringHandlingNulls(sharedState);
-			} else {
-				newSharedState = null;
-			}
-
-		} else if (_state.IsString()) {
-			newState = _state.AsString();
-			newSharedState = null;
+			newState = arr.TryGetValue(0, out var state)
+				? ConvertToStringHandlingNulls(state)
+				: "";
+			newSharedState = arr.TryGetValue(1, out var sharedState)
+				? ConvertToStringHandlingNulls(sharedState)
+				: null;
 		} else {
 			newState = ConvertToStringHandlingNulls(_state);
 			newSharedState = null;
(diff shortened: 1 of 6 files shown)


src/KurrentDB.Projections.JavaScript/Services/Interpreted/JintProjectionStateHandler.cs
```diff
@@ -210,8 +210,8 @@ private void PrepareOutput(out string? newState, out string? newSharedState, out
 		if (_definitionBuilder.IsBiState && _state.IsArray()) {
 			var arr = _state.AsArray();
 			if (arr.TryGetValue(0, out var state)) {
-				if (_state.IsString()) {
-					newState = _state.AsString();
+				if (state.IsString()) {
+					newState = state.AsString();
```
timothycoleman: this is pretty suspicious.

- good: the fix does align the bi-state path with the behaviour of the mono-state path, and is probably what the original author intended to write
 
- bad: the mono-state path (and the new bi-state path) just take the state if it is a string and write that string directly, which is wrong isn't it. the state is supposed to be JSON, so if we have state that is a string, it needs wrapping in quotes, escaping, and so on, which is what bi-state was getting before. i think this is the old V8 equivalent from before we moved to the interpreted js runtime, which always serializes as we'd expect: https://github.com/kurrent-io/KurrentDB/blob/master/src/KurrentDB.Projections.Management/Prelude/Projections.js#L114

i think what we need to do here make both mono-state and bi-state always serialize to json, the net effect of that would be to fix a bug in mon-state projections, which won't currently work if using a string as a state

timothycoleman (approved):

PR #5619, opened 2026-05-07
feat: enable graceful shutdown of persistent subscription consumer

avens19:
This pull request adds a new client message type to the persistent subscription channel. Sending stop from the client will remove the consumer from the strategy pool but will leave the grpc connection alive. This allows the consumer to finish processing in-flight events, ack/nack as needed and then unsubscribe. This avoids the likelihood of double-processing in-flight events as just calling unsubscribe automatically retries in-flight events

NOTE: this PR was created with the help of Claude (Opus 4.7)

### DIFF
--- a/docs/server/features/persistent-subscriptions.md
+++ b/docs/server/features/persistent-subscriptions.md
@@ -33,6 +33,30 @@ Just as with message brokers, processing events in a group of consumers running 
 
 Clients must acknowledge (or not acknowledge) messages as they are handled. If messages aren't acknowledged before they time out on the server, the server will retry them. If a message has been retried more than the `maxRetryCount` setting for the persistent subscription, then the message will be parked and processing will continue.
 
+## Graceful shutdown
+
+When a consumer needs to shut down, simply closing the connection works but is abrupt: any events the consumer was still working on are immediately retried by another consumer in the group, even if the original consumer was about to ack them. This can lead to duplicate processing.
+
+To shut down cleanly, a consumer can send an explicit `Stop` message on its subscription stream. This:
+
+- Removes the consumer from the consumer pool, so the server stops sending it new events.
+- Leaves the connection open so the consumer can still acknowledge or not-acknowledge any events it already received.
+- Does **not** redistribute the consumer's in-flight events to other consumers — they stay assigned to the stopping consumer until it acks/nacks them or the message timeout fires.
+
+The recommended sequence is:
+
+1. **Send `Stop`.** The server marks the consumer as stopped and routes new events to the remaining active consumers in the group.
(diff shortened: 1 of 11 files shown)


cbstns (approved):

timothycoleman:
Hi thanks for this! We'll take a look over the next few days

timothycoleman:
Heya,

Just to clarify, we're looking to improve the graceful termination behaviour with respect to duplicate processing of in-flight/'still being worked on' events, do we mean

1. When unsubscribing, Acks/Nacks that the consumer had _previously_ sent are not being received by the server because the gRPC call is being terminated too soon, resulting in some events being unnecessarily processed twice

2. When unsubscribing, the consumer may be in the middle of processing some events that it will _subsequently_ try to Ack/Nack but be unable to because the gRPC call will have closed, resulting in some events being unnecessarily processed twice

3. When unsubscribing, events that the server had sent to the consumer but that the consumer never processes get redistributed to other consumers but this produces some unideal behaviour

(or more than one of the above)

Thanks!
Tim

avens19:
2 was my primary goal, though 3 is also unideal

avens19:
@timothycoleman do you need any more info?

timothycoleman:
Hi @avens19,

Sorry it's taking a while, it affects the API (albeit in a small way) and therefore the client SDKs and so we're discussing it internally.

What client SDK are you using, and do you also have code changes on the client side? I'm mostly
wondering how, after sending Stop, you're determining when the in-flight messages have been cleared and that the server isn't currently sending some more. Maybe it's approximate and based on a timeout?

It may be that since we're not getting a 'Stopped' message in reponse, it'll behave the same as a client-side only mechanism that:

1. prevents any further events from starting to be processed
2. waits until we've sent an ack or nack for any even currently being processed
3. teminate the subscription

Without involving the server before the termination?

avens19:
I have not implemented any clients, we're using the NodeJS client.

Ya my thinking was that it ultimately doesn't matter since the fallback is the existing behaviour. (IE: if there's a race and the server tries to send me an event while I send stop and have no items in the queue, I will exit immediately, ignore the event and the server will retry it on another client as it does today). I plan to continue working on any events I've received, adding any new events I receive during the drain to the queue and then calling the existing unsubscribe. All of this is best effort with the existing behaviour as the fallback.

You're right that it's similar to a client-only solution which is why I mentioned that this also helps with your # 3 from above. Those events that I'm silently ignoring will end up being delayed and out of order relative to other events. I know there are no guarantees for this in general but ideally we'd do our best to prevent.

avens19:
I did implement the client-side solution on my side which you're right does solve # 2 from above which I said was my primary concern, so fair callout

PR #5642, opened 2026-06-12
Add caller-side DuckDB CPU attribution metric

realtonyyoung:
## What

A new `KurrentDB.DuckDB` meter exposing `kurrentdb.duckdb.cpu.seconds` — a counter of CPU time consumed by DuckDB operations executed on KurrentDB threads, tagged `activity=query|read|commit|checkpoint` and `source=caller`. Today the process-level CPU metric is one opaque number; this is the first step toward decomposing it into "KurrentDB vs DuckDB" on all supported platforms, e.g.:

```
rate(kurrentdb_duckdb_cpu_seconds_total[1m]) / rate(kurrentdb_proc_cpu[1m])
```

## How

- **`ThreadCpuTime`** (KurrentDB.Core/DuckDB): cumulative CPU time of the calling thread — `clock_gettime(CLOCK_THREAD_CPUTIME_ID)` on Linux (clock id 3) and macOS (16), `GetThreadTimes(GetCurrentThread())` on Windows. No-ops on unsupported platforms.
- **`DuckDBCpuMetrics.Measure(activity)`** returns a `ref struct` scope, so the compiler rejects any use across an `await` — per-thread CPU deltas are only valid when start and stop happen on the same thread.
- Instrumented synchronous DuckDB sections: default/user index commits, user index checkpoints, all four index readers (one wrap point in `SecondaryIndexReaderBase`), and `QueryEngine`'s snapshot-capture/parse/plan/execute section.
- The meter is registered in `AddDuckDb()` and listed in `metricsconfig.json`, so it exports through the standard `/metrics` endpoint with no extra wiring.

## Verification

- Two new unit tests: a busy-spin scope records CPU seconds with the right tags; an idle (sleeping) scope records ~nothing. Full SecondaryIndexing suite passes (75/75).
- Live smoke test on a dev server (`CommitBatchSize=1`, one event written):
  ```
  kurrentdb_duckdb_cpu_seconds_total{otel_scope_name="KurrentDB.DuckDB",activity="commit",source="caller"} 0.0011465
  ```

## Known limits

- This measures the **caller-executed share** of DuckDB CPU. DuckDB also schedules work on its own native worker pool (and burns CPU during streaming chunk fetches inside `ConsumeAsync`); that share is invisible from the caller side and is planned as `source=workers` via Quack owning DuckDB's workers (`threads=N, external_threads=N` + `duckdb_execute_tasks_state`) — which also yields a CPU cap. Until then this metric is a lower bound, tightest for commit/checkpoint/read activities which are fully synchronous.
- Windows accrues thread CPU at scheduler-quantum granularity: cumulative rates are statistically sound, but individual short operations under-resolve.


src/KurrentDB.Core/DuckDB/DuckDBCpuMetrics.cs
```diff
@@ -0,0 +1,62 @@
...
+		private readonly long _startNanoseconds;
+
+		internal CpuScope(DuckDBCpuMetrics metrics, string activity) {
+			if (!ThreadCpuTime.IsSupported)
+				return;
+
+			_metrics = metrics;
+			_activity = activity;
+			_startNanoseconds = ThreadCpuTime.CurrentNanoseconds;
+		}
```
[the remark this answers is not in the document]
realtonyyoung: Not changed: since C# 11 (auto-default structs) constructors may return without assigning all fields — they're default-initialized, so this compiles fine on net10.0 (verified: clean local build + CI compile of this file). The unsupported path leaves `_metrics` null and `Dispose` guards on that, so it's a safe no-op as intended.

realtonyyoung (commented):
One actionable issue I found:

The new `query` CPU metric stops before the result fetch loop, but every current Flight SQL consumer uses streaming and spends the actual DuckDB result production in `IQueryResultReader.TryRead()` after `ExecuteAsync` leaves the measurement scope. That means large/expensive streamed queries can report only snapshot/prepare/initial `ExecuteQuery` CPU while omitting the per-chunk `TryFetch` work, and `QueryResultReader.Dispose()` can also drain remaining chunks without attribution. Please thread `DuckDBCpuMetrics` into `QueryResultReader` and measure `_result.TryFetch(...)` in `TryRead()`/`FinalizeEnumeration()`, or otherwise keep the query metric scoped around those synchronous fetch calls too.

realtonyyoung:
Good catch — fixed in ae5a03ec9. You were right that the `query` scope stopped before the streaming fetch loop.

I threaded `DuckDBCpuMetrics` into `QueryResultReader` and now measure the synchronous `_result.TryFetch(...)` calls under the `query` activity in both `TryRead()` and `FinalizeEnumeration()` (the latter covers the drain in `Dispose()`). The setup scope in `ExecuteAsync` stays as-is, so prepare/plan and per-chunk fetch CPU are now both attributed additively to `activity=query, source=caller`.

Added an integration test (`QueryEngineStreamingReadAttributesQueryCpuToCaller`) that runs the streaming `QueryEngine` read and asserts the path emits caller-attributed `query` CPU (filtered against background commit activity on the shared meter). I deliberately assert tagging/wiring rather than a per-fetch measurement *count*: a single sub-vector-size chunk fetch can finish below the Windows `GetThreadTimes` quantum and record zero, so a count assertion would be flaky on the Windows runner. Local suite: 76/76 green.

src/KurrentDB.SecondaryIndexing.Tests/IntegrationTests/ReadTests.cs
```diff
@@ -44,6 +46,50 @@ await engine.ExecuteAsync(
...
+			preparedSql.Memory,
+			consumer,
+			new() { CheckIntegrity = true },
+			TestContext.Current.CancellationToken);
+
+		listener.Dispose(); // stop callbacks before asserting
+
+		Assert.True(consumer.RowCount > 0); // the streaming fetch loop actually ran
+		lock (gate) {
+			Assert.NotEmpty(queryMeasurements);
```
realtonyyoung: This assertion is still timing-dependent. `DuckDBCpuMetrics.Dispose()` only emits when `elapsedNanoseconds > 0`, and this query can be small enough for the setup/fetch scopes to remain at zero CPU ticks, especially on Windows where `GetThreadTimes` is scheduler-quantum granular. In that case the streaming path can be correctly wired but `queryMeasurements` stays empty and the test flakes. Please make the query/consumer burn enough caller CPU to guarantee a non-zero delta, or assert the wiring with a deterministic/fake CPU-time source instead of relying on real per-thread CPU accounting.

realtonyyoung:
### Heads-up: the two vulnerability checks fail on a repo-wide advisory, not this PR

`ci/github/scan-vulnerabilities` (and the build's NU1903 restore audit) fail on **CVE-2025-6965 / GHSA-2m69-gcr7-jv3q**, a high-severity advisory on **`SQLitePCLRaw.lib.e_sqlite3` 2.1.11**, pulled in transitively via `Microsoft.Data.Sqlite`. That package is the latest published version, is now deprecated, and has **no patched release**, so it can't be pinned forward. The advisory post-dates master's last green run (2026-06-12), so it breaks restore + the vuln scan for **every** PR repo-wide, independent of this change.

Per discussion, this is being handled in a **dedicated repo-wide PR** (suppress the audit + allowlist the scan for this advisory, or migrate to the recommended `SourceGear.sqlite3`), not bundled into this feature PR. I briefly added a `NuGetAuditSuppress` here and then **reverted it** — it only silenced the build gate, while `dotnet list package --vulnerable` in the scan job ignores that suppression, so it wasn't a complete fix and didn't belong here.

**This PR's own code is green**: build clean (0 warnings) and the full `KurrentDB.SecondaryIndexing.Tests` suite passes (76/76) locally, including the new streaming-CPU test. Once the SQLite advisory is remediated on master, I'll rebase and the vuln checks will clear.

src/KurrentDB.Core/DuckDB/InjectionExtensions.cs
```diff
@@ -21,6 +23,10 @@ public static IServiceCollection AddDuckDb(this IServiceCollection services) {
 		services.AddSingleton<DuckDBConnectionPool>(sp => sp.GetRequiredService<DuckDBConnectionPoolLifetime>().Shared);
 		services.AddSingleton<DuckDbConnectionPoolMiddleware>();
 		services.AddSingleton<ConnectionInterceptor>(CreatePoolPerConnectionInterceptor);
+		services.AddSingleton(static sp => {
+			var serviceName = sp.GetService<MetricsConfiguration>()?.ServiceName ?? "kurrentdb";
```
realtonyyoung: This lookup never sees the node metrics configuration in the normal startup path. `ClusterVNodeStartup` keeps `_metricsConfiguration` as a private field and `ConfigureNodeServices` only registers `StandardComponents`, not `MetricsConfiguration`, so `sp.GetService<MetricsConfiguration>()` falls back to `"kurrentdb"` even when the node is configured with a different/legacy service name. The result is that this one metric is emitted under `kurrentdb.duckdb.cpu.seconds` while the rest of the metrics pipeline uses `_metricsConfiguration.ServiceName` for naming and views. Please either register the existing `MetricsConfiguration` instance in DI before `AddDuckDb()`, or pass the configured service name into `AddDuckDb` directly.

realtonyyoung: Fixed in 650cfe18e — `AddDuckDb` now takes the service name and `ClusterVNodeStartup` passes `_metricsConfiguration.ServiceName`, so the meter is named consistently with the rest of the pipeline (including legacy `eventstore`). Dropped the `sp.GetService<MetricsConfiguration>()` lookup that never resolved in that collection.

src/KurrentDB.SecondaryIndexing.Tests/IntegrationTests/ReadTests.cs
realtonyyoung: Fixed in 650cfe18e — `DuckDBCpuMetrics` now takes an injectable CPU-time source, and I replaced that integration assertion with deterministic unit tests using a fake clock that verify the exact recorded delta, the activity/source tags, and the zero-delta no-op gate. No reliance on real per-thread CPU, so no Windows-quantum flakiness.

realtonyyoung:
Superseded by the dedicated-executor approach. The caller-side metric fundamentally could not capture DuckDB's parallel worker-thread CPU; it's replaced by the Kurrent.Quack `DuckDBExecutor` (kurrent-io/Kurrent.Quack#52) plus the KurrentDB integration, which now lives on #5655. Closing as superseded.

PR #5662, opened 2026-06-30
[DB-2152]: Kontrol Plane implementation on top of Raft

sakno:
Changed: introduced Kontroller component and Kontrol Plane

Note that this PR contains implementation of the Kontrol Plane on top of the Raft consensus algorithm as a separated assembly (and separated csproj), main `KurrentDB.Core` doesn't reference this assembly, so the PR can be merged safely, there is no impact on the production.


timothycoleman (approved):
Thanks to @shaan1337 for doing most of the review

PR #5675, opened 2026-07-12
Fix disposing ObjectPool after an overflow

Inok:
On overflow, `ObjectPool<>` doesn't roll back the incremented counter. This prevents the pool from ever reaching `_count == 0` and calling `_onPoolDisposed` callback. 

I suspect that this problem has not actually occurred before, given that an overflow in Get is an issue in itself, and current clients are unlikely to ever reach this condition. However, the issue was technically present.

### DIFF
--- a/src/KurrentDB.Core/DataStructures/ObjectPool.cs
+++ b/src/KurrentDB.Core/DataStructures/ObjectPool.cs
@@ -14,8 +14,7 @@ public class ObjectPoolDisposingException : Exception {
 	}
 
 	public ObjectPoolDisposingException(string poolName, Exception innerException)
-		: base(string.Format("Object pool '{0}' is disposing/disposed while Get operation is requested.", poolName),
-			innerException) {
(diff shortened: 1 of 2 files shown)


src/KurrentDB.Core/DataStructures/ObjectPool.cs
```diff
@@ -131,7 +136,8 @@ private void TryDestruct() {
 
 	private void OnPoolDisposed() {
 		// ensure that we call _onPoolDisposed just once
-		if (Interlocked.CompareExchange(ref _poolDisposed, 1, 0) == 0)
-			_onPoolDisposed(this);
+		if (Interlocked.Exchange(ref _onPoolDisposed, value: null) is { } onPoolDisposed) {
```
sakno: `Interlocked.Exchange(...)?.Invoke(this)` rather than pattern matching

Inok: Fixed

sakno (approved):

PR #5677, opened 2026-07-15
Add support for multi-field user-defined indexes

shaan1337:
Extends user-defined secondary indexes from a single field to any number of fields of mixed types.

Each configured field becomes its own DuckDB column.
- A field with no value is stored as `NULL`. If all fields have no values, the record is not indexed.
- A field can optionally set `optimize_lookups` to build a DuckDB ART index for fast equality lookups, at extra write and storage cost.

Query constraints are encoded in the stream name as `field1=value;field2="string"`. Strings must be quoted (with `\"` & `\\` escapes inside quotes).

The legacy form: `:value` is still supported for single-field indexes. There's exactly one case that breaks backward compatibility but it's expected to be rare/unlikely: `$idx-user-orders:country="something"` is ambiguous - it could mean that the field's value is `something` or `country="something"`.  Right now, it is translated as `something` so that single-field indexes also support the new `field=value` form.

## Examples

Create an index over two fields:

```bash
curl -X POST http://<host>:2113/v2/indexes/orders -u admin:changeit \
  -H 'Content-Type: application/json' -d '{
    "filter": "rec => rec.schema.name == '\''OrderCreated'\''",
    "fields": [
      { "name": "country", "selector": "rec => rec.value.country", "type": "INDEX_FIELD_TYPE_STRING" },
      { "name": "total",   "selector": "rec => rec.value.total",   "type": "INDEX_FIELD_TYPE_DOUBLE" }
    ]
  }'
```

Read/subscribe via the filtered `$all` API using the index stream name as the filter:

```
$idx-user-orders                                  # whole index
$idx-user-orders:country="Mauritius"              # one field
$idx-user-orders:country="Mauritius";total=149.99 # both fields (AND), any order
```

Single-field indexes also accept the legacy bare form, e.g. `$idx-user-orders-by-country:Mauritius`.


src/KurrentDB.SecondaryIndexing/Indexes/User/UserIndexHelpers.cs
```diff
@@ -1,53 +1,179 @@
...
-		if (!TryParseQueryStreamName(streamName, out indexName, out field))
+	public static void ParseQueryStreamName(string streamName, out string indexName, out string? suffix) {
+		if (!TryParseQueryStreamName(streamName, out indexName, out suffix))
 			throw new Exception($"Unexpected error: could not parse user index stream name {streamName}");
 	}
 
-	// Parses the user index name [and field] out of the index stream that is being read
-	// $idx-user-<indexname>[:field]
-	public static bool TryParseQueryStreamName(string streamName, out string indexName, out string? field) {
+	public static bool TryParseQueryStreamName(string streamName, out string indexName, out string? suffix) {
```
Inok: We create 1-2 substrings on each query here, but callers seem ready for spans. Maybe  `indexName` and `suffix` shoud become slices instead? 
`ReadOnlySpan<char>` isn't an option for `indexName` because of `_log.LogUserIndex***(indexName)`, but `ReadOnlyMemory<char>` or `StringSegment` should be fine.

src/KurrentDB.Core/Services/Storage/SecondaryIndexReader.cs
```diff
@@ -30,40 +42,41 @@ public SecondaryIndexReaders AddReaders(IEnumerable<ISecondaryIndexReader> reade
...
 	}
 
-	[CanBeNull]
-	private ISecondaryIndexReader FindReader(string indexName) {
+	private bool TryGetReader(string indexName, out ISecondaryIndexReader reader) {
 		for (var i = 0; i < _readers.Length; i++) {
-			var reader = _readers[i];
-			if (reader.CanReadIndex(indexName))
-				return reader;
+			if (_readers[i].CanReadIndex(indexName)) {
```
Inok: Now (and before this change), `_readers[i]` would do a boundary check on each iteration because the field is mutable.

Capturing it into a `var readers = _readers;` or just replacing the loop with a `foreach` would eliminate these extra checks. As I understand, this method is on the hot path, so probably this optimization makes sense.

shaan1337: fixed, thanks!

src/KurrentDB.SecondaryIndexing/Indexes/User/UserIndexHelpers.cs
shaan1337: fixed, thanks!

src/KurrentDB.Api.V2/Modules/Indexes/Validators/CreateIndexValidator.cs
```diff
@@ -20,10 +23,13 @@ private CreateIndexValidator() {
...
 
-		// todo: when we allow multiple fields, their names will need to be unique within the request
 		RuleFor(x => x.Fields)
-			.Must(x => x.Count <= 1)
-			.WithMessage("Currently at most one index field must be provided");
+			.Must(fields => fields.Count <= MaxFields)
+			.WithMessage($"An index can have at most {MaxFields} fields");
+
+		RuleFor(x => x.Fields)
+			.Must(fields => fields.Select(f => f.Name).Distinct().Count() == fields.Count)
```
Inok, line 31: Maybe OrdinalIgnoreCase comparer should be used here? 
I suppose it's wrong to index e.g. "total" and "TOTAL" at the same time.

shaan1337: hey, `FieldValidator -> IndexNameValidator` verifies that only the following characters are allowed: `a-z0-9_-`. So the names are already lower case.

Inok: Got it, thanks!

Inok (approved):

PR #5680, opened 2026-07-15
[DB-2179] Add logging for IHostedService Start and Stop

timothycoleman:
Preventative in case a service were to disrupt graceful shutdown.

### DIFF
--- a/src/KurrentDB/HostedServiceLifecycleDecorator.cs
+++ b/src/KurrentDB/HostedServiceLifecycleDecorator.cs
@@ -1,0 +1,53 @@
+// Copyright (c) Kurrent, Inc and/or licensed to Kurrent, Inc under one or more agreements.
+// Kurrent, Inc licenses this file to you under the Kurrent License v1 (see LICENSE.md).
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using KurrentDB.Core.Time;
+using Microsoft.Extensions.Hosting;
+using Serilog;
+
+namespace KurrentDB;
+
+// Decorates every registered IHostedService and logs entry/exit/duration for
+// StartAsync and StopAsync, so a slow or stalling service stands out during shutdown.
+public sealed class HostedServiceLifecycleDecorator(IHostedService inner) : IHostedService {
+	static readonly ILogger Log = Serilog.Log.ForContext<HostedServiceLifecycleDecorator>();
+
+	readonly string _name = inner.GetType().Name;
+
+	public async Task StartAsync(CancellationToken cancellationToken) {
+		var start = Instant.Now;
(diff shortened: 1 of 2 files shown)


src/KurrentDB/HostedServiceLifecycleDecorator.cs
```diff
@@ -0,0 +1,53 @@
...
+		try {
+			await inner.StartAsync(cancellationToken);
+			Log.Debug("Started {Service} in {Elapsed} ms",
+				_name,
+				Instant.Now.ElapsedTimeSince(start).TotalMilliseconds);
+		} catch (Exception ex) {
+			Log.Debug("Failed to start {Service} after {Elapsed} ms. {Message}",
+				_name,
+				Instant.Now.ElapsedTimeSince(start).TotalMilliseconds,
+				ex.Message);
```
Inok, line 32: Maybe use the special overload for exceptions instead, to preserve more context than just a message?

```suggestion
			Log.Debug(ex, "Failed to start {Service} after {Elapsed} ms",
				_name,
				Instant.Now.ElapsedTimeSince(start).TotalMilliseconds);
```

src/KurrentDB/HostedServiceLifecycleDecorator.cs
```diff
@@ -0,0 +1,53 @@
...
+		try {
+			await inner.StopAsync(cancellationToken);
+			Log.Information("Stopped {Service} in {Elapsed} ms",
+				_name,
+				Instant.Now.ElapsedTimeSince(start).TotalMilliseconds);
+		} catch (Exception ex) {
+			Log.Information("Failed to stop {Service} after {Elapsed} ms. {Message}",
+				_name,
+				Instant.Now.ElapsedTimeSince(start).TotalMilliseconds,
+				ex.Message);
```
Inok, line 49: And here too:

```suggestion
			Log.Information(ex, "Failed to stop {Service} after {Elapsed} ms",
				_name,
				Instant.Now.ElapsedTimeSince(start).TotalMilliseconds);
```

src/KurrentDB/HostedServiceLifecycleDecorator.cs
timothycoleman: it's going to throw the exception again in a sec so i think we'll log the exception more thoroughly further out. this just captures the timing really, which isn't that interesting, its just a counterpart to the shutdown which is what we're actually interested in

Inok: Ok, got it!
I think we could omit the `ex.Message` too in that case, but it doesn't really matter for debug logs I guess :)

Inok (approved):

PR #5687, opened 2026-07-22
[DB-2193] Remove UnbufferedFileStream and replace its usage with a regular pooled stream

Inok:
`UnbufferedFileStream` attempts to do direct IO (NO_BUFFERING / O_DIRECT). However:
- It's used in a single place (PTable verification when it's opened), and only on Windows.
- In that place, replacing it with the Unix's branch (regular FileStream) actually makes loading PTable **2x faster**
- The internals are quite tricky, with a lot of old-style unsafe code, alignment handling, interop and inefficient pointer-based file reads.
- There are a few problems in interop with method signatures, error handling, etc.

So I switched PTable loading to the universal (Unix) path and removed this custom stream with all its internals.


sakno:
Let's wait for @timothycoleman for extra approval

timothycoleman (approved):

PR #5690, opened 2026-07-28
Update Jint to 4.15.3 and adopt the newer embedding surface

lahma:
Updates Jint from **4.0.3** (released 2024-09-23) to **4.15.3**, and adopts the parts of the newer embedding surface that fit what this repository already does. Draft, because a few commits are judgement calls the maintainers should make rather than inherit, and each is flagged below.

Qodo's bot review findings are addressed — see the section at the end.

---

## What changed

Three commits, one per area of the codebase, each independently reviewable:

| Commit | |
|---|---|
| **Update Jint 4.0.3 → 4.15.3 and align the scripting engines with its interop defaults** | the bump, the behaviour-default audit, `EnumConversionMode` replacing the one `IObjectConverter`, and the array-conversion audit test |
| **Rebuild the projection envelope on a shared layout and serialize state with Jint** | the headline: layout-built envelope with lazy parsed members, `JsonSerializer` replacing ~215 hand-written lines, and the amortizable execution constraint |
| **Roll back global pollution between JavaScript validations** | `JsValidationEngine` for the two process-lifetime validation engines |
| **Run Jint's host-contract verifiers in the JavaScript test suites** | an `AppContext` switch per test assembly, each with a positive control |

Net: **+770 / −461** across 16 files.

The second commit groups three changes because they share the per-event path and were measured together; the amortizable constraint sits with it rather than with the bump because it lives in the projection handler. The fourth is separate because it is test-harness policy spanning two assemblies and belongs to none of the other three.

---

## The headline: the envelope is now a shaped object

`EventEnvelope` was a custom `ObjectInstance` subclass. A host subclass carries none of the engine's storage flags, so it reaches **no member-read inline cache at all**: every `event.streamId` in a handler was a virtual call into a property dictionary, on an object rebuilt for every event.

It is now a plain `JsObject` built from a static `JsObjectLayout`, so every envelope in an engine shares one hidden class and a handler's reads stay monomorphic across events even though the object itself is new each time.

**This was not expressible when the branch started.** A layout's slots held values, so a member that must parse a JSON document could only be a raw descriptor — which deopts the object to the dictionary representation and forfeits the point. That is exactly why the envelope was a subclass, and the branch first worked around it with `TryGetOwnPropertyValue` on a still-subclassed envelope. Jint 4.15.1's `JsObjectLayout.Builder.AddLazy` declares a slot whose factory runs on the first read that *observes* its value, so `body`, `data`, `metadata` and `linkMetadata` stay lazy on an object that is hidden-class throughout. The two names for the same document share one memo, so a handler reading both parses once.

Consequences worth reviewing:

- The parsed members are ordinary properties, so `'body' in event` and `Object.keys(event)` answer for them **without parsing anything** — only reading a value parses. The key *set* still matches the old envelope exactly: `body`/`data` appear together and only for JSON events, `metadata`/`linkMetadata` always appear and carry `null` when their document is absent. That needs two layout variants; see the review findings below.
- **Own-key order is now fixed by the layout** instead of by the order members happened to be materialized in. It reproduces the order the previous implementation ended up with, and `when_enumerating_the_event_envelope` pins it.
- **Layout slots are configurable/enumerable/writable**, where the old descriptors were non-writable and non-configurable. A handler assigning to `event.streamId` used to silently do nothing and now mutates the envelope — a per-event object discarded when the handler returns. `Object.freeze` would restore the old attributes exactly, but it drops the object to the dictionary representation and would undo the reason for the change, so this is a deliberate trade rather than an oversight. **This is the change most worth a second opinion.**

Two things fall out. `TryGetOwnPropertyValue` is gone — it existed to make a host subclass cheap to read, and there is no longer a host subclass on this path. And `event.eventType` is no longer round-tripped out of the JavaScript property table to key the CLR handler dictionary; it is read straight off the record, as are `isJson` and `bodyRaw`.

### Why the version moved twice mid-branch

Adopting an API surface is a good way to find out where it does not fit, and the gaps this branch hit were reported upstream rather than worked around permanently. Seven of them shipped while the branch was open, which is why the pin moved from 4.15.0 to 4.15.3:

| What did not fit | Shipped as |
|---|---|
| A lazy member could not live in a shaped object, so the envelope had to be a subclass | `JsObjectLayout.Builder.AddLazy` — **4.15.1** |
| No way to tell whether a CLR array crosses into script, so the `LiveView` audit was prose | `GetInteropConversionDiagnostics` — **4.15.1** |
| `PropertyFlag` had no name for the combinations a host builds | Named combinations + docs — **4.15.1** |
| A lazy slot's entry must be `null`, but the values span's element type was non-nullable, forcing `null!` | `ReadOnlySpan<JsValue?>` — **4.15.2** |
| The documented correctness checker for host contracts required cloning Jint and building it in Debug | Verifiers gated on an `AppContext` switch, live in the shipped Release package — **4.15.3** |
| Restoring a global snapshot correctly meant hand-writing a lock/try/finally recipe each time | `Engine.Advanced.WithRestoredGlobals` — **4.15.3** |
| The shaping a layout is adopted for could be observed once but not asserted | `Engine.Advanced.HasSharedShape` — **4.15.3** |

Two of these close loops this branch opened by name. The `null!` one landed the same week it was reported and removed four null-forgiving operators from the envelope. The verifier one is larger: this PR earlier had to clone Jint, build it in Debug and hand-copy `Jint.dll` over the one in a test output directory to reach the checks that guard host contracts — the report said that made the documented checker unreachable for anyone who was not already a Jint contributor. 4.15.3 gates them on `AppContext.SetSwitch("Jint.EnableHostContractVerification", true)` instead, so the shipped Release package runs them; both JavaScript test assemblies now enable it in a module initializer, each with a positive control proving the switch actually took effect. One further gap remains open by choice rather than oversight: `Engine.Advanced.GetPropertyAccessSemantics` also shipped, but by then the type it would have protected had stopped being a host type — see the verdicts below.

The branch went through an intermediate shape worth knowing about when reading the diff: on 4.15.0 the envelope stayed a host subclass and used `TryGetOwnPropertyValue` to make its reads cheap. That was verified against a Debug build of Jint, whose verifier checks the hook's contract on every read — it passed, and the verifier was confirmed live by deliberately breaking the hook and watching the run fail. `AddLazy` then made the subclass unnecessary, so the hook is gone and none of it survives in the final diff.

---

## The 4.14.0 interop defaults

4.14.0 changed two defaults. Both were audited against our usage and both are **inherited deliberately** rather than pinned back.

**`Options.Interop.ArrayConversion`: `Copy` → `LiveView`.** A CLR `T[]` crossing into script is now a live fixed-size view: `Array.isArray` returns `false` (while `instanceof Array` stays `true`), and `push`/`pop`/`length` writes throw `TypeError`.

Originally this was answered by reading every call site. It is now **asserted** (`InteropArrayConversionAuditTests`): 4.15.1 counts array conversions per engine, and mapping a record then running a filter and a field selector over it — including a selector that walks a JSON array in the payload — converts no CLR array under either mode. A third case is a positive control, so the assertion cannot pass by the counters simply never incrementing.

Scope, stated honestly: the test covers the scripting engines, because `JsRecord` is the only object graph this repository projects into script and therefore the only place a CLR array could appear. The projections engine hands its handlers nothing but `JsValue`s. And the counters deliberately exclude an array crossing under a non-array *declared* type (`IReadOnlyList<T>`), which honours that contract through the ordinary wrapper lane — so this asserts "no array conversion", not "no array-shaped value".

**`Options.Interop.CacheRecentObjectWrappers`: `false` → `true`.** `JsRecordEvaluator` already creates exactly one wrapper and mutates the target behind it, so the cache is a no-op there. Where it does bite is the `JsonNode` dictionary lane, where a fresh `ObjectWrapper` was previously allocated at every level on every access — an improvement, not a hazard: the ring is bounded, `Engine.Dispose()` releases it, and no filter or selector mutates the record.

### Other deltas across the 17 releases

Audited and **not applicable**: `Error.prototype.stack` became an accessor (4.9.3 — we never read it); `Function.prototype.toString()` stopped returning source text (4.11.0 — `JsFunctionValidator` reads `FunctionDeclaration`); static CLR members left instance wrappers (4.1.0 — `JsRecord` has none); writes to read-only CLR members became a strict-mode `TypeError` (4.15.0 — every projected member has a setter); `TimeoutInterval` now replaces instead of accumulating (we call it once); saturated constraint sentinels register nothing (we use none).

Audited and **applicable — behaviour changes, not bugs**:

- **JSON grammar alignment (4.14.0).** `-09` and `1.` are now rejected as in V8, while raw U+2028/U+2029 in strings and escaped control characters in keys are now accepted. Persisted state or event bodies containing those malformed number forms parsed before and now raise `SyntaxError`.
- **`Options.Culture` (4.6.0)** stopped driving `toLocaleString` / `localeCompare`; they route through `Intl`, and `InvariantCulture` resolves to the Intl locale `"en"`. No script in our tests uses them.
- **Runtime error message text was rewritten to match V8/Node (4.7.0)**, stack traces became V8-styled (4.3.0). Our tests assert only on messages scripts throw themselves; anything downstream parsing Jint's own error text will see different strings.

### Blast radius outside this repository

`Kurrent.Surge.Core` carries its own transitive Jint **4.0.3** reference and builds its own engines for `JintRecordTransformer` and `JintSqlReducer`. Central package management lifts it to 4.15.3 **without recompilation**, and because it predates the option it cannot have pinned `ArrayConversion`. The audit test above cannot reach those engines — it asserts for the ones this repository constructs. Worth a look from whoever owns Surge.

---

### Adopted

| Feature | |
|---|---|
| `Constraint.IsAmortizable` | `TimeConstraint.Check()` reads a wall clock and nothing else — it neither counts its own invocations nor budgets a quantity that can grow unboundedly between checks. Being in the exact partition cost a virtual `Check()` per statement and, since 4.13.0, disarmed the interpreter's tight-loop lanes. |
| `Options.Interop.EnumConversion = String` | Deletes a one-type `IObjectConverter` whose mere presence disabled the compiled member-read lane for **every** wrapped member in the engine. |
| `JsonSerializer` (string overload) | Deletes ~215 lines of hand-written JSON. See compatibility below. |
| `JsObjectLayout` + `AddLazy` | The headline, above. |
| `Capture`/`RestoreGlobalSnapshot` | For the two shared validation engines that compile user-supplied JavaScript on a process-lifetime engine and reset nothing. |
| `Engine.Advanced.GetInteropConversionDiagnostics` | Turns the `LiveView` audit into an assertion. |
| `PropertyFlag` named combinations | Replaces a local three-bool shim; `AllForbidden` for `log`, `NonEnumerable` for the functions a projection calls. |

The ten explicit `Constraints.Reset()` calls were reviewed and **kept**: `Engine.Call`/`Execute` reset around themselves, but `GetSourceDefinition()` runs no script and `Load`/`LoadShared` call `JsonParser.Parse`, which checks constraints from inside its own loops without going through `ExecuteWithConstraints`.

### Skipped, with reasons

| Feature | Verdict |
|---|---|
| `JsonSerializer.Serialize(IBufferWriter<byte>)` | **Wrong direction for us.** Every consumer downstream wants a `string`; `PartitionState` even re-parses it with Newtonsoft. The old code produced UTF-8 and immediately transcoded back, so the string overload is the fit. |
| `JsonParser.Parse(ReadOnlySpan<byte>)` | **Skipped for projections, follow-up for indexing.** In projections the transcode is unavoidable: `ResolvedEvent` decodes eagerly into a public `string` field used all over V1, and `bodyRaw` is script-visible. In `KurrentDB.Scripting` it is a real opportunity — `JsRecord.Value`/`Properties` deserialize `evt.Data.Span` into a `JsonNode` tree that then crosses as an `ObjectWrapper` — but that is a model change with its own script-visible surface. |
| `Engine.Advanced.GetPropertyAccessSemantics` | **Skipped — the type it would have protected no longer exists.** The pin guards a *host-defined* type silently flipping `Exotic`↔`Ordinary` when a `Get` override moves. The envelope is now an in-box `JsObject`, for which Jint documents the answer as an internal classification that may be refined in any release. The only remaining host subclass, `InterpreterRuntime`, is definition-phase only and not reachable from a test project. |
| `TryGetOwnPropertyValue` | **Adopted, then removed.** It earned its place while the envelope was a host subclass, and nothing else here is one. |
| `CaptureGlobalSnapshot` for projection handlers | A handler's engine state is *meant* to persist — the accumulated `_state`, the installed globals. Nothing to reset between events. |
| Typed `AddObjectConverter(converter, params Type[])` | Superseded: `EnumConversionMode` removes the converter outright. |
| `Options.AddLazyGlobal` | Would change definition-phase visibility. `emit`/`linkTo`/`linkStreamTo`/`copyTo` are installed *after* `Execute(source)` precisely so the definition phase cannot see them. |
| `Prepared<Script>` / `ReferencedGlobals` | Needs a host-side refactor, not an API adoption. `PrepareScript` is used nowhere, so the V2 partition fan-out reparses per partition and on every restart. Real win, own change. |
| `ArrayLikeObject`, `JsObjectShape` | No host type projects an indexed collection; `JsObjectShape` describes singleton prototypes, not per-item records. |
| `NullPropagatingReferenceResolver` | Changes language semantics for projection authors. |
| Constraint factory overload | Not needed: each handler builds its own `Options` inline with a fresh `TimeConstraint`. |
| `Options.AddImmutableCrossing` (4.15.3) | **Declined — the promise is false where it would pay.** See below. |
| `PropertyDescriptor.CreateLazy` (4.15.3) | **N/A.** It installs a lazy descriptor on a host `ObjectInstance` subclass; our lazy members are layout slots, and there is no host subclass left on this path. |

---

## Serializer compatibility

`when_serializing_state` **already asserted** that our output was byte-identical to Jint's built-in serializer — for every object, array, value, string and number case it covers, plus the `big_state.json` production fixture. Those assertions pass unchanged after the swap. That is the compatibility oracle, and this repository wrote it.

Two deliberate deviations were pinned by their own tests:

- **Top-level `undefined`.** The old serializer wrote `"null"`; Jint reports "no JSON representation". Kept as `"null"`, because every caller treats the result as a JSON document. This now also covers functions, which the old serializer rendered as an object of their non-enumerable members.
- **BigInt.** The old serializer wrote the digits as a JSON string. `JSON.stringify` has no representation for a BigInt and throws, so the projection engines install `BigInt.prototype.toJSON` to restore the old rendering byte for byte. See the review findings below for why, and for the two consequences.

Everything else the swap changes is a correctness gain: a cycle raises a `TypeError` instead of running off a fixed 64-entry stack with `IndexOutOfRangeException`; state nested deeper than 64 levels serializes at all; `toJSON` is honoured; non-enumerable and symbol-keyed own properties are excluded per spec rather than emitted (or, for symbols, throwing).

---

## Measurements

Projection event processing, end to end. **One operation = 100 `ProcessEvent` calls**, each building the event envelope, invoking a handler that reads several members, and serializing the resulting state — the loop a running projection actually executes.

Read the delta for what it is: it spans **fifteen Jint releases plus this migration**, measured from the merge base to the tip of this branch. It is not the effect of any one commit; it is what an upgrader gets.

BenchmarkDotNet v0.15.2, default job, `[MemoryDiagnoser]`, .NET 10.0.10, AMD Ryzen 9 5950X, quiet machine, both sides run serially from an identical harness compiled against each.

| Scenario | | Before (4.0.3) | After (4.15.1) | |
|---|---|---|---|---|
| Handler reads 3 eager envelope members | Mean | 176.4 µs | 103.0 µs | **−41.6%** |
| | Allocated | 367.19 KB | 178.91 KB | **−51.3%** |
| …and one member requiring a body parse | Mean | 275.9 µs | 217.7 µs | **−21.1%** |
| | Allocated | 533.47 KB | 260.89 KB | **−51.1%** |

The pin has since moved to 4.15.3; the table is left as measured rather than re-run, since nothing in 4.15.2 or 4.15.3 touches these paths beyond a nullable annotation, a probe-lane fast path, and additions this PR adopts only in tests.

StdDev was under 1.3% of the mean on every row and one outlier was removed per row; no row was multimodal, so no re-pairing was needed. The body-parse row improves less in time because the JSON parse it adds is work neither version avoids — the envelope and serialization around it are what got cheaper. Gen1 collections also disappear: the before side promoted on both rows, the after side reports none.

The allocation figures were independently confirmed by a deterministic probe (`GC.GetAllocatedBytesForCurrentThread` over the same harness, byte-identical across repeated runs on both sides): 380,713 → 183,913 B/op and 549,898 → 270,778 B/op, agreeing with BenchmarkDotNet at ≈−51%. That is ≈1,968 B and ≈2,791 B saved per event.

**Allocation, state serialization in isolation** (`big_state.json`, 210,509 chars, 500 iterations, deterministic probe):

```
OLD  hand-written + UTF8.GetString :      667,896 B/op
NEW  Jint JsonSerializer           :      403,096 B/op   (-39.6%)
```

Steady state only; excludes the 1 MB `ArrayBufferWriter` the old serializer allocated once per handler — i.e. **per partition worker** under the V2 engine.

The benchmark harness itself is deliberately not part of this PR: it has to compile unchanged against both sides to be a valid A/B, which makes it scaffolding rather than something the repository should carry.

---

## Gates

Full solution build, `dotnet build -c Release KurrentDB.slnx`: **0 errors**, and the only warnings are pre-existing (GitInfo not finding a fork point for a branch named off `master`, and unused protobuf imports).

| Suite | Result |
|---|---|
| `KurrentDB.Projections.Management.Tests` | 523 — 452 passed, 71 skipped, **0 failed** |
| `KurrentDB.Projections.JavaScript.Tests` | 235 — **235 passed** |
| `KurrentDB.Projections.V2.Tests` | 59 — **59 passed** |
| `KurrentDB.Kontext.Tests` | 407 — **407 passed** |
| `KurrentDB.SecondaryIndexing.Tests` | 101 — **101 passed** |

Both JavaScript test assemblies now run with Jint's host-contract verifiers enabled, so those totals are green *with the checks on*.

Not run locally: anything requiring Docker or a live cluster.

While the envelope was still a host subclass, the projections suites were additionally run against a **Debug build of Jint**, whose verifier checks the `TryGetOwnPropertyValue` contract on every read; they passed, and the verifier was confirmed live by deliberately breaking the hook and watching the run fail. That is no longer needed — the envelope is an in-box `JsObject` with no host-implemented read contract to verify. What replaced it as the check is `when_enumerating_the_event_envelope`, which pins each layout variant's observable key set and order.

---

## Why `AddImmutableCrossing` was declined

4.15.3 added it, and this branch's own friction report is where it came from — the walk `record.value.customer.country` re-wrapping every node on every access. Having asked for it, the honest answer on inspection is that it does not fit here.

The memo is rooted on a wrapper, and the root of every such walk is `JsRecord`. `Remap` mutates that object in place for each event — along with its `Schema` and `Position` children — so the promise is simply false at the root and the type cannot be declared. The memo could only be rooted one level in, at the per-event `JsonNode` documents, which genuinely are built fresh per event and never mutated by us.

That is where it stops being worth it. Those documents are reachable and writable from user-supplied filter scripts, and while Jint evicts a key's memo on a write through the wrapper, a wrong promise on this path would surface as a wrong value written to a DuckDB index column and served to queries — silent and persisted. A stale read in an indexing path is a worse outcome than a missed optimization, and nothing here needs the win badly enough to take that trade.

The redundancy it would paper over is also better removed than memoized: a four-field index makes one JS call per field, each re-walking the record from scratch, so fusing the selectors into a single call eliminates the repeated walks outright. And the deeper fix is the one the envelope rework already applies on the projections side — parsing the payload into native JS values, so there is no wrapper in the path at all. The friction turned out to want that, not a memo.

---

## Review findings addressed

Qodo's bot review raised three. All three are resolved; two were real.

**1. `EventType` falls back to `""` (rule violation).** Kept, and made explicit. `ResolvedEvent.EventType` really is null whenever the resolved event carries no event record, and its declaring project has nullable reference types **disabled**, so the `string` annotation asserts nothing — dropping the coalesce would not have been safe. The removed CLR getter read the value back out of the property table through `AsString(...) ?? ""`, so `""` is the dispatch key an untyped event has always produced. Making it throw instead would be a real behaviour change on a path that previously dispatched, which is worse than the style violation. Both halves of the parity were checked: the CLR dispatch key stays `""`, and the JS-visible `event.eventType` is still fed the raw value, so it stays `null` there exactly as before. A comment at the assignment now states this.

**2. Envelope presence semantics changed (real — fixed).** The fixed layout made `body`/`data` exist on non-JSON events, and a projection distinguishing them with `('body' in event)` would have silently changed meaning. The old conditions were reconstructed from the removed code rather than guessed: `metadataRaw` and `linkMetadataRaw` were assigned unconditionally and a null string converts to JS `null`, not `undefined`, so `metadata` and `linkMetadata` were **always** present carrying `null` when absent; only `EnsureBody` additionally required `IsJson`, and `body`/`data` always appeared together. `IsJson` is therefore the sole condition, so **two** layout variants reproduce the old key sets exactly — not the eight a general treatment would need. Each variant is still one shared hidden class, so the cache win survives; a projection handling both kinds of event goes polymorphic at its read sites, which is the honest cost of the key sets genuinely differing.

  One deviation is kept deliberately because a fixed layout cannot express it: the old envelope materialized a parsed member as a side effect of the first read, so `'body' in event` answered **false** before anything had read `event.body` and **true** afterwards. The answer is now the same before and after — it describes the event, not the reading history. `when_enumerating_the_event_envelope` now pins the key set and order per variant, covering a JSON event, a non-JSON event, and an event with no metadata documents.

**3. BigInt serialization now throws (real — fixed).** The blast radius is the point: an already-running projection with a BigInt in its state would fault at serialization and halt checkpointing on upgrade. The old output format was verified from the removed serializer (`WriteStringValue(value.ToString())` — quoted decimal digits) and is restored exactly, through `BigInt.prototype.toJSON`, which `JSON.stringify` consults before deciding a value has no representation. Installed on the projection engines only; the scripting engines never ran that serializer. Not a replacer passed to `Serialize`, because a replacer is invoked for every node of every state document on the per-event path whereas `toJSON` costs nothing until a BigInt is present. Two consequences: round-tripped state turns a BigInt into a string, the same gotcha the `created` comment describes for dates; and a projection's own `JSON.stringify(123n)` now succeeds where it used to throw — strictly more lenient, and confined to projection engines. The test pins the string output again.

---

## Adjacent findings, not changed here

Noticed while auditing; none is in this PR.

1. **`JsFunctionValidator` uses a bare `new Engine()`** — no timeout, no `DisableStringCompilation` — to `Evaluate` user-supplied text, so `(function(){}) + (function(){while(1){}})()` hangs the caller. `JintEngineFactory.CreateEngine()` would fix it, but it is also `Strict()`, which could reject validators accepted today.
2. **`LinkStreamTo` has an empty `if (parameters.Length == 3) { }`**, then reads `parameters.At(4)` inside a second `if (parameters.Length == 3)`. `At` returns `undefined` out of range and `Undefined.AsObject()` throws, so `linkStreamTo(a, b, meta)` always throws.
3. **`LoadCurrentSharedState` assigns `_state = jsValue`** in its non-bi-state branch, overwriting the main state with the shared state.
4. **`LinkTo` uses `|` instead of `||`** in a `TryGetValue` chain.
5. **The projections engine sets no `LimitMemory`, `MaxStatements` or `LimitRecursion`.** Untrusted projection JavaScript is bounded by wall clock alone.
6. **Partitioned projections build two envelopes per event**, so a `partitionBy` that reads `e.body` parses the body twice.
7. **`UserIndexProcessor.TryExtractFieldValues` makes one JS call per field**, each re-walking `record.value.…`: a 4-field index runs 5 JS calls per event.
8. **`JsRecord.Remap` allocates two closures per record** capturing `evt` and `options` even when unused, plus a `Guid.ToString` and a reflective `Enum.Parse`.
9. **Per-JS-call metrics are tagged with the event type**, which is unbounded metric cardinality driven by user data.

### DIFF
--- a/src/Directory.Packages.props
+++ b/src/Directory.Packages.props
@@ -57,7 +57,7 @@
     <PackageVersion Include="IdentityModel" Version="7.0.0" />
     <PackageVersion Include="IgnoresAccessChecksToGenerator" Version="0.8.0" />
     <PackageVersion Include="JetBrains.Annotations" Version="2026.2.0" />
-    <PackageVersion Include="Jint" Version="4.0.3" />
+    <PackageVersion Include="Jint" Version="4.15.3" />
     <PackageVersion Include="Kurrent.Connectors.Elasticsearch" Version="1.1.1-alpha.1.71" />
     <PackageVersion Include="Kurrent.Connectors.Http" Version="1.1.1-alpha.1.71" />
     <PackageVersion Include="Kurrent.Connectors.Kafka" Version="1.1.1-alpha.1.71" />
--- a/src/KurrentDB.Api.V2/Modules/Indexes/Validators/JsFunctionValidator.cs
+++ b/src/KurrentDB.Api.V2/Modules/Indexes/Validators/JsFunctionValidator.cs
@@ -2,21 +2,25 @@
 // Kurrent, Inc licenses this file to you under the Kurrent License v1 (see LICENSE.md).
 
 using Jint;
+using KurrentDB.Scripting;
 
 namespace KurrentDB.Api.Modules.Indexes.Validators;
 
 internal static class JsFunctionValidator {
-	private static readonly Engine Engine = new();
+	// The engine outlives every request and evaluates user-supplied source, so each validation has its
+	// globals rolled back afterwards rather than leaving them for the next one. The engine's own
+	// configuration is unchanged.
+	private static readonly JsValidationEngine Engine = new(new Engine());
 
 	public static bool IsValidFunctionWithOneArgument(string? jsFunction) {
 		if (jsFunction is null)
 			return false;
 
 		try {
-			lock (Engine) {
-				var function = Engine.Evaluate(jsFunction).AsFunctionInstance();
+			return Engine.Validate(engine => {
+				var function = engine.Evaluate(jsFunction).AsFunctionInstance();
 				return function.FunctionDeclaration!.Params.Count == 1;
-			}
+			});
 		} catch {
 			return false;
 		}
--- a/src/KurrentDB.Kontext.Tests/Indexing/InteropArrayConversionAuditTests.cs
+++ b/src/KurrentDB.Kontext.Tests/Indexing/InteropArrayConversionAuditTests.cs
@@ -1,0 +1,87 @@
+// Copyright (c) Kurrent, Inc and/or licensed to Kurrent, Inc under one or more agreements.
+// Kurrent, Inc licenses this file to you under the Kurrent License v1 (see LICENSE.md).
+
+using Jint;
+using Kurrent.Surge.Schema.Serializers.Json;
+using KurrentDB.Scripting;
+using static KurrentDB.Kontext.Tests.Fakes.FakeSystemClient;
+
+namespace KurrentDB.Kontext.Tests.Indexing;
+
+/// <summary>
+/// Jint 4.14 changed the default for <c>Options.Interop.ArrayConversion</c> from <c>Copy</c> to
+/// <c>LiveView</c>, and the two differ in ways a script can observe: a live view writes through to the CLR
+/// array, <c>Array.isArray</c> is false for it, and resizing it throws a <c>TypeError</c>.
+/// <para>
+/// Whether that matters to us is a question about which values actually cross the interop boundary, and it
+/// was previously answerable only by reading every call site. Jint 4.15.1 counts the conversions, so this
+/// asserts the answer instead: mapping a record and running a filter and a field selector over it converts
+/// no CLR array at all, under either mode.
+/// </para>
+/// <para>
+/// Scope. This covers the scripting engines, which is where the question lives: <c>JsRecord</c> is the only
+/// object graph this repository projects into script, so it is the only place a CLR array could appear. The
+/// projections engine hands its handlers nothing but <c>JsValue</c>s and so cannot convert one. Note also
+/// what the counters deliberately exclude: under <c>LiveView</c> an array crossing under a non-array
+/// declared type (<c>IReadOnlyList&lt;T&gt;</c>) honours that declared contract through the ordinary wrapper
+/// lane and is not counted -- so this asserts "no array conversion", not "no array-shaped value".
+/// </para>
+/// </summary>
+public class InteropArrayConversionAuditTests {
+	static (Engine Engine, JsRecordEvaluator Evaluator) Make() {
+		var engine = JintEngineFactory.CreateEngine();
+		return (engine, new JsRecordEvaluator(engine, SystemJsonSchemaSerializerOptions.Default));
+	}
+
+	[Test]
+	public async Task Mapping_And_Filtering_A_Record_Converts_No_Clr_Array() {
+		var (engine, evaluator) = Make();
+		var filter = JsRecordEvaluator.Compile(engine, "rec => rec.value.org === 'acme'");
+		var selector = JsRecordEvaluator.Compile(engine, "rec => rec.value.items.length");
+
+		var evt = MakeEvent("ticket-1", 0, "TicketRaised", data: new { org = "acme", items = new[] { 1, 2, 3 } });
+		evaluator.MapRecord(evt, sequence: 1);
+
+		await Assert.That(evaluator.Match(filter)).IsTrue();
+		await Assert.That(evaluator.Select(selector)!.AsNumber()).IsEqualTo(3d);
+
+		var diagnostics = engine.Advanced.GetInteropConversionDiagnostics();
+		await Assert.That(diagnostics.ArrayLiveViewConversions).IsEqualTo(0L);
+		await Assert.That(diagnostics.ArrayCopyConversions).IsEqualTo(0L);
+	}
+
+	[Test]
+	public async Task Reading_Every_Projected_Member_Converts_No_Clr_Array() {
+		var (engine, evaluator) = Make();
+		var selector = JsRecordEvaluator.Compile(
+			engine,
+			"""
+			rec => [
+			  rec.id, rec.sequence, rec.redacted, rec.timestamp,
+			  rec.schema.name, rec.schema.format, rec.schema.id,
+			  rec.position.stream, rec.position.streamRevision, rec.position.logPosition,
+			  JSON.stringify(rec.value), JSON.stringify(rec.properties)
+			].join('|')
+			""");
+
+		evaluator.MapRecord(MakeEvent("ticket-1", 0, "TicketRaised", data: new { org = "acme" }), sequence: 1);
+		await Assert.That(evaluator.Select(selector)!.IsString()).IsTrue();
+
+		var diagnostics = engine.Advanced.GetInteropConversionDiagnostics();
+		await Assert.That(diagnostics.ArrayLiveViewConversions).IsEqualTo(0L);
+		await Assert.That(diagnostics.ArrayCopyConversions).IsEqualTo(0L);
+	}
+
+	[Test]
+	public async Task The_Counters_Would_Notice_An_Array_Crossing() {
+		// Positive control: without this, the two assertions above would also pass if the counters were
+		// never incremented at all.
+		var engine = JintEngineFactory.CreateEngine();
+		engine.SetValue("numbers", new[] { 1, 2, 3 });
+		engine.Evaluate("numbers[0]");
+
+		var diagnostics = engine.Advanced.GetInteropConversionDiagnostics();
+		await Assert.That(diagnostics.ArrayLiveViewConversions).IsEqualTo(1L);
+		await Assert.That(diagnostics.ArrayCopyConversions).IsEqualTo(0L);
+	}
+}
--- a/src/KurrentDB.Kontext.Tests/Indexing/JintHostContractVerification.cs
+++ b/src/KurrentDB.Kontext.Tests/Indexing/JintHostContractVerification.cs
@@ -1,0 +1,57 @@
+// Copyright (c) Kurrent, Inc and/or licensed to Kurrent, Inc under one or more agreements.
+// Kurrent, Inc licenses this file to you under the Kurrent License v1 (see LICENSE.md).
+
+using System.Runtime.CompilerServices;
+using Jint;
+using Jint.Native;
+using Jint.Native.Object;
+
+namespace KurrentDB.Kontext.Tests.Indexing;
+
+internal static class JintHostContractVerification {
+	/// <summary>
+	/// Turns on Jint's host-contract verifiers for this test assembly.
+	/// </summary>
+	/// <remarks>
+	/// These are the checks that catch host code answering one of the engine's extension points in a way
+	/// that contradicts another. The engine trusts those hooks on its hot paths and cannot re-verify them
+	/// there, so a violation is otherwise silent.
+	/// <para>
+	/// The switch has to be set before the first use of any Jint type: the flag behind it is read once at
+	/// type initialization, and flipping it afterwards does nothing for the rest of the process. A module
+	/// initializer runs before any test in the assembly, which is early enough. There is no cost to leaving
+	/// it off in production -- the guards fold away entirely when the switch is unset -- and it is
+	/// deliberately not set anywhere but here.
+	/// </para>
+	/// </remarks>
+	[ModuleInitializer]
+	internal static void Enable() => AppContext.SetSwitch("Jint.EnableHostContractVerification", true);
+}
+
+/// <summary>
+/// Positive control for the switch above. The scripting engines have no host <c>ObjectInstance</c>
+/// subclass today, so the verifiers currently guard against a future one rather than anything present --
+/// which makes it all the more worth proving the switch is actually live, since nothing else in this
+/// assembly would notice if it were not.
+/// </summary>
+public class JintHostContractVerificationTests {
+	/// <summary>
+	/// Claims every name as its own while <c>GetOwnProperty</c> reports them all absent, which is the
+	/// contradiction <c>TryGetOwnPropertyValue</c>'s contract forbids.
+	/// </summary>
+	sealed class LyingHostObject(Engine engine) : ObjectInstance(engine) {
+		protected override bool TryGetOwnPropertyValue(JsValue property, JsValue receiver, out JsValue value) {
+			value = JsValue.Null;
+			return true;
+		}
+	}
+
+	[Test]
+	public async Task The_verifiers_are_enabled_in_this_assembly() {
+		var engine = new Engine();
+		engine.SetValue("host", new LyingHostObject(engine));
+
+		var ex = Assert.Throws<InvalidOperationException>(() => engine.Evaluate("host.anything"));
+		await Assert.That(ex!.Message).Contains("TryGetOwnPropertyValue");
+	}
+}
--- a/src/KurrentDB.Kontext.Tests/Indexing/JsValidationEngineTests.cs
+++ b/src/KurrentDB.Kontext.Tests/Indexing/JsValidationEngineTests.cs
@@ -1,0 +1,47 @@
+// Copyright (c) Kurrent, Inc and/or licensed to Kurrent, Inc under one or more agreements.
+// Kurrent, Inc licenses this file to you under the Kurrent License v1 (see LICENSE.md).
+
+using Jint;
+using KurrentDB.Scripting;
+
+namespace KurrentDB.Kontext.Tests.Indexing;
+
+public class JsValidationEngineTests {
+	[Test]
+	public async Task Globals_Defined_By_One_Validation_Are_Not_Visible_To_The_Next() {
+		var engine = new JsValidationEngine(JintEngineFactory.CreateEngine());
+
+		// A filter expression is evaluated as `(<expr>)`, and a comma expression lets one both define a
+		// global and produce the function that passes validation. On a shared, process-lifetime engine
+		// that global would otherwise outlive the request that created it.
+		engine.Validate(e => e.Evaluate("(globalThis.leaked = 1, r => true)"));
+
+		var stillThere = engine.Validate(e => e.Evaluate("typeof globalThis.leaked").AsString());
+		await Assert.That(stillThere).IsEqualTo("undefined");
+	}
+
+	[Test]
+	public async Task Globals_Are_Restored_Even_When_A_Validation_Throws() {
+		var engine = new JsValidationEngine(JintEngineFactory.CreateEngine());
+
+		Assert.Throws<Exception>(() => engine.Validate<object>(e => {
+			e.Evaluate("globalThis.leaked = 1");
+			throw new InvalidOperationException("rejected");
+		}));
+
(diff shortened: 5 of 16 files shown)


lahma:
All three findings addressed. Two were real; the first is a parity question rather than a new decision.

**1. `EventType` falls back to `""`.** Kept, with the reasoning now written at the assignment.

I checked whether the coalesce could just be dropped. It cannot: `ResolvedEvent.EventType` is assigned `null` whenever the resolved event carries no event record, and `KurrentDB.Projections.Shared` has nullable reference types disabled, so the `string` annotation asserts nothing about it.

Nor is `""` a new default. The CLR getter this replaced read the value back out of the JavaScript property table through `AsString(...) ?? ""`, so `""` is the dispatch key an untyped event has always produced. Making it throw instead would change behaviour on a path that previously dispatched — a worse outcome than the rule violation, on the event-processing hot path of a database.

Both halves of the parity were verified, since the CLR property and the JS-visible property are fed separately: the dispatch key stays `""`, and the layout is still handed the raw value, so `event.eventType` stays `null` in script exactly as before.

**2. Envelope presence semantics changed.** Agreed, and fixed with conditional layout variants.

The old conditions were reconstructed from the removed code rather than assumed, and they are narrower than they look. `metadataRaw` and `linkMetadataRaw` were assigned unconditionally, and a null string converts to JS `null` rather than `undefined`, so `EnsureMetadata` and `EnsureLinkMetadata` always succeeded — `metadata` and `linkMetadata` were *always* present, carrying `null` when their document was absent, which is distinct from being absent. Only `EnsureBody` additionally required `IsJson`, and `body`/`data` always appeared together sharing one descriptor.

`IsJson` is therefore the sole condition, so two layout variants reproduce the old key sets exactly. Each variant is still one shared hidden class, so the inline-cache win survives; a projection handling both JSON and non-JSON events goes polymorphic at its read sites, which is the honest cost of the key sets genuinely differing and still cheaper than the previous host-subclass path, which reached no cache at all.

One deviation is kept deliberately because a fixed layout cannot express it: the old envelope materialized a parsed member as a side effect of the first read of it, so `'body' in event` answered `false` before anything had read `event.body` and `true` afterwards. It now answers the same before and after — it describes the event rather than the reading history.

`when_enumerating_the_event_envelope` now pins the key set and order per variant, covering a JSON event, a non-JSON event, and an event with no metadata documents.

**3. BigInt serialization now throws.** Agreed, and reverted to the old behaviour. The blast radius is the point — an already-running projection with a BigInt in its state would fault at serialization and halt checkpointing on upgrade, which is not an acceptable way to discover the change.

Option 1 from the suggestion, with the exact old format verified from the removed serializer first (`writer.WriteStringValue(value.ToString())`, i.e. quoted decimal digits). `BigInt.prototype.toJSON` is installed on the projection engines only; the scripting and validation engines never ran that serializer and have no behaviour to preserve. Not a replacer passed to `Serialize`, because a replacer is invoked for every node of every state document on the per-event path, whereas `toJSON` costs nothing until a BigInt is actually present.

Two consequences, both documented in the PR description: round-tripped state turns a BigInt into a string, the same gotcha the existing `created` comment describes for dates; and a projection's own `JSON.stringify(123n)` now succeeds where it previously threw — strictly more lenient, and confined to projection engines. The serialization test pins the string output again.

The branch has also been repackaged from eleven commits into three, since most of the history was version churn while upstream fixes landed mid-review.

PR #5695, opened 2026-08-03
Point at Gaffer and Navigator from the embedded UI

George-Payne:
Surfaces Gaffer from the projections pages, and turns the Tools sidebar's Navigator entry into a deep link at the node you are viewing.

**Sidebar**

image

**Projections Overview**

image

**Projections detail**

image

- **`Components/Projections/Projections.razor`** — a dense 38px ribbon under the page header, the whole band linking out. It sits inside the available-and-leader branch so it never stacks under the "not enabled" or "open projections on the leader" notices, which say something more urgent.
- **`Components/Projections/ProjectionDetail.razor`** — a link beside the Source heading, for user projections only. System projections ship with the server, so there is nothing to author locally and no link to offer.
- **`Components/Layout/NavMenu.razor`** — a Gaffer entry in Tools, and Navigator now hands the browser a `kurrentdb://` connection string built from the address the browser actually reached this node on, rather than the node's advertised address, which can be cluster-internal and unreachable from the client. `+discover` only when gossip reports more than one member.
- **`Components/Pages/Navigator.razor`** — deleted, with `wwwroot/navigator.png` (232K out of the binary). Its content was a sentence, a download link and a screenshot; the sidebar entry does that job directly.
- **`Components/Shared/GafferLink.cs`, `NavigatorLink.cs`** — outbound URL construction. Gaffer links are UTM-tagged with the same source/medium/campaign Navigator sends, so the two admin surfaces aggregate as one referral channel rather than two. The Navigator connection string deliberately carries no credentials — this UI never sees the signed-in user's password — and adds `tls=false` for an http node, without which Navigator attempts TLS against an insecure server and fails to connect.
- **`UI/Theme/KurrentIcons.cs`** — the Gaffer logomark, ported from Navigator with its `327 34 52 52` source viewBox mapped onto MudIcon's 24x24.
- **`KurrentDB.csproj`** — `InternalsVisibleTo` for `KurrentDB.Components.Tests`, so the URL building is testable without making the helpers public. Core, Api.V2 and others already declare the same.
- **Tests** — 12 new in `KurrentDB.Components.Tests`: ribbon presence across the page's available, populated and subsystem-disabled states, the detail link's system/user gating, and the Navigator connection string's scheme selection, tls flag and explicit port.


timothycoleman (approved):
thanks!

PR #5717, opened 2026-08-22
[DB-2208] Upgrade package for GHSA-q939-rpr3-3284

timothycoleman:


src/KurrentDB.Ammeter/KurrentDB.Ammeter.csproj
```diff
@@ -10,6 +10,8 @@
 	<ItemGroup>
 		<PackageReference Include="RestSharp" />
 		<PackageReference Include="Testcontainers" />
+		<!-- Override vulnerable transitive (pulled by Testcontainers). -->
```
Inok, line 13: We can enable [Transitive Pinning](https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management#transitive-pinning) instead; then we won't need to reference transitive packages explicitly. In Directory.Packages.props:
```
<CentralPackageTransitivePinningEnabled>true</CentralPackageTransitivePinningEnabled>
```

The only thing is that it's better to run `dotnet package list --include-transitive` before and after that change and compare the results; enabling transitive pinning may cause unexpected version drift (though likely in the right direction).

timothycoleman: interesting thanks i didn't know about that, i'll have a read about it. i'm a bit wary of anything that might non-explicitly change the version numbers in use. for the minute lets do it the old way

sakno (approved):

PR #5718, opened 2026-08-22
Introduce MemberInfoLite for describing the leader

timothycoleman:
Consumers that only need to know "who is the leader and how do I reach them" were being handed a full MemberInfo, most of whose fields they never read. That made the type hard to produce from anything other than gossip. MemberInfoLite is the subset that is actually consumed, so it can be produced by anything that knows a node's addresses.

Migrates the messages and holders that carry the leader: ElectionsDone, LeaderDiscoveryMessage.LeaderFound, SystemMessage.ReplicaStateMessage, ReplicationMessage, GossipServiceBase, HttpSendService, ReplicaService, LeaderInfoProvider and ClusterVNodeController.

Two fields change meaning rather than just moving:

- The client endpoints are the ADVERTISED addresses. MemberInfo keeps the untranslated address plus separate AdvertiseHostToClientAs and Advertise*PortToClientAs overrides, and LeaderInfoProvider applied them. MemberInfoLite does not carry the overrides, so ToLite applies them and every consumer takes the addresses as given.

- The internal secure/insecure pair collapses to one ReplicationEndPoint. A node configures exactly one of the pair, and a replica dials with its own TLS setting regardless of which the leader advertises, so ReplicaService.GetLeaderEndPoint had nothing left to choose between.

Removing GetLeaderEndPoint loses its "secure connections are required but the leader has no secure endpoint" diagnostic, which fired only when dialling the leader. ClusterMixedInternalTlsLogger replaces it from gossip: it reports every node that disagrees, before any connection is attempted, and does not depend on which node we happen to be following.

Also renames MemberInfoLite.Is to HasReplicationEndPoint, since its callers only ask whether a lost replication connection was the leader's, and parenthesises MemberInfo.Is and VNodeInfo.Is, where `endPoint != null && a || b` left the null guard covering only the first comparison - and MemberInfo.Is would then throw rather than return false, because EndPointEqualityComparer.Equals dereferences both arguments.

### DIFF
--- a/src/Connectors/KurrentDB.Connectors.Tests/System/LeaderNodeBackgroundServiceTests.cs
+++ b/src/Connectors/KurrentDB.Connectors.Tests/System/LeaderNodeBackgroundServiceTests.cs
@@ -72,7 +72,7 @@ public class LeaderNodeBackgroundServiceTests(ITestOutputHelper output, Connecto
 
 			await sut.WaitUntilExecuting();
 
-			MessageBus.Publish(new SystemMessage.BecomeFollower(Guid.NewGuid(), FakeMemberInfo));
+			MessageBus.Publish(new SystemMessage.BecomeFollower(Guid.NewGuid(), FakeMemberInfo.ToLite()));
 
 			TaskCompletionSource<SystemMessage.ComponentTerminated> componentTerminated = new();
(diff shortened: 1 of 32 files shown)


timothycoleman:
can't seem to reopen this against a different branch, replaced by #5723

PR #5731, opened 2026-09-16
[DB-2219]: Removed quorum-based fencing

sakno:
Changed: Removed quorum-based fencing


timothycoleman (approved):
We thought it'd be neat to proceed with the appointment as soon as quorum nodes have responded.

This reverts that optimisation and so we wait for any nodes that respond within the timeout.

It matters because followers ack before flush, so it is possible (though unlikely) for a write to be present on the leader (on disk) and one follower (in memory), then for that follower to be restarted (losing the write) and back in time to participate in an appointment before the leader was able to replicate that write to the other follower, then for the two followers to be the ones who respond to the appointment fence first.

